PR #13, opened 2020-06-27
nv2a: Enable fixed-function sphere map texgen

Voxel9:
Can confirm the previously untested sphere map texgen code in nv2a_shader.c works fine, and should be safe to enable:

image

Had to also fix assertions wrongly comparing against the texture stage index instead of the channel index.


mborgerson:
Great! Thanks for testing and providing a screenshot.

PR #28, opened 2020-10-23
nv2a: fix hangs when polling NV_PCRTC_RASTER

LukeUsher:
This PR implements the NV_PCRTC_RASTER register, albeit as a hack.

This register is used by Direct3D to detect current status (eg. is blank in progress), this implementation switches between three states: start of frame, mid-frame and vblank: This is enough to coax games that use this method to wait for Vblank out of hanging.

With this, Alter Echo no longer hangs on the loading screen, instead, it hangs after the first line of spoken dialog. This is a different (unrelated) audio issue, however.

Before:
Screenshot 2020-10-23 at 21 31 08

After:
Screenshot 2020-10-23 at 21 34 32

### DIFF
--- a/hw/xbox/nv2a/nv2a_regs.h
+++ b/hw/xbox/nv2a/nv2a_regs.h
@@ -603,7 +603,7 @@
 #   define NV_PCRTC_INTR_EN_0_VBLANK                            (1 << 0)
 #define NV_PCRTC_START                                   0x00000800
 #define NV_PCRTC_CONFIG                                  0x00000804
-
+#define NV_PCRTC_RASTER                                  0x00000808
 
 #define NV_PVIDEO_INTR                                   0x00000100
 #   define NV_PVIDEO_INTR_BUFFER_0                              (1 << 0)
--- a/hw/xbox/nv2a/pcrtc.c
+++ b/hw/xbox/nv2a/pcrtc.c
@@ -21,6 +21,8 @@
 
 #include "nv2a_int.h"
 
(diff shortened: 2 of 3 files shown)


Xcedf:
Seems to fix cutscene in Conker, previously it was just stucking when you come ot scarecrow
image

mborgerson:
@LukeUsher thanks for the contribution. Don't worry about rebasing it; there are some things I'd like to look at/change before merging.


@Xcedf

> Seems to fix cutscene in Conker, previously it was just stucking when you come ot scarecrow

Thanks for testing this PR. Have you have modified xemu to bypass `assert`s in the code to get through the menus? You are welcome and encouraged to test PRs, but if you do, please always be specific about what changes if any you have made to the code base in addition to the patch.

Regarding the cutscene, you should be able to trigger this particular cutscene on the master branch (bypassing the associated assertions, of course) if you follow what you have done here and walk into the gated area and stand on the pad. I was able to trigger this cutscene with master as well.

As a side note, based on my current development work, many cutscene issues appear to be caused by lack of proper audio support. I'm actively working on that, and in my testing it has fixed most/all of the associated problems; so stay tuned for updates there.

Xcedf:
@mborgerson Yea i commented out the asserts to make past the menu, but these are the only modifications, thanks for explanation and lots of thanks to you for your beautifull work, best of luck to you.

mborgerson:
@LukeUsher thanks a lot for sharing info and a patch here! I've opted for a simpler implementation which should do the trick via #36, so I'm going to close this PR now.

PR #38, opened 2020-10-27
nv2a: Add support for compressed 3D textures

wilkovatch:
Allows Midtown Madness 3 (and potentially other games using compressed 3d textures) to go ingame, in MM3 they're used mainly for animated particle effects (the failed assert caused xemu to crash when trying to go ingame).


mborgerson:
Great, thanks! This should unblock "Conker: Live and Reloaded" too.

PR #39, opened 2020-10-27
nv2a: Fix compressed 3D textures

wilkovatch:


haxar:
This patch needs additional investigation, as calling `glCompressedTexImage3D()` here with a S3TC compressed texture throws a `GL_INVALID_ENUM` error when going through a menu item in _Conker: Live and Reloaded_.

From _Using Texture Compression in OpenGL_ in `ARB_texture_compression.pdf`:

> Depth  - Only relevant for 3D textures and if not using S3TC (see S3TC note below)

> `<depth>` is irrelevant for S3TC formats since they are only applicable to 2D textures, meaning, `glCompressedTexImage1DARB` and `glCompressedTexImage3DARB` will produce an `INVALID_ENUM` error if `<internalFormat>` is an S3TC format.

wilkovatch:
It turns out that this PR (along with the previous) adds support only on Nvidia GPUs, since it's using this Nvidia-only extension: https://www.khronos.org/registry/OpenGL/extensions/NV/NV_texture_compression_vtc.txt

haxar:
This issue is now addressed in: #40

PR #47, opened 2020-10-29
smbus: Add persistent generic storage device

### LINKED ISSUE #43: EEPROM data should be propagated back to file
EEPROM contents are loaded from a file but are totally lost after xemu closes. It should writeback to the EEPROM file.

@Ernegien is working on fixing this.

Ernegien:
These changes decouple the legacy `smbus_eeprom` from the Xbox machine state and replace it with a generic `smbus_storage` device which is file-backed for persistence. Future work will be required to allow for dynamic generation of new EEPROMs if unspecified, but I consider this a better drop-in replacement to the existing functionality for now.

Fixes: #43


hw/xbox/smbus_cat24wc02.c
```diff
@@ -0,0 +1,193 @@
...
+    .instance_size = sizeof(SMBusCAT24WC02Device),
+    .class_init = smbus_cat24wc02_class_initfn
+};
+
+static void smbus_cat24wc02_register_devices(void)
+{
+    type_register_static(&smbus_cat24wc02_info);
+}
+
+type_init(smbus_cat24wc02_register_devices)
```
dracc: Nit: missing newline

dracc (approved):

dracc:
Tested, works both for writing and reading settings. Amazing job! :muscle:

softmmu/vl.c
```diff
@@ -2917,25 +2917,35 @@ void qemu_init(int argc, char **argv, char **envp)
...
-        char *msg = g_strdup_printf("Failed to open EEPROM file '%s'. Please check machine settings.", eeprom_path);
-        xemu_queue_error_message(msg);
-        g_free(msg);
-        eeprom_path = "";
-    }
-
     int short_animation;
     xemu_settings_get_bool(XEMU_SETTINGS_SYSTEM_SHORTANIM, &short_animation);
 
     fake_argv[fake_argc++] = g_strdup_printf("xbox%s%s%s%s",
```
mborgerson: One `%s` needs to be removed here (otherwise we may print garbage or segfault)

softmmu/vl.c
```diff
@@ -2917,25 +2917,35 @@ void qemu_init(int argc, char **argv, char **envp)
...
+            char *msg = g_strdup_printf("Failed to open EEPROM file '%s'. Please check machine settings.", eeprom_path);
+            xemu_queue_error_message(msg);
+            g_free(msg);
+        }
+        fake_argv[fake_argc++] = strdup("-blockdev");
+        fake_argv[fake_argc++] = g_strdup_printf("raw,node-name=xbeeprom,file.driver=file,file.filename=%s,file.locking=on", eeprom_path);
+        fake_argv[fake_argc++] = strdup("-device");
+        fake_argv[fake_argc++] = strdup("smbus-cat24wc02,drive=xbeeprom");
+    }
+    else
```
mborgerson: nit: Please use single line `} else {`

hw/xbox/smbus_cat24wc02.c
```diff
@@ -0,0 +1,193 @@
...
+
+//#define DEBUG
+#ifdef DEBUG
+# define DPRINTF(format, ...) printf(format, ## __VA_ARGS__)
+#else
+# define DPRINTF(format, ...) do { } while (0)
+#endif
+
+#define SMBUS_EEPROM_SIZE 256
+#define SMBUS_EEPROM_ADDRESS 0x54
```
mborgerson: Ideally this would not be hardcoded, but its okay for now

hw/xbox/smbus_cat24wc02.c
```diff
@@ -0,0 +1,193 @@
...
+        if (blk_set_perm(eeprom->blk, BLK_PERM_CONSISTENT_READ | BLK_PERM_WRITE, BLK_PERM_ALL, &error_fatal) < 0) {
+            error_setg(errp, "%s: incorrect EEPROM file permissions", TYPE_SMBUS_CAT24WC02);
+            return;
+        }
+
+        if (blk_pread(eeprom->blk, 0, eeprom->data, SMBUS_EEPROM_SIZE) != SMBUS_EEPROM_SIZE) {
+            error_setg(errp, "%s: EEPROM file read failure", TYPE_SMBUS_CAT24WC02);
+            return;
+        }
+    }
```
mborgerson: nit: Please format as `} else {`

hw/xbox/smbus_cat24wc02.c
```diff
@@ -0,0 +1,193 @@
...
+        memcpy(eeprom->data, default_eeprom, SMBUS_EEPROM_SIZE);
+    }
+}
+
+static int smbus_cat24wc02_write_data(SMBusDevice *dev, uint8_t *buf, uint8_t len)
+{
+    SMBusCAT24WC02Device *eeprom = SMBUS_CAT24WC02(dev);
+
+    #ifdef DEBUG
+        printf("eeprom_write_cmd: cmd=0x%02x val=0x%02x\n", buf[0], buf[1]);
```
mborgerson: * You can use the `DPRINTF` print macro you defined above here.
* The function prefix in the `printf` statement doesn't match the name of the function; you could use `__func__` here instead of copying the function name too
* Lets also print out the address here

So something like:
```c
DPRINTF("%s: addr=0x%02x cmd=0x%02x val=0x%02x\n", __func__, dev->i2c.address, buf[0], buf[1]);
```

hw/xbox/smbus_cat24wc02.c
```diff
@@ -0,0 +1,193 @@
...
+}
+
+static uint8_t smbus_cat24wc02_receive_byte(SMBusDevice *dev)
+{
+    SMBusCAT24WC02Device *eeprom = SMBUS_CAT24WC02(dev);
+
+    uint8_t val = eeprom->data[eeprom->offset++];
+
+    #ifdef DEBUG
+        printf("eeprom_receive_byte: val=0x%02x\n", val);
```
mborgerson: Same comment as above w.r.t. this print statement.

hw/xbox/smbus_cat24wc02.c
```diff
@@ -0,0 +1,193 @@
...
+    /* len is guaranteed to be > 0 */
+    eeprom->offset = buf[0];
+    buf++;
+    len--;
+
+    for (; len > 0; len--) {
+        eeprom->data[eeprom->offset] = *buf++;
+
+        #ifdef DEBUG
+            printf("eeprom_write_byte: offset=0x%02X, data=0x%02x\n",
```
mborgerson: Same comment as above w.r.t this print statement.

hw/xbox/smbus_cat24wc02.c
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    return 0;
+}
+
+static uint8_t smbus_cat24wc02_receive_byte(SMBusDevice *dev)
+{
+    SMBusCAT24WC02Device *eeprom = SMBUS_CAT24WC02(dev);
+
+    uint8_t val = eeprom->data[eeprom->offset++];
```
mborgerson: `offset` is an 8 bit value so this should be safe as-is and wrap around as the hardware would, but for clarity lets change this to explicitly force it to wrap around (like what is done above):
```c
uint8_t val = eeprom->data[eeprom->offset];
eeprom->offset = (eeprom->offset + 1) % SMBUS_EEPROM_SIZE;
return val;
```

softmmu/vl.c
```diff
@@ -2917,25 +2917,35 @@ void qemu_init(int argc, char **argv, char **envp)
...
         strlen(bootrom_path) > 0 ? g_strdup_printf(",bootrom=%s", bootrom_path) : "", // Leak
-        strlen(eeprom_path) > 0 ? g_strdup_printf(",eeprom=%s", eeprom_path) : "", // Leak
         short_animation ? ",short-animation" : "",
         ",kernel-irqchip=off"
         );
 
+    const char *eeprom_path;
+    xemu_settings_get_string(XEMU_SETTINGS_SYSTEM_EEPROM_PATH, &eeprom_path);
+    if (strlen(eeprom_path) > 0) {
+
```
mborgerson: nit: Drop blank line

softmmu/vl.c
```diff
@@ -2917,25 +2917,35 @@ void qemu_init(int argc, char **argv, char **envp)
...
-        strlen(eeprom_path) > 0 ? g_strdup_printf(",eeprom=%s", eeprom_path) : "", // Leak
         short_animation ? ",short-animation" : "",
         ",kernel-irqchip=off"
         );
 
+    const char *eeprom_path;
+    xemu_settings_get_string(XEMU_SETTINGS_SYSTEM_EEPROM_PATH, &eeprom_path);
+    if (strlen(eeprom_path) > 0) {
+        
+        if (xemu_check_file(eeprom_path)) {
```
mborgerson: Can you replace this with `get_image_size(filename)` (defined in include/hw/loader.h, header should already be included above) and check the EEPROM size here too. We've had issues with people passing in large files here and xemu catches it too late and just crashes without telling the user.

hw/xbox/smbus_cat24wc02.c
```diff
@@ -0,0 +1,193 @@
...
+    .parent = TYPE_SMBUS_DEVICE,
+    .instance_size = sizeof(SMBusCAT24WC02Device),
+    .class_init = smbus_cat24wc02_class_initfn
+};
+
+static void smbus_cat24wc02_register_devices(void)
+{
+    type_register_static(&smbus_cat24wc02_info);
+}
+
```
mborgerson: This new device does not inclued a `VMStateDescription`, so EEPROM contents will not be preserved in a snapshot. Is this an intentional decision? I think loading a snapshot should restore the EEPROM contents at time of creation to keep machine state consistent.

hw/xbox/smbus_cat24wc02.c
```diff
@@ -0,0 +1,193 @@
...
+static void smbus_cat24wc02_realize(DeviceState *dev, Error **errp)
+{
+    qdev_prop_set_uint8(dev, "address", SMBUS_EEPROM_ADDRESS);
+
+    SMBusCAT24WC02Device *eeprom = SMBUS_CAT24WC02(dev);
+    eeprom->offset = 0;
+
+    if (eeprom->blk) {
+        // NOTE: returns the minimum sector size of 512; let the UI handle additional validation or just use first 256 bytes regardless
```
mborgerson: nit: Please wrap all lines at column 80

hw/xbox/smbus_cat24wc02.c
```diff
@@ -0,0 +1,193 @@
...
+
+    for (; len > 0; len--) {
+        eeprom->data[eeprom->offset] = *buf++;
+
+        #ifdef DEBUG
+            printf("eeprom_write_byte: offset=0x%02X, data=0x%02x\n",
+                eeprom->offset, eeprom->data[eeprom->offset]);
+        #endif
+
+        if (eeprom->blk) {
```
mborgerson: re: Above comment about read-only support, condition here would change to `if (eeprom->blk && !eeprom->ro)`

hw/xbox/smbus_cat24wc02.c
```diff
@@ -0,0 +1,193 @@
...
+            if (blk_pwrite(eeprom->blk, eeprom->offset, eeprom->data + eeprom->offset, 1, 0) != 1) {
```
mborgerson: Let's move this condition/block driver call out of this loop and just flush the whole 256 bytes back down to the device after this loop has handled wrap around to reduce our time in the block driver.

hw/xbox/smbus_cat24wc02.c
```diff
@@ -0,0 +1,193 @@
...
+            error_setg(errp, "%s: incorrect EEPROM file size", TYPE_SMBUS_CAT24WC02);
+            return;
+        }
+
+        if (blk_set_perm(eeprom->blk, BLK_PERM_CONSISTENT_READ | BLK_PERM_WRITE, BLK_PERM_ALL, &error_fatal) < 0) {
+            error_setg(errp, "%s: incorrect EEPROM file permissions", TYPE_SMBUS_CAT24WC02);
+            return;
+        }
+
+        if (blk_pread(eeprom->blk, 0, eeprom->data, SMBUS_EEPROM_SIZE) != SMBUS_EEPROM_SIZE) {
```
mborgerson: This appears to succeed even if the EEPROM file is 1 byte long, leaving the rest of the EEPROM data uninitialized. The above mentioned check in the UI should get this, but just good to be aware of.

softmmu/vl.c
```diff
@@ -2917,25 +2917,35 @@ void qemu_init(int argc, char **argv, char **envp)
...
+            char *msg = g_strdup_printf("Failed to open EEPROM file '%s'. Please check machine settings.", eeprom_path);
+            xemu_queue_error_message(msg);
```
mborgerson, line 2945: This error message will never actually be displayed, because the block driver will cause xemu to bail out before it can be displayed. In this case, set eeprom_path=NULL so we go through the default EEPROM path and the UI can come up. e.g.

```
get eeprom path
if (eeprom_path)
    check eeprom file sanity
    if not sane
        queue error message
        eeprom_path = NULL

if (eeprom_path)
   ...
else
   ...

**Comment by Ernegien** [comment]
Windows build was segfaulting with "Block protocol 'file' doesn't support the option 'locking'". Need to investigate locking further at the blockdev level.

**Comment by mborgerson** [comment]
Nice work! I am able to successfully update my EEPROM with this patch.

Unfortunately, this appears to break snapshots with a raw EEPROM image. Can you look into a possible solution for this? If this is actually the case (which is news to me), then we may need to revert to simple libc file functions and forego the block layer :(

**Comment by Ernegien** [comment]
Well, blockdev was certainly a learning experience; unfortunate about the [lack of out-of-the-box snapshot support](https://bugs.launchpad.net/qemu/+bug/1860759) however so simple files will have to do for now. I've opted to go a slightly different route making the implementation a bit more generic which will also allow for adding mod-chip EEPROM devices of different addresses/sizes etc. for those that may need it.

The default EEPROM stuff can be considered temporary and will be pulled out in a future PR where they'll be dynamically generated upstream if unspecified.

File: hw/xbox/smbus_storage.c [hunk 18]
```diff
@@ -0,0 +1,213 @@
...
+#else
+# define DPRINTF(format, ...) do { } while (0)
+#endif
+
+typedef struct SMBusStorageDevice {
+    SMBusDevice smbusdev;
+    char* file;
+    uint8_t *data;
+    uint32_t size;
+    uint16_t addr;
```
Ernegien: nit: move addr after offset for better packing?

hw/xbox/smbus_storage.c
```diff
@@ -0,0 +1,213 @@
...
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00
+};
+
+static void smbus_storage_realize(DeviceState *dev, Error **errp)
+{
+    SMBusStorageDevice *s = SMBUS_STORAGE(dev);
+    qdev_prop_set_uint8(dev, "address", s->addr);
+    s->data = g_malloc0(s->size);
```
Ernegien, line 96: Not sure if/how this gets free'd, referencing `hw/nvram/eeprom_at24c.c` does the same thing...

softmmu/vl.c
```diff
@@ -2917,25 +2917,40 @@ void qemu_init(int argc, char **argv, char **envp)
...
         strlen(bootrom_path) > 0 ? g_strdup_printf(",bootrom=%s", bootrom_path) : "", // Leak
-        strlen(eeprom_path) > 0 ? g_strdup_printf(",eeprom=%s", eeprom_path) : "", // Leak
         short_animation ? ",short-animation" : "",
         ",kernel-irqchip=off"
         );
 
+    const char *eeprom_path;
+    xemu_settings_get_string(XEMU_SETTINGS_SYSTEM_EEPROM_PATH, &eeprom_path);
+    fake_argv[fake_argc++] = strdup("-device");
+    if (strlen(eeprom_path)) {
```
Ernegien: nit: can eliminate this and just directly check file size

hw/xbox/smbus_storage.c
```diff
@@ -0,0 +1,213 @@
...
+        DPRINTF("%s: addr=0x%02x off=0x%02x, data=0x%02x\n",
+            __func__, s->addr, s->offset, s->data[s->offset]);
+        s->offset = (s->offset + 1) % s->size;
+    }
+
+    if (changed && s->file && s->persist) {
+        int fd = open(s->file, O_WRONLY | O_BINARY);
+        if (fd < 0) {
+            DPRINTF("%s: file '%s' could not be opened\n", __func__, s->file);
+            close(fd);
```
mborgerson: Don't need to `close` in this case

hw/xbox/smbus_storage.c
```diff
@@ -0,0 +1,213 @@
...
+    .parent = TYPE_SMBUS_DEVICE,
+    .instance_size = sizeof(SMBusStorageDevice),
+    .class_init = smbus_storage_class_initfn
+};
+
+static void smbus_storage_register_devices(void)
+{
+    type_register_static(&smbus_storage_info);
+}
+
```
mborgerson, line 223: Please add `VMStateDescription` structure so snapshots can persist EEPROM state

hw/xbox/smbus_storage.c
```diff
@@ -0,0 +1,213 @@
...
+            error_setg(errp, "%s: file '%s' size of %d, expected %d\n",
+                __func__, s->file, size, s->size);
+            return;
+        }
+
+        int fd = open(s->file, O_RDONLY | O_BINARY);
+        if (fd < 0) {
+            error_setg(errp, "%s: file '%s' could not be opened\n",
+                __func__, s->file);
+            close(fd);
```
mborgerson: Don't need to `close` in this case

hw/xbox/smbus_storage.c
mborgerson: It'll be around for the lifetime of the device. If you wanted to you could implement an `unrealize` to clean up, but it's not a huge issue since we expect these devices to always be here

hw/xbox/smbus_storage.c
mborgerson: No worries, can leave ordering as is. Let's make `addr` a `uint8_t` though since addresses in SMBus/I2C are 7 bits (or was there a reason you chose 16b here?)

hw/xbox/smbus_storage.c
```diff
@@ -0,0 +1,213 @@
...
+    if (changed && s->file && s->persist) {
+        int fd = open(s->file, O_WRONLY | O_BINARY);
+        if (fd < 0) {
+            DPRINTF("%s: file '%s' could not be opened\n", __func__, s->file);
+            close(fd);
+            return -1;
+        }
+
+        int wc = write(fd, s->data, s->size);
+        if (wc != s->size) {
```
mborgerson, line 156: `write` can be successful and have written fewer than desired number of bytes (manpage explains this for more info). This is fine for now, but in the future we can make this more robust

hw/xbox/smbus_storage.c
```diff
@@ -0,0 +1,213 @@
...
+        int fd = open(s->file, O_RDONLY | O_BINARY);
+        if (fd < 0) {
+            error_setg(errp, "%s: file '%s' could not be opened\n",
+                __func__, s->file);
+            close(fd);
+            return;
+        }
+
+        int rc = read(fd, s->data, s->size);
+        if (rc != s->size) {
```
mborgerson, line 115: `read` can be successful and have read fewer than desired number of bytes (manpage explains this for more info). This is fine for now, but in the future we can make this more robust

hw/xbox/smbus_storage.c
Ernegien: Unrealize isn't called unless a hotplug event occurs iirc, so I assume this one just stays until app exit by default. I assume there's not much else we can do about this unfortunately.

hw/xbox/smbus_storage.c
Ernegien: I was somehow anchoring off of IO ports being 16-bit, but not SMBus ones; will switch to 8-bit.

hw/xbox/smbus_storage.c
Ernegien: This was desired; any write failure (even partial) should in my opinion be deemed a critical error. The question is whether you want to report differently on partial write failure vs. a complete one.

hw/xbox/smbus_storage.c
Ernegien: This was desired; any read failure (even partial) should in my opinion be deemed a critical error. The question is whether you want to report differently on partial read failure vs. a complete one.

hw/xbox/smbus_storage.c
Ernegien: Not sure what additional changes might be required but it seems good enough?

hw/xbox/smbus_storage.c
```diff
@@ -0,0 +1,228 @@
...
+
+    uint8_t val = s->data[s->offset];
+    DPRINTF("%s: addr=0x%02x off=0x%02x val=0x%02x\n",
+        __func__, s->addr, s->offset, val);
+    s->offset = (s->offset + 1) % s->size;
+
+    return val;
+}
+
+static const VMStateDescription vmstate_smbus_storage = {
```
Ernegien, line 179: Not sure what additional changes might be required but it seems good enough?

hw/xbox/smbus_storage.c
Ernegien: Perhaps replace "failure" with "issue"?

hw/xbox/smbus_storage.c
Ernegien: Perhaps replace "failure" with "issue"?

PR #48, opened 2020-10-30
nv2a: Software decompression for compressed 3D textures

### LINKED ISSUE #40: NV_texture_compression_vtc is assumed to be present
When uploading compressed 3D textures, the [`NV_texture_compression_vtc`](https://www.khronos.org/registry/OpenGL/extensions/NV/NV_texture_compression_vtc.txt) extension is mistakenly assumed to be present. If this extension is not present, `GL_INVALID_ENUM` will be generated and xemu will crash. A software fallback is required to handle the case when this extension is not available.

Note that this is not just a simple array of DXT1/S3TC textures: the texels of the volume slices are swizzled:
```
    Each VTC compression format is similar to a corresponding S3TC
    compression format, but where an S3TC block encodes a 4x4 block of
    texels, a VTC block encodes a 4x4x1, 4x4x2, or 4x4x4 (or 4x4x3 if
    non-power-of-two textures are supported) block of texels.  If the
    depth of the image is four or greater, 4x4x4 blocks are used, and
    if the depth is 1 or 2, 4x4x1 or 4x4x2 blocks are used.
```

---

Special thanks to @haxar for discovering this issue, and to @wilkovatch for identifying this extension and volunteering to fix the issue!

wilkovatch:
This should resolve #40


hw/xbox/nv2a/pgraph.c
```diff
@@ -5257,6 +5257,207 @@ static uint8_t* convert_texture_data(const TextureShape s,
     }
 }
 
+static inline void decode_bc1_colors(uint16_t c0,
```
mborgerson: Can we move these decompression functions into a new file, e.g. s3tc.{c|h}

hw/xbox/nv2a/pgraph.c
```diff
@@ -5257,6 +5257,207 @@ static uint8_t* convert_texture_data(const TextureShape s,
...
+                                           unsigned int height,
+                                           unsigned int depth)
+{
+    int block_depth = MIN(depth, 4);
+    int num_blocks_x = width/4,
+        num_blocks_y = height/4,
+        num_blocks_z = depth/block_depth;
+    uint8_t* converted_data = (uint8_t*)g_malloc(width * height * depth * 4);
+
+    if (color_format == GL_COMPRESSED_RGBA_S3TC_DXT1_EXT) {
```
mborgerson: Can we refactor this a bit into separate respective functions for each block decompression algorithm? We can then handle the block swizzling separately (and only once) from the block decompression logic, and reduce the amount of nested loops/repeated code here.

mborgerson (changes requested):

wilkovatch:
Is it better now? Anyway, this is what I meant yesterday on Discord when I was talking about code duplication/deduplication (in the end I was just being paranoid about the possible performance overhead, since in fact after refactoring all I added to the main loop are just a couple of ifs) EDIT: Hmm, it's failing on macOS, I'll have to check...

dracc:
Why closed? :(

wilkovatch:
I did something stupid and the build fails on macOS, I'll create a new pull request when I sort this out

PatrickvL:
Perhaps change the decoder interface to allow reuse of an externally allocated buffer, to reduce strain on memory allocater?

mborgerson:
@wilkovatch 
> Is it better now? Anyway, this is what I meant yesterday on Discord when I was talking about code duplication/deduplication (in the end I was just being paranoid about the possible performance overhead

Much better! Thank you for the changes. I will try to review it as soon as possible.

> I did something stupid and the build fails on macOS, I'll create a new pull request when I sort this out

No need to create a new pull request. You can rebase+force-push to your original branch it will update the PR accordingly.

@PatrickvL
> Perhaps change the decoder interface to allow reuse of an externally allocated buffer, to reduce strain on memory allocater?

If necessary the code can be changed to allocate a scratch buffer upon startup, but I'd prefer to keep it simple like this for now.

wilkovatch:
Ok, I'll try to update this one...

hw/xbox/nv2a/s3tc.h
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
```
mborgerson: Please include conventional C pre-processor header guards, e.g.:

```c
#ifndef S3TC_H
#define S3TC_H

...

#endif
```

hw/xbox/nv2a/s3tc.c
```diff
@@ -0,0 +1,199 @@
...
+    const uint8_t* block = data + 16 * (block_index * block_depth + slice);
+    *separate_alpha = true;
+
+    uint16_t c0 = *(uint16_t*)(block + 8),
+             c1 = *(uint16_t*)(block + 10);
+
+    decode_bc1_colors(c0, c1, r, g, b, a, false);
+
+    uint64_t alpha = *(uint64_t*)block;
+    uint8_t a0 = *block;
```
mborgerson: nit: Could be written `block[0]` and `block[1]` respectively

hw/xbox/nv2a/s3tc.c
```diff
@@ -0,0 +1,199 @@
...
+        a_palette[7] = (1*a0+6*a1)/7;
+    } else {
+        a_palette[2] = (4*a0+1*a1)/5;
+        a_palette[3] = (3*a0+2*a1)/5;
+        a_palette[4] = (2*a0+3*a1)/5;
+        a_palette[5] = (1*a0+4*a1)/5;
+        a_palette[6] = 0;
+        a_palette[7] = 255;
+    }
+    for(int a_i = 0; a_i < 16; a_i++){
```
mborgerson: nit: Space following `for` keyword

hw/xbox/nv2a/s3tc.c
```diff
@@ -0,0 +1,199 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#include "qemu/osdep.h"
+#include "s3tc.h"
+
+static inline void decode_bc1_colors(uint16_t c0,
+                                     uint16_t c1,
+                                     uint8_t* r,
```
mborgerson: nit: In general the code base follows a convention of having the `*` character next to the variable name, e.g. `uint8_t *r`. In this case however, as we know the dimensions of the array, let's go with `uint8_t r[4]` etc. Cases repeated below for other functions.

hw/xbox/nv2a/s3tc.c
```diff
@@ -0,0 +1,199 @@
...
+                                     uint8_t* g,
+                                     uint8_t* b,
+                                     uint8_t* a,
+                                     bool transparent)
+{
+    r[0] = ((c0 & 0xF800) >> 8) * 0xFF / 0xF8,
```
mborgerson: Looks great here, I like how you broke this color decoder out.

hw/xbox/nv2a/s3tc.c
```diff
@@ -0,0 +1,199 @@
...
+    }
+    *indices = *(uint32_t*)(block + 12);
+}
+
+uint8_t* decompress_3d_texture_data(GLint color_format,
+                                    const uint8_t *data,
+                                    unsigned int width,
+                                    unsigned int height,
+                                    unsigned int depth)
+{
```
mborgerson: To be safe, can we include some checks on the input here to safeguard against crazy situations like we saw previously; e.g.

```c
assert((width > 0) && (width % 4 == 0));
assert((height > 0) && (height % 4 == 0));
assert(depth > 0);
...
```

In case of Xbox, these dimensions must also be power-of-two, but we don't need such a restriction here if your decoder can handle NPOT all the same.

mborgerson:
Hi @wilkovatch, I'm not quite finished reviewing yet (still need to look over the block swizzling) but I've left some initial feedback for you. So far things are looking great with only minor fixes requested. Thanks for making updates so far!

hw/xbox/nv2a/s3tc.c
```diff
@@ -0,0 +1,199 @@
...
+        a[2] = 255;
+
+        r[3] = (r[0]+2*r[1])/3;
+        g[3] = (g[0]+2*g[1])/3;
+        b[3] = (b[0]+2*b[1])/3;
+        a[3] = 255;
+    }
+}
+
+static inline void decompress_dxt1_block(const uint8_t* data,
```
mborgerson: Let's simplify this function by pre-calculating the `data` pointer and remove the `block_index`, `block_depth`, and `slice` arguments here. Further, let us change the `const uint8_t *data` to the known block size here, like `const uint8_t block_data[8]`, and add an output pointer/stride parameter for the output pixel data block (see comment below). Likewise for DXT3/5 functions below.

hw/xbox/nv2a/s3tc.c
```diff
@@ -0,0 +1,199 @@
...
+                    }
+
+                    int x0 = i * 4,
+                        y0 = j * 4,
+                        z_pos_factor = (k * block_depth + slice) * width * height;
+
+                    int x1 = x0 + 4,
+                        y1 = y0 + 4;
+
+                    for (int y = y0; y < y1; y++) {
```
mborgerson: Would you refactor this segment into a separate function which is then called from the corresponding decompress_dxtX_block s.t. when we decompress a block it actually populates a pixel array instead of just producing the color palette/indices.

wilkovatch:
@mborgerson Fixes done, let me know if I forgot anything.

mborgerson:
Thank you @wilkovatch! There is some minor refactoring I'd like to see done, but this is ready to merge now and we can iterate on it in the future.

I've tested your code with [some simple scripts here](https://github.com/mborgerson/test-xemu-s3tc-volume-compress), and I've tested my compressor against hardware so everything appears to check out. I've also made sure this works with the compressed 3D texture Conker uses in its main menu and it looks to work just fine.

Thanks for the work!

mborgerson (approved):

mborgerson:
It looks like I force-pushed upstream master to your branch and accidentially wiped out your commit in the process of rebasing. Sorry about that. However, your work is safe and has been integrated into xemu's master branch.

PR #64, opened 2020-11-04
ui: Add EEPROM generation

Ernegien:
This adds automatic EEPROM generation in the case of an unspecified or invalid file within the settings. Currently it signs based on a version 1.0 Xbox but can be expanded upon for other versions if needed. Most importantly, this will prevent conflicts caused by matching MAC addresses if networking with others as well as offer persistence now by default, although that can be made configurable via future UI changes with the `persist=off` argument.


hw/xbox/eeprom_generation.c
```diff
@@ -0,0 +1,262 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "eeprom_generation.h"
+
+void xbox_eeprom_crc(void *crc, uint8_t *data, uint32_t len) {
+	uint32_t high = 0;
```
mborgerson, line 23: Nit: spacing

hw/xbox/eeprom_generation.c
```diff
@@ -0,0 +1,262 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "eeprom_generation.h"
+
+void xbox_eeprom_crc(void *crc, uint8_t *data, uint32_t len) {
```
mborgerson: * What do you think of changing this function to return the CRC value instead of using this output pointer void*/cast thing inherited from XKUtils design?
* We should probably use conventional `size_t` type for `len` argument
* Can you make any function not used outside of this file `static` to let the compiler inline the functions

hw/xbox/eeprom_generation.c
```diff
@@ -0,0 +1,262 @@
...
+	*(uint32_t*)crc = ~(high + low);
+}
+
+void xbox_rc4_init(RC4Context *ctx, uint8_t *data, uint32_t len) {
+    ctx->x = 0;
+    ctx->y = 0;
+    uint32_t i1 = 0;
+    uint32_t i2 = 0;
+
+    for (uint32_t i = 0; i < XBOX_RC4_KEY_LENGTH; i++) {
```
mborgerson: * nit: For readability, let's just use `int`s for these iterators
* RC4 KSA specifies this array should be 256 bytes in size, but the key length can vary. Let's rename this macro to something else because it's not necessarily the length of the key

hw/xbox/eeprom_generation.c
```diff
@@ -0,0 +1,262 @@
...
+		low += val;
+	}
+
+	*(uint32_t*)crc = ~(high + low);
+}
+
+void xbox_rc4_init(RC4Context *ctx, uint8_t *data, uint32_t len) {
+    ctx->x = 0;
+    ctx->y = 0;
+    uint32_t i1 = 0;
```
mborgerson: * nit: Can we rename `i1`, `i2` to `i`, `j` respectively to align with common algorithm listings
* nit: Let's just make these `int`s unless you think they need to be unsigned 32b values
* Let's move these defs down by where they are used

hw/xbox/eeprom_generation.c
```diff
@@ -0,0 +1,262 @@
...
+
+		high = (uint32_t)((sum + val) >> 32);
+		low += val;
+	}
+
+	*(uint32_t*)crc = ~(high + low);
+}
+
+void xbox_rc4_init(RC4Context *ctx, uint8_t *data, uint32_t len) {
+    ctx->x = 0;
```
mborgerson: Let's rename `x`,`y` vars to `prga_i`,`prga_j` or something to match [common RC4 algorithm listings](https://en.wikipedia.org/wiki/RC4#Key-scheduling_algorithm_(KSA))

hw/xbox/eeprom_generation.c
```diff
@@ -0,0 +1,262 @@
...
+    ctx->y = 0;
+    uint32_t i1 = 0;
+    uint32_t i2 = 0;
+
+    for (uint32_t i = 0; i < XBOX_RC4_KEY_LENGTH; i++) {
+        ctx->key[i] = (uint8_t)(i % XBOX_RC4_KEY_LENGTH);
```
mborgerson: We should rename this `key` field, which is a bit misleading. It is not the key itself, but the keystream array. In the common algorithm listings it's called `S`

hw/xbox/eeprom_generation.c
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    for (uint32_t i = 0; i < XBOX_RC4_KEY_LENGTH; i++) {
+        i2 = (data[i1] + ctx->key[i] + i2) % XBOX_RC4_KEY_LENGTH;
```
mborgerson: This `data` field is actually the RC4 key and `ctx->key` is `S` array. We should rename these fields to match common algorithm listings per above

hw/xbox/eeprom_generation.c
```diff
@@ -0,0 +1,262 @@
...
+    uint8_t *data, uint32_t len, uint8_t *hash) {
+    xbox_sha1_reset(ctx, ver, true);
+    xbox_sha1_input(ctx, data, len);
+    xbox_sha1_result(ctx, ctx->msg_blk);
+    xbox_sha1_reset(ctx, ver, false);
+    xbox_sha1_input(ctx, ctx->msg_blk, XBOX_SHA1_HASH_LENGTH);
+    xbox_sha1_result(ctx, hash);
+}
+
+bool xbox_eeprom_generate(const char* file, XboxEEPROMVersion ver) {
```
mborgerson: nit: `*` should be on the var side

hw/xbox/eeprom_generation.c
```diff
@@ -0,0 +1,262 @@
...
+    XboxEEPROM e;
+    memset(&e, 0, sizeof(e));
+
+    // set default North American and NTSC-M region settings
+    e.region = cpu_to_le32(1);
+    e.video_standard = cpu_to_le32(0x00400100);
+
+    // TODO: set default video flags in user section for 480p and 720p?
```
mborgerson: We should probably just default to what an Xbox from the factory would do

hw/xbox/eeprom_generation.c
```diff
@@ -0,0 +1,262 @@
...
+    RC4Context rctx;
+    SHA1Context sctx;
+    uint8_t seed[20];
+    xbox_sha1_compute(&sctx, ver, e.confounder, 0x1C, e.hash);
+    xbox_sha1_compute(&sctx, ver, e.hash, sizeof(e.hash), seed);
+    xbox_rc4_init(&rctx, seed, sizeof(seed));
+    xbox_rc4_crypt(&rctx, e.confounder, 0x1C);
+
+    // save to file
+    int fd = open(file, O_CREAT | O_RDWR | O_BINARY,
```
mborgerson: Let's simplify this and just use `fopen`

hw/xbox/eeprom_generation.c
```diff
@@ -0,0 +1,262 @@
...
+        S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH);
+    if (fd < 0) {
+        return false;
+    }
+    if (write(fd, &e, sizeof(e)) != sizeof(e)) {
```
mborgerson: Can simplify this to something like:

```c
bool success = fwrite(&e, sizeof(e), 1, fd) == 1;
fclose(fd);
return success;
```

hw/xbox/eeprom_generation.c
```diff
@@ -0,0 +1,262 @@
...
+    qcrypto_random_bytes(e.online_key, sizeof(e.online_key), &error_fatal);
+    qcrypto_random_bytes(e.mac, sizeof(e.mac), &error_fatal);
+    qcrypto_random_bytes(e.serial, sizeof(e.serial), &error_fatal);
+    memcpy(e.mac, "\x00\x50\xF2", 3);
+    for (int i=0; i<sizeof(e.serial); i++) {
+        e.serial[i] = '0' + (e.serial[i] % 10);
+    }
+    
+    // update checksums
+    xbox_eeprom_crc(&e.checksum, e.serial, 0x2C);
```
mborgerson: * See note above about possibly changing this function to return the CRC
* Potential endianness hazard

hw/xbox/eeprom_generation.c
```diff
@@ -0,0 +1,262 @@
...
+
+    // TODO: set default video flags in user section for 480p and 720p?
+
+    // randomize hardware information
+    qcrypto_random_bytes(e.confounder, sizeof(e.confounder), &error_fatal);
+    qcrypto_random_bytes(e.hdd_key, sizeof(e.hdd_key), &error_fatal);
+    qcrypto_random_bytes(e.online_key, sizeof(e.online_key), &error_fatal);
+    qcrypto_random_bytes(e.mac, sizeof(e.mac), &error_fatal);
+    qcrypto_random_bytes(e.serial, sizeof(e.serial), &error_fatal);
+    memcpy(e.mac, "\x00\x50\xF2", 3);
```
mborgerson, line 217: * nit: Move one line up, keep MAC init together.
* Please make sure [we set/clear any necessary bits](https://en.wikipedia.org/wiki/MAC_address)

hw/xbox/eeprom_generation.h
```diff
@@ -0,0 +1,93 @@
...
+    uint32_t padding2;
+    uint32_t user_checksum;
+    uint8_t user_section[156];
+} XboxEEPROM;
+
+typedef enum {
+    // debug kernels
+    XBOX_EEPROM_DEBUG,
+    // v1.0 kernels
+    XBOX_EEPROM_FIRST,
```
mborgerson: Can we rename these to something more explicit like `XBOX_EEPROM_VERSION_1`, `_2`, `_3`

hw/xbox/eeprom_generation.h
```diff
@@ -0,0 +1,93 @@
...
+    // v1.1-1.4 kernels
+    XBOX_EEPROM_MIDDLE,
+    // v1.6 kernels
+    XBOX_EEPROM_LAST
+} XboxEEPROMVersion;
+
+void xbox_eeprom_crc(void *crc, uint8_t *data, uint32_t len);
```
mborgerson: See comment above about possibly making some of these functions `static` if they are not necessary outside of the respective source file

softmmu/vl.c
```diff
@@ -45,6 +45,7 @@
 #include "hw/isa/isa.h"
 #include "hw/scsi/scsi.h"
 #include "hw/display/vga.h"
+#include "hw/xbox/eeprom_generation.h"  // TODO: rename?
```
mborgerson: Rename it to what?

ui/xemu-settings.c
```diff
@@ -216,6 +216,31 @@ const char *xemu_settings_get_path(void)
 	return settings_path;
 }
 
+const char *xemu_settings_get_default_eeprom_path(void)
+{
+	char *base = SDL_GetPrefPath("xemu", "xemu");
+	assert(base != NULL);
+	size_t base_len = strlen(base);
+
+	const char* name = "eeprom.bin\0";
```
mborgerson: * Don't need to specify `\0` here, it'll be included already
* nit: `*` goes next to var

hw/xbox/eeprom_generation.c
Ernegien: I force a common version 1.0 Xbox prefix of `00:50:F2` with only the highest octet at play with status flags so I think we're covered here.

softmmu/vl.c
```diff
@@ -118,6 +118,7 @@
 #include "ui/xemu-notifications.h"
 #include "ui/xemu-net.h"
 #include "ui/xemu-input.h"
+#include "hw/xbox/eeprom_generation.h"
```
Ernegien, line 121: I'm still questioning where this might better fit in. I suppose it's more on the `ui` side than `hw` which is what I've been prefixing my commits as?

hw/xbox/eeprom_generation.c
```diff
@@ -0,0 +1,242 @@
...
+ * License along with this library; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "eeprom_generation.h"
+
+static uint32_t xbox_eeprom_crc(uint8_t *data, size_t len) {
+	uint32_t high = 0;
+    uint32_t low = 0;
+	for (int i = 0; i < len / 4; i++) {
+		uint32_t val = ((uint32_t*)data)[i];
```
mborgerson: Possible endianness issue here. If we are calculating CRC over EEPROM contents, when we lift this 32b value out and calculate the sum it will be wrong. Consider using `ldl_le_p`

hw/xbox/eeprom_generation.c
```diff
@@ -0,0 +1,242 @@
...
+    ctx->intermediate[2] += c;
+    ctx->intermediate[3] += d;
+    ctx->intermediate[4] += e;
+    ctx->msg_blk_index = 0;
+}
+
+static void xbox_sha1_input(SHA1Context *ctx, uint8_t *data, size_t len) {
+    ctx->length += len << 3;
+    for (int i = 0; i < len; i++) {
+        ctx->msg_blk[ctx->msg_blk_index++] = data[i];
```
mborgerson, line 161: ~~Is it possible to overflow `msg_blk` here?~~ Nevermind, I see that `xbox_sha1_process` resets the counter every 64 bytes.

hw/xbox/eeprom_generation.c
```diff
@@ -0,0 +1,242 @@
...
+    qcrypto_random_bytes(e.online_key, sizeof(e.online_key), &error_fatal);
+    memcpy(e.mac, "\x00\x50\xF2", 3);
+    qcrypto_random_bytes(e.mac + 3, sizeof(e.mac) - 3, &error_fatal);
+    qcrypto_random_bytes(e.serial, sizeof(e.serial), &error_fatal);
+    for (int i = 0; i < sizeof(e.serial); i++) {
+        e.serial[i] = '0' + (e.serial[i] % 10);
+    }
+    
+    // update checksums
+    e.checksum = xbox_eeprom_crc(e.serial, 0x2C);
```
mborgerson: Possible endianness issue here

softmmu/vl.c
mborgerson: Well it's not hardware, but it's related to the EEPROM. You can leave it here for now, it's not a huge issue

hw/xbox/eeprom_generation.c
```diff
@@ -0,0 +1,242 @@
...
+    xbox_sha1_reset(ctx, ver, false);
+    xbox_sha1_input(ctx, ctx->msg_blk, XBOX_SHA1_HASH_LENGTH);
+    xbox_sha1_result(ctx, hash);
+}
+
+bool xbox_eeprom_generate(const char *file, XboxEEPROMVersion ver) {
+    XboxEEPROM e;
+    memset(&e, 0, sizeof(e));
+
+    // set default North American and NTSC-M region settings
```
mborgerson, line 209: Is this really the only init we need to do (and dashboard handles the rest of the fields on its OOBE startup?)

Does this also work with dummy dashboard? I seem to recall getting a hang with dummy dash when it fails to init EEPROM like MS dash does

hw/xbox/eeprom_generation.h
```diff
@@ -0,0 +1,74 @@
...
+
+typedef struct SHA1Context {
+    uint32_t intermediate[XBOX_SHA1_HASH_LENGTH / sizeof(uint32_t)];
+    uint8_t msg_blk[64];
+    uint32_t msg_blk_index;
+    uint32_t length;
+    bool computed;
+} SHA1Context;
+
+#pragma pack(1)
```
mborgerson: Let's restore the packing settings after this struct.

Like this:
```c
#pragma pack(push,1)
typedef struct XboxEEPROM {
...
} XboxEEPROM;
#pragma pack(pop)
```

hw/xbox/eeprom_generation.c
```diff
@@ -0,0 +1,242 @@
...
+	return ~(high + low);
+}
+
+static void xbox_rc4_swap(RC4Context *ctx, uint32_t first, uint32_t second) {
+    uint8_t temp = ctx->s[first];
+    ctx->s[first] = ctx->s[second];
+    ctx->s[second] = temp;
+}
+
+static void xbox_rc4_init(RC4Context *ctx, uint8_t *data, size_t len) {
```
mborgerson, line 41: This looks much better, ty for the refactor

hw/xbox/eeprom_generation.c
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    for (int i = 16; i < 80; i++) {
+        w[i] = xbox_sha1_rotate(1, w[i - 3] ^ w[i - 8] ^ w[i - 14] ^ w[i - 16]);
+    }
+
+    for (int i = 0; i < 80; i++) {
+        uint32_t temp = xbox_sha1_rotate(5, a);
+        switch (i / 20) {
+        case 0: temp += ((b & c) | ((~b) & d)) + e + w[i] + k[0]; break;
```
mborgerson: nit: You could move the `e + w[i]` term up to the `temp` definition for all of these cases

hw/xbox/eeprom_generation.c
```diff
@@ -0,0 +1,242 @@
...
+        while (ctx->msg_blk_index < 64) {
+            ctx->msg_blk[ctx->msg_blk_index++] = 0;
+        }
+        xbox_sha1_process(ctx);
+    }
+    while (ctx->msg_blk_index < 56) {
+        ctx->msg_blk[ctx->msg_blk_index++] = 0;
+    }
+    *(uint32_t*)&ctx->msg_blk[56] = 0;
+    *(uint32_t*)&ctx->msg_blk[60] = cpu_to_be32(ctx->length);
```
mborgerson, line 180: How/where can I verify this padding function is correct?

hw/xbox/eeprom_generation.c
Ernegien: Yup, my testbed was flawed, thanks for catching this.

hw/xbox/eeprom_generation.c
Ernegien: Yup, my testbed was flawed, thanks for catching this.

hw/xbox/eeprom_generation.c
Ernegien: This produces the same EEPROM a stock Xbox would have where the user section is blank. Whether or not all hacked dashboards handle this is questionable however, but I believe they should.

hw/xbox/eeprom_generation.c
Ernegien: It stems from the [XKUtils stuff](https://github.com/mborgerson/xbeeprom/blob/master/XKSHA1.cpp#L288) except I've refactored out the high 32-bits of the length field to simplify things since we aren't working with 4GB+ EEPROMs ;) You can also use [this](https://gist.github.com/Ernegien/5874af40c89eae6034c5e4f97148fc65) for quickly verifying endianness and then use [my tool](https://github.com/Ernegien/XboxEepromEditor) to open the generated EEPROMs to verify that they've been encrypted and can be decrypted properly.

mborgerson: I meant where in the SHA1 spec can I find this or is it Xbox specific. I found my answer in [RFC3174](https://tools.ietf.org/html/rfc3174#page-4):

```
4. Message Padding

   SHA-1 is used to compute a message digest for a message or data file
   that is provided as input.  The message or data file should be
   considered to be a bit string.  The length of the message is the
   number of bits in the message (the empty message has length 0).  If
   the number of bits in a message is a multiple of 8, for compactness
   we can represent the message in hex.  The purpose of message padding
   is to make the total length of a padded message a multiple of 512.
   SHA-1 sequentially processes blocks of 512 bits when computing the
   message digest.  The following specifies how this padding shall be
   performed.  As a summary, a "1" followed by m "0"s followed by a 64-
   bit integer are appended to the end of the message to produce a
   padded message of length 512 * n.  The 64-bit integer is the length
   of the original message.  The padded message is then processed by the
   SHA-1 as n 512-bit blocks.
```

mborgerson:
I have rebased, squashed the commits on this branch, amended the commit message, and fixed a small bug in `xbox_eeprom_generate` to check `fd` before calling `fwrite`. I've also confirmed there are no issues with the "dummy dash," and with retail dash it will correctly display the OOBE setup.

This is ready to merge now. There are some edge cases where we might abruptly exit, for example if the EEPROM file is made write-only the device will create successfully but fail to read and cause exit. We can make handling of these types of cases more robust in the future.

Thanks for the work!

PR #68, opened 2020-11-06
Provide license with distribution

dracc:
Closes https://github.com/mborgerson/xemu/projects/8#card-39811457

Untested, thus Draft status.


mborgerson:
Thanks for the PR. The LICENSE file you've referenced is for QEMU. I'll need to update it for xemu specifically as the provided license can be confusing. I'll close this for now and we can address it soon.

PR #69, opened 2020-11-06
nv2a: fix DS/DT texture offset shader operation

wilkovatch:
This fixes the DS/DT texture offset shader operations (BUMPENVMAP and BUMPENVMAP_LUM, for non-linear and linear textures).
There were a few mistakes before:
- DS/DT textures use the green and blue channels on hardware (plus the red channel for scale in BUMPENVMAP_LUM), in the code the red and green channels were used instead (plus blue for the scale).
- The matrix was used incorrectly. The swizzling was correct, but the offset vector had to be pre-multiplied by it, not post-multiplied (post-multiplying is the same as pre-multiplying by the transpose).
- On hardware, if the wrong format is used, the ds/dt values are still interpreted as signed, even if the texture is unsigned (this on hardware causes the boost dash effect in Jet Set Radio Future to look differently than intended).
- In the case of linear textures nothing special happens on hardware, so the asserts were not needed.

Taking into account those things results in [this sample](https://github.com/Voxel9/xbox-xgu-examples/tree/master/bumpenv_ripple) (see the screenshots below) and effects in other games working exactly as on hardware.

Removing the unneeded asserts also fixes crashes in Jet Set Radio Future (49470018), Halo 2 (4d530064), Otogi (46530002) and possibly other titles.

Big thanks to @Voxel9 for his sample, playing with it allowed me to figure out what was going on on hardware.

The aforementioned sample on hardware:
bumpenv_ripple

The same sample on xemu with this PR:
xemu_bumpenv_ripple

The boost dash effect in Jet Set Radio Future with this PR (same as on hardware, the rays look sharp instead of smooth like the ds/dt texture because the unsigned ds/dt texture is interpreted as signed):
jsrf

### DIFF
--- a/hw/xbox/nv2a/pgraph.c
+++ b/hw/xbox/nv2a/pgraph.c
@@ -3312,6 +3312,10 @@ static void pgraph_bind_shaders(PGRAPHState *pg)
             GET_MASK(pg->regs[NV_PGRAPH_TEXFMT0 + i*4],
                      NV_PGRAPH_TEXFMT0_COLOR);
 
+        /* This is needed to detect cases where the wrong format for ds/dt textures is used
+         * and a remapping of the unsigned values to signed is needed (e.g. the boost dash in JSRF) */
+        state.psh.dsdt_tex[i] = kelvin_color_format_map[color_format].gl_type == GL_BYTE;
+
         if (enabled && kelvin_color_format_map[color_format].linear) {
             state.psh.rect_tex[i] = true;
         }
--- a/hw/xbox/nv2a/psh.c
+++ b/hw/xbox/nv2a/psh.c
@@ -685,27 +685,55 @@ static QString* psh_convert(struct PixelShader *ps)
         }
         case PS_TEXTUREMODES_BUMPENVMAP:
             assert(i >= 1);
-            assert(!ps->state.rect_tex[i]);
-            sampler_type = "sampler2D";
+
+            if (!ps->state.dsdt_tex[ps->input_tex[i]]) {
+                /* Fix for cases using improper formats for the ds/dt offset texture,
+                 * like the boost dash effect in JSRF */
+                /* FIXME: Figure out why the green and blue channels arrive reversed */
+                qstring_append_fmt(vars, "t%d.bg = dotmap_minus1_to_1_gl(t%d.rgb).gb;\n",
+                                   ps->input_tex[i], ps->input_tex[i]);
+            }
+
+            sampler_type = ps->state.rect_tex[i] ? "sampler2DRect" : "sampler2D";
             qstring_append_fmt(preflight, "uniform mat2 bumpMat%d;\n", i);
+
             /* FIXME: Do bumpMat swizzle on CPU before upload */
-            qstring_append_fmt(vars, "vec4 t%d = texture(texSamp%d, pT%d.xy + t%d.rg * mat2(bumpMat%d[0].xy,bumpMat%d[1].yx));\n",
-                               i, i, i, ps->input_tex[i], i, i);
+            qstring_append_fmt(vars, "t%d.gb = mat2(bumpMat%d[0].xy, bumpMat%d[1].yx) * t%d.gb;\n",
+                               ps->input_tex[i], i, i, ps->input_tex[i]);
+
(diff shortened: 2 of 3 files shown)


hw/xbox/nv2a/pgraph.c
```diff
@@ -3312,6 +3312,13 @@ static void pgraph_bind_shaders(PGRAPHState *pg)
             GET_MASK(pg->regs[NV_PGRAPH_TEXFMT0 + i*4],
                      NV_PGRAPH_TEXFMT0_COLOR);
 
+        state.psh.signed_tex[i] = kelvin_color_format_map[color_format].gl_type == GL_BYTE ||
+                                  kelvin_color_format_map[color_format].gl_type == GL_FLOAT;
+
```
wilkovatch, line 3317: Here I checked in the map at the beginning of the file which output formats were signed and which with only two channels, so I might have missed something.

mborgerson:
Hi @wilkovatch, I have not forgotten about this. I will be reviewing it shortly, have been busy but would like to confirm some things with the test case you shared previously. Thanks for another contribution!

hw/xbox/nv2a/psh.c
```diff
@@ -685,27 +685,55 @@ static QString* psh_convert(struct PixelShader *ps)
         }
         case PS_TEXTUREMODES_BUMPENVMAP:
             assert(i >= 1);
-            assert(!ps->state.rect_tex[i]);
-            sampler_type = "sampler2D";
+
+            if (!ps->state.dsdt_tex[ps->input_tex[i]]) {
+                /* Fix for cases using improper formats for the ds/dt offset texture,
+                 * like the boost dash effect in JSRF */
+                /* FIXME: Figure out why the green and blue channels arrive reversed */
```
mborgerson, line 692: Turns out the channels were mistakenly flipped in our texture channel swizzling map

hw/xbox/nv2a/psh.c
```diff
@@ -685,27 +685,55 @@ static QString* psh_convert(struct PixelShader *ps)
...
+                                   ps->input_tex[i], ps->input_tex[i]);
+            }
+
+            sampler_type = ps->state.rect_tex[i] ? "sampler2DRect" : "sampler2D";
             qstring_append_fmt(preflight, "uniform mat2 bumpMat%d;\n", i);
+
             /* FIXME: Do bumpMat swizzle on CPU before upload */
-            qstring_append_fmt(vars, "vec4 t%d = texture(texSamp%d, pT%d.xy + t%d.rg * mat2(bumpMat%d[0].xy,bumpMat%d[1].yx));\n",
-                               i, i, i, ps->input_tex[i], i, i);
+            qstring_append_fmt(vars, "t%d.gb = mat2(bumpMat%d[0].xy, bumpMat%d[1].yx) * t%d.gb;\n",
```
mborgerson, line 701: Good finds on the multiplication order being swapped and the wrong channels being used!

Prefer to not overwrite the input texture color channels, I've opted for a temp variable

hw/xbox/nv2a/psh.c
```diff
@@ -685,27 +685,55 @@ static QString* psh_convert(struct PixelShader *ps)
...
-                               i, i, i, ps->input_tex[i], i);
+
+            if (!ps->state.dsdt_tex[ps->input_tex[i]]) {
+                /* Fix for cases using improper texture formats for the ds/dt offset texture */
+                /* FIXME: Figure out why the channels arrive reversed */
+                qstring_append_fmt(vars, "t%d.gbr = vec3(dotmap_minus1_to_1_gl(t%d.abg).gb, t%d.r);\n",
+                                   ps->input_tex[i], ps->input_tex[i], ps->input_tex[i]);
+            } else {
+                /* FIXME: Figure out why the green and blue channels arrive reversed */
+                qstring_append_fmt(vars, "t%d.rgb = vec3(t%d.rbg);\n",
```
mborgerson, line 720: See above for note about swapped channels. Also we want the luminance value to be unsigned

mborgerson (commented):
Nice job finding and fixing the bugs here! I've edited your commit message a bit and have done some additional bug fixing and refactoring on top of this after merge instead of going back and forth on the review. I've left comments in review about the important stuff.

Also thank you (and @Voxel9) for providing a test case for this, it was helpful in doing some additional research. I actually found some minor bugs in nxdk/tests along the way.

PR #72, opened 2020-11-12
hw: Add cached flash write workaround

### LINKED ISSUE #71: MCPX 1.1 kernels fail to boot when treating flash memory like RAM
For MCPX 1.1-based kernels, early bootstrap code will attempt to write-then-read a value from memory-mapped flash. Based on a trace of memory accesses during boot, there appears to be only one instance of this case which occurs 5 times during boot.

image

I assume this works on hardware due to CPU writeback caching policy (confirmed w/ check against MTRR config), which could permit the immediate store and read back from this read-only memory region. Because QEMU does not support emulating the behavior of these CPU caching policies, and because this memory is marked as ROM, the value is not updated in xemu memory and these kernels will fail to boot. I also assume that this is a bug in MS code.

In testing, by permitting writes to flash memory (that is, not marking it as read-only), these later kernel revisions with corresponding MCPX 1.1 boot ROM and EEPROM will boot as expected in xemu.

In absence of accurate cache policy emulation, marking this memory region as read/write may be best immediate course of action, as it unblocks these later kernels without apparent side effects.

Big thanks to @Ernegien for his efforts in discovering the issue!

Ernegien:
Quoting @mborgerson from [here](https://github.com/mborgerson/xemu/issues/71)

> For MCPX 1.1-based kernels, early bootstrap code will attempt to write-then-read a value from memory-mapped flash. Based on a trace of memory accesses during boot, there appears to be only one instance of this case which occurs 5 times during boot.
> 
> image
> 
> I assume this works on hardware due to CPU writeback caching policy (confirmed w/ check against MTRR config), which could permit the immediate store and read back from this read-only memory region. Because QEMU does not support emulating the behavior of these CPU caching policies, and because this memory is marked as ROM, the value is not updated in xemu memory and these kernels will fail to boot. I also assume that this is a bug in MS code.
> 
> In testing, by permitting writes to flash memory (that is, not marking it as read-only), these later kernel revisions with corresponding MCPX 1.1 boot ROM and EEPROM will boot as expected in xemu.
> 
> In absence of accurate cache policy emulation, marking this memory region as read/write may be best immediate course of action, as it unblocks these later kernels without apparent side effects.
> 
> Big thanks to @Ernegien for his efforts in discovering the issue!

Fixes #71

### DIFF
--- a/hw/xbox/xbox.c
+++ b/hw/xbox/xbox.c
@@ -143,12 +143,26 @@ static void xbox_flash_init(MemoryRegion *rom_memory)
         g_free(filename);
     }
 
+    /* XBOX_FIXME: The "memory_region_set_readonly" calls below have been
+    * temporarily commented out due to MCPX 1.1-based kernels hanging
+    * in the first bootloader stage when doing RSA signature verification.
+    * 
+    * This is caused by code incorrectly using the flash memory range to
+    * store the following computation; luckily real hardware's writeback
+    * cache policy (verified against MTRR config) appears to allow this
+    * to succeed, but qemu's emulation of such isn't capable of this yet
+    * so the value is never updated in ROM unless readonly is unspecified.
+    * 
+    *   sub ds:0FFFFD52Ch, eax
+    *   mov eax, ds:0FFFFD52Ch
+    */
+
     /* Create BIOS region */
     MemoryRegion *bios;
     bios = g_malloc(sizeof(*bios));
     assert(bios != NULL);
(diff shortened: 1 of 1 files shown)


mborgerson:
Fix and comment looks good. I had some nits about commit message, so I just fixed it instead of back-and-forth, hope you don't mind. These nits were: generally `xbox` prefix based on file history (we are not changing all `hw` so it makes sense to narrow prefix scope), avoid referencing GitHub in commit messages/code if you can avoid it, be more specific about the actual change in the header if possible).

Issue #80, opened 2020-11-17
Rumble not working on Xbox Controller Windows 10


theboy181:
Rumble is not working in Windows 10, and there has been discussions on discord that Xbox controllers are working in linux, and that this could be an odd issue with SDL and Windows.

PR #81, opened 2020-11-18
input: Fix SDL haptic effect length and magnitude

theboy181:
Stumbled into a a solution, guided by the experts.


Ernegien:
Fixes #80

mborgerson:
Thanks!

PR #82, opened 2020-11-20
ui: Set swap interval after creating GL context

haxar:
A call to `SDL_GL_SwapWindow()` will sleep when vertical retrace synchronization is enabled by default. We need to disable this since we already have finer grained control of when to sleep and swap again.

### DIFF
--- a/ui/xemu.c
+++ b/ui/xemu.c
@@ -803,7 +803,6 @@ static void sdl2_display_very_early_init(DisplayOptions *o)
         SDL_GL_CONTEXT_PROFILE_MASK,
         SDL_GL_CONTEXT_PROFILE_CORE);
     SDL_GL_SetAttribute(SDL_GL_DOUBLEBUFFER, 1);
-    SDL_GL_SetSwapInterval(0);
 
     // Create main window
     m_window = SDL_CreateWindow(
@@ -853,6 +852,7 @@ static void sdl2_display_early_init(DisplayOptions *o)
     display_opengl = 1;
 
     SDL_GL_MakeCurrent(m_window, m_context);
+    SDL_GL_SetSwapInterval(0);
     xemu_hud_init(m_window, m_context);
     blit = create_decal_shader(SHADER_TYPE_BLIT);
 }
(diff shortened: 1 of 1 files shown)


haxar:
Commenting out the throttle code at the end of `sdl2_gl_refresh()` without this patch will exhibit that `SDL_GL_SwapWindow()` does in fact sleep by default.

Adding `SDL_GL_SetSwapInterval(0)` will make swaps happen immediately.

haxar:
Looks like `sdl2_display_very_early_init()` already has `SDL_GL_SetSwapInterval(0)`, but when calling `SDL_GL_GetSwapInterval()` before `SDL_GL_SwapWindow()`, without this patch, it returns `1` indicating that driver vsync is on, which may create additional unnecessary delay in the thread.

haxar:
The call that already exists in early init actually throws an error:
```
1: SDL_GetError=''
2: SDL_GetError='No OpenGL context has been made current'
```

So, presumably, the call would have no effect if a context was not made current.

```diff
diff --git a/ui/xemu.c b/ui/xemu.c
index a191120e6d..a8f3f9517d 100644
--- a/ui/xemu.c
+++ b/ui/xemu.c
@@ -801,7 +801,9 @@ static void sdl2_display_very_early_init(DisplayOptions *o)
         SDL_GL_CONTEXT_PROFILE_MASK,
         SDL_GL_CONTEXT_PROFILE_CORE);
     SDL_GL_SetAttribute(SDL_GL_DOUBLEBUFFER, 1);
+fprintf(stderr,"1: SDL_GetError='%s'\n",SDL_GetError());
     SDL_GL_SetSwapInterval(0);
+fprintf(stderr,"2: SDL_GetError='%s'\n",SDL_GetError());
 
     // Create main window
     m_window = SDL_CreateWindow(
```
If this call needs to be moved so it is in a current context, let me know, so I can update the PR.

This could be implemented as a "disable driver vsync" UI option, but not sure if that would be necessary.

mborgerson:
Good find! Yes, apparently the call to `SDL_GL_SetSwapInterval` needs to happen *after* the context is initialized. Makes sense. Let's move the existing call down to just after the `SDL_GL_CreateContext`. We can make it configurable in the UI later--for now lets just go with this.

While you're at it, would you please change the commit message to something which reflects the updated change/intent. For example (of course, feel free to use your own words):

```
ui: Set swap interval after creating GL context
```

Thanks

PR #89, opened 2020-11-23
input: Set SDL ALLOW_BACKGROUND_EVENTS hint

theboy181:


theboy181:
Not a complete solution, but its a start. I would like to see the ability for the user to select it in the input menu.

ui/xemu-input.c
```diff
@@ -56,6 +56,9 @@ void xemu_input_init(void)
         exit(1);
     }
 
+    // set background input event, eventual coding for an option is needed.(todo)
+    SDL_SetHint(SDL_HINT_JOYSTICK_ALLOW_BACKGROUND_EVENTS, "1");
```
mborgerson: Lets move this line up to the start of this function, and drop the comment which doesn't really give any extra info

mborgerson:
Thanks

PR #91, opened 2020-12-02
nv2a: Support line and poly anti-aliasing

mborgerson:
A marked reduction in aliasing artifacts in MS dashboard, possibly elsewhere:

MS-0
MS-0-Smooth

Settings0
Settings1

PR #93, opened 2020-12-02
nv2a: Add Gaussian convolution filter for AA

mborgerson:
Improved texture filtering, used e.g. in AA for dashboard. Cleans up most (all?) of the remaining aliasing artifacts.

After:
Settings-Gaussian
Before:
Settings1


haxar:
Running xemu with this PR, with an Intel HD 3000 iGPU (Sandy Bridge) that does not support line & poly AA, that could have been fixed by PR #91.

After (just Gaussian convolution filter):
Screenshot at 2020-12-02 02-59-14
Before (with PR #91 line & poly AA as a no-operation):
Screenshot at 2020-12-02 03-02-58

PR #95, opened 2020-12-03
nv2a: Refactor VGA init

haxar:
This PR supports loading xbox-linux with 128 MiB of VRAM.

When Cromwell sets the framebuffer location to the end of VRAM while loading the Linux kernel, the xemu display remains stuck at the Cromwell Linux loading screen, but execution continues normally.

The refactor prevents `vga_draw_graphic()` from setting `force_shadow` with a previously hardcoded VRAM size of 64 MiB, allowing Cromwell & `simplefb` to update the xemu display.

After:
Screenshot at 2020-12-03 03-53-20
Before:
Screenshot at 2020-12-03 04-34-45

### DIFF
--- a/hw/xbox/nv2a/nv2a.c
+++ b/hw/xbox/nv2a/nv2a.c
@@ -255,19 +255,34 @@ static void nv2a_init_memory(NV2AState *d, MemoryRegion *ram)
     memory_region_set_log(d->vram, true, DIRTY_MEMORY_NV2A_TEX);
     memory_region_set_dirty(d->vram, 0, memory_region_size(d->vram));
 
-    /* hacky. swap out vga's vram */
-    memory_region_destroy(&d->vga.vram);
-    // memory_region_unref(&d->vga.vram); // FIXME: Is ths right?
-    memory_region_init_alias(&d->vga.vram, OBJECT(d), "vga.vram",
-                             d->vram, 0, memory_region_size(d->vram));
-    d->vga.vram_ptr = memory_region_get_ram_ptr(&d->vga.vram);
-    vga_dirty_log_start(&d->vga);
-
     pgraph_init(d);
 
     /* fire up pfifo */
     qemu_thread_create(&d->pfifo.thread, "nv2a.pfifo_thread",
                        pfifo_thread, d, QEMU_THREAD_JOINABLE);
+}
+
+static void nv2a_init_vga(NV2AState *d)
+{
+    VGACommonState *vga = &d->vga;
+    vga->vram_size_mb = memory_region_size(d->vram) / MiB;
+
+    vga_common_init(vga, OBJECT(d));
+    vga->get_bpp = nv2a_get_bpp;
+    vga->get_offsets = nv2a_get_offsets;
+    // vga->overlay_draw_line = nv2a_overlay_draw_line;
+
+    d->hw_ops = *vga->hw_ops;
+    d->hw_ops.gfx_update = nv2a_vga_gfx_update;
+    vga->con = graphic_console_init(DEVICE(d), 0, &d->hw_ops, vga);
+
+    /* hacky. swap out vga's vram */
+    memory_region_destroy(&vga->vram);
+    // memory_region_unref(&vga->vram); // FIXME: Is ths right?
+    memory_region_init_alias(&vga->vram, OBJECT(d), "vga.vram",
+                             d->vram, 0, memory_region_size(d->vram));
+    vga->vram_ptr = memory_region_get_ram_ptr(&vga->vram);
+    vga_dirty_log_start(vga);
 }
 
 static void nv2a_lock_fifo(NV2AState *d)
@@ -340,18 +355,6 @@ static void nv2a_realize(PCIDevice *dev, Error **errp)
     pci_set_word(dev->config + PCI_SUBSYSTEM_VENDOR_ID, 0);
     pci_set_word(dev->config + PCI_SUBSYSTEM_ID, 0);
     dev->config[PCI_INTERRUPT_PIN] = 0x01;
-
-    /* legacy VGA shit */
-    VGACommonState *vga = &d->vga;
-    vga->vram_size_mb = 64;
-
-    vga_common_init(vga, OBJECT(dev));
-    vga->get_bpp = nv2a_get_bpp;
-    vga->get_offsets = nv2a_get_offsets;
-
-    d->hw_ops = *vga->hw_ops;
-    d->hw_ops.gfx_update = nv2a_vga_gfx_update;
-    vga->con = graphic_console_init(DEVICE(dev), 0, &d->hw_ops, vga);
 
     /* mmio */
     memory_region_init(&d->mmio, OBJECT(dev), "nv2a-mmio", 0x1000000);
@@ -610,5 +613,6 @@ void nv2a_init(PCIBus *bus, int devfn, MemoryRegion *ram)
     PCIDevice *dev = pci_create_simple(bus, devfn, "nv2a");
     NV2AState *d = NV2A_DEVICE(dev);
     nv2a_init_memory(d, ram);
+    nv2a_init_vga(d);
     qemu_add_vm_change_state_handler(nv2a_vm_state_change, d);
 }
--- a/hw/xbox/nv2a/nv2a_int.h
+++ b/hw/xbox/nv2a/nv2a_int.h
@@ -25,6 +25,7 @@
 #include <assert.h>
 
 #include "qemu/osdep.h"
+#include "qemu/units.h"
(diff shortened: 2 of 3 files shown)


hw/xbox/nv2a/nv2a.c
```diff
@@ -427,7 +427,7 @@ static void nv2a_realize(PCIDevice *dev, Error **errp)
 
     /* legacy VGA shit */
     VGACommonState *vga = &d->vga;
-    vga->vram_size_mb = 64;
+    vga->vram_size_mb = (d->vram_size_mb == 128) ? 128 : 64;
```
GXTX: Why not?

```
vga->vram_size_mb = d->vram_size_mb;
```

haxar: I have no objection to this.

hw/xbox/xbox_pci.c
```diff
@@ -121,9 +121,9 @@ static int xbox_lpc_map_irq(PCIDevice *pci_dev, int intx)
     case 30: /* agp bridge -> PIRQC? */
         return XBOX_NUM_INT_IRQS + 2;
     default:
-        /* don't actually know how this should work */
-        assert(false);
-        return XBOX_NUM_INT_IRQS + ((slot + intx) & 3);
+        /* FIXME: xbox-linux queries in this range, so just return an unused irq? for now */
```
mborgerson: Huh? What device is it trying to bring up here?

hw/xbox/nv2a/nv2a.c
```diff
@@ -640,6 +640,11 @@ static const VMStateDescription vmstate_nv2a = {
     },
 };
 
+static Property nv2a_properties[] = {
```
mborgerson: Let's avoid adding this new property and instead refactor the vga init inside `realize` into a new function that gets called after memory init and then you can `memory_region_size(d->vram)`

mborgerson (changes requested):

hw/xbox/xbox_pci.c
haxar: I assume it's Linux enumerating all the PCI devices, but not able to confirm since `simplefb` is not initialized at that point & can't see dmesg (black screen), so not sure what this function is trying to return, as it does not match physical IRQ. Didn't want to return a value that was used by another Xbox PCI device, such as the AGP bridge above.

hw/xbox/nv2a/nv2a.c
haxar: Good call on the refactor.

hw/xbox/xbox_pci.c
haxar: Linux PCI probe enumerates `xbox-pci`, `xbox-lpc` & `xbox-smbus` which don't have mappable IRQs & will cause us to use the default case & hit this assert.

~~See `xbox_pci: Remove assert for pci_update_irq_disabled()` for more info.~~

haxar: There's also this previous convo regarding the above comment to give more context:

> https://discord.com/channels/428359196719972353/616286065355128832/619425788776808448
JayFoxRox 09/05/2019
I never understood the XQEMU / Xbox PCI IRQ remapping stuff btw; and the comments suggest it's hardcoded based on trial and error, too :stuck_out_tongue:
it's also one of those things that's hard to test on hardware [to me]
https://github.com/xqemu/xqemu/blob/dd1248a563af6cfb024a68d21250069b5c41e25e/hw/xbox/xbox_pci.c#L122 here's that comment for the XQEMU code - also the function above probably would have deserved some explanations

hw/xbox/nv2a/nv2a_int.h
```diff
@@ -25,6 +25,7 @@
 #include <assert.h>
 
 #include "qemu/osdep.h"
+#include "qemu/units.h"
```
mborgerson, line 28: Is this necessary

haxar: This provides `MiB` & is used for `memory_region_size(d->vram) / MiB`.

hw/xbox/xbox_pci.c
haxar: The above devices don't have interrupts to set. My latest force-push should be a better solution.

hw/xbox/xbox_pci.c
```diff
@@ -121,9 +123,8 @@ static int xbox_lpc_map_irq(PCIDevice *pci_dev, int intx)
     case 30: /* agp bridge -> PIRQC? */
         return XBOX_NUM_INT_IRQS + 2;
     default:
-        /* don't actually know how this should work */
-        assert(false);
-        return XBOX_NUM_INT_IRQS + ((slot + intx) & 3);
+        /* don't set an interrupt if a device does not have one */
+        return XBOX_NUM_INT_IRQS + XBOX_NUM_PIRQS;
```
haxar: OK, this won't work since this can cause an out of bounds access with `bus->irq_count[irq_num]` in `pci_change_irq_level()` when the array is this large (`bus->nirq == XBOX_NUM_INT_IRQS + XBOX_NUM_PIRQS`). This function (`xbox_lpc_map_irq()`) is returning an index to a nibble at the below mentioned LPC PCI configuration register.

Just PCI slot 0 & 1 need to be mapped somewhere (for Linux to enumerate the PCI devices) & they may have to be a device on the internal bus. The LPC PCI configuration register at `lpc->dev.config + XBOX_LPC_INT_IRQ_ROUT` provides the actual IRQ mappings & only 2 nibbles are actually `0`, which may map respectively & don't have to set an interrupt when `pic_irq == 0`.

Here's xemu's LPC PCI configuration register dump at that location:
```
[    0.241000] pci 0000:00:01.0: config space:
...
[    0.244000] 00000060: 00 04 00 00 0c 0b 00 00 00 00 03 00 91 54 06 0e
```

Each nibble of `91 54 06 0e` (offset `XBOX_LPC_INT_IRQ_ROUT`) corresponds to an actual IRQ number of a device.

hw/xbox/xbox_pci.c
```diff
@@ -109,17 +107,37 @@ static void xbox_lpc_set_irq(void *opaque, int pirq, int level)
...
+    /*
+     * Return the index of a nibble at LPC PCI config
+     * register XBOX_LPC_INT_IRQ_ROUT for the actual
+     * IRQ number of a given PCI device slot.
+     *
+     * This register is hardcoded on hardware as:
+     *
+     * 0x0e065491 @ XBOX_LPC_INT_IRQ_ROUT
+     */
+    case 0: return 5; /* hostbridge, no IRQ */
```
mborgerson, line 122: Is this a guess based on unassigned routes, or did you verify that these are the correct routes for these particular slots?

Either way it should be okay

hw/xbox/xbox_pci.c
```diff
@@ -109,17 +107,37 @@ static void xbox_lpc_set_irq(void *opaque, int pirq, int level)
...
-    case 30: /* agp bridge -> PIRQC? */
-        return XBOX_NUM_INT_IRQS + 2;
+    /*
+     * Return the index of a nibble at LPC PCI config
+     * register XBOX_LPC_INT_IRQ_ROUT for the actual
+     * IRQ number of a given PCI device slot.
+     *
+     * This register is hardcoded on hardware as:
+     *
+     * 0x0e065491 @ XBOX_LPC_INT_IRQ_ROUT
```
mborgerson, line 120: Confirmed on my box as well

hw/xbox/xbox_pci.c
```diff
@@ -109,17 +107,37 @@ static void xbox_lpc_set_irq(void *opaque, int pirq, int level)
...
+    case 9: return 6; /* ide, IRQ 14 */
+    /* pirqs */
+    /*
+     * Return the index of a byte at LPC PCI config
+     * register XBOX_LPC_PIRQ_ROUT for the actual
+     * IRQ number of a given PCI device slot.
+     *
+     * This register is hardcoded on hardware as:
+     *
+     * 0x00031000 @ XBOX_LPC_PIRQ_ROUT
```
mborgerson, line 138: Confirmed on my box as well

hw/xbox/xbox_pci.c
haxar: This is a guess & not able to verify if this is correct. These slots should not have a mappable interrupt & should safely not assert one.

GXTX:
Just pinging to bring attention to the maintainer that this is still open.

MasonT8198:
> Just pinging to bring attention to the maintainer that this is still open.

@mborgerson are there any outstanding issues with this being merged

mborgerson:
Thanks! Sorry about the long merge time

PR #98, opened 2020-12-12
build.sh: guess job_count when possible, better way to guess project directory, fix out of tree build

illwieckz:
When I rewrote `build.sh` long time ago (2918fb9), it was told me it had to be very simple so I removed `job_count` computation from my PR at the time. But since the script became more complex and maybe I may suggest this little feature again ?

Also in the meantime I [reused my script](https://gitlab.com/xonotic/netradiant/-/blob/master/easy-builder) on another unrelated project so I updated and improved that `job_count` compute since that time. It is meant to work on Linux, FreeBSD, macOS, MSYS2, otherwise it falls back on the default number. Explicit command line option `-j` can still override it of course.

I also wrote for that `build.sh` derivative a better way to guess the project directory so I import that change as well.

I then used that project directory variable to fix imgui out of tree build and out of tree packaging.

While I was at it I added for free some quote around some strings to make them less error prone, I also silenced the `command -v greadlink` error since there is an explicit one right after that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -164,6 +164,6 @@ trace-ust-all.h
 trace-ust-all.c
 /target/arm/decode-sve.inc.c
 
-
+/dist
 /xemu-version.c
 /xemu.ini
--- a/build.sh
+++ b/build.sh
@@ -88,8 +88,43 @@ package_linux() {
 postbuild=''
 debug_opts=''
 build_cflags='-O3'
-job_count='12'
+default_job_count='12'
 sys_ldflags=''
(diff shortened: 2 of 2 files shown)


dracc (approved):

illwieckz:
Wait a moment before merging this because I just noticed the special `readlink` thing was also the part I rewrote, so I'll do some testing to see if we can just drop it entirely or something may be done differently.

mborgerson:
Thanks

PR #100, opened 2020-12-13
xbox: Fix MCPX overlay and allow disablement

Ernegien:
This commit reverts back to the ROM address space being read-only with exception to the last alias instance during boot which has an overlay containing the BIOS + MCPX and honors the xbox's request to disable it.

These changes should allow for greater compatibility with retail kernels that rely on write-caching in the upper ROM address space during boot as well as hacked kernels that use the upper 512 bytes of its image (MCPX territory) to either 1.) store settings which have been previously getting overwritten by the MCPX being applied to all aliases, or 2.) fall through to custom code when the MCPX gets disabled via x-codes.

### DIFF
--- a/hw/xbox/xbox_pci.c
+++ b/hw/xbox/xbox_pci.c
@@ -146,6 +146,7 @@ void xbox_pci_init(qemu_irq *pic,
                    MemoryRegion *address_space_io,
                    MemoryRegion *pci_memory,
                    MemoryRegion *ram_memory,
+                   MemoryRegion *rom_memory,
                    PCIBus **out_host_bus,
                    ISABus **out_isa_bus,
                    I2CBus **out_smbus,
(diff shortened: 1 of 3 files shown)


hw/xbox/xbox_pci.c
```diff
@@ -315,6 +317,27 @@ static void xbox_lpc_reset(DeviceState *dev)
...
+    XBOX_LPCState *s = XBOX_LPC_DEVICE(dev);
+
+    pci_default_write_config(dev, addr, val, len);
+
+    if ((addr == 0x80) && (val & 2)) {
+        XBOXPCI_DPRINTF("DEACTIVATING BOOT ROM\n");
+        MemoryRegion *subregion;
+        QTAILQ_FOREACH(subregion, &s->rom_memory->subregions, subregions_link) {
+            if (subregion->name != NULL && strcmp(subregion->name, "xbox.mcpx") == 0) {
+                memory_region_set_enabled(subregion, false);
```
mborgerson: Did you check reset behavior? We will need to re-enable the boot ROM upon reset; I don't think it happens automatically

Ernegien: Good catch, it definitely needed to be re-enabled upon reset. Luckily we're able to do so cleanly on PCI reset and not somewhere else upstream.

mborgerson:
Thank you sir

PR #105, opened 2020-12-21
smbus: Add missing error flag on SMBus transactions

LoveMHz:
The current implementation always returns success on SMBus transactions even if the requested device does not exist.

This PR attempts to use the same logic as in pm_smbus.c to add the error flag to the SMBus transaction's status when amd756_smb_transaction() fails to read/write data to a device.

**This PR is not ready to be merged. It has only been briefly tested.**

While not fully investigated this appears to cause the following problems and possibly more. 
- The m8plus BIOS, and possibly others, attempt to read registers from the video encoder to determine which encoder the system has. With the m8plus BIOS this happens at 0x800b1b76, during kernel initialization, and incorrectly determines the system has an xcalibur encoder since it's the first one tested.
- Homebrew apps that require HalReadSMBusValue/HalWriteSMBusValue to fail to determine hardware. This can be seen in the app for the XboxHDMI when it attempts to check for the device before reading the values for the device's firmware. 
https://github.com/MakeMHz/xbox-hdmi-app/blob/master/source/scene0.cpp#L30 

Current xemu:
Screenshot_from_2020-12-21_14-50-18

Expected:
Screenshot from 2020-12-21 17-41-28

### DIFF
--- a/hw/xbox/amd_smbus.c
+++ b/hw/xbox/amd_smbus.c
@@ -86,55 +86,80 @@ static void amd756_smb_transaction(AMD756SMBus *s)
     uint8_t cmd  = s->smb_cmd;
     uint8_t addr = (s->smb_addr >> 1) & 0x7f;
     I2CBus *bus  = s->smbus;
+    int ret;
 
     SMBUS_DPRINTF("SMBus trans addr=0x%02x prot=0x%02x\n", addr, prot);
 
     switch (prot) {
     case AMD756_QUICK:
-        smbus_quick_command(bus, addr, read);
+        ret = smbus_quick_command(bus, addr, read);
+        goto done;
         break;
     case AMD756_BYTE:
         if (read) {
-            s->smb_data0 = smbus_receive_byte(bus, addr);
+            ret = smbus_receive_byte(bus, addr);
+            goto data8;
         } else {
-            smbus_send_byte(bus, addr, cmd);
+            ret = smbus_send_byte(bus, addr, cmd);
+            goto done;
         }
         break;
     case AMD756_BYTE_DATA:
         if (read) {
-            s->smb_data0 = smbus_read_byte(bus, addr, cmd);
+            ret = smbus_read_byte(bus, addr, cmd);
+            goto data8;
         } else {
-            smbus_write_byte(bus, addr, cmd, s->smb_data0);
+            ret = smbus_write_byte(bus, addr, cmd, s->smb_data0);
+            goto done;
         }
         break;
     case AMD756_WORD_DATA:
         if (read) {
-            uint16_t val;
-            val = smbus_read_word(bus, addr, cmd);
-            s->smb_data0 = val;
-            s->smb_data1 = val >> 8;
+            ret = smbus_read_word(bus, addr, cmd);
+            goto data16;
         } else {
-            smbus_write_word(bus, addr, cmd, (s->smb_data1 << 8) | s->smb_data0);
+            ret = smbus_write_word(bus, addr, cmd, (s->smb_data1 << 8) | s->smb_data0);
+            goto done;
         }
         break;
     case AMD756_BLOCK_DATA:
         if (read) {
-            s->smb_data0 = smbus_read_block(bus, addr, cmd, s->smb_data,
+            ret = smbus_read_block(bus, addr, cmd, s->smb_data,
                                             sizeof(s->smb_data), true, true);
+            goto data8;
         } else {
-            smbus_write_block(bus, addr, cmd, s->smb_data, s->smb_data0,
+            ret = smbus_write_block(bus, addr, cmd, s->smb_data, s->smb_data0,
                               true);
+            goto done;
         }
         break;
     default:
         goto error;
     }
+    abort();
 
+data16:
+    if (ret < 0) {
+        goto error;
+    }
+    s->smb_data1 = ret >> 8;
+data8:
+    if (ret < 0) {
+        goto error;
+    }
+    s->smb_data0 = ret;
+done:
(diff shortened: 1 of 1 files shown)


haxar:
This PR provides the correct I2C detection behavior in `i2cdetect` when used in xbox-linux via xemu.

After:
```
/ # i2cdetect -y 0
     0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
00:          -- -- -- -- -- -- -- -- -- -- -- -- -- 
10: 10 -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
20: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
30: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
40: -- -- -- -- -- 45 -- -- -- -- -- -- 4c -- -- -- 
50: -- -- -- -- 54 -- -- -- -- -- -- -- -- -- -- -- 
60: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
70: -- -- -- -- -- -- -- --
```
Before:
```
/ # i2cdetect -y 0
     0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
00:          03 04 05 06 07 08 09 0a 0b 0c 0d 0e 0f 
10: 10 11 12 13 14 15 16 17 18 19 1a 1b 1c 1d 1e 1f 
20: 20 21 22 23 24 25 26 27 28 29 2a 2b 2c 2d 2e 2f 
30: 30 31 32 33 34 35 36 37 38 39 3a 3b 3c 3d 3e 3f 
40: 40 41 42 43 44 45 46 47 48 49 4a 4b 4c 4d 4e 4f 
50: 50 51 52 53 54 55 56 57 58 59 5a 5b 5c 5d 5e 5f 
60: 60 61 62 63 64 65 66 67 68 69 6a 6b 6c 6d 6e 6f 
70: 70 71 72 73 74 75 76 77
```

hw/xbox/amd_smbus.c
```diff
@@ -86,55 +86,80 @@ static void amd756_smb_transaction(AMD756SMBus *s)
...
+        goto error;
+    }
+    s->smb_data0 = ret;
+done:
+    if (ret < 0) {
+        goto error;
+    }
     s->smb_stat |= GS_HCYC_STS;
+    goto out;
+out:
```
haxar, line 157: nit: `out` label is only used once here by the above `goto`.

haxar:
nit: Commit is missing `smbus:` module prefix.

GXTX:
Can we get rid of the `goto`'s? Or am I missing something? And set the `default: ret = -1;`

```c
    if (ret < 0) {
        s->smb_stat |= GS_PRERR_STS;
        return;
    }

    if (!read) {
        if (prot == AMD756_WORD_DATA) {
            s->smb_data1 = ret >> 8;
        }
        s->smb_data0 = ret;
    }

    s->smb_stat |= GS_HCYC_STS;

    return;
```

LoveMHz:
> Can we get rid of the `goto`'s? Or am I missing something? And set the `default: ret = -1;`
> 
> ```c
>     if (ret < 0) {
>         s->smb_stat |= GS_PRERR_STS;
>         return;
>     }
> 
>     if (!read) {
>         if (prot == AMD756_WORD_DATA) {
>             s->smb_data1 = ret >> 8;
>         }
>         s->smb_data0 = ret;
>     }
> 
>     s->smb_stat |= GS_HCYC_STS;
> 
>     return;
> ```

I'm impartial to either implementation. The submitted PR follows the style of pm_smbus.c and the extra goto logic was kept to keep it inline with that.

GXTX:
It just feels dirty to use them here.

mborgerson:
> The submitted PR follows the style of pm_smbus.c

Consistency with existing code is preferable for reviewing+integrating with upstream changes

mborgerson:
@haxar thanks for comparing with hardware!

GXTX:
Just pinging to bring attention to the maintainer that this is still open.

mborgerson:
> Just pinging to bring attention to the maintainer that this is still open.

This is not merged because the PR description claims it is untested and that there's evidence it causes problems. Until whatever needs to be fixed is fixed and tested, this will remain unmerged. I encourage you to get involved if you're interested, @GXTX

GXTX:
@mborgerson 

>there's evidence it causes problems

This patch fixes the underlying issues brought up in the description.

>untested

haxar has already confirmed that this lines up with what should be expected via XboxLinux.

I did some quick tests and this patch lines up with hardware (1.0) by attempting to read from the XCal encoder.

On hardware `HalReadSMBusValue` will return `C0000185` when the request fails and the return variable will have what ever was last in the query *before* the failed request. Applying this patch brings xemu along with hardware.

Before:
image

After:
image


Perhaps @LoveMHz could touch base on this again if there's still concerns.

hw/xbox/amd_smbus.c
```diff
@@ -86,55 +86,80 @@ static void amd756_smb_transaction(AMD756SMBus *s)
...
     return;
 
 error:
     s->smb_stat |= GS_PRERR_STS;
+    return;
```
mborgerson, line 162: nit: `return;` statement at end of function is superfluous

mborgerson:
> Perhaps @LoveMHz could touch base on this again if there's still concerns.

If you're convinced it's good to go, I'm fine to merge

mborgerson:
> @mborgerson
> 
> > there's evidence it causes problems
> 
> This patch fixes the underlying issues brought up in the description.

@GXTX Oops. Thanks for bringing that up, and testing

PR #127, opened 2021-01-01
i386: Change default pentium3 cpuid to match 1.0 Xbox

GXTX:
Small mostly cosmetic change, however the original xbox does seem to actually have `CPUID_PN`, enabling this feature causes a warning
`xemu: warning: TCG doesn't support requested feature: CPUID.01H:EDX.pn [bit 18]` which is strange because if you set the CPU to be the `athlon` this actually has one set as well so the feature does seem to work?

retail box:
2021-00-01_00-30-13-ed

xemu with change:
after


mborgerson:
Confirmed this CPU model/stepping on my Xbox. Thanks!

PR #131, opened 2021-01-02
smbus: Fix Xbox SMC version detection

### LINKED ISSUE #128: Reading SMC Version is broken
Reading the SMC version appears broken - letting it retry multiple times appears to "fix" it. Looking in to how it works it seems it uses some index system and is getting messed up there somehow.

https://github.com/mborgerson/xemu/blob/master/hw/xbox/smbus_xbox_smc.c#L167

xbox:
xbox

xemu (each line represents the machine "reloading" the xbe - 4th is actually 5th - 1st was a blank):
xemu-smc

```C
    char smc_rev[3];
    for (int i = 0; i < 3; i++) {
        HalReadSMBusValue(0x20, 0x01, FALSE, &smc_rev[i]);
    }
    debugPrint("SMC Revision: %s\n", smc_rev);
```

GXTX:
This fixes the version detection, I think it is probably required for some stuff as the kernels check it during boot, first they check `[0]` then reset the index then check the full string. So probably doing something backend.

Eventually I'd like to have this configurable at run-time so I moved it into the device struct.

xemu:
image

Fixes #128

### DIFF
--- a/hw/xbox/smbus_xbox_smc.c
+++ b/hw/xbox/smbus_xbox_smc.c
@@ -91,7 +91,7 @@
 #define SMC_REG_SCRATCH             0x1b
 #define     SMC_REG_SCRATCH_SHORT_ANIMATION 0x04
 
-static const char *smc_version_string = "P01";
+static const char smc_version_string[] = "P01";
 
 typedef struct SMBusSMCDevice {
     SMBusDevice smbusdev;


hw/xbox/smbus_xbox_smc.c
```diff
@@ -91,10 +91,9 @@
 #define SMC_REG_SCRATCH             0x1b
 #define     SMC_REG_SCRATCH_SHORT_ANIMATION 0x04
 
-static const char *smc_version_string = "P01";
```
mborgerson, line 94: Nice catch! The reason this was broken is that `sizeof(smc_version_string)` below returned the size of a `char` pointer (8 if you're on x86-64), not the statically allocated character array size.

While your patch is not bad by design, I do not think the added complexity of this patch is necessary right now; instead, this problem could be fixed by a 3 character delta. If you drop all other changes and simply change this line to:
```c
static const char smc_version_string[] = "P01";
```
It should work as expected; that is, the `sizeof(smc_version_string)` below will return the length of the string + 1. So, let's not move the version array into state until it becomes necessary for some reason.

GXTX: No problem, I can switch it out. As I said I had hoped to eventually make this something selectable at run-time but we can come back it when the time comes.

PR #158, opened 2021-01-19
nv2a: Change stepping to match retail 1.0 Xbox

GXTX:
xbox:
image

xemu:
image


mborgerson:
Matches what my Xbox reports. Tested working. Thanks

PR #171, opened 2021-01-29
Initial Configurable Keyboard Scancode Settings

camden-smallwood:
This adds input settings to the `xemu.ini` file to allow user specification of keyboard controller emulation SDL scancodes. It uses the previous control scheme out of the box, but allows a user to set their own scancodes if they so choose.


mborgerson:
Thanks for the PR! I agree this should be a feature but have planned to do button mapping in a slightly different way. I'll hold off on merging this for now but will come back to it soon and we can work together to get it merged. Thanks for your patience.

PR #181, opened 2021-02-09
settings: Read xemu.ini from binary path first

### LINKED ISSUE #138: Implement portable mode
This is a (basic and must-have) feature request:

Please consider to implement portable mode, that is the **xemu.ini** and any other files must be read/written in root directory of **xemu.exe**. 

One easy way for all users to set portable mode probably would be to let them create a "portable.txt" file in the root folder of **xemu.exe**, and then **xemu** should do the magic to recognize the installation as a portable mode. Or through an option in the emulator UI.

I cannot think of a good emulator that is not portable itself or does not provide the user a functionality to make it portable, like PCSX2, RPCS3, Dolphin, Cemu, Citra and of course Cxbx-R just to mention some of them.

I'm in love with **Xemu**'s UI and its functionality but this is a thing that stops me from using it as my main choice instead of others (well, the only other working xbox emulator) because I need emulators to be portable, not to save settings in the user roaming folder like **xemu** does :(

Just please consider this feature request. Thanks in advance.

dracc:
Closes #138.

### DIFF
--- a/ui/xemu-settings.c
+++ b/ui/xemu-settings.c
@@ -21,7 +21,9 @@
 #include <SDL_filesystem.h>
 #include <string.h>
 #include <stddef.h>
+#include <stdbool.h>
 #include <assert.h>
+#include <glib.h>
 
 #include "xemu-settings.h"
 #include "inih/ini.c" // FIXME
@@ -189,6 +191,20 @@ int xemu_settings_get_enum(enum xemu_settings_keys key, int *val)
 	return 0;
 }
 
+bool xemu_settings_detect_portable_mode(void)
+{
+	bool val = false;
+	char *portable_path = g_strdup_printf("%s%s", SDL_GetBasePath(), filename);
+	FILE *tmpfile;
+	if ((tmpfile = fopen(portable_path, "r"))) {
+		fclose(tmpfile);
+		val = true;
+	}
+
+	free(portable_path);
+	return val;
+}
+
 const char *xemu_settings_get_path(void)
 {
 	if (settings_path != NULL) {
@@ -197,6 +213,13 @@ const char *xemu_settings_get_path(void)
 
 	char *base = SDL_GetPrefPath("xemu", "xemu");
 	assert(base != NULL);
+
+	// Check for xemu.ini in xemu binary directory, "portable mode"
+	if (xemu_settings_detect_portable_mode()) {
+		base = SDL_GetBasePath();
+		assert(base != NULL);
+	}
+
 	size_t base_len = strlen(base);
 
 	size_t filename_len = strlen(filename);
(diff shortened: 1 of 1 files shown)


ui/xemu-settings.c
```diff
@@ -197,6 +197,17 @@ const char *xemu_settings_get_path(void)
 
 	char *base = SDL_GetPrefPath("xemu", "xemu");
 	assert(base != NULL);
+
+	// Check for xemu.ini in xemu binary directory, "portable mode"
+	char portable_path[260];
+	sprintf(portable_path, "%s%s", SDL_GetBasePath(), "xemu.ini");
```
haxar: I always avoid using `sprintf()` & instead use `snprintf(portable_path, sizeof(portable_path), ...)` to avoid any buffer overflow.

dracc: Thanks, I opted to follow your example. :+1:

ui/xemu-settings.c
```diff
@@ -197,6 +197,17 @@ const char *xemu_settings_get_path(void)
 
 	char *base = SDL_GetPrefPath("xemu", "xemu");
 	assert(base != NULL);
+
+	// Check for xemu.ini in xemu binary directory, "portable mode"
+	char portable_path[260];
```
mborgerson: - Why 260? How about just using some automatically allocating method as I have done before, e.g. glib `g_strdup_printf` so we don't arbitrarily limit path length here
- Please refactor so this exact code isn't duplicated here and below

dracc: * My bad, didn't read existing code enough to spot that. Thanks!
* Code refactored.

ui/xemu-settings.c
```diff
@@ -189,6 +190,20 @@ int xemu_settings_get_enum(enum xemu_settings_keys key, int *val)
 	return 0;
 }
 
+int xemu_settings_look_for_settings_file(void)
```
mborgerson: Some minor changes requested:
- Let's change the name of this function to better reflect what it does: `xemu_settings_detect_portable_mode`, or something
- This should return boolean type
- Below where we further determine base path based on this; probably that should be extracted into a new function `xemu_get_settings_path` or something.

I'm not too worried about this last point right now--we can refactor later if you don't want to. But please do fix up the first two.

BiatuAutMiahn: `int getXemuCfg(void)`
You could also simply look for a portable.txt like a few programs out there and if it exists create a directory like `UserCfg` which would have everything that would be in AppData.

dracc: @mborgerson Good points, I'll try to address them tomorrow.

@BiatuAutMiahn Not a bad name, but doesn't fit the existing naming scheme.

dracc: @mborgerson First two points have been taken care of now.

mborgerson:
Thanks

PR #203, opened 2021-02-18
vl: Allow overriding settings path from command line

thrimbor:
This allows to override the settings file path from the command line by using `-config_path`, similar to what's possible with `-dvd_path`. It's mainly a qol improvement for myself, to make it easier to switch between different hdd and eeprom image combinations for testing.


mborgerson:
Thanks

PR #231, opened 2021-03-05
Add workaround for OpenGL Core profiles (macOS)

### LINKED ISSUE #214: glDrawElements GL error
Regression from 0d9ff0872ac233bc633ae6815f31197cc4d142ec on 2013 MBP, crashing on startup. Should be an easy fix.

camden-smallwood:
This is a workaround for the crashing on macOS that was introduced after the recent revisions to the nv2a code. Tested successfully on a 2020 MacBook Pro 13-inch model with Big Sur 11.2.2 installed.

References:
https://community.khronos.org/t/gldrawelements-causes-gl-invalid-operation/65185
https://github.com/ocornut/imgui/pull/278/files
https://stackoverflow.com/a/12618739

Fixes #214

### DIFF
--- a/hw/xbox/nv2a/nv2a_int.h
+++ b/hw/xbox/nv2a/nv2a_int.h
@@ -341,6 +341,7 @@ typedef struct PGRAPHState {
 
     unsigned int inline_elements_length;
     uint32_t inline_elements[NV2A_MAX_BATCH_LENGTH];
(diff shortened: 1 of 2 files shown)


mborgerson:
Thanks! We actually used to do this before I botched things up on a recent update. I've taken the liberty to fix some style nits to expedite this.

PR #233, opened 2021-03-06
nv2a: Factor Kelvin methods into functions

mborgerson:


PatrickvL:
What is the benefit of this?

PR #253, opened 2021-03-12
nv2a: Round inline array attr offsets to natural boundary

### LINKED ISSUE #104: Assertion for non-dword aligned inline array offset (Conker L&R)
When you try to catch the key , you must have weapon. After conker having a Frying Pan , the emu crash.

Here's the scene:

And here's the error:

Titles: 4d530051, 4d57001a

mborgerson:
Partially fixes the issue present in Conker allowing it to run; does need more investigation, but this is objectively better.

Closes #104


Xpmaster2002:
seems to be crashing at certain points right before of after something loads, like an ingame event starts then it crashes or an in game event finishes then it crashes... Seems to still have a problem with inline array amount on current release later on in game, locked at this save point for now... Entering the Dung Mound....
Untitled

PR #258, opened 2021-03-14
xemu.c: Window Adjusts to Display Size

MasonT8198:
This change is because without it there is an issue where the menu settings disappear because of a smaller screens

### DIFF
--- a/ui/xemu.c
+++ b/ui/xemu.c
@@ -883,16 +883,10 @@ static void sdl2_display_init(DisplayState *ds, DisplayOptions *o)
     SDL_SysWMinfo info;
 
     SDL_DisplayMode current;
-
+    // Query Window Size and Print The Display Mode
     for(i = 0; i < SDL_GetNumVideoDisplays(); ++i){
-
         int window_query = SDL_GetCurrentDisplayMode(i, &current);
-
-        if(window_query != 0)
-            SDL_Log("Couldn't get display mode on your display #%d: %s", i, SDL_GetError());
-
-        else
-            SDL_Log("Display #%d: current display mode is %dx%dpx @ %dhz. ", i, current.w, current.h, current.refresh_rate);
+        SDL_Log("Display #%d: current display mode is %dx%dpx @ %dhz. ", i, current.w, current.h, current.refresh_rate);
     }
     
     assert(o->type == DISPLAY_TYPE_XEMU);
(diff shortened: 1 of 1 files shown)


GXTX:
Are people using 800x600 these days? Looking at Steams hardware survey (although a bit biased towards "gamers") 1024x768 is 0.2% of the population, `statcounter.com` puts the same at 1% globally.

MasonT8198:
> Are people using 800x600 these days? Looking at Steams hardware survey (although a bit biased towards "gamers") 1024x768 is 0.2% of the population, `statcounter.com` puts the same at 1% globally.

It's the way it starts, on some laptop screens and some tv screens it'll start without the hud, and you have to fiddle around with it to make the hud reappear.

medievil1:
experience this myself.. 1920x1080 here but my dpi on a 50" display is 150%
dunno it is related as well... without setting it to use application size in the override(win 10/file properties) it will start top and bottom touching the full 1080 res and with the top cut off.....
click the side to resize and boom, snaps into correct range

MasonT8198:
> experience this myself.. 1920x1080 here but my dpi on a 50" display is 150%
> 
> dunno it is related as well... without setting it to use application size in the override(win 10/file properties) it will start top and bottom touching the full 1080 res and with the top cut off.....
> 
> click the side to resize and boom, snaps into correct range

It has to do with the way windows app scaling is. I have it happen on my 4k tv as well, it gets super big so it breaks xemu. Because scaling issues.

mborgerson:
The preferred fix is to query maximum window size (size of screen) and create a window no larger than current screen

MasonT8198:
> The preferred fix is to query maximum window size (size of screen) and create a window no larger than current screen

how would you go about doing this

mborgerson:
See https://wiki.libsdl.org/SDL_GetCurrentDisplayMode

mborgerson (changes requested):
Please check host resolution first as suggested

ui/xemu.c
```diff
@@ -882,6 +882,13 @@ static void sdl2_display_init(DisplayState *ds, DisplayOptions *o)
     int i;
     SDL_SysWMinfo info;
 
+    SDL_DisplayMode current;
+    // Query Window Size and Print The Display Mode
```
mborgerson, line 886: You have queried the screen size, now we need to make sure that we don't create a window larger than that size.

MasonT8198: Yes, my trouble is trying to find the command to limit those bounds, in SDL there is a command to adjust the screen size to those bounds but not limit them @mborgerson

mborgerson: You can limit them when we create the window where we specify window dimensions

mborgerson:
This is unfinished. I'll close for now and we can return with a follow up PR.

PR #261, opened 2021-03-18
video: Default to HDTV avpack

dracc:
Small change.
Most of us are interested in the emulation looking as good as possible - something unlocked by using the HDTV avpack.

Allowing the user to set the avpack to use themselves through the GUI has been suggested in the Discord server previously.
I believe this is a worthy stopgap solution until that is implemented.

### DIFF
--- a/hw/xbox/smbus_xbox_smc.c
+++ b/hw/xbox/smbus_xbox_smc.c
@@ -230,7 +230,7 @@ bool xbox_smc_avpack_to_reg(const char *avpack, uint8_t *value)
 
 void xbox_smc_append_avpack_hint(Error **errp)
 {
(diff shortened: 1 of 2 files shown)


MasonThompson8198:
This is definitely a good addition, especially for the end user, since they won't have to append a launch argument onto a shortcut (since you can't do it to the main.exe file) that would be more user friendly of a choice especially since all would have to do now would be attach a dash and turn allow use of 720p.

7oxicshadow:
Wondering if the default EEPROM that Xemu creates should enable 480p and 720p by default if this change is approved.
Also a note in the readme might be required as enabling 1080i in the dash can cause graphics corruption in some games (Dragons lair 3D).

theboy181:
This PR just sets HDTV to default? Will the command-line switching still work if needed? 

Thanks for doing this too! Great Idea!

PR #266, opened 2021-03-21
xbox: Fix for MCPX ROM overlay when not specified

### LINKED ISSUE #243: Debug Kernels broken after MCPX overlay fix
Appears #100 broke using real debug kernels.

Ernegien:
This allows earlier DVT4 debug BIOS images to boot when an MCPX ROM is not specified in the settings due to the BIOS image already having what it needs embedded within. Fixes https://github.com/mborgerson/xemu/issues/243.

### DIFF
--- a/hw/xbox/xbox.c
+++ b/hw/xbox/xbox.c
@@ -162,13 +162,13 @@ static void xbox_flash_init(MemoryRegion *rom_memory)
         assert(rc == bootrom_size);
         close(fd);
         g_free(filename);
-
-        MemoryRegion *mcpx = g_malloc(sizeof(MemoryRegion));
-        memory_region_init_ram(mcpx, NULL, "xbox.mcpx", bios_size,
-                               &error_fatal);
-        rom_add_blob_fixed("xbox.mcpx", bios_data, bios_size, -bios_size);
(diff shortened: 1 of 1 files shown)


GXTX:
Appears to fix the issue I was having, thank you! Still broken Cromwell but we should make that a separate issue all together.

Ernegien:
> Appears to fix the issue I was having, thank you! Still broken Cromwell but we should make that a separate issue all together.

The Cromwell issue is exactly that, an issue with Cromwell. It's dependent on CPU cache for functionality, and I suspect qemu will not easily be able to emulate this for a while. I believe my outstanding [PR](https://github.com/XboxDev/cromwell/pull/53) is a fair solution to this problem, and any existing Cromwell images that need to run within xemu may have to have the end of the image modified as discussed in https://github.com/mborgerson/xemu/issues/243.

GXTX:
> The Cromwell issue is exactly that, an issue with Cromwell.

I'll just make the comment I had here public but we shouldn't continue the discussion here. I disagree - it's an issue with xemu. Maybe the scope changed from when the fork happened but I believe the idea is to be compatible with a 1.0 Xbox.

PR #271, opened 2021-03-26
ui: Add Support for Higher UI Scaling

MasonT8198:
This is so on bigger 4K TVs on a HTPC setup (or just bigger 4K TVs in general) are easier to see in xemu, because even at 2x UI scaling it is hard to see, I have good vision so I could make it out rather easily, but I could see someone without good vision being able to spot it.


MasonT8198:
Second Add-In, sorry for the PR spam

dracc:
> Second Add-In, sorry for the PR spam

Do not worry about that, PRs touching five different things *are* worse.

Re changes: nit: bad commit titles (try prepending `ui: ` and modifying the rest of them to fit).

Aerocatia:
Perhaps it would be better to not have a hard-coded size menu for this but instead draw at a consistent scale that increases with resolution much like game UI's do.

MasonT8198:
> Perhaps it would be better to not have a hard-coded size menu for this but instead draw at a consistent scale that increases with resolution much like game UI's do.

I could see this being a feature implemented in the feature, I just appended the changes onto the already existing system, as when I use a bigger sized TV, 2x doesn't really help much, granted for me I can see it, but my dad couldn't or my stepmom, so I figured the feature would be a nice addition to an already existing system

ui/xemu-hud.cc
```diff
@@ -1733,14 +1733,14 @@ static void ShowMainMenu()
         {
             int ui_scale_combo = g_ui_scale - 1.0;
             if (ui_scale_combo < 0) ui_scale_combo = 0;
-            if (ui_scale_combo > 1) ui_scale_combo = 1;
-            if (ImGui::Combo("UI Scale", &ui_scale_combo, "1x\0" "2x\0")) {
+            if (ui_scale_combo > 1) ui_scale_combo = ui_scale_combo;
```
mborgerson: This statement now has no effect

ui/xemu-hud.cc
```diff
@@ -1733,14 +1733,14 @@ static void ShowMainMenu()
...
+            if (ImGui::Combo("UI Scale", &ui_scale_combo, "1x\0" "2x\0" "3x\0" "4x\0")) {
                 g_ui_scale = ui_scale_combo + 1;
                 xemu_settings_set_int(XEMU_SETTINGS_DISPLAY_UI_SCALE, g_ui_scale);
                 xemu_settings_save();
                 g_trigger_style_update = true;
             }
-
```
mborgerson: Please restore blank line

In general when submitting patches please try to follow the existing style

ui/xemu-hud.cc
```diff
@@ -1733,14 +1733,14 @@ static void ShowMainMenu()
...
+            ImGui::SameLine(); HelpMarker("Controls how the UI is scaled, not Surface Scaling");
```
mborgerson: I'm not sure this is necessary

mborgerson (changes requested):
There is also a check in `xemu_hud_init` that would need to be updated

mborgerson:
> Perhaps it would be better to not have a hard-coded size menu for this but instead draw at a consistent scale that increases with resolution much like game UI's do.

@Vaporeon  I welcome a more intuitive/automatic scaling method. Can you prototype this?

ui/xemu-hud.cc
```diff
@@ -1732,24 +1732,24 @@ static void ShowMainMenu()
...
-            if (ui_scale_combo > 1) ui_scale_combo = 1;
-            if (ImGui::Combo("UI Scale", &ui_scale_combo, "1x\0" "2x\0")) {
+            if (ImGui::Combo("UI Scale", &ui_scale_combo, "1x\0" "2x\0" "3x\0" "4x\0")) {
                 g_ui_scale = ui_scale_combo + 1;
                 xemu_settings_set_int(XEMU_SETTINGS_DISPLAY_UI_SCALE, g_ui_scale);
                 xemu_settings_save();
                 g_trigger_style_update = true;
             }
-
+
```
mborgerson: Please avoid whitespace changes like this a 2 more below.

ui/xemu-hud.cc
```diff
@@ -1732,24 +1732,24 @@ static void ShowMainMenu()
         if (ImGui::BeginMenu("View"))
         {
             int ui_scale_combo = g_ui_scale - 1.0;
-            if (ui_scale_combo < 0) ui_scale_combo = 0;
-            if (ui_scale_combo > 1) ui_scale_combo = 1;
-            if (ImGui::Combo("UI Scale", &ui_scale_combo, "1x\0" "2x\0")) {
+            if (ImGui::Combo("UI Scale", &ui_scale_combo, "1x\0" "2x\0" "3x\0" "4x\0")) {
```
mborgerson, line 2107: Don't forget to fix bounds checking when we initialize g_ui_scale at startup

MasonT8198:
@mborgerson got around to fixing up this one, it's been a while, so something may have been done incorrectly.

MasonT8198:
@mborgerson

MasonT8198:
This one has gone stale, I think it's clean now however I am closing this

PR #273, opened 2021-04-05
texture dumbing replacing 

kurdischerjapane:
can someone please add this function? maybe i also pay it


poobarfoob:
Not sure what you're asking for exactly, and why you made it a pull request, but you can always donate to Matt's [patreon](https://www.patreon.com/mborgerson) if you'd like to support him. However it wont make, what im assuming you want is, texture dumping / replacing a for sure thing.

PR #275, opened 2021-05-02
nv2a: implement FFP LIGHT_SPOT

### LINKED ISSUE #162: FFP LIGHT_SPOT needs to be implemented
https://github.com/mborgerson/xemu/blob/aab63105fce0b8f2699b520a157d2b0e95539e93/hw/xbox/nv2a/shaders.c#L569

Titles: 4c410005, 55530013, 5454008f, 47560004, 5655000d, 4b420001, 58490004, 47560003, 49460001, 53430001, 4947007b, 43430003, 564e0004, 55530038, 56550048

wilkovatch:
This closes #162.

Tested on Conflict: Desert Storm (53430001), Hunter: The Reckoning (564e0004) and Batman: Dark Tomorrow (4b420001).

It works mostly as described [here](https://docs.microsoft.com/en-us/windows/win32/direct3d9/attenuation-and-spotlight-factor#spotlight-factor).
However, there are a couple of things that had to be figured out.

The first is the order of the parameters `falloff`, `phi` and `theta`, since the input for all of them is a single `vec3` (`lightSpotFalloff`)
The values used by the 3 games I tested were:
`0.000000, -0.494592, 1.494592` (Conflict: Desert Storm and Hunter: The Reckoning)
`-0.581921, -2.174225, 2.592304` (Batman: Dark Tomorrow)

Now, the requirements are:
* `abs(phi) >= abs(theta)` (in theory `phi` and `theta` should be non-negative, but in fact using the absolute value is enough, since `cos(x)==cos(-x)`)
* `abs(phi) < pi` and `abs(theta) < pi` (but all the values above satisfy this requirement, so this does not help)
* `falloff > 0` (else the falloff curve would not be decreasing)

Considering the values above, the only order that satisfies these requirements (and indeed produces the same result as on hardware) is: `x, y, z` => `theta, phi, falloff`

The second thing is the direction vector. Normalizing it (as described in the documentation linked above) does not produce the same result as on hardware.
After looking at the values used by the games, I noticed that the `w` component is always close to the difference between 1 and the vector length.
So, dividing the vector by `1-w` is the same as normalizing it and scaling it a bit (which in turn scales the subsequent dot product, effectively changing the spotlight radius), and in fact doing this produces the same result as on hardware.

Some examples:
(from Hunter: The Reckoning):
* direction: `0.000000, -62.513657, -46.423523, -75.486740`
* `length = 77.865916`
* `1 - length = -76.865916`
* `length / (1 - w) = 1.018032`

(from Conflict: Desert Storm):
* direction: `1.720397, 1.274108, -1.382891, -1.461833`
* `length = 2.548628`
* `1 - length = -1.548628`
* `length / (1 - w) = 1.035256`

(from Batman: Dark Tomorrow):
* direction: `-0.389671, 0.743980, -0.574882, -0.017762`
* `length = 1.017762`
* `1 - length = -0.017762`
* `length / (1 - w) = 1.0`

A spotlight in the intro of Conflict: Desert Storm with this PR (same as on hardware):
cds

Here (Hunter: The Reckoning, in the subway station at the beginning of the game) spotlights are used on the characters, matching the ground illumination (same as on hardware):
htr

### DIFF
--- a/hw/xbox/nv2a/shaders.c
+++ b/hw/xbox/nv2a/shaders.c
@@ -566,8 +566,31 @@ GLSL_DEFINE(materialEmissionColor, GLSL_LTCTXA(NV_IGRAPH_XF_LTCTXA_CM_COL) ".xyz
                 /* Everything done already */
                 break;
             case LIGHT_SPOT:
-                assert(false);
-                /*FIXME: calculate falloff */
+                /* From: https://docs.microsoft.com/en-us/windows/win32/direct3d9/attenuation-and-spotlight-factor#spotlight-factor */
+                /* Note: normalizing the direction does not produce the same result as on hardware.
+                 *       Instead, dividing the direction by 1-w produces the same result as on hardware.
+                 *       Rationale: the w component is always close (but not equal to) the difference
+                 *       between 1 and the the direction vector length (e.g. 1.0-2.6=-1.6, with w=-1.5).
+                 *       So, dividing by 1-w is the same as normalizing the vector and scaling it a bit.
+                 *       In the example: 1-(-1.5) = 2.5, 2.6/2.5 = 1.04. This also scales the subsequent
+                 *       dot product by the same amount, effectively increasing the spotlight radius.
+                 */
+                qstring_append_fmt(body,
+                    "  vec3 dir = lightSpotDirection(%d).xyz / (1.0 - lightSpotDirection(%d).w);\n"
+                    "  float rho = dot(dir, VP);\n"
+                    "  float theta = lightSpotFalloff(%d).x;\n"
+                    "  float phi = lightSpotFalloff(%d).y;\n"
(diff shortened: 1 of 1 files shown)


mborgerson:
Great, this will unblock several titles. I plan to review the logic before merge, but it may take a few days. It would be helpful to have an isolated test case that exercises this if you have one

wilkovatch:
Unfortunately, apart from the two games in the screenshots above, I don't have any isolated test case for this (I don't know if it's possible to make one with nxdk).

mborgerson:
Thanks for working on this, and for your patience while I review it! I've rebased your patch and added some improvements on the  `feat/light_spot` branch, which I do intend to merge imminently. The main difference is that we can derive cos(phi/2) and cos(theta/2) from the direction vector, which simplifies the math/code a bit. This does appear to line up more closely to hardware. The falloff curve parameters do still need to be researched, but that can be deferred.

PR #372, opened 2021-06-26
build: Don't strip Windows debug binaries

thrimbor:
This came up in a Discord discussion yesterday, where a user wanted to debug the xemu process on Windows, but gdb complained about missing debug information despite it being the debug build.
This was caused by running the strip tool on the debug builds, so this commit disables that.

Not stripping the debug binary approximately doubles its size.


MasonT8198:
I had been doing this on my personal builds of xemu when I'm testing, as it makes it easier to work with on a non-stripped build, however I figured there was a reason why we strip debug builds

MasonT8198:
Addendum: I just tested this on Windows, it connected with gdb and symbols worked, fixing any issues a Windows user should have connected via GDB. Looks Good To Me @thrimbor !

PR #404, opened 2021-07-11
nv2a: Support texture format LC_IMAGE_YB8CR8YA8CB8 (UYVY)

Ernegien:
This matches up against real hardware using my [testbed](https://github.com/Ernegien/nxdk/tree/test/texture-formats/samples/texture-formats). I've also verified (not necessarily looking for pixel-perfectness) the existing YUY2 implementation for correctness. Something worth mentioning however, nv2a requires a configuration register to be set allowing for color space conversions before displaying properly; ideally we should assert here or just disregard as any games using the official XDK should already be following the rules. Currently xemu happily renders (incorrectly) YUV/UYVY textures without this set.


mborgerson:
Great! Thanks

PR #415, opened 2021-07-19
nv2a: Support NV062_SET_COLOR_FORMAT_LE_Y32 blitting

Ernegien:
Harry Potter and the Sorcerer's Stone asserts [here](https://github.com/mborgerson/xemu/blob/cc6d19e0990ab94aa55632302afb76b22f3ca09d/hw/xbox/nv2a/pgraph.c#L869) after finishing the wand shop tutorials. I'm fairly certain surface format 0xB is Y32 as the rest of [these](https://cgit.freedesktop.org/nouveau/xf86-video-nouveau/tree/src/hwdefs/nv01_2d.xml.h?id=e70d801ae9287eab5e82f4d467dc8cd4be1b31a8#n265) values line up with what we already have, but this needs more testing to be 100% sure.


mborgerson:
Awesome, thanks!

Issue #375, opened 2021-06-28
Meta: Crashes from Malloc Issues


Triticum0:
### Title Information

* Titles: 
*  ~~Crime Life: Gang Wars~~  (4b4e001c)
*  ~~Digimon World X~~ Fixed (42410002)
* ~~MotoGP: Ultimate Racing Technology 3~~ (54510088)
* ~~Second Sight~~ (434d0047)
* Phantom Dust (4d530046)
* ~~Links 2004~~  (4d53005d)
* ~~Torino 2006~~  (545400a9)
* ~~Ultimate Beach Soccer~~ (57450001)
* ~~Fahrenheit~~ (Needs to be tested on linux to check for assertion) (4947007b)

https://xemu.app/titles/4b4e001c/#Crime-Life-Gang-Wars
https://xemu.app/titles/42410002/#Digimon-World-X
https://xemu.app/titles/54510088/#MotoGP-Ultimate-Racing-Technology-3
https://xemu.app/titles/434d0047/#Second-Sight
https://xemu.app/titles/4d53005d/#Links-2004
https://xemu.app/titles/545400a9/#Torino-2006
https://xemu.app/titles/57450001/#Ultimate-Beach-Soccer
https://xemu.app/titles/4947007b/#Fahrenheit
https://xemu.app/titles/4d530046/#Phantom-Dust
### Bug Description
When trying to go in-game and performing certain actions in-game the game crashes with a malloc assertion.
Digimon World X and MotoGP: Ultimate Racing Technology 3: 
malloc(): corrupted top size

Crime LifeGang Wars, Second Sight and Links 2004:
 malloc(): invalid size (unsorted)

Torino 2006, Phantom Dust(while loading maps) and Second Sight (computer terminal crash):
xemu: malloc.c:2379: sysmalloc: Assertion `(old_top == initial_top (av) && old_size == 0) || ((unsigned long) (old_size) >= MINSIZE && prev_inuse (old_top) && ((unsigned long) old_end & (pagesize - 1)) == 0)' failed.

Ultimate Beach Soccer:
malloc(): mismatching next->prev_size (unsorted)

Steps to reproduce this issue:
1. Insert disk
2. start a game mode from the menu
3. before reaching in-game the game crashes
4. Press Start or go to the computer terminal in the first mission(Second sight) 

### Expected Behavior

The game should go in-game and not crash during game-play.

### Version

Version: 0.6.2
Branch: master
Commit: bbcd576
[...]

PR #426, opened 2021-07-25
pgraph.c: refactor clip dimensions

MasonT8198:
After experimenting with a few different solutions for issue #375 this solution (partially suggested by xbox7887) corrects the malloc(); corrupted top size issue seen in Digimon World 4/X, fixing #375


MasonT8198:
Also Seems to fix Predator: Concrete Jungle

hw/xbox/nv2a/pgraph.c
```diff
@@ -5319,8 +5319,8 @@ static void pgraph_populate_surface_binding_entry(NV2AState *d, bool color,
         /* Since we determine surface dimensions based on the clipping
          * rectangle, make sure to include the surface offset as well.
          */
-        width += pg->surface_shape.clip_x;
-        height += pg->surface_shape.clip_y;
+        width -= width % ( 1 << pg->surface_shape.clip_x);
+        height -= height % ( 1 << pg->surface_shape.clip_y);
```
mzji:
```suggestion
        width &= ~((1 << pg->surface_shape.clip_x) - 1);
        height &= ~((1 << pg->surface_shape.clip_y) - 1);
```

MasonT8198: Built and tested, also works as well implementation would most likely be up to @mborgerson

mborgerson:
The logic here is not right, but I've fixed it elsewhere and will merge the fix soon. I'll close this now. Thank you for the PR!

PR #437, opened 2021-07-31
Add the ability to remember window size across restarts.

7oxicshadow:
Fairly small change that users who prefer windowed mode usage over fullscreen might find useful.

Upon exit of xemu the current window size is captured and stored in the ini so that it can be recovered next time the emulator is restarted.

It has a default size of 1024 x 768 will be used as a fall back should the settings fail to load for any reason.

This will benefit from fixed aspect ratio support in a future pull request


ui/xemu-settings.c
```diff
@@ -119,6 +121,8 @@ struct config_offset_table {
 	[XEMU_SETTINGS_DISPLAY_SCALE]        = { CONFIG_TYPE_ENUM, "display", "scale",        offsetof(struct xemu_settings, scale),        { .default_int = DISPLAY_SCALE_SCALE }, display_scale_map },
 	[XEMU_SETTINGS_DISPLAY_UI_SCALE]     = { CONFIG_TYPE_INT,  "display", "ui_scale",     offsetof(struct xemu_settings, ui_scale),     { .default_int = 1                   }                    },
 	[XEMU_SETTINGS_DISPLAY_RENDER_SCALE] = { CONFIG_TYPE_INT,  "display", "render_scale", offsetof(struct xemu_settings, render_scale), { .default_int = 1                   }                    },
+	[XEMU_SETTINGS_DISPLAY_WIN_WIDTH] = { CONFIG_TYPE_INT,  "display", "win_width", offsetof(struct xemu_settings, win_width),    { .default_int = 1024                  }                    },
+	[XEMU_SETTINGS_DISPLAY_WIN_HEIGHT] = { CONFIG_TYPE_INT,  "display", "win_height", offsetof(struct xemu_settings, win_height), { .default_int = 768                   }                    },
```
dracc: Formatting; `=` sign is not alligned with the others.

7oxicshadow: should be resolved in the latest update

softmmu/vl.c
```diff
@@ -2700,7 +2700,11 @@ void qemu_init(int argc, char **argv, char **envp)
             break;
         }
     }
-    xemu_settings_load();
+
+    /* Loading the settings here has been moved in favor of parsing them before the
+       window is created so that window size can be maintained across reboots. */
+    /*xemu_settings_load();*/
```
mborgerson, line 2810: We should opt to remove the code here (not comment it out)

ui/xemu.c
```diff
@@ -840,9 +842,19 @@ static void sdl2_display_very_early_init(DisplayOptions *o)
 #endif
                                   , xemu_version);
 
+
+    int local_ini_win_width = 1024;
+    int local_ini_win_height = 768;
+
+    if(FALSE == xemu_settings_did_fail_to_load())
```
mborgerson: Settings api should always provide sane defaults here, don't need to hardcode it

ui/xemu.c
```diff
@@ -1538,3 +1552,8 @@ void xemu_load_disc(const char *path)
 
     xbox_smc_update_tray_state();
 }
+
+SDL_Window* xemu_get_winid(void)
+{
+    return(m_window);
```
mborgerson: nit: Formatting

ui/xemu-hud.cc
```diff
@@ -2023,6 +2023,7 @@ static void action_reset(void)
 
 static void action_shutdown(void)
 {
+    xemu_save_window_size();
```
mborgerson: Doesn't take into account ~~closing window~~, Ctrl-C, shutdown from monitor

ui/xemu.c
```diff
@@ -1538,3 +1552,8 @@ void xemu_load_disc(const char *path)
 
     xbox_smc_update_tray_state();
 }
+
+SDL_Window* xemu_get_winid(void)
```
mborgerson: nit: `*` goes next to function name

ui/xemu-hud.cc
```diff
@@ -2493,3 +2494,19 @@ void xemu_queue_error_message(const char *msg)
 {
     g_errors.push_back(strdup(msg));
 }
+
+void xemu_save_window_size(void)
+{
+    int win_h = 0;
+    int win_w = 0;
+
+    if(xemu_is_fullscreen() == FALSE)
```
mborgerson: nit: Formatting. Can be simply `!xemu_is_fullscreen()`

ui/xemu.c
```diff
@@ -840,9 +842,19 @@ static void sdl2_display_very_early_init(DisplayOptions *o)
...
+    if(FALSE == xemu_settings_did_fail_to_load())
+    {
+        xemu_settings_get_int(XEMU_SETTINGS_DISPLAY_WIN_WIDTH, &local_ini_win_width);
+        xemu_settings_get_int(XEMU_SETTINGS_DISPLAY_WIN_HEIGHT, &local_ini_win_height);
+    }
+
     // Create main window
     m_window = SDL_CreateWindow(
-        title, SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED, 1024, 768,
+        title, SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED, local_ini_win_width, local_ini_win_height,
```
mborgerson, line 878: Parameters are not validated (what happens with <= 0?)

7oxicshadow:
Updated to reflect changes requested from review.

The only curious one is the ctrl-c scenario. From what I can see, Xemu does not capture ctrl-c as it interprets it as a copy request (proven in the monitor window).

If ctrl-c did work I believe it would be captured anyway as SDL converts a ctrl-c into an SDL_QUIT event.

mborgerson:
I didn't mean Ctrl-C as a GUI shortcut for quitting, what I meant was Ctrl-C after launching from terminal will kill the app

How about updating the saved window size by handling the resize event in xemu.c (e.g. SDL_WINDOWEVENT_RESIZED)

binarymaster:
Looking forward for this one, really wanted feature for me.

7oxicshadow:
> Looking forward for this one, really wanted feature for me.

I did start updating this and then real life got in the way. I will try to sort it this weekend

7oxicshadow:
> I didn't mean Ctrl-C as a GUI shortcut for quitting, what I meant was Ctrl-C after launching from terminal will kill the app
> 
> How about updating the saved window size by handling the resize event in xemu.c (e.g. SDL_WINDOWEVENT_RESIZED)

Updated to use SDL_WINDOWEVENT_RESIZED as a trigger and as such have removed the functions/calls that are no longer required.

I decided to add a time element before performing a save as the SDL_WINDOWEVENT_RESIZED event happens 100's of times during a resize and I didn't want to spam the ini writes as this would be damaging for any devices in the future that might be using flash storage.

If you see no issue with the ini spam then I am happy to simplify this even further.

My original intention was to exploit the SDL_MOUSEBUTTONUP event to detect when the resize has finished but unfortunately SDL doesn't trigger this event after a releasing from a window resize.

mborgerson:
Thanks for the patch! I'll close this since it was finished up in #773

PR #438, opened 2021-07-31
Add 16:9 fixed ratio support

7oxicshadow:
Another small change to allow forced 16:9 screen ratio regardless of screen size


ui/xemu-hud.cc
```diff
@@ -2118,7 +2118,7 @@ static void ShowMainMenu()
                 nv2a_set_surface_scale_factor(rendering_scale+1);
             }
 
-            if (ImGui::Combo("Scaling Mode", &scaling_mode, "Center\0Scale\0Stretch\0")) {
+            if (ImGui::Combo("Scaling Mode", &scaling_mode, "Center\0Scale\0Stretch\0Override 16:9\0")) {
```
mborgerson: I think this would make more sense placed after `Scale` and simply stated as `Scale (Widescreen 16:9)`. Thoughts?

ui/xemu.c
```diff
@@ -1165,8 +1165,19 @@ void sdl2_gl_refresh(DisplayChangeListener *dcl)
         scale[0] = (float)tw/(float)ww;
         scale[1] = (float)th/(float)wh;
     } else {
-        // Scale to fit
-        float t_ratio = (float)tw/(float)th;
+        
+        //default the t_ratio variable to 4:3
+        float t_ratio = 1.333333333;
```
mborgerson: Assignment and comment not necessary, it will always be set below

ui/xemu-hud.cc
7oxicshadow: I agree. Updated

ui/xemu.c
7oxicshadow: Default value and comment removed. Thanks for review

mborgerson:
I fixed a couple minor formatting nits and rebased+squashed.

Thanks for the contribution!

PR #445, opened 2021-08-16
Make keys configurable

Zakru:
This PR adds a `[keyboard]` section to `xemu.ini` containing `key_button_*` and `key_axis_*` fields with keyboard scancode values. These configure the keyboard keys that map to the virtual controller. Default values are the old hardcoded keys.

Draft because untested.


Zakru:
I cannot build this myself because of technical issues with WSL.

GXTX:
You don't have a WSL issue, your changes are breaking.

Zakru:
@GXTX No, as in, the build needs Docker with WSL, but WSL literally does not work even after some troubleshooting and reinstalling -> no builds for me currently

GXTX:
It doesn't need docker, and your changes break - you're missing a header.

j8r:
> It doesn't need docker, and your changes break - you're missing a header.
You can be right, but he has no way to easily know that without testing.
@Zakru if WSL does not work for you, you can create a VM with something like VirtualBox, and make a shared directory.

j8r:
> It doesn't need docker, and your changes break - you're missing a header.

You can be right, but he has no way to easily know that without testing.
@Zakru if WSL does not work for you, you can create a VM with something like VirtualBox, and make a shared directory.

Bittermens:
I Suggest not only making Keys configurable but also add Mouse support for Right Analog serve as mouse aim, it will be handy for fps games and even the Starcraft Ghost dump.

octopoulos:
I just saw this PR. I was already working on a way to configure the buttons and keys, but mine has several advantages:
1) it can be configured independently for each player, whereas your changes are global => how do you have player1 vs player2 if both want to use the keyboard?
2) gamepad can be reconfigured too, not only the keyboard;
3) it can be done in the GUI, which means you don't need to know the name/code of the keys, which is crucial => you just click on the button to reconfigure then push the key or button and it will be used;
4) it doesn't create so many keys in the .ini file, only 2 keys per player (1 for the gamepad config, 1 for the keyboard config) => 8 extra lines.
5) last but not least, the rest of the code is almost unaffected, so it's a more streamlined solution, it doesn't need to add any new #define or variables, just 2 members in the ControllerState and that's it!

WIP:
image

MasonT8198:
@Zakru will you still be maintaining this? If not this should be closed.

MasonT8198:
> I just saw this PR. I was already working on a way to configure the buttons and keys, but mine has several advantages:
> 
> 1. it can be configured independently for each player, whereas your changes are global => how do you have player1 vs player2 if both want to use the keyboard?
> 2. gamepad can be reconfigured too, not only the keyboard;
> 3. it can be done in the GUI, which means you don't need to know the name/code of the keys, which is crucial => you just click on the button to reconfigure then push the key or button and it will be used;
> 4. it doesn't create so many keys in the .ini file, only 2 keys per player (1 for the gamepad config, 1 for the keyboard config) => 8 extra lines.
> 5. last but not least, the rest of the code is almost unaffected, so it's a more streamlined solution, it doesn't need to add any new #define or variables, just 2 members in the ControllerState and that's it!
> 
> WIP: image

@octopoulos would you be willing to join the discord and perhaps we could try this approach if you're willing to maintain it

PR #447, opened 2021-08-27
pgraph.c: Clear Surfaces and Shader Cache

MasonT8198:
This should finish the pgraph_destroy function and clear out the surface and shader cache properly when pgraph_destroy is called.


hw/xbox/nv2a/pgraph.c
```diff
@@ -1,4 +1,4 @@
-/*
+sudo apt update/*
```
GXTX: ?

MasonT8198: Hmm, I pasted in two places at the same time apparently. 2 am programming tends to extend horrible results

MasonT8198:
There may be a pointer that is being used incorrectly but it should be correct.

hw/xbox/nv2a/pgraph.c
```diff
@@ -3550,12 +3550,13 @@ void pgraph_destroy(PGRAPHState *pg)
...
-    // TODO: clear out surfaces
-
+    // Clear out surfaces
+    lru_flush(&pg->surfaces);
     glDeleteFramebuffers(1, &pg->gl_framebuffer);
 
-    // TODO: clear out shader cached
-
+    // Clear out shader cache
+    lru_flush(&pg->shader_cache);
```
mborgerson: shader_cache is not LRU type

hw/xbox/nv2a/pgraph.c
```diff
@@ -3550,12 +3550,13 @@ void pgraph_destroy(PGRAPHState *pg)
 
     glo_set_current(g_nv2a_context_render);
 
-    // TODO: clear out surfaces
-
+    // Clear out surfaces
+    lru_flush(&pg->surfaces);
```
mborgerson, line 3554: surfaces is not lru type

mborgerson (changes requested):
See mentioned

hw/xbox/nv2a/pgraph.c
MasonT8198: Excuse my ineptitude but what type is shader_cache

hw/xbox/nv2a/pgraph.c
MasonT8198: What type is surfaces

Googling didn't seem to get me anywhere meaningful

hw/xbox/nv2a/pgraph.c
GXTX: What type is `pg`? And you'll find your answer.

mborgerson:
Let's return to this with a follow-up PR

Issue #462, opened 2021-09-29
OS, CPU model not shown in log file


mborgerson:
### Feature Request

These data should be shown.

PR #463, opened 2021-09-30
xemu.c: Print OS Version to Log

MasonT8198:
Partially solves #462


MasonT8198:
I’ll get to cpu model later, I decided to try and support arm cpus, but that’ll be a separate PR

mborgerson:
Thanks!

PR #474, opened 2021-10-06
nv2a: Set frequency to 233MHz

GXTX:
No clue about the coefficients and such but the frequency is supposed to be 233MHz according to `XBOverclock`. This might not even matter if it gets set in `pramdac_write` correctly.

Unrelated has anyone seen if nxdk apps work recently? I get a black screen, or a broken Xbox logo, and if I reset I can see what was supposed to be shown as if the surface wasn't updated.

### DIFF
--- a/hw/xbox/nv2a/nv2a.c
+++ b/hw/xbox/nv2a/nv2a.c
@@ -306,8 +306,8 @@ static void nv2a_reset(NV2AState *d)
     memset(d->pgraph.regs, 0, sizeof(d->pgraph.regs));
 
     d->pcrtc.start = 0;
-    d->pramdac.core_clock_coeff = 0x00011c01; /* 189MHz...? */
-    d->pramdac.core_clock_freq = 189000000;
+    d->pramdac.core_clock_coeff = 0x00011C01; /* 189MHz...? */
+    d->pramdac.core_clock_freq = 233333324;
     d->pramdac.memory_clock_coeff = 0;
     d->pramdac.video_clock_coeff = 0x0003C20D; /* 25182Khz...? */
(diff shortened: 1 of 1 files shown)


MasonT8198:
> Unrelated has anyone seen if nxdk apps work recently? I get a black screen, or a broken Xbox logo, and if I reset I can see what was supposed to be shown as if the surface wasn't updated.

I haven't had this issue no, I've also asked a few other people, what bios and mcpx combo are you using?

hw/xbox/nv2a/nv2a.c
```diff
@@ -307,7 +307,7 @@ static void nv2a_reset(NV2AState *d)
 
     d->pcrtc.start = 0;
     d->pramdac.core_clock_coeff = 0x00011c01; /* 189MHz...? */
-    d->pramdac.core_clock_freq = 189000000;
+    d->pramdac.core_clock_freq = 233000000;
```
mborgerson: According to the expression in `pramdac_write` a `core_clock_coeff` of `0x11c01` actually gives 233333324. Let's update this value to be consistent

mborgerson (changes requested):
See inline

hw/xbox/nv2a/nv2a.c
GXTX: Fixed

mborgerson:
Thanks!

GXTX:
Coming back to this... I've been playing around with "overclocking" xemu, I don't think we should actually be setting these in reset, and if we do we should probably just zero the values. The x-codes should be what's setting this.

PR #495, opened 2021-10-13
xbox: Allow manually setting SMC version

GXTX:


MasonT8198:
You should be able to rerun your checks now, build works, macos runner was broken for a bit

PR #499, opened 2021-10-17
Improve Unicode filename handling on Windows

### LINKED ISSUE #479: File paths containing non-ASCII characters are unsupported
### Bug Description

If file path contains non-ascii characters (e.g. Cyrillic), the user interface will not display the characters properly (glyphs are not in the font) and, if the user's name contains such characters creating xemu.ini file and EEPROM file will fail with default paths. This is a major usability issue.

The path management issues for config and EEPROM appear possibly localized to Windows, cannot repro on Linux. macOS state unknown. Unclear if this is a bug in SDL2 or xemu. Path rendering affects all platforms.

A workaround is to install xemu to C:\xemu, and create C:\xemu\xemu.ini manually.

Issue has been reported multiple times. #99 #470 #476

### Expected Behavior

xemu should work on all platforms, regardless of locale, and displaying paths correctly.

### xemu Version

v0.6.1-12-ga8c73197da

### System Information

N/A

### Additional Context

_No response_

CookiePLMonster:
All identified UTF-8/Unicode issues have been fixed, now xemu can:
* Run from a directory with Unicode characters
* Load/save settings to a directory with Unicode characters (typically non-ASCII user names)
* Load BIOS/hard disc from directories with Unicode characters
* Run games from directories with Unicode characters

image
Here it's been tested with nonsense paths, but in practice this means that if those paths work, then e.g. paths with Russian, Japanese or Chinese characters will also work fine.

There may be more places that need fixing, but basing on my testing of the above, all known broken cases are now handled properly.

Fixes #479

### DIFF
--- a/block/file-win32.c
+++ b/block/file-win32.c
@@ -337,6 +337,7 @@ static int raw_open(BlockDriverState *bs, QDict *options, int flags,
     QemuOpts *opts;
     Error *local_err = NULL;
     const char *filename;
+    wchar_t *wfilename;
     bool use_aio;
     OnOffAuto locking;
     int ret;
@@ -391,9 +392,14 @@ static int raw_open(BlockDriverState *bs, QDict *options, int flags,
         snprintf(s->drive_path, sizeof(s->drive_path), "%c:\\", buf[0]);
     }
 
-    s->hfile = CreateFile(filename, access_flags,
+    wfilename = g_utf8_to_utf16(filename, -1, NULL, NULL, NULL);
+    if (!filename) {
+        goto fail;
+    }
+    s->hfile = CreateFileW(wfilename, access_flags,
                           FILE_SHARE_READ, NULL,
                           OPEN_EXISTING, overlapped, NULL);
+    g_free(wfilename);
     if (s->hfile == INVALID_HANDLE_VALUE) {
         int err = GetLastError();
 
@@ -556,27 +562,35 @@ static int64_t raw_getlength(BlockDriverState *bs)
 
 static int64_t raw_get_allocated_file_size(BlockDriverState *bs)
 {
-    typedef DWORD (WINAPI * get_compressed_t)(const char *filename,
+    typedef DWORD (WINAPI * get_compressed_t)(const wchar_t *filename,
                                               DWORD * high);
+    int64_t size = -1;
     get_compressed_t get_compressed;
     struct _stati64 st;
-    const char *filename = bs->filename;
+    wchar_t *wfilename = g_utf8_to_utf16(bs->filename, -1, NULL, NULL, NULL);
+    if (!wfilename) {
+        goto done;
+    }
     /* WinNT support GetCompressedFileSize to determine allocate size */
     get_compressed =
         (get_compressed_t) GetProcAddress(GetModuleHandle("kernel32"),
-                                            "GetCompressedFileSizeA");
+                                            "GetCompressedFileSizeW");
     if (get_compressed) {
         DWORD high, low;
-        low = get_compressed(filename, &high);
+        low = get_compressed(wfilename, &high);
         if (low != 0xFFFFFFFFlu || GetLastError() == NO_ERROR) {
-            return (((int64_t) high) << 32) + low;
+            size = (((int64_t) high) << 32) + low;
+            goto done;
         }
     }
 
-    if (_stati64(filename, &st) < 0) {
-        return -1;
+    if (_wstati64(wfilename, &st) >= 0) {
+        size = st.st_size;
     }
-    return st.st_size;
+
+done:
+    g_free(wfilename);
+    return size;
 }
 
 static int raw_co_create(BlockdevCreateOptions *options, Error **errp)
@@ -752,6 +766,7 @@ static int hdev_open(BlockDriverState *bs, QDict *options, int flags,
 
     Error *local_err = NULL;
     const char *filename;
+    wchar_t *wfilename;
     bool use_aio;
 
     QemuOpts *opts = qemu_opts_create(&raw_runtime_opts, NULL, 0,
@@ -795,9 +810,14 @@ static int hdev_open(BlockDriverState *bs, QDict *options, int flags,
 
     create_flags = OPEN_EXISTING;
 
-    s->hfile = CreateFile(filename, access_flags,
-                          FILE_SHARE_READ, NULL,
-                          create_flags, overlapped, NULL);
+    wfilename = g_utf8_to_utf16(filename, -1, NULL, NULL, NULL);
+    if (!filename) {
+        goto done;
+    }
+    s->hfile = CreateFileW(wfilename, access_flags,
+                           FILE_SHARE_READ, NULL,
+                           create_flags, overlapped, NULL);
+    g_free(wfilename);
     if (s->hfile == INVALID_HANDLE_VALUE) {
         int err = GetLastError();
 
--- a/hw/core/loader.c
+++ b/hw/core/loader.c
@@ -70,7 +70,7 @@ int64_t get_image_size(const char *filename)
 {
     int fd;
     int64_t size;
-    fd = open(filename, O_RDONLY | O_BINARY);
+    fd = qemu_open(filename, O_RDONLY | O_BINARY, NULL);
     if (fd < 0)
         return -1;
     size = lseek(fd, 0, SEEK_END);
@@ -84,7 +84,7 @@ ssize_t load_image_size(const char *filename, void *addr, size_t size)
     int fd;
     ssize_t actsize, l = 0;
 
-    fd = open(filename, O_RDONLY | O_BINARY);
+    fd = qemu_open(filename, O_RDONLY | O_BINARY, NULL);
     if (fd < 0) {
         return -1;
     }
@@ -229,7 +229,7 @@ int load_aout(const char *filename, hwaddr addr, int max_sz,
     struct exec e;
     uint32_t magic;
 
-    fd = open(filename, O_RDONLY | O_BINARY);
+    fd = qemu_open(filename, O_RDONLY | O_BINARY, NULL);
     if (fd < 0)
         return -1;
 
@@ -359,7 +359,7 @@ void load_elf_hdr(const char *filename, void *hdr, bool *is64, Error **errp)
     }
     e_ident = hdr;
 
-    fd = open(filename, O_RDONLY | O_BINARY);
+    fd = qemu_open(filename, O_RDONLY | O_BINARY, NULL);
     if (fd < 0) {
         error_setg_errno(errp, errno, "Failed to open file: %s", filename);
         return;
@@ -456,7 +456,7 @@ int load_elf_ram_sym(const char *filename,
     int fd, data_order, target_data_order, must_swab, ret = ELF_LOAD_FAILED;
     uint8_t e_ident[EI_NIDENT];
 
-    fd = open(filename, O_RDONLY | O_BINARY);
+    fd = qemu_open(filename, O_RDONLY | O_BINARY, NULL);
     if (fd < 0) {
         perror(filename);
         return -1;
@@ -613,7 +613,7 @@ static int load_uboot_image(const char *filename, hwaddr *ep, hwaddr *loadaddr,
     int ret = -1;
     int do_uncompress = 0;
 
-    fd = open(filename, O_RDONLY | O_BINARY);
+    fd = qemu_open(filename, O_RDONLY | O_BINARY, NULL);
     if (fd < 0)
         return -1;
 
@@ -962,7 +962,7 @@ int rom_add_file(const char *file, const char *fw_dir,
         rom->path = g_strdup(file);
     }
 
-    fd = open(rom->path, O_RDONLY | O_BINARY);
+    fd = qemu_open(rom->path, O_RDONLY | O_BINARY, NULL);
     if (fd == -1) {
         fprintf(stderr, "Could not open option rom '%s': %s\n",
                 rom->path, strerror(errno));
--- a/include/qemu/osdep.h
+++ b/include/qemu/osdep.h
@@ -593,6 +593,21 @@ int qemu_lock_fd_test(int fd, int64_t start, int64_t len, bool exclusive);
 bool qemu_has_ofd_lock(void);
 #endif
 
+#ifdef _WIN32
+FILE *qemu_fopen(const char *filename, const char *mode);
+int qemu_access(const char *pathname, int mode);
+#else
+static inline FILE *qemu_fopen(const char *filename, const char *mode)
+{
+    return fopen(filename, mode);
+}
+
+static inline int qemu_access(const char *pathname, int mode)
+{
+    return access(pathname, mode);
+}
+#endif
+
 #if defined(__HAIKU__) && defined(__i386__)
 #define FMT_pid "%ld"
 #elif defined(WIN64)
--- a/ui/noc_file_dialog.h
+++ b/ui/noc_file_dialog.h
@@ -164,48 +164,91 @@ const char *noc_file_dialog_open(int flags,
 
 #include <windows.h>
 #include <commdlg.h>
+#include <glib.h>
+#include <stdio.h>
 
 const char *noc_file_dialog_open(int flags,
                                  const char *filters,
                                  const char *default_path,
                                  const char *default_name)
 {
-    OPENFILENAME ofn;       // common dialog box structure
-    char szFile[_MAX_PATH]; // buffer for file name
-    char initialDir[_MAX_PATH];
-    char drive[_MAX_DRIVE];
-    char dir[_MAX_DIR];
-    char fname[_MAX_FNAME];
-    char ext[_MAX_EXT];
-    int ret;
+    OPENFILENAMEW ofn;         // common dialog box structure
+    wchar_t szFile[_MAX_PATH]; // buffer for file name
+    wchar_t initialDir[_MAX_PATH];
+    wchar_t drive[_MAX_DRIVE];
+    wchar_t dir[_MAX_DIR];
+    wchar_t fname[_MAX_FNAME];
+    wchar_t ext[_MAX_EXT];
+    int ret = 0;
+    wchar_t *wfilters = NULL;
+    wchar_t *wdefault_path = NULL;
+    wchar_t *wdefault_name = NULL;
+    size_t filters_length = 0;
+
+    if (filters) {
+        // 'filters' is a null-terminated list of null-terminated strings,
+        // so the buffer length must be provided explicitly
+        while (filters[filters_length]) {
+            filters_length += strlen(filters + filters_length) + 1;
+        }
+        wfilters = (wchar_t*)g_convert(filters, filters_length + 1, "UTF-16", "UTF-8", NULL, NULL, NULL);
+        if (!wfilters) {
+            fprintf(stderr, "Failed to convert UTF-8 string to UTF-16\n");
+            goto done;
+        }
+    }
+    if (default_path) {
+        wdefault_path = g_utf8_to_utf16(default_path, -1, NULL, NULL, NULL);
+        if (!wdefault_path) {
+            fprintf(stderr, "Failed to convert UTF-8 string to UTF-16\n");
+            goto done;
+        }
+    }
+    if (default_name) {
+        wdefault_name = g_utf8_to_utf16(default_name, -1, NULL, NULL, NULL);
+        if (!wdefault_name) {
+            fprintf(stderr, "Failed to convert UTF-8 string to UTF-16\n");
+            goto done;
+        }
+    }
 
     // init default dir and file name
-    _splitpath_s(default_path, drive, _MAX_DRIVE, dir, _MAX_DIR, fname,
-                 _MAX_FNAME, ext, _MAX_EXT );
-    _makepath_s(initialDir, _MAX_PATH, drive, dir, NULL, NULL);
-    if (default_name)
-        strncpy(szFile, default_name, sizeof(szFile) - 1);
+    _wsplitpath_s(wdefault_path, drive, G_N_ELEMENTS(drive), dir, G_N_ELEMENTS(dir), fname,
+                 G_N_ELEMENTS(fname), ext, G_N_ELEMENTS(ext) ); 
+    _wmakepath_s(initialDir, G_N_ELEMENTS(initialDir), drive, dir, NULL, NULL);
+    if (wdefault_name)
+        wcscpy_s(szFile, G_N_ELEMENTS(szFile), wdefault_name);
     else
-        _makepath_s(szFile, _MAX_PATH, NULL, NULL, fname, ext);
+        _wmakepath_s(szFile, G_N_ELEMENTS(szFile), NULL, NULL, fname, ext);
 
     ZeroMemory(&ofn, sizeof(ofn));
     ofn.lStructSize = sizeof(ofn);
     ofn.lpstrFile = szFile;
-    ofn.nMaxFile = sizeof(szFile);
-    ofn.lpstrFilter = filters;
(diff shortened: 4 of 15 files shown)


hw/core/loader.c
```diff
@@ -359,9 +359,9 @@ void load_elf_hdr(const char *filename, void *hdr, bool *is64, Error **errp)
     }
     e_ident = hdr;
 
-    fd = open(filename, O_RDONLY | O_BINARY);
+    fd = open_utf8(filename, O_RDONLY | O_BINARY);
     if (fd < 0) {
-        error_setg_errno(errp, errno, "Failed to open file: %s", filename);
+        error_setg_errno(errp, errno, "Failed to open_utf8 file: %s", filename);
```
PatrickvL: Perhaps this replacement of `open` by `open_utf8` is not necessary?

hw/core/loader.c
```diff
@@ -962,9 +962,9 @@ int rom_add_file(const char *file, const char *fw_dir,
         rom->path = g_strdup(file);
     }
 
-    fd = open(rom->path, O_RDONLY | O_BINARY);
+    fd = open_utf8(rom->path, O_RDONLY | O_BINARY);
     if (fd == -1) {
-        fprintf(stderr, "Could not open option rom '%s': %s\n",
+        fprintf(stderr, "Could not open_utf8 option rom '%s': %s\n",
```
PatrickvL: Perhaps this replacement of `open` by `open_utf8` is not necessary?

PatrickvL (changes requested):
A few suggestions

hw/core/loader.c
CookiePLMonster: Good catch on this and on the above, a Find & Replace fail.

GXTX:
I have no real review for this. It seems there are issues with base QEMU that are also being fixed. Can this be upstreamed?

block/file-win32.c
```diff
@@ -337,6 +337,7 @@ static int raw_open(BlockDriverState *bs, QDict *options, int flags,
     QemuOpts *opts;
     Error *local_err = NULL;
     const char *filename;
+    wchar_t* wfilename;
```
mborgerson: nit: `*` goes next to name

block/file-win32.c
```diff
@@ -752,6 +766,7 @@ static int hdev_open(BlockDriverState *bs, QDict *options, int flags,
 
     Error *local_err = NULL;
     const char *filename;
+    wchar_t* wfilename;
```
mborgerson: nit: `*` goes next to identifier

block/file-win32.c
```diff
@@ -795,9 +810,14 @@ static int hdev_open(BlockDriverState *bs, QDict *options, int flags,
 
     create_flags = OPEN_EXISTING;
 
-    s->hfile = CreateFile(filename, access_flags,
+    wfilename = g_utf8_to_utf16(filename, -1, NULL, NULL, NULL);
+    if (!filename) {
+        goto done;
+    }
+    s->hfile = CreateFileW(wfilename, access_flags,
                           FILE_SHARE_READ, NULL,
```
mborgerson: nit: Indentation

ui/noc_file_dialog.h
```diff
@@ -164,48 +164,83 @@ const char *noc_file_dialog_open(int flags,
...
+    wchar_t* wdefault_name = NULL;
+    size_t filters_length = 0;
+
+    if (filters) {
+        // 'filters' is a null-terminated list of null-terminated strings,
+        // so the buffer length must be provided explicitly
+        while (filters[filters_length]) {
+            filters_length += strlen(filters + filters_length) + 1;
+        }
+        wfilters = (wchar_t*)g_convert(filters, filters_length + 1, "UTF-16", "UTF-8", NULL, NULL, NULL);
```
mborgerson, line 194: * Why the use of `g_convert` here, and `g_utf8_to_utf16` elsewhere?
* Can you confirm this handles all string terminating `\x00` correctly for number of strings > 1
* If conversion fails, `wfilters` and other converted strings will be `NULL` and no filters will be displayed. I think in this case an error should be generated.

ui/noc_file_dialog.h
```diff
@@ -164,48 +164,83 @@ const char *noc_file_dialog_open(int flags,
...
+        ret = GetSaveFileNameW(&ofn);
+
+    g_free(wdefault_name);
+    g_free(wdefault_path);
+    g_free(wfilters);
 
     free(g_noc_file_dialog_ret);
-    g_noc_file_dialog_ret = ret ? strdup(szFile) : NULL;
+    if (ret)
+    {
```
mborgerson: nit: braces goes on line with `if`, `else` keywords

util/oslib-win32.c
```diff
@@ -813,11 +819,19 @@ bool qemu_write_pidfile(const char *filename, Error **errp)
     HANDLE file;
     OVERLAPPED overlap;
     BOOL ret;
+    wchar_t* wfilename;
```
mborgerson: nit: `*` goes next to identifier

util/oslib-win32.c
```diff
@@ -813,11 +819,19 @@ bool qemu_write_pidfile(const char *filename, Error **errp)
...
     memset(&overlap, 0, sizeof(overlap));
 
-    file = CreateFile(filename, GENERIC_WRITE, FILE_SHARE_READ, NULL,
+    wfilename = g_utf8_to_utf16(filename, -1, NULL, NULL, NULL);
+    if (!wfilename) {
+        error_setg(errp, "Failed to create PID file");
```
mborgerson: Error message should be different, or these should be grouped below like other error handlers

util/oslib-win32.c
```diff
@@ -856,3 +870,65 @@ size_t qemu_get_host_physmem(void)
...
     return 0;
 }
+
+int open_utf8(const char *filename, int oflag)
+{
+    wchar_t *wfilename = NULL;
+    int fd = -1;
+
+    wfilename = g_utf8_to_utf16(filename, -1, NULL, NULL, NULL);
+    if (!wfilename)
```
mborgerson: nit: Braces should be on line with `if`, `else`

util/oslib-win32.c
```diff
@@ -856,3 +870,65 @@ size_t qemu_get_host_physmem(void)
     }
     return 0;
 }
+
+int open_utf8(const char *filename, int oflag)
+{
+    wchar_t *wfilename = NULL;
```
mborgerson: Initialization of variable unnecessary

util/oslib-win32.c
```diff
@@ -856,3 +870,65 @@ size_t qemu_get_host_physmem(void)
...
+    fd = _wopen(wfilename, oflag);
+
+done:
+    g_free(wfilename);
+    return fd;
+}
+
+FILE *fopen_utf8(const char *filename, const char *mode)
+{
+    wchar_t *wfilename = NULL;
```
mborgerson: Initialization of variable unnecessary

util/oslib-win32.c
```diff
@@ -856,3 +870,65 @@ size_t qemu_get_host_physmem(void)
...
+    wchar_t *wmode = NULL;
+    FILE *file = NULL;
+
+    wfilename = g_utf8_to_utf16(filename, -1, NULL, NULL, NULL);
+    if (!wfilename)
+    {
```
mborgerson: nit: Braces

util/oslib-win32.c
```diff
@@ -856,3 +870,65 @@ size_t qemu_get_host_physmem(void)
...
+        goto done;
+    }
+
+    ret = _waccess(wpathname, mode);
+
+done:
+    g_free(wpathname);
+    return ret;
+}
```
mborgerson: nit: Newline here

util/oslib-win32.c
```diff
@@ -856,3 +870,65 @@ size_t qemu_get_host_physmem(void)
...
+    return file;
+}
+
+int access_utf8(const char *pathname, int mode)
+{
+    wchar_t *wpathname = NULL;
+    int ret = -1;
+
+    wpathname = g_utf8_to_utf16(pathname, -1, NULL, NULL, NULL);
+    if (!wpathname)
```
mborgerson: nit: Braces

CookiePLMonster:
> Can this be upstreamed?

Technically it probably can, but from what I know qemu's barrier of entry for contributions is rather steep and I'm not sure I want to deal with the entire ritual right now.

simi:
@CookiePLMonster if you can prepare separate patch against QEMU only, I can try to help contribute this to QEMU directly.

CookiePLMonster:
> @CookiePLMonster if you can prepare separate patch against QEMU only, I can try to help contribute this to QEMU directly.

Sounds good! Let's wait until this is reviewed though to avoid going back with possible additional fixes.

hw/core/loader.c
```diff
@@ -70,7 +70,7 @@ int64_t get_image_size(const char *filename)
 {
     int fd;
     int64_t size;
-    fd = open(filename, O_RDONLY | O_BINARY);
+    fd = open_utf8(filename, O_RDONLY | O_BINARY);
```
mborgerson: Hmm.. Preferably all `open`, `fopen`, `access`, etc. would be replaced with `qemu_*` prefix versions and the encoding conversion would be handled there.

This is okay for now, but something I'd like to see refactored. I really want to avoid having the code base mixing `X` and `X_utf8` variants in different spots, and using the prefixed version would make it easier to detect usages of these encoding issue-prone versions.

If you are planning to send a patch to upstream QEMU, I suggest doing this.

ui/noc_file_dialog.h
```diff
@@ -164,48 +164,83 @@ const char *noc_file_dialog_open(int flags,
...
                                  const char *default_name)
 {
-    OPENFILENAME ofn;       // common dialog box structure
-    char szFile[_MAX_PATH]; // buffer for file name
-    char initialDir[_MAX_PATH];
-    char drive[_MAX_DRIVE];
-    char dir[_MAX_DIR];
-    char fname[_MAX_FNAME];
-    char ext[_MAX_EXT];
+    OPENFILENAMEW ofn;       // common dialog box structure
```
mborgerson: nit: comment alignment

ui/noc_file_dialog.h
```diff
@@ -164,48 +164,83 @@ const char *noc_file_dialog_open(int flags,
...
-    if (default_name)
-        strncpy(szFile, default_name, sizeof(szFile) - 1);
+    _wsplitpath_s(wdefault_path, drive, G_N_ELEMENTS(drive), dir, G_N_ELEMENTS(dir), fname,
+                 G_N_ELEMENTS(fname), ext, G_N_ELEMENTS(ext) ); 
+    _wmakepath_s(initialDir, _MAX_PATH, drive, dir, NULL, NULL);
+    if (wdefault_name)
+        wcscpy_s(szFile, G_N_ELEMENTS(szFile), wdefault_name);
     else
-        _makepath_s(szFile, _MAX_PATH, NULL, NULL, fname, ext);
+        _wmakepath_s(szFile, _MAX_PATH, NULL, NULL, fname, ext);
```
mborgerson: Why have some buffer sizes been replaced with `G_N_ELEMENTS(buffer)` element count wrappers and others remain to be their original definition constant value (e.g. `_MAX_PATH`)?

ui/noc_file_dialog.h
```diff
@@ -164,48 +164,83 @@ const char *noc_file_dialog_open(int flags,
...
+    g_free(wdefault_name);
+    g_free(wdefault_path);
+    g_free(wfilters);
 
     free(g_noc_file_dialog_ret);
-    g_noc_file_dialog_ret = ret ? strdup(szFile) : NULL;
+    if (ret)
+    {
+        gszFile = g_utf16_to_utf8(szFile, -1, NULL, NULL, NULL);
+        g_noc_file_dialog_ret = strdup(gszFile);
```
mborgerson: Why duplicate and free the converted string? We should already have an allocation to keep (`free` above will need to be `g_free`)

util/oslib-win32.c
```diff
@@ -294,17 +294,23 @@ char *
 qemu_get_local_state_pathname(const char *relative_pathname)
 {
     HRESULT result;
-    char base_path[MAX_PATH+1] = "";
+    char* ret;
```
mborgerson: nit: `*` goes next to identifier

util/oslib-posix.c
```diff
@@ -858,3 +858,18 @@ size_t qemu_get_host_physmem(void)
 #endif
     return 0;
 }
+
+int open_utf8(const char *filename, int oflag)
```
mborgerson: Prefer these passthru functions to be in header

util/oslib-win32.c
```diff
@@ -294,17 +294,23 @@ char *
...
+    wchar_t wbase_path[MAX_PATH] = L"";
+    char* base_path;
 
-    result = SHGetFolderPath(NULL, CSIDL_COMMON_APPDATA, NULL,
-                             /* SHGFP_TYPE_CURRENT */ 0, base_path);
+    result = SHGetFolderPathW(NULL, CSIDL_COMMON_APPDATA, NULL,
+                             /* SHGFP_TYPE_CURRENT */ 0, wbase_path);
```
mborgerson: nit: param alignment

util/oslib-win32.c
```diff
@@ -294,17 +294,23 @@ char *
...
     if (result != S_OK) {
         /* misconfigured environment */
         g_critical("CSIDL_COMMON_APPDATA unavailable: %ld", (long)result);
         abort();
     }
-    return g_strdup_printf("%s" G_DIR_SEPARATOR_S "%s", base_path,
+    base_path = g_utf16_to_utf8(wbase_path, -1, NULL, NULL, NULL);
```
mborgerson, line 308: What if conversion fails?

ui/noc_file_dialog.h
```diff
@@ -164,48 +164,83 @@ const char *noc_file_dialog_open(int flags,
...
+    OPENFILENAMEW ofn;       // common dialog box structure
+    wchar_t szFile[_MAX_PATH]; // buffer for file name
+    wchar_t initialDir[_MAX_PATH];
+    wchar_t drive[_MAX_DRIVE];
+    wchar_t dir[_MAX_DIR];
+    wchar_t fname[_MAX_FNAME];
+    wchar_t ext[_MAX_EXT];
     int ret;
+    char* gszFile;
+    wchar_t* wfilters = NULL;
```
mborgerson: nit: `*` placements

util/oslib-win32.c
```diff
@@ -350,8 +356,8 @@ void qemu_init_exec_dir(const char *argv0)
         p--;
```
mborgerson, line 361: Comparison with path terminator `'\\'` should probably have `L` prefix

util/oslib-win32.c
```diff
@@ -331,15 +337,15 @@ static const char *exec_dir;
...
+    wchar_t *p;
+    wchar_t buf[MAX_PATH];
     DWORD len;
 
     if (exec_dir) {
         return;
     }
 
-    len = GetModuleFileName(NULL, buf, sizeof(buf) - 1);
+    len = GetModuleFileNameW(NULL, buf, G_N_ELEMENTS(buf) - 1);
```
mborgerson, line 353: Apparently it is possible to have a path longer than `MAX_PATH`, in which case the module path may be truncated. This consideration also complicates below `MAX_PATH` cases--although, we can leave it for follow up as its an existing issue.

util/oslib-win32.c
```diff
@@ -856,3 +870,65 @@ size_t qemu_get_host_physmem(void)
...
+
+done:
+    g_free(wfilename);
+    g_free(wmode);
+    return file;
+}
+
+int access_utf8(const char *pathname, int mode)
+{
+    wchar_t *wpathname = NULL;
```
mborgerson: Initialization unnecessary

ui/xemu-settings.c
```diff
@@ -24,6 +24,7 @@
 #include <stdbool.h>
 #include <assert.h>
 #include <glib.h>
+#include "qemu/osdep.h"
```
mborgerson: osdep.h should come first in includes

ui/noc_file_dialog.h
```diff
@@ -164,48 +164,83 @@ const char *noc_file_dialog_open(int flags,
 
 #include <windows.h>
 #include <commdlg.h>
+#include <glib.h>
+#include <stdio.h>
```
mborgerson, line 168: Why is stdio.h added?

ui/noc_file_dialog.h
CookiePLMonster: `g_convert` handles embedded nulls, `g_utf8_to_utf16` does not (even when you specify length explicitly).

hw/core/loader.c
CookiePLMonster: The reason I decided against it is because `qemu_open` already means something else and it doesn't even return a fd or `FILE*`. If you have other ideas for a function name, I'll happily rename those.

ui/noc_file_dialog.h
CookiePLMonster: Looks like an oversight.

ui/noc_file_dialog.h
CookiePLMonster: If the above is the only place where `g_noc_file_dialog_ret` is freed, then I can simplify this. I just didn't want to mix the allocators.

util/oslib-posix.c
CookiePLMonster: Is there a posix specific oslib header?

ui/noc_file_dialog.h
CookiePLMonster: Leftover from an earlier test, will remove.

util/oslib-win32.c
CookiePLMonster: In this case, I'd like to keep it for consistency with `wmode` that must be initialized.

util/oslib-win32.c
CookiePLMonster: It's generally very annoying to account for that in `GetModuleFileName`, since the usual "call the first time to get the buffer size, then again to fill the buffer" idiom of WinAPI does not apply to this function.

hw/core/loader.c
mborgerson: > The reason I decided against it is because `qemu_open` already means something else and it doesn't even return a fd or `FILE*`.

`qemu_open` is a thin wrapper around `open`, and returns a file descriptor. See util/osdep.c

ui/noc_file_dialog.h
```diff
@@ -164,48 +164,77 @@ const char *noc_file_dialog_open(int flags,
...
-    g_noc_file_dialog_ret = ret ? strdup(szFile) : NULL;
+    g_free(wdefault_name);
+    g_free(wdefault_path);
+    g_free(wfilters);
+
+    g_free(g_noc_file_dialog_ret);
+    if (ret) {
+        g_noc_file_dialog_ret = g_utf16_to_utf8(szFile, -1, NULL, NULL, NULL);
+    }
+    else {
```
mborgerson: nit: `else {` goes on line with preceding bracket

ui/noc_file_dialog.h
mborgerson: > `g_convert` handles embedded nulls, `g_utf8_to_utf16` does not (even when you specify length explicitly).

Interesting; thanks for clarifying.

Did you consider my other feedback regarding possible failed conversion?

util/oslib-posix.c
mborgerson: You can use the header that you've defined them in already, with a macro for Windows platform detection.

ui/noc_file_dialog.h
CookiePLMonster: I'm not sure how to gracefully display an error in this context? If you have an example of that, I'll amend.

hw/core/loader.c
CookiePLMonster: You're right, I must've missed it. Do you think in this case I should modify `qemu_open` to handle UTF-8 gracefully and also add `qemu_fopen` that calls `qemu_open` + `fdopen` internally?

util/oslib-win32.c
mborgerson: In this case the approach could be to call `GetModuleFileName` with a reasonable buffer size, then double the buffer size and retry every time it throws a truncation error. Anyway, it's a problem to be solved later.

util/oslib-win32.c
```diff
@@ -856,3 +876,62 @@ size_t qemu_get_host_physmem(void)
...
+    FILE *file = NULL;
+
+    wfilename = g_utf8_to_utf16(filename, -1, NULL, NULL, NULL);
+    if (!wfilename) {
+        goto done;
+    }
+    
+    wmode = g_utf8_to_utf16(mode, -1, NULL, NULL, NULL);
+    if (!wmode)
+    {
```
mborgerson: nit: Brace should be moved up

util/oslib-win32.c
mborgerson: I think it is clear in the code, as the assignment happens 3 lines later. Not a huge deal, just redundant

hw/core/loader.c
mborgerson: That would be fine

ui/noc_file_dialog.h
mborgerson: How about we simply print a message to `stderr` then return NULL

util/oslib-win32.c
```diff
@@ -294,17 +294,29 @@ char *
...
+                              /* SHGFP_TYPE_CURRENT */ 0, wbase_path);
     if (result != S_OK) {
         /* misconfigured environment */
         g_critical("CSIDL_COMMON_APPDATA unavailable: %ld", (long)result);
         abort();
     }
-    return g_strdup_printf("%s" G_DIR_SEPARATOR_S "%s", base_path,
+    base_path = g_utf16_to_utf8(wbase_path, -1, NULL, NULL, NULL);
+    if (!base_path) {
+        /* this shouldn't happen, so something is very wrong if it does */
```
mborgerson: Comment isn't necessary (aborting is unusual)

util/oslib-win32.c
CookiePLMonster: I agree, let's solve it in a separate PR.

include/qemu/osdep.h
```diff
@@ -593,6 +593,21 @@ int qemu_lock_fd_test(int fd, int64_t start, int64_t len, bool exclusive);
 bool qemu_has_ofd_lock(void);
 #endif
 
+#ifdef _WIN32
+FILE *qemu_fopen(const char *filename, const char *mode);
+int qemu_access(const char *pathname, int mode);
+#else
+static FILE *qemu_fopen(const char *filename, const char *mode)
```
mborgerson: Let's add the `inline` keyword here to silence compiler complaining about unused static functions. Same with below function.

util/osdep.c
```diff
@@ -292,6 +292,26 @@ int qemu_lock_fd_test(int fd, int64_t start, int64_t len, bool exclusive)
...
     int ret;
+#ifdef _WIN32
+    wchar_t *wname;
+
+    wname = g_utf8_to_utf16(name, -1, NULL, NULL, NULL);
+    if (!wname) {
+        return -1;
+    }
+
+#ifdef O_CLOEXEC
```
mborgerson: We can simplify this function by moving the path,mode encoding into a thin open wrapper function (just static define it above this function), then we can keep original `qemu_open_cloexec` logic as it was before, but replace its `open` call with your new encoding handle version of `open`.

mborgerson (changes requested):

mborgerson:
Thank you for this contribution!

I have amended your original commit message to more closely align with repo conventions and make more explicit the changes of the patch.

PR #510, opened 2021-10-29
cpu: Makes cpuid match hardware behavior.

### LINKED ISSUE #509: Return To Castle Wolfenstein: Crash in VM due to null pointer exception, possibly triggered by difference in `cpuid` handling
### Title

https://xemu.app/titles/41560010/#Return-to-Castle-Wolfenstein-Tides-of-War

### Bug Description

RTCW crashes shortly after boot while displaying the copyright screen. The underlying reason for the crash is that the client init function is never called due to the fact that it ends up having an unexpectedly changed value for the EBX register after calling `cpuid` at 0x00140c39. This leads to the init being skipped which ends up leading to a crash awhile later after initialization.

On XEMU before `cpuid`:
```
# eax            0x1                 1
# ecx            0x0                 0
# edx            0x7f355fba          2134204346
# ebx            0x34b9000           55283712
```
  After the call:
```
# eax            0x68a               1674
# ecx            0x80000000          -2147483648
# edx            0x383f9fd           58980861
# ebx            0x800               2048
```

On XBOX hardware:
```
# eax            0x1                 1
# ecx            0x0                 0
# edx            0x7f355fba          2134204346
# ebx            0x35ff000           56619008
```
After:
```
# eax            0x68a               1674
# ecx            0x0                 0
# edx            0x383f9ff           58980863
# ebx            0x0                 0
```

### Expected Behavior

The guest machine should not crash and `cpuid` should emulate XBOX hardware in all modified registers.

### xemu Version

0.6.1-24-gc6e05f51b2

### System Information

OS: Ubuntu (same crash happens on macOS and Windows)
GPU: GTX 1070

### Additional Context

_No response_

abaire:
This improves the crashing behavior of Return To Castle Wolfenstein as noted in #509 by causing the `cpuid` instruction to return the same values as real hardware.

As noted in the [datasheet](http://datasheets.chipdb.org/Intel/x86/CPUID/24161821.pdf), `cpuid` 0 returns the max supported index in `eax`, which is `2` on the hardware I tested with. Since only 0-2 are supported, this change simplifies the method by eliding all of the other cases and returning the equivalent of `cpuid(2)` for all of them.

Test code to generate these values is available at https://github.com/abaire/nxdk_check_cpuid

Fixes #509 (though the game still crashes later on if run from an iso, surprisingly it seems to work if copied to the E:)

### DIFF
--- a/target/i386/cpu.c
+++ b/target/i386/cpu.c
@@ -5112,6 +5112,38 @@ static void x86_register_cpudef_types(const X86CPUDefinition *def)
 
 }
 
+#if XBOX
+void cpu_x86_cpuid(CPUX86State *env, uint32_t index, uint32_t count,
+                   uint32_t *eax, uint32_t *ebx,
+                   uint32_t *ecx, uint32_t *edx)
+{
+  // See http://datasheets.chipdb.org/Intel/x86/CPUID/24161821.pdf
+  // These values were retrieved from a v1.0 XBOX.
+  switch(index) {
+    case 0x00000000:
+      *eax = 0x00000002;
+      *ebx = 0x756E6547;
+      *ecx = 0x6C65746E;
+      *edx = 0x49656E69;
+      break;
+
+    case 0x00000001:
+      *eax = 0x0000068A;
+      *ebx = 0x00000000;
+      *ecx = 0x00000000;
+      *edx = 0x0383F9FF;
+      break;
+
+    case 0x00000002:
+    default:
+      *eax = 0x03020101;
+      *ebx = 0x00000000;
+      *ecx = 0x00000000;
+      *edx = 0x0C040841;
+      break;
+  }
+}
+#else
 void cpu_x86_cpuid(CPUX86State *env, uint32_t index, uint32_t count,
                    uint32_t *eax, uint32_t *ebx,
                    uint32_t *ecx, uint32_t *edx)
@@ -5630,6 +5662,7 @@ void cpu_x86_cpuid(CPUX86State *env, uint32_t index, uint32_t count,
         break;
     }
 }
+#endif // #if XBOX
 
 static void x86_cpu_reset(DeviceState *dev)
 {


mborgerson:
Thank you for debugging this issue, for providing a patch, and for providing a test case to reproduce real data from hardware! Your patch indeed does fix #509 

However, I would prefer for xemu to retain the ability to support CPU model configuration in QEMU style. For this reason, I have pushed an alternative approach to my [fix/cpuid](https://github.com/mborgerson/xemu/tree/fix/cpuid) branch. It will reproduce `cpuid` results retrieved on a real Xbox, well.. save for 1 feature bit. The title in question also still boots with my version of the patch.

I invite you to review the alternative implementation. Additionally, I have added you as a co-author on the alternative implementation--please confirm you accept this co-author tag.

Thanks

abaire:
Totally makes sense.

Your change looks good to me, with a couple notes:

1. I don't know that a non-XBOX P3 would be limited to leaf 3 (I think P3 was when the processor serial numbers were introduced). I wonder if it might be better to ifdef the entire definition of `builtin_x86_defs` and just have a custom one for XBOX since it's a non-standard P3? Otherwise maybe `#ifdef` the level in the same way as model and stepping?
2. I'm not sure if it would ever matter in practice, but I think the `default` handling in `cpu_x86_cpuid` would do the wrong (or at least different from XBOX HW) thing in a case where `cpuid` was blindly called with a leaf > 2. In my testing and as far as I can tell from the docs, it should return the results of leaf 2 (the highest supported) rather than 0's.

Thanks for taking care of the proper integration, and very appreciative of the co-author tag!

mborgerson:
> Totally makes sense.
> 
> Your change looks good to me, with a couple notes:
> 
>     1. I don't know that a non-XBOX P3 would be limited to leaf 3 (I think P3 was when the processor serial numbers were introduced). I wonder if it might be better to ifdef the entire definition of `builtin_x86_defs` and just have a custom one for XBOX since it's a non-standard P3? Otherwise maybe `#ifdef` the level in the same way as model and stepping?
>

Not sure actually. We can change it later if necessary

>     2. I'm not sure if it would ever matter in practice, but I think the `default` handling in `cpu_x86_cpuid` would do the wrong (or at least different from XBOX HW) thing in a case where `cpuid` was blindly called with a leaf > 2. In my testing and as far as I can tell from the docs, it should return the results of leaf 2 (the highest supported) rather than 0's.
>

`cpu_x86_cpuid` will clip index to the highest available level like hardware does https://github.com/mborgerson/xemu/blob/7e4ca2d31984e768dc6737624118e9a33deb79dd/target/i386/cpu.c#L5161
 
> Thanks for taking care of the proper integration, and very appreciative of the co-author tag!

Thanks!

PR #550, opened 2021-11-16
Add USB daughterboard

Ryzee119:
This adds emulation for the USB Daugherboard which caused the USB input bug described here: https://github.com/mborgerson/xemu/pull/223

Additionally this adds the internal hubs for the game controllers in a separate commit. This is the initial step to add XMU support etc. Note the PID/VIDs for the hubs are different from retail, only because I am using QEMUs usb-hub device which does not have configurable IDs. This does not seem to impact functionality.

I was able to reproduce the input issue when the MCPX rev was set for 1.0 Xbox (no input at all), and adding the daughterboard fixed it.

Based on testing on real hardware, the USB daughterboard is connected to Port 1 of the root hub. The controllers are connected to ports 3,4,1,2 respectively on the daughterboard.

This PR vs real HW when connecting then disconnecting a controller from controller port 1:
image

Output from all 4 controllers:
image

Tested on a few retail titles and MS Dashboard. 4 players tested on Halo CE, all work as expected.

### DIFF
--- a/ui/xemu-input.c
+++ b/ui/xemu-input.c
@@ -431,7 +431,19 @@ void xemu_input_bind(int index, ControllerState *state, int save)
         bound_controllers[index]->bound = index;
 
         const int port_map[4] = {3, 4, 1, 2};
+        char *tmp;
 
+        // Create controller's internal USB hub.
+        QDict *usbhub_qdict = qdict_new();
+        qdict_put_str(usbhub_qdict, "driver", "usb-hub");
+        tmp = g_strdup_printf("1.%d", port_map[index]);
+        qdict_put_str(usbhub_qdict, "port", tmp);
+        qdict_put_int(usbhub_qdict, "ports", 3);
+        QemuOpts *usbhub_opts = qemu_opts_from_qdict(qemu_find_opts("device"), usbhub_qdict, &error_abort);
+        DeviceState *usbhub_dev = qdev_device_add(usbhub_opts, &error_abort);
+        g_free(tmp);
+
+        // Create XID controller. This is connected to Port 1 of the controller's internal USB Hub
         QDict *qdict = qdict_new();
 
         // Specify device driver
@@ -439,13 +451,13 @@ void xemu_input_bind(int index, ControllerState *state, int save)
 
         // Specify device identifier
         static int id_counter = 0;
-        char *tmp = g_strdup_printf("gamepad_%d", id_counter++);
+        tmp = g_strdup_printf("gamepad_%d", id_counter++);
         qdict_put_str(qdict, "id", tmp);
         g_free(tmp);
 
         // Specify index/port
         qdict_put_int(qdict, "index", index);
-        tmp = g_strdup_printf("1.%d", port_map[index]);
+        tmp = g_strdup_printf("1.%d.1", port_map[index]);
         qdict_put_str(qdict, "port", tmp);
         g_free(tmp);
 
@@ -455,10 +467,12 @@ void xemu_input_bind(int index, ControllerState *state, int save)
         assert(dev);
 
         // Unref for eventual cleanup
+        qobject_unref(usbhub_qdict);
+        object_unref(OBJECT(usbhub_dev));
         qobject_unref(qdict);
         object_unref(OBJECT(dev));
 
-        state->device = dev;
+        state->device = usbhub_dev;
     }
 }
 
--- a/ui/xemu-input.c
+++ b/ui/xemu-input.c
@@ -79,6 +79,15 @@ void xemu_input_init(void)
     new_con->name = "Keyboard";
     new_con->bound = -1;
 
+    // Create USB Daughterboard for 1.0 Xbox. This is connected to Port 1 of the Root hub.
+    QDict *usbhub_qdict = qdict_new();
+    qdict_put_str(usbhub_qdict, "driver", "usb-hub");
+    qdict_put_int(usbhub_qdict, "port", 1);
+    qdict_put_int(usbhub_qdict, "ports", 4);
+    QemuOpts *usbhub_opts = qemu_opts_from_qdict(qemu_find_opts("device"), usbhub_qdict, &error_fatal);
+    DeviceState *usbhub_dev = qdev_device_add(usbhub_opts, &error_fatal);
+    assert(usbhub_dev);
+
     // Check to see if we should auto-bind the keyboard
     int port = xemu_input_get_controller_default_bind_port(new_con, 0);
     if (port >= 0) {
@@ -436,7 +445,9 @@ void xemu_input_bind(int index, ControllerState *state, int save)
 
         // Specify index/port
         qdict_put_int(qdict, "index", index);
-        qdict_put_int(qdict, "port", port_map[index]);
+        tmp = g_strdup_printf("1.%d", port_map[index]);
+        qdict_put_str(qdict, "port", tmp);
+        g_free(tmp);
 
         // Create the device
         QemuOpts *opts = qemu_opts_from_qdict(qemu_find_opts("device"), qdict, &error_abort);
--- a/hw/xbox/xbox_pci.c
+++ b/hw/xbox/xbox_pci.c
@@ -419,8 +419,7 @@ static void xbox_lpc_class_init(ObjectClass *klass, void *data)
     k->config_write = xbox_lpc_config_write;
     k->vendor_id = PCI_VENDOR_ID_NVIDIA;
     k->device_id = PCI_DEVICE_ID_NVIDIA_NFORCE_LPC;
-    /* FIXME - correct revision for this is 0xB2 (178) for retail 1.0 Xbox, causes known USB bug */
-    k->revision = 212;
+    k->revision = 178;
     k->class_id = PCI_CLASS_BRIDGE_ISA;
 
     dc->desc = "nForce LPC Bridge";


GXTX:
> Note the PID/VIDs for the hubs are different from retail

For more completeness this could be changed in `hw/usb/dev-hub.c#L143`. But that's rather nitpicky, and there's other instances where IDs aren't matching such as the root hub reporting as some Apple thing.

I had a feeling the hub was required when kl0wn tried directly connecting to her 1.0 box without it.

Thank you Ryzee!

mborgerson:
Nice job, @Ryzee119! Does this break compatibility with any BIOS revs? Eventually I would like for xemu to support selection of different hardware steppings, which would change any necessary configuration and IDs accordingly. If this does break compat at all (which would not surprise me), we can prioritize stepping selection.

@GXTX 

> For more completeness this could be changed in hw/usb/dev-hub.c#L143. But that's rather nitpicky, and there's other instances where IDs aren't matching such as the root hub reporting as some Apple thing.

Preferably IDs would match hardware. Can you find out what's remaining?

GXTX:
@mborgerson [GXTX/xemu/tree/fix/more_pci_ids](https://github.com/GXTX/xemu/tree/fix/more_pci_ids) Should be enough to fix the remaining (although untested, can't find my linux drive image). I just never PRed, and I don't think it's correct to go and change USB devices in that branch.

Ryzee119:
Ill investigate adding the ability to configure the hub IDs as we would need to handle the daughterboard and the controller hub which are different. 

Ill also check a few other bioses and report back!

Ryzee119:
Ive pushed a change that allows usb hub IDs to be set at creation. Although Im not 100% happy with it. The only alternative I can think of is if we copy the entire USB hub code and make our own HUBs with custom IDs, but this seems even worse :)

In anycase, the PID/VID now match USB descriptor dumps I have on hand for the daughterboard and my controller's internal hub.

image

I also tested this on a range of retail and 'scene' bioses. My findings below. In general this seems to improve compatibility with earlier retail BIOS revisions.

Testing conducted booting into MS Dash version 5960, NTSC-M North America EEPROM Signed for the correct version.
| BIOS Rev | MCPX Rev | EEPROM Type | Current Master (a3e507860e831d5a4670d5e8d217d8cdfabb35d6) | This Branch |
|--|--|--|--|--|
| 3944 | MCPX1.0 | EEPROM 1.0 | Boots but input does not work | Boots ok |
| 4034 | MCPX1.0 | EEPROM 1.0 | Boots but input does not work | Boots ok |
| 4627 | MCPX1.0 | EEPROM 1.0 | Boots ok | Boots ok |
| 4817 | MCPX1.1 | EEPROM 1.1 - 1.4 | Boots ok | Boots ok |
| 5101 | MCPX1.1 | EEPROM 1.1 - 1.4 | Boots ok | Boots ok |
| 5713 | MCPX1.1 | EEPROM 1.1 - 1.4 | Boots ok | Boots ok |
| 5836 | MCPX1.1 | EEPROM 1.6 | Boots ok | Boots ok |
| 'Scene' Bioses | MCPX1.0 | EEPROM 1.0 | All seem to work | All seem to work |

GXTX:
@Ryzee119 5101+ require MCPX 1.1 and a differently 'signed' eeprom and should boot, xemu wont hold your hands there.

Ryzee119:
Oh yes right ofcourse. Ill create a newer eeprom and update the table

hw/usb/dev-hub.c
```diff
@@ -49,6 +49,8 @@ struct USBHubState {
     bool port_power;
     QEMUTimer *port_timer;
     USBHubPort ports[MAX_PORTS];
+    uint16_t pid;
+    uint16_t vid;
```
binarymaster: It seems this file belongs to QEMU, so you probably want `#ifdef XBOX` in all places (unless you plan to upstream).

hw/usb/dev-hub.c
```diff
@@ -674,6 +685,8 @@ static const VMStateDescription vmstate_usb_hub = {
 static Property usb_hub_properties[] = {
     DEFINE_PROP_UINT32("ports", USBHubState, num_ports, 8),
     DEFINE_PROP_BOOL("port-power", USBHubState, port_power, false),
+    DEFINE_PROP_UINT16("vid", USBHubState, vid, 0x0409),
+    DEFINE_PROP_UINT16("pid", USBHubState, pid, 0x55ab),
```
binarymaster: Would be nice to have these device ID constants defined somewhere, just like nForce PCI ids are defined here: https://github.com/mborgerson/xemu/blob/933d34ec35e3efd9bb68f5bfbd817911280b4280/include/hw/pci/pci_ids.h#L286

ui/xemu-input.c
```diff
@@ -79,6 +79,17 @@ void xemu_input_init(void)
     new_con->name = "Keyboard";
     new_con->bound = -1;
 
+    //Create USB Daughterboard for 1.0 Xbox. This is connected to Port 1 of the Root hub.
+    QDict *usbhub_qdict = qdict_new();
+    qdict_put_str(usbhub_qdict, "driver", "usb-hub");
+    qdict_put_int(usbhub_qdict, "port", 1);
+    qdict_put_int(usbhub_qdict, "ports", 4);
+    qdict_put_int(usbhub_qdict, "vid", 0x0451);
+    qdict_put_int(usbhub_qdict, "pid", 0x2046);
```
binarymaster: And these ids as well.

ui/xemu-input.c
```diff
@@ -422,32 +433,50 @@ void xemu_input_bind(int index, ControllerState *state, int save)
...
+        char *tmp;
+
+        //Create controller's internal USB hub.
+        QDict *usbhub_qdict = qdict_new();
+        qdict_put_str(usbhub_qdict, "driver", "usb-hub");
+        tmp = g_strdup_printf("1.%d", port_map[index]);
+        qdict_put_str(usbhub_qdict, "port", tmp);
+        qdict_put_int(usbhub_qdict, "ports", 3);
+        qdict_put_int(usbhub_qdict, "vid", 0x045E);
+        qdict_put_int(usbhub_qdict, "pid", 0x0288);
```
binarymaster: And these ids too.

hw/usb/dev-hub.c
mborgerson: Agreed. Let's define these VID/PIDs as constant values [repeated below]

ui/xemu-input.c
```diff
@@ -79,6 +79,17 @@ void xemu_input_init(void)
     new_con->name = "Keyboard";
     new_con->bound = -1;
 
+    //Create USB Daughterboard for 1.0 Xbox. This is connected to Port 1 of the Root hub.
```
mborgerson: Nit: Space between // and comment text, repeated below

ui/xemu-input.c
```diff
@@ -422,32 +433,50 @@ void xemu_input_bind(int index, ControllerState *state, int save)
         bound_controllers[index]->bound = index;
 
         const int port_map[4] = {3, 4, 1, 2};
+        char *tmp;
+
+        //Create controller's internal USB hub.
+        QDict *usbhub_qdict = qdict_new();
```
mborgerson, line 437: Eventually we probably want to move this hub+controller logic out to it's own unit, but this is good for now

hw/usb/dev-hub.c
mborgerson: Let's not make pid/vid a property, but rather just define the desc properly for Xbox controller, daughter-board hub

mborgerson:
@Ryzee119 if the USB VID/PIDs don't gate booting of any kernels, let's strip out the configurable VID/PID for now and simplify this commit please. I'd like the configurable IDs to be done a little differently, but don't want to hold up this PR if there's little to no practical benefit to it

hw/usb/dev-hub.c
Ryzee119: Agreed but I dropped this commit as per other review comments. So this is indirectly resolved

hw/usb/dev-hub.c
Ryzee119: Dropped commit as per other review comments. No longer applicable.

ui/xemu-input.c
Ryzee119: Thanks. Fixed

ui/xemu-input.c
Ryzee119: > @Ryzee119 if the USB VID/PIDs don't gate booting of any kernels, let's strip out the configurable VID/PID for now and simplify this commit please. I'd like the configurable IDs to be done a little differently, but don't want to hold up this PR if there's little to no practical benefit to it

It appears to be purely superficial change. All testing ive done does not impact the functionality. I stripped it out for now. We can revisit later for completeness.

mborgerson:
Thanks a lot for testing on real hardware, and for testing many bios versions--the compat matrix is great! We will need to do some additional testing to make sure that port mapping here is correct on all bios revs (in particular I want to be sure later retail versions do work as expected). We can do that and fix in the future if required.

Eventually I would like for a user to be able to easily select their hardware stepping, which may control the presence of this daughter board, device revs, etc. Future work as well.

I'm very happy that input is working on early kernel revs now. Looking forward to having well integrated xmu and xblc support next :)

Thanks a lot for fixing this!

Issue #322, opened 2021-06-02
[Assertion] Zeta surface clear for Z24S8 (float) is untested


Triticum0:
### Title Information
https://xemu.app/titles/454d0001/#Antz-Extreme-Racing
https://xemu.app/titles/4143000a/#ATV-Quad-Power-Racing-2
https://xemu.app/titles/494c0002/#Braveknight-Leverant-Eiyuuden
https://xemu.app/titles/4b4e0003/#ESPN-NFL-PrimeTime-2002
https://xemu.app/titles/5553000b/#Largo-Winch-Empire-Under-Threat
https://xemu.app/titles/4253000a/#Malice
https://xemu.app/titles/494c0001/#Metal-Dungeon
https://xemu.app/titles/54530005/#RoboCop
https://xemu.app/titles/41560009/#Rally-Fusion-Race-of-Champions
https://xemu.app/titles/49470004/#TransWorld-Snowboarding
https://xemu.app/titles/4553000e/#The-Italian-Job
https://xemu.app/titles/4d530048/#The-Wild-Rings
https://xemu.app/titles/41540001/#Touge-R

### Bug Description
The untested code is here:
https://github.com/mborgerson/xemu/blob/bbcd5763e6204d6d1e109c2d19cfd794e3abf8e2/hw/xbox/nv2a/pgraph.c#L3056-L3057

In particular, test this function:
https://github.com/mborgerson/xemu/blob/bbcd5763e6204d6d1e109c2d19cfd794e3abf8e2/hw/xbox/nv2a/pgraph.c#L6314-L6321

Steps to reproduce this issue:
1.Insert disk
2. Game crashes

### Expected Behavior

Should not hit assertion and make it into menus

<h3>Version</h3>
<ul>
<li>Version: build-202106020756( <a href="https://github.com/mborgerson/xemu/commit/3884b8d44b4bfaed73f26227cd88d7946e2dfaa5"><tt>3884b8d</tt></a>)</li>
<li>Last-known Working: Unknown</li>
</ul>
<h3>System Information</h3>

Field | Value
-- | --
OS | Windows 10 64-bit (21H1)
CPU | AMD Ryzen 5 2600 Six-Core Processor @ 3.40GHz
Graphics Device | AMD Radeon RX 5700 XT
Graphics Driver | 4.0.148202 Core Profile Context 21.3.2 27.20.15003.5017

###Additional Context
[...]

PR #552, opened 2021-11-18
pgraph: Enables 16-bit floating point z-buffer mode.

### LINKED ISSUE #60: NV097_SET_SURFACE_FORMAT_ZETA_Z16 needs verification
The Voodoo Vince demo on the Crimson Skies disc crashes upon trying to go in-game after selecting play on the main menu. This is the resulting error:

ERROR:D:/a/xemu/xemu/hw/xbox/nv2a/pgraph.c:2482:pgraph_method: assertion failed: (0)
Bail out! ERROR:D:/a/xemu/xemu/hw/xbox/nv2a/pgraph.c:2482:pgraph_method: assertion failed: (0)

Titles: 4d53005f, 4d530021, 43430002, 43430009

https://xemu.app/titles/4156000f/#Star-Wars-Obi-Wan
https://xemu.app/titles/494c0002/#Braveknight-Leverant-Eiyuuden
https://xemu.app/titles/4b4e0003/#ESPN-NFL-PrimeTime-2002
https://xemu.app/titles/43430005/#Group-S-Challenge
https://xemu.app/titles/4b4e0019/#Dance-Dance-Revolution-Ultramix
https://xemu.app/titles/41430003/#Legends-of-Wrestling
https://xemu.app/titles/4143000f/#Legends-of-Wrestling-II
https://xemu.app/titles/4143001d/#Showdown-Legends-of-Wrestling
https://xemu.app/titles/4d570029/#Mortal-Kombat-Shaolin-Monks
https://xemu.app/titles/45530016/#Backyard-Wrestling-2-There-Goes-the-Neighborhood
https://xemu.app/titles/4d530034/#Inside-Pitch-2003
https://xemu.app/titles/424d0002/#Reign-of-Fire
https://xemu.app/titles/53450029/#Spikeout-Battle-Street
https://xemu.app/titles/41540001/#Touge-R

abaire:
Removes the verification assert when clearing 16-bit floating point z-buffer.

Tested via https://github.com/abaire/nxdk_pgraph_tests and for the most part the output produced by xemu very closely matches the output from a 1.0 XBOX (there are very minor differences in color blending that are also observable in normal texturing and do not appear to be related to z-buffering). Output of running the test against hardware can be found here: https://github.com/abaire/nxdk_pgraph_tests_golden_results

The main difference I observed is that on the XBOX, z values < `convert_f16_to_float(0x1000) == 0.0156250f` will all result in 0 being written to the depth buffer, whereas xemu seems to support all mappable values down to zero. This results in clear values < 0x1000 still rendering anything < `0.0156250f` on hardware but content being properly cleared on xemu. I'm not sure if this matters in practice as it only affects a small range of z-values with clipping disabled or with a 0 near-plane.

Fixes #60

### DIFF
--- a/hw/xbox/nv2a/pgraph.c
+++ b/hw/xbox/nv2a/pgraph.c
@@ -3043,7 +3043,6 @@ DEF_METHOD(NV097, CLEAR_SURFACE)
             /* FIXME: Remove bit for stencil clear? */
             if (pg->surface_shape.z_format) {
                 gl_clear_depth = convert_f16_to_float(z) / f16_max;
-                assert(false); /* FIXME: Untested */
             } else {
                 gl_clear_depth = z / (float)0xFFFF;
             }


hw/xbox/nv2a/vsh.c
```diff
@@ -776,9 +776,15 @@ void vsh_translate(uint16_t version,
          * Note this makes the values in the depth buffer wrong. This should be
          * handled with gl_ClipDistance instead, but that has performance issues
          * on OS X.
+         *
+         * This conversion is performed at double precision to handle max depth
```
abaire: I think this is technically not necessary for the 16-bit buffer, but the z24 will fail with the clip range at maximum precision. I can throw this into a future PR to address #322 if you'd prefer (it is necessary for anybody who wants to compare the full set of results against the hardware).

mborgerson: I thought I had resolved this precision issue already, but apparently not everywhere. I know it's still an issue on M1 at least. I'm okay to switch up to DP. Let's make this a separate commit and correct typo (z25->z24)

mborgerson:
Thanks for the unit tests! Minor fixup and I'll merge

mborgerson:
> The main difference I observed is that on the XBOX, z values < convert_f16_to_float(0x1000) == 0.0156250f will all result in 0 being written to the depth buffer, whereas xemu seems to support all mappable values down to zero. This results in clear values < 0x1000 still rendering anything < 0.0156250f on hardware but content being properly cleared on xemu. I'm not sure if this matters in practice as it only affects a small range of z-values with clipping disabled or with a 0 near-plane.

We should clamp the values if necessary to replicate HW behavior. We can do this in a follow up

hw/xbox/nv2a/vsh.c
abaire: sg, will split it out.

abaire:
> > The main difference I observed is that on the XBOX, z values < convert_f16_to_float(0x1000) == 0.0156250f will all result in 0 being written to the depth buffer, whereas xemu seems to support all mappable values down to zero. This results in clear values < 0x1000 still rendering anything < 0.0156250f on hardware but content being properly cleared on xemu. I'm not sure if this matters in practice as it only affects a small range of z-values with clipping disabled or with a 0 near-plane.
> 
> We should clamp the values if necessary to replicate HW behavior. We can do this in a follow up

Sounds good. I'm somewhat wary that there may be some dumb mistake with my pgraph config that is causing the xbox to clamp the exponent-less values, but I guess it makes sense that if my test is doing something wrong in setting up the hardware it's doing the same wrong thing setting up xemu so the behavior should probably match. If there's anybody out there with more knowledge of nv2a configuration that would like to take a look, it'd be very much appreciated :)

mborgerson:
FYI I've edited commit message.

Thanks for the contribution!

PR #553, opened 2021-11-18
vsh: Uses double precision when doing clip range remapping.

abaire:
Fixes failure when mapping a full range floating point z24s8 clip range into OpenGL due to lack of precision.


mborgerson:
Minor commit msg edit

Thanks

mborgerson:
I've tested this merged change from yesterday, and while it does fix precision issues, using doubles causes xemu on M1 to use unacceptably (seconds per frame) slow software vertex processing. Had I known that was the case I would not have merged. Oops.

abaire:
Makes some sense, I figured moving to double would have some hw-dependent performance cost. Given that float z24 has other, bigger problems it probably doesn't makes sense to make any precision changes to the shader until there's resolution for #322

PR #556, opened 2021-11-19
nv2a/vsh: Precompute clip range to avoid double in shader.

abaire:
It looks like clip range is only used in the vertex shader so we can avoid hopping into (potentially slow) doubles and precompute the depth buffer mapping at double precision once on the CPU instead of on every frame in the shader.

I'm not sure if it's also worth renaming `clipRange` to something like `clipMapping` or `clipRangeMapping` to better capture the fact that it's not actually a range. I added a comment where it's declared to describe it (I'm assuming anybody confused by the behavior would search for `clipRange` and see it there).

### DIFF
--- a/hw/xbox/nv2a/shaders.c
+++ b/hw/xbox/nv2a/shaders.c
@@ -682,6 +682,8 @@ static MString *generate_vertex_shader(const ShaderState state,
     MString *header = mstring_from_str(
 "#version 400\n"
 "\n"
+// Holds an offset (the near plane) and a precomputed 1/2 (far - near) suitable
+// for converting Xbox coordinates into GL.
 "uniform vec2 clipRange;\n"
 "uniform vec2 surfaceSize;\n"
 "\n"
--- a/hw/xbox/nv2a/pgraph.c
+++ b/hw/xbox/nv2a/pgraph.c
@@ -3770,7 +3770,8 @@ static void pgraph_shader_update_constants(PGRAPHState *pg,
     }
 
     if (binding->clip_range_loc != -1) {
-        glUniform2f(binding->clip_range_loc, zclip_min, zclip_max);
+        double precomputed_depth = (double)(zclip_max - zclip_min) * 0.5;
+        glUniform2f(binding->clip_range_loc, zclip_min, (float)precomputed_depth);
     }
 
     /* Clipping regions */
--- a/hw/xbox/nv2a/vsh.c
+++ b/hw/xbox/nv2a/vsh.c
@@ -781,10 +781,7 @@ void vsh_translate(uint16_t version,
          * clipping in the Xbox z24s8 floating point mode.
          */
         "  if (clipRange.y != clipRange.x) {\n"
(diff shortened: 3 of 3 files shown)


abaire:
Note: This has a negative impact on the floating point z24s8 depth buffer in #322 so we may end up having to go back to doing the actual division as a double depending on how 24-bit floating point depth buffers are solved (but could still precompute the range when binding shaders).

mborgerson:
I've tested the merged doubles change from yesterday, and while it does fix precision issues, using doubles causes xemu on M1 to use unacceptably (seconds per frame) slow software vertex processing. Had I known that was the case I would not have merged. Oops.

hw/xbox/nv2a/shaders.c
```diff
@@ -682,6 +682,8 @@ static MString *generate_vertex_shader(const ShaderState state,
     MString *header = mstring_from_str(
 "#version 400\n"
 "\n"
+// Holds an offset (the near plane) and a precomputed 1/2 (far - near) suitable
```
mborgerson, line 685: Don't need this comment here

JohnGodgames:
Just tested the first random game V-rally 3 and this commit brokes menu text

Master:
vrally master

PR:
vrally clipping

mborgerson:
> It looks like clip range is only used in the vertex shader so we can avoid hopping into (potentially slow) doubles and precompute the depth buffer mapping at double precision once on the CPU instead of on every frame in the shader.

It is not computed for every frame, rather for every vertex.

> I'm not sure if it's also worth renaming clipRange to something like clipMapping or clipRangeMapping to better capture the fact that it's not actually a range. I added a comment where it's declared to describe it (I'm assuming anybody confused by the behavior would search for clipRange and see it there).

The comment within the vert shader gen makes clear exactly what it is doing

> Note: This has a negative impact on the floating point z24s8 depth buffer in #322 so we may end up having to go back to doing the actual division as a double depending on how 24-bit floating point depth buffers are solved (but could still precompute the range when binding shaders).

Unfortunately moving to double precision causes an unacceptable performance regression an Apple M1 at least, possibly elsewhere (https://github.com/mborgerson/xemu/pull/553#issuecomment-974402327), so the change has been reverted. One of the issues here is that we do not have the same level of precision on all machines; however we can mitigate some of the issues caused by this (e.g. clamping).

> Just tested the first random game V-rally 3 and this commit brokes menu text

@JohnGodgames Thanks for testing.

abaire:
> Unfortunately moving to double precision causes an unacceptable performance regression an Apple M1 at least, possibly elsewhere ([#553 (comment)](https://github.com/mborgerson/xemu/pull/553#issuecomment-974402327)), so the change has been reverted. One of the issues here is that we do not have the same level of precision on all machines; however we can mitigate some of the issues caused by this (e.g. clamping).

Makes sense. I think this PR should be parked until there's a solution for #322 since it triggers a regression and I don't think it provides a substantial benefit outside the float z24 case.

@JohnGodgames ++ to thanks for testing :)

PR #557, opened 2021-11-19
nv2a: Treat 0 SET_FRONT_FACE as CCW to match hw.

### LINKED ISSUE #321: Unknown front face
### Title Information
Titles: 
Whiplash (45530011)
Project: Snowblind (45530019)
https://xemu.app/titles/45530011/#Whiplash
https://xemu.app/titles/45530019/#Project-Snowblind

### Bug Description
Game crashes with this assertion before the intro
https://github.com/mborgerson/xemu/blob/bbcd5763e6204d6d1e109c2d19cfd794e3abf8e2/hw/xbox/nv2a/pgraph.c#L1707-L1708

1. Insert disk
2. Game crashes

### Expected Behavior

It should make it into the intro

### Version
<ul>
</ul>
<h3>System Information</h3>

Field | Value
-- | --
OS | Windows 10 64-bit (21H1)

abaire:
Treats a `SET_FRONT_FACE` of `0` as `CCW` to match observed behavior on HW (Xbox 1.0).

Test case: https://github.com/abaire/nxdk_pgraph_tests/blob/main/tests/front_face_tests.cpp

Fixes #321

### DIFF
--- a/hw/xbox/nv2a/pgraph.c
+++ b/hw/xbox/nv2a/pgraph.c
@@ -1704,9 +1704,8 @@ DEF_METHOD(NV097, SET_FRONT_FACE)
     case NV097_SET_FRONT_FACE_V_CCW:
         ccw = true; break;
     default:
-        fprintf(stderr, "Unknown front face: 0x%x\n", parameter);
-        assert(false);
-        break;
+        NV2A_DPRINTF("Unknown front face: 0x%08x\n", parameter);
+        return; /* discard */
     }
     SET_MASK(pg->regs[NV_PGRAPH_SETUPRASTER],
              NV_PGRAPH_SETUPRASTER_FRONTFACE,


hw/xbox/nv2a/pgraph.c
```diff
@@ -1701,6 +1701,12 @@ DEF_METHOD(NV097, SET_FRONT_FACE)
...
         ccw = false; break;
+
+    case 0:
+        // Treat 0 as CCW to match Xbox behavior.
+        // This is known to be set by the following titles:
+        //   Whiplash (45530011)
+        //   Project Snowblind (45530019)
     case NV097_SET_FRONT_FACE_V_CCW:
         ccw = true; break;
     default:
```
abaire, line 1706: From playing with a few random values, I suspect all invalid params are treated as CCW on hardware. Does it make sense to have the default warn and return CCW rather than assert? 

It seems like you'd probably have to jump through some hoops to get DirectX to send an invalid winding, so maybe these `0` sending titles are the only ones to worry about?

mborgerson: Can you confirm hardware actually does this? Set the value and read it back and see what you get. For other methods I've seen hardware drop the invalid value and retain whatever was programmed previously.

e.g. Set CW, then set some invalid value. Then check that it was set to CCW as you suggested, or whether it remains CW.

abaire: Yeah, I had the same thought and (manually) ran the tests out of order to verify that it seems to always force CCW.

I haven't dug into pbkit enough to know how to read the value back, do you happen to know off the top of your head? Otherwise I can modify the test to explicitly set CW and flush before setting the random value so what I did manually is automated.

abaire: And I take it back. I forgot that pbkit is reinitializing things under the hood between tests so my validation was invalid. I'm now forcibly setting CW prior to the value under test and you're correct, it just preserves the previously set value.

In light of this, maybe we should just remove the assert and have the default be a no-op?

mborgerson: Yes, here's an example of how this is handled elsewhere: https://github.com/mborgerson/xemu/blob/50a5368f473ea95482f94fd4c7edd1b2b9709179/hw/xbox/nv2a/pgraph.c#L1491

Let's follow the model here too

abaire: Cool, changed

mborgerson:
Thanks

Issue #324, opened 2021-06-02
Unhandled NV_IMAGE_BLIT methods


Triticum0:
### Title Information

https://xemu.app/titles/41560005/#Blade-II - Crashes before Intros
https://xemu.app/titles/4d570011/#Dr-Muto - Crashes when going in-game
https://xemu.app/titles/4b420003/#Rogue-Ops  - Crashes After Intros
https://xemu.app/titles/56560029/#Die-Hard-Vendetta - Crashes when going in-game

### Bug Description
The games crash with an assertion:
Unknown blit operation: 0x2
https://github.com/mborgerson/xemu/blob/ebcc77c6c82e6457df9f7f769820fd5da0d6a0ce/hw/xbox/nv2a/pgraph.c#L915

### Expected Behavior
Should make it in-game

<h3>Version</h3>
<ul>
<li>Version: build-202106020756( <a href="https://github.com/mborgerson/xemu/commit/3884b8d44b4bfaed73f26227cd88d7946e2dfaa5"><tt>3884b8d</tt></a>)</li>
<li>Last-known Working: Unknown</li>
</ul>
<h3>System Information</h3>

Field | Value
-- | --
OS | Windows 10 64-bit (21H1)
CPU | AMD Ryzen 5 2600 Six-Core Processor @ 3.40GHz
Graphics Device | AMD Radeon RX 5700 XT
Graphics Driver | 4.0.148202 Core Profile Context 21.3.2 27.20.15003.5017

abaire:
For Dr. Muto, `image_blit->operation` == `0x02` (as of Xemu 0.6.2-5, the only operation supported by `NV095_SIZE` is `NV09F_SET_OPERATION_SRCCOPY` which == `0x03`).

I'm assuming by context that this is `BLEND_AND` (https://github.com/torvalds/linux/blob/master/drivers/gpu/drm/nouveau/include/nvhw/class/cl902d.h#L172)

Triticum0:
@abaire Tested all the games on the list and there bilt operation are the same as Dr Muto

abaire:
> @abaire Tested all the games on the list and there bilt operation are the same as Dr Muto

Perfect, this should be straightforward to fix then (at least an initial inefficient fix). Thanks for testing!

Triticum0:
@abaire I retested and unfortunately Blade 2 uses 2 Blitting surface formats,So as well 0x2 it also has 0x6 blit surface format.

abaire:
[...]

PR #565, opened 2021-11-23
hud: Add nv2a state debug window.

abaire:
In investigating #324 I found that it'd be useful to understand which `ramin` context objects are involved in blitting, so I added an additional window to the debug menu to visualize the entries in the nv2a `ramht` table.

Screen Shot 2021-11-22 at 19 01 11


abaire:
Closing this out, I'm writing a more full featured debugger for nv2a (ability to step frames and begin/end calls, ability to examine textures, probably states, etc...) that will incorporate this change and be more generally useful.

PR #569, opened 2021-11-24
nv2a: Download destination surface for partial SRCCOPY blits.

abaire:
While getting set up to test #324 I found that SRCCOPY of an image that is smaller than the full framebuffer leads to graphical corruption as it ends up blitting the image on top of stale data instead of the pending frame. This change initiates a download if the image is smaller than the destination surface and the surface is dirty.

Test at https://github.com/abaire/nxdk_pgraph_tests/blob/main/tests/image_blit_tests.cpp

Note: I think we might need to reevaluate this approach and leverage HW acceleration for the blit since we'll likely want to use it for more complex ops (like `BLEND_AND`).

### DIFF
--- a/hw/xbox/nv2a/pgraph.c
+++ b/hw/xbox/nv2a/pgraph.c
@@ -854,97 +854,111 @@ int pgraph_method(NV2AState *d, unsigned int subchannel,
         image_blit->width = parameter & 0xFFFF;
         image_blit->height = parameter >> 16;
 
-        /* I guess this kicks it off? */
-        if (image_blit->operation == NV09F_SET_OPERATION_SRCCOPY) {
+        if (image_blit->width && image_blit->height) {
+            if (image_blit->operation == NV09F_SET_OPERATION_SRCCOPY) {
 
-            NV2A_GL_DPRINTF(true, "NV09F_SET_OPERATION_SRCCOPY");
+                NV2A_GL_DPRINTF(true, "NV09F_SET_OPERATION_SRCCOPY");
 
-            pgraph_update_surface(d, false, true, true);
+                pgraph_update_surface(d,
+                                      false,
+                                      true,
+                                      true);
 
-            ContextSurfaces2DState *context_surfaces = context_surfaces_2d;
-            assert(context_surfaces->object_instance
-                    == image_blit->context_surfaces);
+                ContextSurfaces2DState *context_surfaces = context_surfaces_2d;
+                assert(context_surfaces->object_instance
+                               == image_blit->context_surfaces);
 
-            unsigned int bytes_per_pixel;
-            switch (context_surfaces->color_format) {
-            case NV062_SET_COLOR_FORMAT_LE_Y8:
-                bytes_per_pixel = 1;
-                break;
-            case NV062_SET_COLOR_FORMAT_LE_R5G6B5:
-                bytes_per_pixel = 2;
-                break;
-            case NV062_SET_COLOR_FORMAT_LE_A8R8G8B8:
-            case NV062_SET_COLOR_FORMAT_LE_X8R8G8B8:
-            case NV062_SET_COLOR_FORMAT_LE_Y32:
-                bytes_per_pixel = 4;
-                break;
-            default:
-                fprintf(stderr, "Unknown blit surface format: 0x%x\n", context_surfaces->color_format);
+                unsigned int bytes_per_pixel;
+                switch (context_surfaces->color_format) {
+                    case NV062_SET_COLOR_FORMAT_LE_Y8:bytes_per_pixel = 1;
+                        break;
+                    case NV062_SET_COLOR_FORMAT_LE_R5G6B5:bytes_per_pixel = 2;
+                        break;
+                    case NV062_SET_COLOR_FORMAT_LE_A8R8G8B8:
+                    case NV062_SET_COLOR_FORMAT_LE_X8R8G8B8:
+                    case NV062_SET_COLOR_FORMAT_LE_Y32:bytes_per_pixel = 4;
+                        break;
+                    default:
+                        fprintf(stderr,
+                                "Unknown blit surface format: 0x%x\n",
+                                context_surfaces->color_format);
+                        assert(false);
+                        break;
+                }
+
+                hwaddr source_dma_len, dest_dma_len;
+                uint8_t *source, *dest;
+
+                source = (uint8_t *) nv_dma_map(
+                        d, context_surfaces->dma_image_source, &source_dma_len);
+                assert(context_surfaces->source_offset < source_dma_len);
+                source += context_surfaces->source_offset;
+
+                dest = (uint8_t *) nv_dma_map(d,
+                                              context_surfaces->dma_image_dest,
+                                              &dest_dma_len);
+                assert(context_surfaces->dest_offset < dest_dma_len);
+                dest += context_surfaces->dest_offset;
+
+                NV2A_DPRINTF("  - 0x%tx -> 0x%tx\n", source - d->vram_ptr,
+                             dest - d->vram_ptr);
+
+
+                // FIXME: Blitting from part of a surface
+
+                SurfaceBinding *surf_src = pgraph_surface_get(
+                        d, source - d->vram_ptr);
+                if (surf_src) {
+                    pgraph_download_surface_data(d, surf_src, true);
+                }
+
+                SurfaceBinding *surf_dest = pgraph_surface_get(
+                        d, dest - d->vram_ptr);
+                if (surf_dest) {
+                    if (image_blit->height < surf_dest->height ||
+                            image_blit->width < surf_dest->width) {
+
+                        pgraph_download_surface_data_if_dirty(d, surf_dest);
+                    }
+                    surf_dest->upload_pending = true;
+                }
+
+                uint32_t bytes_per_row = image_blit->width * bytes_per_pixel;
+                int y;
+                unsigned int source_pitch = context_surfaces->source_pitch;
+                unsigned int dest_pitch = context_surfaces->dest_pitch;
+                for (y = 0; y < image_blit->height; y++) {
+                    uint8_t *source_row = source
+                            + (image_blit->in_y + y) * source_pitch
+                            + image_blit->in_x * bytes_per_pixel;
+
+                    uint8_t *dest_row = dest
+                            + (image_blit->out_y + y) * dest_pitch
+                            + image_blit->out_x * bytes_per_pixel;
+
+                    memmove(dest_row, source_row, bytes_per_row);
+                }
+
+                hwaddr dest_start = image_blit->out_x +
+                        image_blit->out_y * context_surfaces->dest_pitch;
+                hwaddr dirty_start = dest - d->vram_ptr + dest_start;
+                uint32_t dirty_size = image_blit->width * bytes_per_pixel +
+                        (image_blit->height - 1) * context_surfaces->dest_pitch;
+                memory_region_set_client_dirty(d->vram,
+                                               dirty_start,
+                                               dirty_size,
+                                               DIRTY_MEMORY_VGA);
+                memory_region_set_client_dirty(d->vram,
+                                               dirty_start,
+                                               dirty_size,
+                                               DIRTY_MEMORY_NV2A_TEX);
+            } else {
+                fprintf(stderr,
+                        "Unknown blit operation: 0x%x\n",
+                        image_blit->operation);
                 assert(false);
-                break;
             }
-
-            hwaddr source_dma_len, dest_dma_len;
-            uint8_t *source, *dest;
-
-            source = (uint8_t*)nv_dma_map(d, context_surfaces->dma_image_source,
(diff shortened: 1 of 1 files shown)


hw/xbox/nv2a/pgraph.c
```diff
@@ -923,25 +927,19 @@ int pgraph_method(NV2AState *d, unsigned int subchannel,
...
-                                           (dest - d->vram_ptr),
-                                           image_blit->height
-                                           * image_blit->width
-                                           * bytes_per_pixel,
-                                           DIRTY_MEMORY_NV2A_TEX);
-
+                memmove(dest_row, source_row, bytes_per_row);
 
+                memory_region_set_client_dirty(d->vram,
+                                               (dest_row - d->vram_ptr),
```
mborgerson: Let's move these calls back out of the loop

mborgerson:
> While getting set up to test #324 I found that SRCCOPY of an image that is smaller than the full framebuffer leads to graphical corruption as it ends up blitting the image on top of stale data instead of the pending frame. This change initiates a download if the image is smaller than the destination surface and the surface is dirty.

Great! Yes destination surface should be synchronized in this case. Is there a game that exhibited this behavior?

> Test at https://github.com/abaire/nxdk_pgraph_tests/blob/main/tests/image_blit_tests.cpp
> 
> Note: I think we might need to reevaluate this approach and leverage HW acceleration for the blit since we'll likely want to use it for more complex ops (like `BLEND_AND`).

Yeah eventually this will be fixed to allow partial surface updates; for now we accept the hit

hw/xbox/nv2a/pgraph.c
abaire: How important is it that the dirty region be accurate? If we take them out of the loop I think they'll only be correct for full-width blits (or single row blits).

E.g., if the image is 10x10x4 and the framebuffer is 640x480x4 and the subimage is rendered at 0 it'll mark the first 100 pixels dirty instead of the 10x10 subtexture.
I briefly looked to see if there was a stride-aware dirty region method but didn't see one. That'd at least get rid of a bunch of function call overhead (might also be interesting to allow multiple clients to be set at once, though I haven't looked at the details of the `memory_region_set_client_dirty` call to see if there's a reason that would be hard).

I'm happy to add a method to do stride-able dirty marking if you think it's worthwhile.

abaire:
> > While getting set up to test #324 I found that SRCCOPY of an image that is smaller than the full framebuffer leads to graphical corruption as it ends up blitting the image on top of stale data instead of the pending frame. This change initiates a download if the image is smaller than the destination surface and the surface is dirty.
> 
> Great! Yes destination surface should be synchronized in this case. Is there a game that exhibited this behavior?

I'm not sure if there's a game that hits the issue, I just ran into it while trying to build out a test bed for the `BLEND_AND` and noticed it was behaving incorrectly.

> > Test at https://github.com/abaire/nxdk_pgraph_tests/blob/main/tests/image_blit_tests.cpp
> > Note: I think we might need to reevaluate this approach and leverage HW acceleration for the blit since we'll likely want to use it for more complex ops (like `BLEND_AND`).
> 
> Yeah eventually this will be fixed to allow partial surface updates; for now we accept the hit

Is there work in flight already to do this? I assume doing per-pixel alpha on the CPU for `BLEND_AND` is going to be a non-starter for lower end hardware at least. Happy to look into it if nobody is already working on it, but will take some poking around to understand the interaction w/ GL and emulated vram.

hw/xbox/nv2a/pgraph.c
mborgerson: Dirty tracking is page level so we do not require such precision here

mborgerson: On second thought, we can keep this partial tracking as you have implemented it here

abaire: Sounds good :)

Do you think it's worth having a stride-able set_client_dirty rather than do potentially hundreds of calls per blit? I'm not sure how often the `IMAGE_BLIT` stuff is used in practice (besides the xbox logo and the crash in Dr. Muto that started me looking into this).

mborgerson:
> Is there work in flight already to do this? I assume doing per-pixel alpha on the CPU for `BLEND_AND` is going to be a non-starter for lower end hardware at least. Happy to look into it if nobody is already working on it, but will take some poking around to understand the interaction w/ GL and emulated vram.

Kind of, but I've deferred it for later. Instead of working on this I'd prefer we focus on fixing game-blocking issues, as you have been doing

abaire:
> > Is there work in flight already to do this? I assume doing per-pixel alpha on the CPU for `BLEND_AND` is going to be a non-starter for lower end hardware at least. Happy to look into it if nobody is already working on it, but will take some poking around to understand the interaction w/ GL and emulated vram.
> 
> Kind of, but I've deferred it for later. Instead of working on this I'd prefer we focus on fixing game-blocking issues, as you have been doing

Definitely, hoping to get all the red ones into orange at least :)

This will be a game blocker for #324 since you can't get past the menu. From Dr. Muto it looks like it only does a single `BLEND_AND` blit, so maybe a CPU-based blend will be OK just to get things working and I can `FIXME: Optimize` it?
I'm also hoping @Triticum0 might be able to re-test those #324 games w/ this PR since it should now print out the missing blit op before asserting to tell us which ones need to be implemented.

mborgerson:
> Definitely, hoping to get all the red ones into orange at least :)

We will get there! It's closer than it looks.

> 
> This will be a game blocker for #324 since you can't get past the menu. From Dr. Muto it looks like it only does a single `BLEND_AND` blit, so maybe a CPU-based blend will be OK just to get things working and I can `FIXME: Optimize` it?

Exactly. Let's defer such optimization until necessary and take the simple, slow path first

> I'm also hoping @Triticum0 might be able to re-test those #324 games w/ this PR since it should now print out the missing blit op before asserting to tell us which ones need to be implemented.

As for testing, you are welcome to join the discord server and request some testing from registered testers

hw/xbox/nv2a/pgraph.c
mborgerson: I wouldn't sorry about it. I don't think it's worth doing unless profiling shows significant cost

Typical surfaces would need to be very large to benefit from row granularity like this, which is why I originally suggested keeping the calls to mark regions dirty outside this loop. If you already have the code done, let's move them back out. Thanks for the patience on that

mborgerson: I wouldn't worry about it. I don't think it's worth doing unless profiling shows significant cost

I think row-level tracking like this is unlikely to be beneficial in most cases, and adds extra overhead in dirty bit tracking. If you would move it back out of the loop, we can put it back if we discover a case for this. Thanks for your patience on it

Triticum0:
Tested Blade 2 and Rogue Ops and Die Hard Vendetta and there bilt operation is 0x2

hw/xbox/nv2a/pgraph.c
abaire: No problem, reverted. I left a TODO to verify that it's sufficient as a hint in case somebody finds a graphical corruption case. I'm guessing that 2D HW isn't used much in practice so hopefully this will never be an issue.

abaire:
> Tested Blade 2 and Rogue Ops and Die Hard Vendetta and there bilt operation is 0x2

Thanks again!

hw/xbox/nv2a/pgraph.c
```diff
@@ -923,10 +927,12 @@ int pgraph_method(NV2AState *d, unsigned int subchannel,
                     + (image_blit->out_y + y) * context_surfaces->dest_pitch
                     + image_blit->out_x * bytes_per_pixel;
 
-                memmove(dest_row, source_row,
-                        image_blit->width * bytes_per_pixel);
+                memmove(dest_row, source_row, bytes_per_row);
             }
 
+            // TODO: Verify that this dirty tracking is sufficient.
```
mborgerson: Yes, this is incorrect. The actual dirtied memory range should account for destination pitch. Ideally the only problem we should have here is over-dirty. Under-dirty is definitely unacceptable. Let's fix the size calculation and drop the comment

abaire: sgtm, done (but please double check my math, if you don't mind).

abaire: (I think I had an off by one, hopefully correct now)

abaire:
The build failure doesn't immediately seem to be related to this change; I've seen this happen with macOS deps building on an M1 and had to dump the downloaded libs and re-fetch. Is it possible the cache is in a bad state?

```
ERROR: sizeof(size_t) doesn't match GLIB_SIZEOF_SIZE_T.
       You probably need to set PKG_CONFIG_LIBDIR
       to point to the right pkg-config files for your
       build target
```

mborgerson:
> The build failure doesn't immediately seem to be related to this change; I've seen this happen with macOS deps building on an M1 and had to dump the downloaded libs and re-fetch. Is it possible the cache is in a bad state?

Yeah, it's not an issue with your PR; just need to fix the build

hw/xbox/nv2a/pgraph.c
```diff
@@ -923,25 +927,22 @@ int pgraph_method(NV2AState *d, unsigned int subchannel,
                     + (image_blit->out_y + y) * context_surfaces->dest_pitch
                     + image_blit->out_x * bytes_per_pixel;
 
-                memmove(dest_row, source_row,
-                        image_blit->width * bytes_per_pixel);
+                memmove(dest_row, source_row, bytes_per_row);
             }
 
+            uint32_t dest_start = image_blit->out_y * context_surfaces->dest_pitch + image_blit->out_x;
```
mborgerson: Nit: type for this should be `hwaddr`

hw/xbox/nv2a/pgraph.c
```diff
@@ -910,9 +910,13 @@ int pgraph_method(NV2AState *d, unsigned int subchannel,
 
             SurfaceBinding *surf_dest = pgraph_surface_get(d, dest - d->vram_ptr);
             if (surf_dest) {
+                if (image_blit->height < surf_dest->height || image_blit->width < surf_dest->width) {
```
mborgerson: nit: hard wrap at column 80

hw/xbox/nv2a/pgraph.c
```diff
@@ -923,25 +927,22 @@ int pgraph_method(NV2AState *d, unsigned int subchannel,
                     + (image_blit->out_y + y) * context_surfaces->dest_pitch
                     + image_blit->out_x * bytes_per_pixel;
 
-                memmove(dest_row, source_row,
-                        image_blit->width * bytes_per_pixel);
+                memmove(dest_row, source_row, bytes_per_row);
             }
 
+            uint32_t dest_start = image_blit->out_y * context_surfaces->dest_pitch + image_blit->out_x;
+            uint32_t dirty_size = ((image_blit->height - 1) * context_surfaces->dest_pitch)
```
mborgerson: * Nit: unnecessary parens
* We should probably check that height is never 0

mborgerson:
> > The build failure doesn't immediately seem to be related to this change; I've seen this happen with macOS deps building on an M1 and had to dump the downloaded libs and re-fetch. Is it possible the cache is in a bad state?
> 
> Yeah, it's not an issue with your PR; just need to fix the build

Build should be fixed now if you rebase this patch. I'll merge after the nits are fixed

hw/xbox/nv2a/pgraph.c
```diff
@@ -923,25 +927,22 @@ int pgraph_method(NV2AState *d, unsigned int subchannel,
...
+                    + image_blit->width * bytes_per_pixel;
             memory_region_set_client_dirty(d->vram,
-                                           (dest - d->vram_ptr),
-                                           image_blit->height
-                                           * image_blit->width
-                                           * bytes_per_pixel,
+                                           (dest - d->vram_ptr + dest_start),
```
mborgerson: nit: While refactoring this, may as well move the dest addr calculation out

hw/xbox/nv2a/pgraph.c
abaire: Fixed.

hw/xbox/nv2a/pgraph.c
abaire: Done

hw/xbox/nv2a/pgraph.c

hw/xbox/nv2a/pgraph.c
abaire: Done. The hardware gracefully handles this and the `width == 0` case (seems equivalent to a nop), so I added a check that bypasses the copy entirely if appropriate.

abaire:
@mborgerson - just checking to make sure you're not still waiting on any additional changes that I might've missed

mborgerson:
> @mborgerson - just checking to make sure you're not still waiting on any additional changes that I might've missed

Thanks for your patience. A bit under the weather now, but I'll return to this and other PRs within the next couple of days or so

mborgerson:
Thanks!

Merged with #624

PR #570, opened 2021-11-24
nv2a: Implement BLEND_AND image blit operation.

### LINKED ISSUE #324: Unhandled NV_IMAGE_BLIT methods

abaire:
CPU-based implementation of BLEND_AND hardware op.

Fixes #324 for opcode 0x02 (which appears to be the case for all known failures)

### DIFF
--- a/hw/xbox/nv2a/pgraph.c
+++ b/hw/xbox/nv2a/pgraph.c
@@ -854,97 +854,111 @@ int pgraph_method(NV2AState *d, unsigned int subchannel,
         image_blit->width = parameter & 0xFFFF;
         image_blit->height = parameter >> 16;
 
-        /* I guess this kicks it off? */
-        if (image_blit->operation == NV09F_SET_OPERATION_SRCCOPY) {
+        if (image_blit->width && image_blit->height) {
+            if (image_blit->operation == NV09F_SET_OPERATION_SRCCOPY) {
 
-            NV2A_GL_DPRINTF(true, "NV09F_SET_OPERATION_SRCCOPY");
+                NV2A_GL_DPRINTF(true, "NV09F_SET_OPERATION_SRCCOPY");
 
-            pgraph_update_surface(d, false, true, true);
+                pgraph_update_surface(d,
+                                      false,
+                                      true,
+                                      true);
 
-            ContextSurfaces2DState *context_surfaces = context_surfaces_2d;
-            assert(context_surfaces->object_instance
-                    == image_blit->context_surfaces);
+                ContextSurfaces2DState *context_surfaces = context_surfaces_2d;
+                assert(context_surfaces->object_instance
+                               == image_blit->context_surfaces);
 
-            unsigned int bytes_per_pixel;
-            switch (context_surfaces->color_format) {
-            case NV062_SET_COLOR_FORMAT_LE_Y8:
-                bytes_per_pixel = 1;
-                break;
-            case NV062_SET_COLOR_FORMAT_LE_R5G6B5:
-                bytes_per_pixel = 2;
-                break;
-            case NV062_SET_COLOR_FORMAT_LE_A8R8G8B8:
-            case NV062_SET_COLOR_FORMAT_LE_X8R8G8B8:
-            case NV062_SET_COLOR_FORMAT_LE_Y32:
-                bytes_per_pixel = 4;
-                break;
-            default:
-                fprintf(stderr, "Unknown blit surface format: 0x%x\n", context_surfaces->color_format);
+                unsigned int bytes_per_pixel;
+                switch (context_surfaces->color_format) {
+                    case NV062_SET_COLOR_FORMAT_LE_Y8:bytes_per_pixel = 1;
+                        break;
+                    case NV062_SET_COLOR_FORMAT_LE_R5G6B5:bytes_per_pixel = 2;
+                        break;
+                    case NV062_SET_COLOR_FORMAT_LE_A8R8G8B8:
+                    case NV062_SET_COLOR_FORMAT_LE_X8R8G8B8:
+                    case NV062_SET_COLOR_FORMAT_LE_Y32:bytes_per_pixel = 4;
+                        break;
+                    default:
+                        fprintf(stderr,
+                                "Unknown blit surface format: 0x%x\n",
+                                context_surfaces->color_format);
+                        assert(false);
+                        break;
+                }
+
+                hwaddr source_dma_len, dest_dma_len;
+                uint8_t *source, *dest;
+
+                source = (uint8_t *) nv_dma_map(
+                        d, context_surfaces->dma_image_source, &source_dma_len);
+                assert(context_surfaces->source_offset < source_dma_len);
+                source += context_surfaces->source_offset;
+
+                dest = (uint8_t *) nv_dma_map(d,
+                                              context_surfaces->dma_image_dest,
+                                              &dest_dma_len);
+                assert(context_surfaces->dest_offset < dest_dma_len);
+                dest += context_surfaces->dest_offset;
+
+                NV2A_DPRINTF("  - 0x%tx -> 0x%tx\n", source - d->vram_ptr,
+                             dest - d->vram_ptr);
+
+
+                // FIXME: Blitting from part of a surface
+
+                SurfaceBinding *surf_src = pgraph_surface_get(
+                        d, source - d->vram_ptr);
+                if (surf_src) {
+                    pgraph_download_surface_data(d, surf_src, true);
+                }
+
+                SurfaceBinding *surf_dest = pgraph_surface_get(
+                        d, dest - d->vram_ptr);
+                if (surf_dest) {
+                    if (image_blit->height < surf_dest->height ||
+                            image_blit->width < surf_dest->width) {
+
+                        pgraph_download_surface_data_if_dirty(d, surf_dest);
+                    }
+                    surf_dest->upload_pending = true;
+                }
+
+                uint32_t bytes_per_row = image_blit->width * bytes_per_pixel;
+                int y;
+                unsigned int source_pitch = context_surfaces->source_pitch;
+                unsigned int dest_pitch = context_surfaces->dest_pitch;
+                for (y = 0; y < image_blit->height; y++) {
+                    uint8_t *source_row = source
+                            + (image_blit->in_y + y) * source_pitch
+                            + image_blit->in_x * bytes_per_pixel;
+
+                    uint8_t *dest_row = dest
+                            + (image_blit->out_y + y) * dest_pitch
+                            + image_blit->out_x * bytes_per_pixel;
+
+                    memmove(dest_row, source_row, bytes_per_row);
+                }
+
+                hwaddr dest_start = image_blit->out_x +
+                        image_blit->out_y * context_surfaces->dest_pitch;
+                hwaddr dirty_start = dest - d->vram_ptr + dest_start;
+                uint32_t dirty_size = image_blit->width * bytes_per_pixel +
+                        (image_blit->height - 1) * context_surfaces->dest_pitch;
+                memory_region_set_client_dirty(d->vram,
+                                               dirty_start,
+                                               dirty_size,
+                                               DIRTY_MEMORY_VGA);
+                memory_region_set_client_dirty(d->vram,
+                                               dirty_start,
+                                               dirty_size,
+                                               DIRTY_MEMORY_NV2A_TEX);
+            } else {
+                fprintf(stderr,
+                        "Unknown blit operation: 0x%x\n",
+                        image_blit->operation);
                 assert(false);
-                break;
             }
-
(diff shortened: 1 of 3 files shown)


hw/xbox/nv2a/pgraph.c
```diff
@@ -741,6 +741,121 @@ static const MethodFunc pgraph_kelvin_method_handlers[0x800] = {
...
+                    + image_blit->in_x * bytes_per_pixel;
+
+            uint8_t *dest_row = dest
+                    + (image_blit->out_y + y) * context_surfaces->dest_pitch
+                    + image_blit->out_x * bytes_per_pixel;
+
+            for (int x = 0; x < bytes_per_row; ++x, ++source_row, ++dest_row) {
+                uint32_t a = *source_row * beta_mult;
+                uint32_t b = *dest_row * inv_beta_mult;
+                *dest_row = (a + b) / max_beta_mult;
```
mborgerson: I believe you have a test case for image blitting, right? Does your test confirm this behavior? I'm not fully confident about this blending logic and would like to double check if you have it handy

hw/xbox/nv2a/pgraph.c
```diff
@@ -741,6 +741,121 @@ static const MethodFunc pgraph_kelvin_method_handlers[0x800] = {
...
+        surf_dest->upload_pending = true;
+    }
+
+    if (image_blit->operation == NV09F_SET_OPERATION_SRCCOPY) {
+        NV2A_GL_DPRINTF(true, "NV09F_SET_OPERATION_SRCCOPY");
+        NV2A_DPRINTF("  - 0x%tx -> 0x%tx\n", source - d->vram_ptr,
+                     dest - d->vram_ptr);
+
+        uint32_t bytes_per_row = image_blit->width * bytes_per_pixel;
+        int y;
```
mborgerson: nit: While we are here, let's merge the `int y` declaration into the `for` below. Repeated for next op.

hw/xbox/nv2a/pgraph.c
```diff
@@ -741,6 +741,121 @@ static const MethodFunc pgraph_kelvin_method_handlers[0x800] = {
...
+static void perform_image_blit(NV2AState *d, ContextSurfaces2DState *context_surfaces, ImageBlitState *image_blit, BetaState *beta)
+{
+    pgraph_update_surface(d, false, true, true);
+
+    assert(context_surfaces->object_instance
+                   == image_blit->context_surfaces);
+
+    unsigned int bytes_per_pixel;
+    switch (context_surfaces->color_format) {
+        case NV062_SET_COLOR_FORMAT_LE_Y8:bytes_per_pixel = 1;
```
mborgerson: Formatting got messed up here, please retain newline after `:`, repeated below

hw/xbox/nv2a/pgraph.c
```diff
@@ -741,6 +741,121 @@ static const MethodFunc pgraph_kelvin_method_handlers[0x800] = {
     } \
     *num_words_consumed = param_iter;
 
+static void perform_image_blit(NV2AState *d, ContextSurfaces2DState *context_surfaces, ImageBlitState *image_blit, BetaState *beta)
+{
+    pgraph_update_surface(d, false, true, true);
+
+    assert(context_surfaces->object_instance
```
mborgerson: nit: Let's unwrap this line

hw/xbox/nv2a/pgraph.c
```diff
@@ -799,6 +915,23 @@ int pgraph_method(NV2AState *d, unsigned int subchannel,
...
+        break;
+
+    case NV012_SET_BETA:
+        if (parameter & 0x80000000) {
+            beta->beta = 0;
+        } else {
+            // The parameter is a signed fixed-point number with a sign bit and
+            // 31 fractional bits. Note that negative values are clamped to 0,
+            // and only 8 fractional bits are actually implemented in hardware.
+            beta->beta = parameter & 0x7f800000;
```
mborgerson, line 989: Does your test confirm these claims? I would like to double check that:
* Indeed only 8 fractional bits are available
* Negatives are indeed clamped
* Sign bit is actually a sign bit or a signed integer component (+/- or 0,-1)

hw/xbox/nv2a/pgraph.c
```diff
@@ -741,6 +741,121 @@ static const MethodFunc pgraph_kelvin_method_handlers[0x800] = {
     } \
     *num_words_consumed = param_iter;
 
+static void perform_image_blit(NV2AState *d, ContextSurfaces2DState *context_surfaces, ImageBlitState *image_blit, BetaState *beta)
```
mborgerson: Let's rename to something like `pgraph_image_blit`. Let's also simplify this function prototype to just retrieve these pointers from `NV2AState *d`.

hw/xbox/nv2a/pgraph.c
```diff
@@ -741,6 +741,121 @@ static const MethodFunc pgraph_kelvin_method_handlers[0x800] = {
...
+                    + image_blit->out_x * bytes_per_pixel;
+
+            memmove(dest_row, source_row, bytes_per_row);
+        }
+    } else if (image_blit->operation == NV09F_SET_OPERATION_BLEND_AND) {
+        NV2A_GL_DPRINTF(true, "NV09F_SET_OPERATION_BLEND_AND");
+        NV2A_DPRINTF("  - 0x%tx -> 0x%tx\n", source - d->vram_ptr,
+                     dest - d->vram_ptr);
+
+        // TODO: Optimize. Ideally this should be done via OpenGL.
```
mborgerson: Comment applies to the whole function

hw/xbox/nv2a/pgraph.c
```diff
@@ -741,6 +741,121 @@ static const MethodFunc pgraph_kelvin_method_handlers[0x800] = {
...
+        fprintf(stderr, "Unknown blit operation: 0x%x\n", image_blit->operation);
+        assert(false);
+    }
+
+    // TODO: Verify that this dirty tracking is sufficient.
+    // For sub-images of < surface width this can flag regions that are
+    // not actually dirty and miss regions that are.
+    memory_region_set_client_dirty(d->vram,
+                                   (dest - d->vram_ptr),
+                                   image_blit->height
```
mborgerson: Because this moved, please update these calls to reflect requested change from your other PR, w.r.t. size of dirtied region and comment, when you rebase this second patch

hw/xbox/nv2a/pgraph.c

hw/xbox/nv2a/pgraph.c

hw/xbox/nv2a/pgraph.c
abaire: Oops, sorry. CLion likes to reformat automatically on paste and when modifying nearby lines :(

hw/xbox/nv2a/pgraph.c
abaire: Happily done :)

hw/xbox/nv2a/pgraph.c
abaire: Moved.

hw/xbox/nv2a/pgraph.c
abaire: Yup, will do as soon as the first PR is merged.

hw/xbox/nv2a/pgraph.c
abaire: I actually capped the values in the test code as well as I assumed the envytools doc was correct. I'll adjust and re-test (once I'm back with my hardware) and update if necessary.

hw/xbox/nv2a/pgraph.c
abaire: The macro level behavior does match the hardware (i.e., this is a simple blend with the beta as the source factor). See response to your comment below wrt. the capping, which would change the math if more than 8 bits are utilized.

mborgerson: I agree this does simple blending, but what I'm not sure about is, for instance:
* whether or not this must take into account source and/or destination alpha component
* whether or not negative values are actually supported, contrary to previous assumptions

Can you please test this and find out

hw/xbox/nv2a/pgraph.c
abaire: Also as some evidence that the doc may be correct, Dr. Muto uses `0x7d400000` as beta.

`nv2a: pgraph method (4): 0x12 -> 0x0300 (0x7d400000)`

mborgerson: `0x7d400000` suggests at least 9 fractional bits

abaire: Ha, good call. I'm going to WIP this until I can update the tests on HW.

hw/xbox/nv2a/pgraph.c
abaire: Some initial findings:

- As far as I can see, the hardware throws an exception when trying to configure the surface context with an alpha-enabled color format (`NV062_SET_COLOR_FORMAT_LE_A8R8G8B8`) when using operation `BLEND_AND`. I've tried both blitting directly to channel 11 (framebuffer) and into an interim memory buffer, both fail with a "format exception error" as reported by the pbkit handler. Interestingly/annoyingly the exception is only raised when the `NV09F_SIZE` parameter is pushed, but this matches the comment in xemu's pgraph.c - I'm guessing the failure is only detected once the operation is actually kicked off.
- `SRCCOPY` seems to ignore the source alpha but works in both XRGB and ARGB surface modes as well as when bouncing through the compositing buffer I tried for the `BLEND_AND` testing. 
- Using a negative beta (high bit set) does result in a fully transparent image on HW.

I still need to do a bit more work to verify whether the HW actually supports more than 8 bits, will try to get back around to that later today.

abaire: Found a bit more time, it looks like anything past 8 bits is indeed ignored. Results from hardware are at https://github.com/abaire/nxdk_pgraph_tests_golden_results

Key tests:
- [beta `0x007FFFFF`](https://github.com/abaire/nxdk_pgraph_tests_golden_results/blob/main/ImageBlt_BLENDAND_XRGB_B007FFFFF.png) - all pixels are the same color (the clear color).
- [beta `0x00800000`](https://github.com/abaire/nxdk_pgraph_tests_golden_results/blob/main/ImageBlt_BLENDAND_XRGB_B00800000.png) - most of the boxes from the [test image](https://github.com/abaire/nxdk_pgraph_tests/blob/main/resources/image_blit/TestImage.png) (note, only a subimage from this image is used, slightly larger than the green grid) have slightly different color from the background.
- [`0x44400000`](https://github.com/abaire/nxdk_pgraph_tests_golden_results/blob/main/ImageBlt_BLENDAND_XRGB_B44400000.png) and [`0x444FFFFF`](https://github.com/abaire/nxdk_pgraph_tests_golden_results/blob/main/ImageBlt_BLENDAND_XRGB_B444FFFFF.png) have 0 pixel differences.

The differences can be confirmed w/ Gimp or imagemagick's `compare` utility (or presumably any image diff that does pixel comparison and not perceptual).

E.g.: `compare -verbose -metric AE ImageBlt_BLENDAND_XRGB_B44400000.png ImageBlt_BLENDAND_XRGB_B444FFFFF.png difference.png`

abaire: Did a bunch more testing while looking into #578.

- It looks like `X8R8G8B8` always sets alpha to `0xFF` on hardware. This is true for `BLEND_AND` and `SRCCOPY` operations. This means that `SRCCOPY` is slightly incorrect (I'll just roll that fix into this PR unless you feel strongly about pulling it out).
- I still have not been able to get a `BLEND_AND` with an `ARGB` surface context working, so I'll skip the blend on the alpha channel since it'll be set to 0xFF (or 0x00 for `ZRGB`, will do that in a followup).

abaire: Updated. I patch up `SRCCOPY` after the `memmove` but added a separate loop for the `BLEND_AND` to avoid useless divides. Let me know if you'd prefer to just have one patch-after path instead. Given that these ops don't seem to be used often I guess it's arguably premature optimization.

mborgerson:
Thanks!

Merged with #624

Issue #164, opened 2021-01-24
[Feature Request] I/O Throttling


BiatuAutMiahn:
Can we limit the rate or the IOPS for the disc drive and possily the HDD, with a configurable option in the settings?

GXTX:
What's the use case?

BiatuAutMiahn:
Faithful to OG xbox experience. IE, when booting Xbox logo shows before xbe is loaded from disc, then microsoft logo appears when title is executed. Halo 2 for example normally has a loading screen before main menu, and before each level.

mborgerson:
Thanks, you may be the first to have requested that I make xemu slower. :) It is unlikely that xemu will ever match the exact timing characteristics of the mechanical DVD/HDD drives, but I do think this is an interesting idea and we can try. QEMU does support such throttling, so we can experiment with it and consider adding an option.

abembritishav:
I did a test with PAL version of Halo CE. Since the release already has issue with PAL 60Hz mode even on real console, the game seems to have some timing issue where speeches may advance too quickly with no pauses in-between. This can be noticed right away on the start with Chief in the cryobay. And throttling IO speed fixed the issue, making the dialogue paced normally.

During the test and playthrough, both the ISO and hard drive images are on the very fast M.2 SSD. Nonetheless, I believe the issue can even occur with spinning hard disk drives.

PR #589, opened 2021-12-05
vl: Add settings to throttle disc and HDD speeds

antangelo:
Adds two config options (one for HDD and one for DVD-ROM) to throttle the bytes/sec IO speed of the drives. Default values are set to 0, which is unthrottled.

Solves #164


BiatuAutMiahn:
Can you initialize defaults to what you would expect on real hw? ie 20/6 HDD/DVDROM

antangelo:
> Can you initialize defaults to what you would expect on real hw? ie 20/20 HDD/DVDROM

20 isn't accurate to real hardware. There are plans to measure real hw performance and offer defaults based on the results.

mborgerson:
Thanks for the patch! A few things before merge:

* The config system has been replaced since this patch was submitted, which I know is a bit of a pain; but I think you'll find using the new system to be less painful. Can you please update the patch to use the new config system?
* I've added a stub config item in config_spec.yml/mainmenu.cc for this feature. I think reducing it to a simple on/off would be preferred for user experience, with the actual throttle rates being configurable in the config file but perhaps not necessary in the main menu.
* I'm not sure what defaults would be good approximations of hardware limits, but I'd like for us to figure that out before merging.

PR #130, opened 2021-01-02
xbox: Fix some PCI revs to match retail 1.0 Xbox

GXTX:
While looking into some stuff with Cromwell I noticed that during init it does some checking with some PCI stuff. This byte on a retail 1.0 box reports as `0xB2` while xemu reports this as `0xD4`. If I follow the logic right in Cromwell it thinks xemu is a 1.1+ system.

Not sure if anything else checks for this.

https://github.com/XboxDev/cromwell/blob/master/drivers/pci/pci.c#L325

Test code:
```
#include <hal/debug.h>
#include <hal/video.h>
#include <windows.h>
#include <hal/io.h>

uint8_t PciReadByte(unsigned int bus, unsigned int dev,
                    unsigned int func, unsigned int reg_off)
{
    uint32_t base_addr = 0x80000000;
    base_addr |= ((bus & 0xFF) << 16);      // bus #
    base_addr |= ((dev & 0x1F) << 11);      // device #
    base_addr |= ((func & 0x07) << 8);      // func #

    IoOutputDword(0xcf8, (base_addr + (reg_off & 0xfc)));
    return IoInputByte(0xcfc + (reg_off & 3));
}

int main(void)
{
    XVideoSetMode(640, 480, 32, REFRESH_DEFAULT);

    uint8_t rev;
    rev = PciReadByte(0, 1, 0, 0x8);
    debugPrint("Mystery 1: 0x%2X\n", rev);

    Sleep(10000);
    return 0;
}
```

xbox:
image

xemu before:
image

PR #591, opened 2021-12-06
xbox: Change PCI IDs for USB and IDE controllers to match Xbox

GXTX:
I don't have my Xbox-Linux drive image (nor a working Xbox) any longer so I can't 100% confirm but going by my past tests should be enough. Although we are still missing 3 devices on the bus (memory controller, AC97 modem, pci-to-pci bridge).

Checking out my original patch (#130) on IDs, Matt had suggested getting a proper dump of the config space of all of the devices, someone should do that *for completeness*.


hw/usb/hcd-ohci-pci.c
```diff
@@ -134,8 +134,14 @@ static void ohci_pci_class_init(ObjectClass *klass, void *data)
...
+    k->vendor_id = PCI_VENDOR_ID_NVIDIA;
+    k->device_id = PCI_DEVICE_ID_NVIDIA_NFORCE_USB;
+    k->revision  = 0xB1;
+#else
     k->vendor_id = PCI_VENDOR_ID_APPLE;
     k->device_id = PCI_DEVICE_ID_APPLE_IPID_USB;
+#endif
     k->class_id = PCI_CLASS_SERIAL_USB;
     set_bit(DEVICE_CATEGORY_USB, dc->categories);
     dc->desc = "Apple USB Controller";
```
binarymaster, line 150:
```suggestion
#ifdef XBOX
    dc->desc = "nForce USB Controller";
#else
    dc->desc = "Apple USB Controller";
#endif
```

GXTX: Is this presented to the OS? If so is that the correct description?

binarymaster: Not sure, it might be some internal QEMU thing. Just suggested it for the consistence.

GXTX: I dumped the USB reg area and didn't see any plain text like this. I assume this isn't passed onto the OS.

mborgerson: It is used in device infrastructure (e.g. when printing out available devices). I agree with @binarymaster's suggested patch; Let's update the device description string while we are updating the id for consistency.

mborgerson (changes requested):
See inline

hw/usb/hcd-ohci-pci.c
GXTX: Fixed

PR #593, opened 2021-12-06
ui: 4:3 scale option

octopoulos:
Many games run in 4:3 mode and not 16:9, so having this option allows to have a proper aspect ratio.

image


7oxicshadow:
Does this differ from "Scale" then? In other words, are there some games that are still the wrong aspect when "Scale" is selected? If I remember correctly the "Scale" option uses the framebuffer ratio which I assume is 4:3 by default?

octopoulos:
In Ninja Gaiden Black, the framebuffer is 720x480, so that's a 1.5 AR, not 4:3, here I'm showing it in the window title:
image

Notice how Ryu looks fat (or too short)? However, the UI is good, but in this game, the UI is made to look good in WS.

This is with the 4:3 scaling, now he looks normal. Trust me, I have played 300+ hours of this game on the original Xbox :) And also the Sigma versions etc, so I know how he should look.
image

octopoulos:
2 more examples: those things should be circles, but with "scale", they're not because of the 1.5 AR.
image

Now with 4:3 scaling, they are proper circles:
image

7oxicshadow:
Nice, that's a good spot.

One for Matt I guess but it kind of looks like the "Scale" output isn't required any-more if this is accepted.

mborgerson:
Nice, thanks

Issue #598, opened 2021-12-12
C.A.T: Cyber Attack Team: Texture sampling Issue when upscaling


Triticum0:
### Title

https://xemu.app/titles/4d450001/#C-A-T-Cyber-Attack-Team

### Bug Description

When playing this game the animated parts of the characters are clearly viable when there not meant to be visible.

Xbox:
![Screenshot from 2021-12-12 18-46-48](https://user-images.githubusercontent.com/71127352/145725390-846ab4ca-7839-4d8c-a294-4d3cdc74f4a7.png)

Xemu(Look at the face):
![Screenshot from 2021-11-24 14-12-40](https://user-images.githubusercontent.com/71127352/145725382-3bb1060d-42df-4d7a-8ee3-566e708138f6.png)

### Expected Behavior

This should not occurred. 

### xemu Version

Version: 0.6.2
Branch: master
Commit: 69ceec4
Date: Wed Nov 24 02:59:42 UTC 2021
Last-known Working: Unknown

### System Information

Field | Value
-- | --
OS | Ubuntu 20.04.3 LTS
CPU | Intel(R) Pentium(R) CPU 3825U @ 1.90GHz
Graphics Device | Mesa Intel(R) HD Graphics (BDW GT1)
Graphics Driver | Drivers: 4.6 (Core Profile) Mesa 21.0.3

### Additional Context
This games should be playable now
Edit: Energy Indicator was fixed a while back

Triticum0:
pr implement screen coordinate rounding to 4 bit fractional precision Fixes the Issue at 1x resolution.

![Image](https://github.com/user-attachments/assets/3004b3ce-58f1-4b39-a40c-176e3d3c421b)

Triticum0:
Close as metioned as part of #979 so now a dupe

Issue #469, opened 2021-10-03
Title Name: Elder Scrolls III, The Morrowind - Game of the Year Edition (USA)


databoose:
### Title

https://xemu.app/titles/42530005/#The-Elder-Scrolls-III-Morrowind-Game-of-the-Year-Edition

### Bug Description

Cannot see anything in the game, audio is fully working, and the game is still otherwise running, but there is a black overlay with red corners (presumably the red edges are supposed to be when the player takes damage?)

As far as i can tell, this bug never goes away and remains the entire game, rendering it effectively unplayable.

![Screenshot at 2021-10-03 10-59-23](https://user-images.githubusercontent.com/7876395/135769328-35b2285a-1de4-4c0b-912a-933540ee7211.png)

Steps to reproduce : 

1. Start a new game
2. Skip intro with start button
3. Bug will appear

### Expected Behavior

Red corners are not supposed to appear at all, the screen is supposed to fade in from black from the video intro to fully visible (it is very quick but can be seen in the video below)

OG xbox footage for reference : https://www.youtube.com/watch?v=nhP49PlN6HY&t=45s

### xemu Version

0.6.1-12-ga8c73197da

### System Information

OS: Ubuntu 20.04.3 LTS 64 bit
CPU : i5 10600k 4.8GHz (stock clock)
GPU : GTX 1660ti
GPU Driver : 460.91.03

abaire:
I think the behavior may have improved a bit since this was filed, I can see the game world being rendered with some significant artifacts (a red poly instead of what I guess is a shadow / stencil buffer behind the NPC?)

![Screenshot_20211109_125259](https://user-images.githubusercontent.com/448413/141003568-e82706b3-0d99-4787-b3ce-11fcab7b10b4.png)


OS: Kubuntu 21.04
CPU: i7 6700K @ 4GHz
GPU: GTX 1070
GPU Driver: 470.82.00

Version: 0.6.1-33-gc8c8cacc6e
Branch:  fixes_yuv_colorspace_conversion
Commit:  c8c8cacc6eb0cf0effcf888fd0d2ef1e4e1822aa
Date:    Tue Nov  9 08:45:20 PM UTC 2021

Triticum0:
Yeah shadow / stencil buffer implementation is broken in a lot of games.

Triticum0:
@abaire I asked OP to retest it on the discord and behave hasn't change so it might be version differences causing the bug.

databoose:
[...]

Issue #584, opened 2021-12-02
Rolling menus are broken


Triticum0:
### Title

https://xemu.app/titles/53430007/#Rolling

### Bug Description

When pausing the game you end up getting a black screen.
I tested on my windows and Linus laptops to confirm it not an Intel specific issue.

https://user-images.githubusercontent.com/71127352/144518222-0d89356f-ede0-4e0d-8ea7-f3fe25ff31c5.mov

### Expected Behavior

The pause menu should be visable

### xemu Version

Version: 0.6.2
Branch: master
Commit: 69ceec4
Date: Wed Nov 24 02:59:42 UTC 2021
Last-known Working: Unknown

### System Information

Field | Value
-- | --
OS | Ubuntu 20.04.3 LTS
CPU | Intel(R) Pentium(R) CPU 3825U @ 1.90GHz
Graphics Device | Mesa Intel(R) HD Graphics (BDW GT1)
Graphics Driver | Drivers: 4.6 (Core Profile) Mesa 21.0.3

### Additional Context
This also affect the main menu after finishing a mission making the game unplayable on-till this issue fixed. Here a renderdoc https://drive.google.com/file/d/1qE3EBNy7gEXjmXY40yY7dI8aWZqDItAe/view?usp=sharing

abaire:
Hey @Triticum0 would you mind grabbing an nv2a log from this game? I'm curious as to whether it may be the same problem as #670 (I'd be especially interested in any "unhandled" log lines)

Triticum0:
Here [log.zip](https://github.com/mborgerson/xemu/files/7947210/log.zip)

Triticum0:
Look at the log it seems like it

Triticum0:
I used the wrong build when creating the log. Here [log.zip](https://github.com/mborgerson/xemu/files/7947248/log.zip)

abaire:
Thanks! I don't see the same unhandled commands in that log so it seems that they may be unrelated issues.

SlyCooperReloadCoded:
This is still an issue on v0.7.134.  Something to note is that this isn't just a pause menu issue.  If you start the game in Practice mode, you'll see the game for 1 frame then the screen will turn black both during gameplay in the pause menu:

![Capture2](https://github.com/user-attachments/assets/b399f018-692a-48ce-97d7-afad0afd2487)

However, if you start the game in Career mode, it's not broken during gameplay, but it's still broken in the pause menu:

[...]

Issue #547, opened 2021-11-12
Seablade and Nba jam menu is not fully rendered


Triticum0:
### Title

https://xemu.app/titles/544d0009/#SeaBlade
https://xemu.app/titles/41430013/#NBA-Jam

### Bug Description

The text on the menu screen for both games are missing.
xemu:
![Screenshot from 2021-11-02 22-48-01](https://user-images.githubusercontent.com/71127352/141383159-517f2d88-1f77-4cdd-b5e9-e6b9758cc28f.png)

xemu:
![Screenshot from 2021-11-02 22-48-27](https://user-images.githubusercontent.com/71127352/141383256-5a0d826a-32e4-4ad8-9d59-3d530d54fdde.png)

xemu:
![Screenshot from 2021-11-02 22-48-31](https://user-images.githubusercontent.com/71127352/141383263-6c3a3dc9-a4d9-41ca-8b81-edeced543d21.png)

xemu:
![Screenshot from 2021-11-11 23-30-03](https://user-images.githubusercontent.com/71127352/141383478-fc8dd2d9-80f4-4f6e-9fb1-e83364794bca.png)

### Expected Behavior

Text should be rendered 
xbox:
![Screenshot from 2021-11-11 23-52-06](https://user-images.githubusercontent.com/71127352/141385456-3669e142-d431-4c3a-840d-55b3bc44f0d4.png)

xbox:
![Screenshot from 2021-11-11 23-52-14](https://user-images.githubusercontent.com/71127352/141385470-337b8c37-e27a-4502-a47d-6ecb2af7239f.png)

xbox:
![Screenshot from 2021-11-11 23-52-35](https://user-images.githubusercontent.com/71127352/141385486-8ef52b53-da25-49bd-95bb-c5bfb37a8625.png)

xbox:
![Screenshot from 2021-11-11 23-50-26](https://user-images.githubusercontent.com/71127352/141385769-4ba49ca6-fa89-48d7-a78e-22db0c2efc90.png)

### xemu Version

Version: 0.6.2
Branch: master
Commit: bbcd576
Date: Sat Oct 30 23:09:11 UTC 2021
Last-known Working: N/A

### System Information

Field | Value
-- | --
OS | Windows 10 64-bit (21H1)
CPU | AMD Ryzen 5 2600 Six-Core Processor @ 3.40GHz
Graphics Device | AMD Radeon RX 5700 XT
[...]

PR #599, opened 2021-12-18
nv2a: Implement SET_MATERIAL_ALPHA.

### LINKED ISSUE #242: [Bug] Some textures not rendered properly in MagiDeathFight!
Xemu   

Cxbx reloaded

Xemu

Cxbx reloaded

### xemu Version

Version: 0.6.2
Branch: master
Commit: bbcd576
Date: Sat Oct 30 23:09:11 UTC 2021
Last-known Working: N/A

### System Information

Field | Value
-- | --
OS | Windows 10 64-bit (21H1)

### LINKED ISSUE #397: Jet Set Radio Future: Map Bug
### Title Information

* JSRF: Jet Set Radio Future IG-024 [49470018]
* https://xemu.app/titles/49470018/

### Bug Description
The maps on the pause screen are filled in gray blocks in the shape of the map instead of showing elevations, graffiti, and such.

1. Start game.
2. Load into a level.
3. Pause the game and look at the map.

JSRF map 0

### Expected Behavior
The maps should look like this...

https://giantbomb1.cbsistatic.com/uploads/original/1/15777/2143183-img_20120304_054531.jpg

### Version

* Version: 0.5.4-21-gf03015e20e

### System Information

| Field | Value |
|-------|-------|
| OS			|Windows 10 Pro| 
| CPU			| AMD FX(tm)-8350 Eight-Core Processor| 
| Graphics Device	|Radeon RX 580| 
| Graphics Driver	| Radeon 21.4.1| 

### Additional Context
This isn't a huge issue during early levels since they are pretty small or linear. But later maps can be big sprawling multi-tiered places where finding things without the map might turn into a "needle in a haystack" situation.

### LINKED ISSUE #700: nv2a 0x97 -> 0x03b4 (3D_MATERIAL_FACTOR_FRONT_A?) unhandled
### Title

### Bug Description

According to the debug output, 0x97 -> 0x03b4 is unhandled.
In the [NV20 nouveau sources](https://gitlab.freedesktop.org/mesa/mesa/-/blob/184a690fca700aa9ef2ae2e58b0f4bd86562aa6d/src/mesa/drivers/dri/nouveau/nv20_3d.xml.h#L868), this instruction is `3D_MATERIAL_FACTOR_FRONT_A` (A as in "Alpha").

### Expected Behavior

The NV2A instruction 0x03b4 should be handled.

### xemu Version

0.6.2-55-gcaf84884c4

### System Information

N/A

### Additional Context

_No response_

abaire:
Adds handling for the `SET_MATERIAL_ALPHA` command and the `diffuse_src` in the fixed function lighting model.

The test suite is here: https://github.com/abaire/nxdk_pgraph_tests/blob/main/tests/material_alpha_tests.cpp

Fixes #397
Fixes #700 
Screenshot_20211217_232350
 
(From @Triticum0)  Fixes #242


abaire:
I don't fully understand the `NV_IGRAPH_XF_*` registers in `nv2a_regs.h` so I'm not sure if there's actually a HW register that should be used for the alpha value. If you can point me at docs or how to read back the register from HW I can probably figure it out. For now I just created a standalone value and piped it through to the FF shader.

Triticum0:
I did some testing and this pr also fixes #242. Thank abaire. Once #331 code is tested to be correct this game should be fully playable on Xemu. Great job!!!

Triticum0:
This could also fix #598,#584 and  #547 but have tested yet to confirm. Will get back to you in the coming days

abaire:
> This could also fix #598,#584 and #547 but have tested yet to confirm. Will get back to you in the coming days

Awesome, fingers crossed that this fixes a bunch of glitches and thank you very much for testing, as always!

Shoegzer:
Fantastic find @abaire!  Out of curiosity - the fixed screenshot shows the progress indicator (lower-left on screen) with a black background rather than light blue as it should be.  Is that part of a separate issue?

abaire:
> Fantastic find @abaire! Out of curiosity - the fixed screenshot shows the progress indicator (lower-left on screen) with a black background rather than light blue as it should be. Is that part of a separate issue?

@Shoegzer I believe it is separate. As far as I can see from the actual HW screens the light blue doesn't have any transparency. The fact that the icons also do not display makes me think there's something interesting about the way that's rendered.

It's probably worth filing a bug if you feel like it, though @Triticum0 might know if there's already a bug tracking similar black-out issues with UI elements (I'm inclined to guess this happens in more than one game).

Shoegzer:
Thanks, and that makes sense even if unfortunate.  Let's see what @Triticum0 comes up with but I'm happy to file an issue for this.

I tested JSRF against this PR myself and found the map issue is indeed fixed.  I tested a few other issues too that I thought had some possibility of being fixed (transparent floor near beginning of Halo 2 and menu elements in Star Wars: Knights of the Old Republic) but no change so I assume separate issues.  Still, this is nice progress.

Triticum0:
It only fixes the #598 where the colour of the characters energy gauge is missing. The other issues I mentioned aren't fixed. A bit off-topic but looking at the code, I think this issue from [#281](https://github.com/xqemu/xqemu/issues/281) may be useful in making the implementation accurate implementation as JayFoxRox never finished the work on it. It's mostly a hunch as I don't have any knowledge on the Xbox inter-workings, But either way, [jayfoxrox](https://github.com/JayFoxRox/xqemu) and [xqemu](https://github.com/xqemu/xqemu) repository has a lot of useful information that was never transferred to Xemu as issues/pr when it was forked. So you might find something useful Or ask matt as he is most knowledgeable with the code base.

abaire:
After extending my test cases a bit more, the behavior this introduces is not exactly right. I think this is tied to the fact that the current shader doesn't respect the diffuse source selection - if it's set to material, then this PR is correct, but if it's set to diffuse or specular the material alpha is erroneously applied. I think we should fix that before merging to avoid introducing regressions.

abaire:
Updated to respect the `diffuse_src` selection, which also fixes the output to more closely resemble what I see on HW (it's still not exactly right, but is much closer).

@Triticum0 if you feel up to another test pass - there's a reasonably good chance that this second change will either fix some more issues or break things that seemed to work before, totally unrelated to the alpha issue. In particular, surfaces that were untextured are the most likely to be obviously affected.

Note: This additional fix does not change the info panel in JSRF, it's still erroneously black and should be tracked as a separate issue.

Triticum0:
I got around to retesting and there are no regressions for #242 and #598 both are still fixed. Still leave #598 open as it still plaged with another issue. The rest of the games haven't changed so are properly unrelated to this issue.

abaire:
> I got around to retesting and there are no regressions for #242 and #598 both are still fixed. Still leave #598 open as it still plaged with another issue. The rest of the games haven't changed so are properly unrelated to this issue.

Perfect, thanks for your help!

abaire:
> I don't fully understand the `NV_IGRAPH_XF_*` registers in `nv2a_regs.h` so I'm not sure if there's actually a HW register that should be used for the alpha value. If you can point me at docs or how to read back the register from HW I can probably figure it out. For now I just created a standalone value and piped it through to the FF shader.

Digging around docs some more, I suspect that this may be `MATERIAL_FACTOR_A` in the LTC registers listed https://envytools.readthedocs.io/en/latest/hw/graph/xf/ctx.html?highlight=LTCTX#ltctx (I'm assuming the `NV_IGRAPH_XF_LTC1_rX` and `NV_IGRAPH_XF_LTC1_LX` #defines map to `LIGHT_X_LOCAL_RANGE` and `LIGHT_X_SPOT_CUTOFF_0` resp). It looks like none of the LTC3 values are currently used so I'm not sure if it'd be desirable to put in similar handling to LTC1 or just keep a separate var until more than one thing out of LTC3 is going to be used.

FrozenFish24:
This also fixes an issue in Ninja Gaiden Black where Ryu would incorrectly have his fiend textures applied at all times (#270).

Before:
ngb-wrong

After:
ngb

abaire:
> This also fixes an issue in Ninja Gaiden Black where Ryu would incorrectly have his fiend textures applied at all times (#270).

Great to hear, thanks for testing @FrozenFish24 !

abaire:
> Digging around docs some more, I suspect that this may be `MATERIAL_FACTOR_A` in the LTC registers listed https://envytools.readthedocs.io/en/latest/hw/graph/xf/ctx.html?highlight=LTCTX#ltctx (I'm assuming the `NV_IGRAPH_XF_LTC1_rX` and `NV_IGRAPH_XF_LTC1_LX` #defines map to `LIGHT_X_LOCAL_RANGE` and `LIGHT_X_SPOT_CUTOFF_0` resp). It looks like none of the LTC3 values are currently used so I'm not sure if it'd be desirable to put in similar handling to LTC1 or just keep a separate var until more than one thing out of LTC3 is going to be used.

Hey @mborgerson do you have a preference on ^^? 
I can update this to also add LTC3 (but I'd only be using the `MATERIAL_FACTOR_A` and don't know how to definitively prove that it actually is LTC3 that's being used).

mborgerson:
> > Digging around docs some more, I suspect that this may be `MATERIAL_FACTOR_A` in the LTC registers listed https://envytools.readthedocs.io/en/latest/hw/graph/xf/ctx.html?highlight=LTCTX#ltctx (I'm assuming the `NV_IGRAPH_XF_LTC1_rX` and `NV_IGRAPH_XF_LTC1_LX` #defines map to `LIGHT_X_LOCAL_RANGE` and `LIGHT_X_SPOT_CUTOFF_0` resp). It looks like none of the LTC3 values are currently used so I'm not sure if it'd be desirable to put in similar handling to LTC1 or just keep a separate var until more than one thing out of LTC3 is going to be used.
> 
> Hey @mborgerson do you have a preference on ^^? I can update this to also add LTC3 (but I'd only be using the `MATERIAL_FACTOR_A` and don't know how to definitively prove that it actually is LTC3 that's being used).

We can keep it in a special var for now

pigeon4195:
> Adds handling for the `SET_MATERIAL_ALPHA` command and the `diffuse_src` in the fixed function lighting model.
> 
> The test suite is here: https://github.com/abaire/nxdk_pgraph_tests/blob/main/tests/material_alpha_tests.cpp
> 
> Fixes #397 Screenshot_20211217_232350
> 
> (From @Triticum0) Fixes #242

The test suite link gives me a 404

abaire:
I refactored the test suite sources since this PR was created. New location is https://github.com/abaire/nxdk_pgraph_tests/blob/7643b865548572c7750575d25e6c85b69283d647/src/tests/material_alpha_tests.cpp#L1

hw/xbox/nv2a/shaders.c
```diff
@@ -488,12 +488,23 @@ GLSL_DEFINE(materialEmissionColor, GLSL_LTCTXA(NV_IGRAPH_XF_LTCTXA_CM_COL) ".xyz
 
         //FIXME: Do 2 passes if we want 2 sided-lighting?
```
MasonT8198, line 489: Would it be possible to also add this into the PR, seems easy enough

dracc: 2 sided lighting requires more NV2A methods to be implemented AFAICT. Please keep scope of PR limited. :)

MasonT8198: OIC, It looked simpler than that to me

abaire: > 2 sided lighting requires more NV2A methods to be implemented AFAICT. Please keep scope of PR limited. :)

+1, even if it was trivial it wouldn't really be related to material alpha, so it'd belong in a separate PR.

@MasonT8198 I'm curious if you know of a test case where 2-sided lighting is actually used? I've never seen it enabled in any of the nv2a logs I've examined (not that I've always paid a ton of attention). 
My copy of Forza will hopefully arrive next week and from #708 I see that @dracc came across it while log diving that game - would be a great help to add a screenshot or repro case to that bug if you know of one.

mborgerson:
Awesome, thanks!

hw/xbox/nv2a/shaders.c
MasonT8198: Forza is the only one I can think of off the top of my head, perhaps one of the games that used FFP_LIGHTSPOT would have 2 sided lighting, we can continue this discussion on the discord server

abaire:
Sounds like this fixes #469 as well

LazyTheHedgehog:
Hey this will sound extremely stupid, but because I know nothing about coding or whatever HOW the heck do I use this? I really wanna see the damn map in JSRF already and I genuinely got no idea how to do it

abaire:
The easiest thing to do would be to wait until the next xemu build, which
will contain this fix.

If you don't want to wait, you will need to learn to build from source:
https://xemu.app/docs/building-from-source/ Since the fix is already merge
to master you can follow those instructions exactly and it should be
included. I'd also suggest you ask these types of questions on Discord,
GitHub is not the best place to ask for this type of support.

ko81e24wy:
test NGB in chapter 12,fiend texture were fixed,stream effect still has issue
before
2022-02-07 02-06-27 的屏幕截图
after
2022-02-07 02-07-35 的屏幕截图

PR #600, opened 2021-12-19
macOS: Fix codesign after rpath changes.

abaire:
After upgrading Xcode to 13.2 I could no longer launch my dev build of xemu due to signature errors stemming from the patchups made to the library dependencies. This change re-signs the modified dependencies and re-signs the xemu binary after the app bundle is fully formed such that `codesign --verify --deep --strict --verbose=4  dist/xemu.app` will pass.

I also noticed that the `libsamplerate.dylib` min OS target is 11.3 when building for arm64, so I bumped the min version and dropped support for the 11.1 SDK to avoid any potential issues.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -55,12 +55,14 @@ package_macos() {
       echo "Fixing $exe_path dependency $dep_basename -> $new_path"
       install_name_tool -change "$dep" "$new_path" "$exe_path"
     done
+
     for lib_path in ${lib_path}/*.dylib; do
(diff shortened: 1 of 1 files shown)


MasonT8198:
This seems fine, dropping 11.1 should also be fine as on arm it'll cause issues, where on 64bit macs it shouldn't cause issues unless your mac is rather old.

mborgerson:
Minor nit: the convention for for commit messages in this repo is to not have a trailing period. I'll fixup here, but in the future you can simply leave off the period. Moreover, please try to use prefix consistent with file history when possible

Issue #601, opened 2021-12-20
Jet Set Radio Future [JSRF]: Incorrect rendering of progress indicator panel


Shoegzer:
### Bug Description

The progress indicator in the lower-left panel of the map screen in JSRF is black (it should be light-blue) and icons are not visible.  Borrowing screenshots because laziness:

Incorrect:
![124822410-db4a2280-df3d-11eb-9566-f58de8b48250](https://user-images.githubusercontent.com/12475371/146827688-9b0e1734-889b-48b4-b288-b8a3ca462c09.png)
Correct:
![146447186-d617a735-a2f3-485c-9f6f-3d21f95f26f8](https://user-images.githubusercontent.com/12475371/146827702-20d7824f-137d-44b1-9876-a4f61f906d51.png)

Note: The issue with the map itself being incorrectly rendered as solid grey is being tracked in #599.

### Expected Behavior

Panel should be light-blue and icons should be visible.

### xemu Version

0.6.2-10-gaa7acc8f64

### System Information

OS: Linux Mint 20.2 / kernel 5.13
CPU: Intel i7-10870H
GPU: Nvidia GTX 3070
GPU Driver: 470.86

GXTX:
Dupe of #397

Shoegzer:
It isn't though.  See note above, the map issue is being tracked separately and @abaire requested a separate issue report for the info panel.

abaire:
Tracking some initial findings:

- I see some unhandled pgraph messages that I think are related to specular lighting, but I wouldn't expect that to affect this. To more or less rule it out, I modified the shader to set `oD0`'s red channel to 0.8 and the resultant image is still black.
- The icons are missing/black as well as the blue background so, coupled with the lack of effect when hardcoding the vertex colors, I suspect this may be a texturing issue. I'm guessing the info panel may be a textured quad with dynamic text rendered on top of it, and that there may be some issue with the fixed function shader that is preventing the correct texture from being utilized (perhaps due to some state leakage since textures clearly work elsewhere on this screen?). Also possible that the texture is never loaded properly in the first place or is in some unusual format that is mishandled, but given that other textures on the screen work it seems unlikely that the developers would've used a special format just for this element.

abaire:
[...]

PR #603, opened 2021-12-24
ui: Add frame step debugger.

abaire:
Adds some very basic debugger functionality (not compiled in by default) allowing the user to step through a couple interesting nv2a states (namely buffer swaps and `NV097_SET_BEGIN_END_OP_END`). Also adds a very basic texture viewer.

I anticipate that this will help track down things like the blacked out textures in #601 and will potentially be a starting point for more advanced debugging in the future.

Screenshot_20211225_101926
r

### DIFF
--- a/hw/xbox/nv2a/debug.c
+++ b/hw/xbox/nv2a/debug.c
@@ -132,4 +132,229 @@ void gl_debug_frame_terminator(void)
     glFrameTerminatorGREMEDY();
 }
 
-#endif
+#endif // DEBUG_NV2A_GL
+
+#ifdef ENABLE_NV2A_DEBUGGER
+#include <memory.h>
+#include "nv2a_int.h"
+#include "qemu/osdep.h"
+#include "qemu/thread.h"
+#include "sysemu/runstate.h"
+
+enum NV2A_DBG_STATE {
+    NV2A_DBG_RUNNING = 0,
+    NV2A_DBG_STOPPED_FRAMEBUFFER_SWAP,
+    NV2A_DBG_STOPPED_BEGIN_END
+};
+
+typedef struct NV2ADebuggerVMState {
+    // Indicates that the VM should be paused at the next framebuffer swap.
+    uint32_t frame_break_requested;
+    // Indicates that the VM should be paused after the next begin_end::end op.
+    uint32_t draw_end_break_requested;
+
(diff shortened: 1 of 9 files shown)


abaire:
Putting back into draft. My initial approach allowed multiple BEGIN_END's to be processed after hitting the step button. I'm not attempting to halt pfifo processing, which appears to work, but eventually gets into a state where it doesn't resume.

abaire:
OK, I think the livelock issue is resolved (or happens so rarely that I haven't seen it in a day of messing around).
I added a display of the current backbuffer which, combined with `Step DrawArrays` lets you see the results of each BEGIN_END call. This lets the person debugging see exactly which set of pgraph commands is applicable to whatever they're trying to trace down.

mborgerson:
Building a graphics debugger is an ambitious project; props for digging into it.

The reason xemu does not have a graphics debugger is that there are already [several debugging tools out there](https://www.khronos.org/opengl/wiki/Debugging_Tools). Such tools can capture a trace of a frame, show all draw operations, shaders, buffer contents, textures, etc. You can enable the GL debug macro in hw/xbox/nv2a/debug.h and additional debug information will be inserted into the command stream to logically group commands and add associated object labels.

I appreciate the effort put into this so far, but I do not see how in current form this provides any utility over existing graphics debugging tools to justify the added code and complexity. For that reason, I think it is better to not merge this into master at this time.

abaire:
> Building a graphics debugger is an ambitious project; props for digging into it.
> 
> The reason xemu does not have a graphics debugger is that there are already [several debugging tools out there](https://www.khronos.org/opengl/wiki/Debugging_Tools). Such tools can capture a trace of a frame, show all draw operations, shaders, buffer contents, textures, etc. You can enable the GL debug macro in hw/xbox/nv2a/debug.h and additional debug information will be inserted into the command stream to logically group commands and add associated object labels.
> 
> I appreciate the effort put into this so far, but I do not see how in current form this provides any utility over existing graphics debugging tools to justify the added code and complexity. For that reason, I think it is better to not merge this into master at this time.

No worries, it's inarguable that this is invasive. 

From my perspective an included debugger lowers the bar for less technical users to participate in debugging and, selfishly, gives me a convenient place to expose pgraph state when trying to work out what various commands do (e.g., the ramht entries), so I'm happy to just keep it in my fork.

PR #604, opened 2021-12-24
nv2a: Add option to suppress irq and dma_map debug messages.

abaire:
When trying to debug pgraph issues, I find that the log gets very long due to irq and dma_map messages that aren't always immediately useful for the task at hand. This adds a preprocessor flag to suppress those messages.

### DIFF
--- a/hw/xbox/nv2a/nv2a.c
+++ b/hw/xbox/nv2a/nv2a.c
@@ -21,6 +21,9 @@
 
 #include "hw/xbox/nv2a/nv2a_int.h"
 
+#define DBG_IRQ 0
+#define DBG_DMA 0
+
 void nv2a_update_irq(NV2AState *d)
 {
     /* PFIFO */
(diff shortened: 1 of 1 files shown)


hw/xbox/nv2a/nv2a.c
```diff
@@ -45,7 +45,9 @@ void nv2a_update_irq(NV2AState *d)
     }
 
     if (d->pmc.pending_interrupts && d->pmc.enabled_interrupts) {
+#ifndef DEBUG_NV2A_SUPPRESS_IRQ_AND_DMA_MESSAGES
         NV2A_DPRINTF("raise irq\n");
```
mborgerson: Let's just use the `NV2A_XPRINTF` macro to avoid extra ifdefs here

abaire: Done. 

I added a new control for this, if we want to come closest to the existing behavior I should probably set it to 1 if `DEBUG_NV2A` is set - do you feel strongly about that versus having it be disabled by default despite `DEBUG_NV2A`?

hw/xbox/nv2a/debug.h
```diff
@@ -36,6 +36,9 @@
 # define NV2A_DPRINTF(format, ...)       do { } while (0)
 #endif
 
+// Enable raise irq and dma_map mesages.
+#define DBG_NV2A_IRQ_AND_DMA 0
```
mborgerson: * Let them be separate macros (DBG_IRQ, DBG_DMA)
* Have them be defined locally in nv2a.c to avoid rebuilds of the other files that include debug.h

abaire: Done.

PR #605, opened 2021-12-27
Adds dbg shader invalidate

abaire:
In combination with the DrawArrays stepper I've found that it's useful to see exactly what shader code is being used. This button lets you trash the shader cache, forcing the applicable shaders to be recompiled on the next draw call.


abaire:
Closing along with #603.

Issue #140, opened 2021-01-06
SpongeBob SquarePants: BFBB: pgraph_render_surface_to_texture assertion


irllzrd:
In Spongebob: Battle for Bikini Bottom, assertion error in pgraph.c occurs in function pgraph_render_surface_to_texture when you use the X/Y button bubble-bash and bubble-stomp moves. Removing assert(glGetError() == GL_NO_ERROR) from pgraph_render_surface_to_texture and rebuilding shows the specific error is GL_INVALID_FRAMEBUFFER_OPERATION. It occurs multiple times during the moves, and ends when the bubble models disappear. The game continues to be very playable.

The bubble foot/helmet model generated may use a framebuffer as a texture to show a 'reflection' of the world around it. I tried moving the assert around to pinpoint where the errors begin, but it seems to start right on function call. 

Titles: 5451001a

medievil1:
```GL_INVALID_FRAMEBUFFER_OPERATION is generated if the specified read and draw framebuffers are not framebuffer complete. ```

Triticum0:
It seems that Batman: Rise Of Sin Tzu(https://xemu.app/titles/5553001c/#Batman-Rise-of-Sin-Tzu) also hit the same assertion even though it gives the assertion: Bail out! ERROR:D:/a/xemu/xemu/hw/xbox/nv2a/pgraph.c:3671:pgraph_gl_sync: assertion failed: (epoxy_glGetError() == 0)  looking at the assertion on github commit it give this: https://github.com/mborgerson/xemu/blob/b9df39749346d83ebd3be8884ead5c75fdb8261f/hw/xbox/nv2a/pgraph.c#L3668-L3671 So I think it should go hear but feel free to correct me if I'm wrong.

antangelo:
Seems this assertion failure is different in newer versions. #635 tracks the current assertion failure message.

abaire:
If this is no longer valid and #635 tracks the new issue, can we close this bug?

PR #607, opened 2021-12-28
pgraph.c: Only mark surface dirty if parameter has been changed

### LINKED ISSUE #503: SpongeBob: Battle for Bikini Bottom: Error pg-> color_binding || pg-> zeta_binding
### Title

https://xemu.app/titles/5451001a/#Nickelodeon-SpongeBob-SquarePants-in-Battle-for-Bikini-Bottom

NTSC game version 1.00

### Bug Description

If you do a jump attack(not spin) or the helmet attack, it throws this C++ error for some reason.

### Expected Behavior

What's supposed to happen on the not spin jump attack is spongebob makes a lot of bubbles that attack an enemy instead of **crashing the whole game with a C++ error.**

### xemu Version

 0.6.1-23-g5a80cfa2d9

### System Information

OS: Windows 11 Home
CPU: AMD Athlon Gold 3150U 
GPU: AMD Radeon
GPU Driver: AMD Radeon Driver v27.20.21030.1005

### Additional Context

_No response_

antangelo:
Surfaces are no longer marked dirty when the parameters are unchanged in calls to SET_SURFACE_PITCH, SET_SURFACE_COLOR_OFFSET, and SET_SURFACE_ZETA_OFFSET.

In my testing this has resolved #503 and #140 for title 5451001a. I have not tested whether this resolves similar assertion failures for the other games mentioned on those issues.

### DIFF
--- a/hw/xbox/nv2a/pgraph.c
+++ b/hw/xbox/nv2a/pgraph.c
@@ -1180,30 +1180,28 @@ DEF_METHOD(NV097, SET_SURFACE_FORMAT)
 DEF_METHOD(NV097, SET_SURFACE_PITCH)
 {
     pgraph_update_surface(d, false, true, true);
+    unsigned int color_pitch = GET_MASK(parameter, NV097_SET_SURFACE_PITCH_COLOR);
+    unsigned int zeta_pitch  = GET_MASK(parameter, NV097_SET_SURFACE_PITCH_ZETA);
 
-    pg->surface_color.pitch =
-        GET_MASK(parameter, NV097_SET_SURFACE_PITCH_COLOR);
-    pg->surface_zeta.pitch =
-        GET_MASK(parameter, NV097_SET_SURFACE_PITCH_ZETA);
+    pg->surface_color.buffer_dirty |= (pg->surface_color.pitch != color_pitch);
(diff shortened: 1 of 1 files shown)


mborgerson:
Thanks

abaire:
This change seems to break trivial nxdk binaries. It looks like it can lead to a state where the buffers are never marked as dirty, leading to stale content being displayed.

Specifically, the [nxdk `triangle` sample](https://github.com/XboxDev/nxdk/tree/master/samples/triangle) displays this behavior (as does my [pgraph tester](https://github.com/abaire/nxdk_pgraph_tests)).

Note: it also seems the latest nxdk generates binaries that trigger segfaults in xemu (seems like it broke with commit fe47efb), I synced back to bc79f005 when building the triangle test.

This can be worked around by modifying pbkit's [`set_draw_buffer` around here](https://github.com/XboxDev/nxdk/blob/master/lib/pbkit/pbkit.c#L1940) and calling it twice; once with some non-zero `SET_SURFACE_COLOR_OFFSET` and then again with the correct value.

For clarity, this behavior is not exhibited in hardware, and my old test binaries that used to work fail with this commit and work if I revert the commit (i.e., it is not due to changes in the nxdk).

PR #616, opened 2022-01-04
xemu: Update Copyright Year

MasonT8198:
Same as commit 6e564854fc03266cabb59e01010ea5d20ebd03c5 but updating the year again!

### DIFF
--- a/hw/xbox/acpi_xbox.c
+++ b/hw/xbox/acpi_xbox.c
@@ -2,7 +2,7 @@
  * Xbox ACPI implementation
  *
  * Copyright (c) 2012 espes
- * Copyright (c) 2020-2021 Matt Borgerson
+ * Copyright (C) 2020-2022 Matt Borgerson
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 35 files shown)


mborgerson:
Based on `grep -rPe "202\d\s+Matt"` there should be at least 70 files that need an update. Can you update the others?

Also no need to change the case of the 'C' in (C). Just the year number range can be updated (never changing the start year, only adding a -2022 or updating a -2021 to -2022)

MasonT8198:
> Based on `grep -rPe "202\d\s+Matt"` there should be at least 70 files that need an update. Can you update the others?
> 
> Also no need to change the case of the 'C' in (C). Just the year number range can be updated (never changing the start year, only adding a -2022 or updating a -2021 to -2022)

Yeah my bad on changing the cases, I just ran a quick find and replace command

mborgerson:
If you can, please update the remaining files and we will merge this

MasonT8198:
Closing this on the account of seemingly not having enough spare time to finish this, however will open with this completed if I find the time

FTR; maybe just biting the bullet and automating this process is the best idea

Issue #329, opened 2021-06-03
[Assertion] pg->inline_elements_length == 0


Triticum0:
### Title Information

https://xemu.app/titles/4e4d0009/#Breakdown
https://xemu.app/titles/53450033/#ESPN-College-Hoops-2K5
https://xemu.app/titles/54540091/#College-Hoops-2K6
https://xemu.app/titles/545400af/#College-Hoops-2K7
https://xemu.app/titles/53450023/#ESPN-NBA-Basketball
https://xemu.app/titles/53450030/#ESPN-NFL-2K5
https://xemu.app/titles/53450032/#ESPN-NHL-2K5
https://xemu.app/titles/53450031/#ESPN-NBA-2K5
https://xemu.app/titles/545400ae/#NBA-2K7
https://xemu.app/titles/54540090/#NBA-2K6
https://xemu.app/titles/5454008d/#NHL-2K6
https://xemu.app/titles/545400ad/#NHL-2K7
https://xemu.app/titles/4e4d0008/#R-Racing
https://xemu.app/titles/5553001c/#Batman-Rise-of-Sin-Tzu

### Bug Description
The game crash as soon as try to go in-game with the Assertion:
https://github.com/mborgerson/xemu/blob/6f477f02a653a0417c82dc3a1473b1bd3e96bca0/hw/xbox/nv2a/pgraph.c#L2308

Steps to reproduce this issue:
1. Insert disk
2. go to the main menu
3. start new game/race
4. game crashes

### Expected Behavior
you should be able to go in-game

<h3>Version</h3>
<ul>
<li>Version: build-202106020756( <a href="https://github.com/mborgerson/xemu/commit/3884b8d44b4bfaed73f26227cd88d7946e2dfaa5"><tt>3884b8d</tt></a>)</li>
<li>Last-known Working: Unknown</li>
</ul>
<h3>System Information</h3>

Field | Value
-- | --
OS | Windows 10 64-bit (21H1)
CPU | AMD Ryzen 5 2600 Six-Core Processor @ 3.40GHz
Graphics Device | AMD Radeon RX 5700 XT
Graphics Driver | 4.0.148202 Core Profile Context 21.3.2 27.20.15003.5017

MattPonton:
As of v0.6.1-20-g21718399f2

Issue also occurs with TItle:
https://xemu.app/titles/53450030/#ESPN-NFL-2K5

Triticum0:
Thanks. I Will add it to the list

[...]

Issue #238, opened 2021-03-07
Compressed disc image support


rtissera:
Hello,

I am working on libchdr and batocera.linux amongst others.
First, congrats for the great progress done on xemu, it's really impressive.

I would like to reconsider this issue : https://github.com/mborgerson/xemu/issues/129
HDD games converted back to ISO (which is basically, similar to the original ISO stripped from random junk data if I am not wrong), could be compressed very efficiently with CHD.

I can help with CHD implementation in xemu if needed.
I am open to discussion on some other kind of solution like a custom format (like compressed iso block compression based on zstd for instance).

Regards, keep up the good work !

liberodark:
Indeed CHD is the most widespread format and surely also the best format to compress iso.
Especially when we are already making xiso (so iso from an HDD version) to save space.
It would be really great to be able to make CHD which could still compress the game losslessly. 

Proof :
- 007 - From Russia With Love 6.21 Go (iso Standard)
- 007 - From Russia With Love 2.0 Go (xiso)
- 007 - From Russia With Love 2.0 Go (hdd)
- 007 - From Russia With Love 1.7 Go (chd)

This game is one of the worst examples.
Because in some games we earn more than 1 Go

Darknior:
WOAW yes if possible CHD like on all other console will help a lot to save hard drive space :)
The MAME chd library can be a good start point :)

Veuks:
Good initiative

zognic:
Great idea!!

Redemp:
That would be a great idea 💡! Saves slot of space!

KevinLeenders:
One format to rule them all!

mborgerson:
@rtissera I agree that there is opportunity to reduce disc image size through introduction of support for some format which enables a block compression scheme. Although I do not anticipate extreme compression ratios, I do expect a 10-20% reduction in disc image size is possible (varies by game, it is not uncommon for games to already compress their data); that is, if the user wishes to exclude other parts of the disc image such as the randomized data stream and video partition, including these parts would not compress well and diminish the overall compression ratio considerably.

I'm not convinced that CHD is the ideal format yet, but that's only because I'm not familiar with it. @Ernegien and I have been considering introducing a custom disc image format to support all features of the Xbox disc security system for accurate emulation, but this work has stalled temporarily due to higher priority tasks taking precedence. Such a format would ideally also include optional opportunistic compression.

[...]

PR #623, opened 2022-01-08
nv2a: Update `inline_value`s as rendering side-effect

### LINKED ISSUE #601: Jet Set Radio Future [JSRF]: Incorrect rendering of progress indicator panel

abaire:
This updates the default `inline_value` for the normal vertex attribute whenever valid values for the normal are passed, mirroring the hardware behavior observed in https://github.com/abaire/nxdk_pgraph_tests/blob/main/tests/lighting_normal_tests.cpp.

Additional tests for all attributes:
* [test](https://github.com/abaire/nxdk_pgraph_tests/blob/4e2776729ed6a3ea86b4eb15247bc9009ed7cff9/src/tests/attribute_carryover_tests.cpp#L1)
* [results](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-Attrib_carryover)

I'm not entirely confident that this resolves all possible paths, but it does fix #601 

Screenshot_20220110_213034

(note: screenshot was from my dev branch that also includes fix for #599, without which the minimap would be opaque but the lower left panel would still be fixed)

### DIFF
--- a/hw/xbox/nv2a/pgraph.c
+++ b/hw/xbox/nv2a/pgraph.c
@@ -399,7 +399,7 @@ static void pgraph_apply_anti_aliasing_factor(PGRAPHState *pg, unsigned int *wid
 static void pgraph_apply_scaling_factor(PGRAPHState *pg, unsigned int *width, unsigned int *height);
 static void pgraph_get_surface_dimensions(PGRAPHState *pg, unsigned int *width, unsigned int *height);
 static void pgraph_update_memory_buffer(NV2AState *d, hwaddr addr, hwaddr size, bool quick);
-static void pgraph_bind_vertex_attributes(NV2AState *d, unsigned int min_element, unsigned int max_element, bool inline_data, unsigned int inline_stride);
+static void pgraph_bind_vertex_attributes(NV2AState *d, unsigned int min_element, unsigned int max_element, bool inline_data, unsigned int inline_stride, unsigned int provoking_element);
 static unsigned int pgraph_bind_inline_array(NV2AState *d);
 static float convert_f16_to_float(uint16_t f16);
 static float convert_f24_to_float(uint32_t f24);
@@ -2677,7 +2677,9 @@ DEF_METHOD(NV097, SET_BEGIN_END)
             assert(pg->inline_elements_length == 0);
 
             pgraph_bind_vertex_attributes(d, pg->draw_arrays_min_start,
-                                          pg->draw_arrays_max_count, false, 0);
+                                          pg->draw_arrays_max_count - 1,
+                                          false, 0,
+                                          pg->draw_arrays_max_count - 1);
             glMultiDrawArrays(pg->shader_binding->gl_primitive_mode,
                               pg->gl_draw_arrays_start,
                               pg->gl_draw_arrays_count,
@@ -2707,6 +2709,9 @@ DEF_METHOD(NV097, SET_BEGIN_END)
                     glVertexAttribPointer(i, 4, GL_FLOAT, GL_FALSE, 0, 0);
                     glEnableVertexAttribArray(i);
                     attr->inline_buffer_populated = false;
+                    memcpy(attr->inline_value,
+                           attr->inline_buffer + (pg->inline_buffer_length - 1) * 4,
+                           sizeof(attr->inline_value));
                 } else {
                     glDisableVertexAttribArray(i);
                     glVertexAttrib4fv(i, attr->inline_value);
@@ -2744,7 +2749,8 @@ DEF_METHOD(NV097, SET_BEGIN_END)
             }
 
             pgraph_bind_vertex_attributes(
-                d, min_element, max_element, false, 0);
+                d, min_element, max_element, false, 0,
+                pg->inline_elements[pg->inline_elements_length - 1]);
 
             VertexKey k;
             memset(&k, 0, sizeof(VertexKey));
@@ -6538,11 +6544,74 @@ static void pgraph_update_memory_buffer(NV2AState *d, hwaddr addr, hwaddr size,
     }
 }
 
+static void pgraph_update_inline_value(VertexAttribute *attr,
+                                       const uint8_t *data)
+{
+    assert(attr->count <= 4);
+    attr->inline_value[0] = 0.0f;
+    attr->inline_value[1] = 0.0f;
+    attr->inline_value[2] = 0.0f;
+    attr->inline_value[3] = 1.0f;
+
+    switch (attr->format) {
+        case NV097_SET_VERTEX_DATA_ARRAY_FORMAT_TYPE_UB_D3D:
+        case NV097_SET_VERTEX_DATA_ARRAY_FORMAT_TYPE_UB_OGL:
+            for (uint32_t i = 0; i < attr->count; ++i) {
+                attr->inline_value[i] = (float)data[i] / 255.0f;
+            }
+            break;
+        case NV097_SET_VERTEX_DATA_ARRAY_FORMAT_TYPE_S1: {
+            const int16_t *val = (const int16_t *) data;
+            for (uint32_t i = 0; i < attr->count; ++i, ++val) {
+                attr->inline_value[i] = MAX(-1.0f, (float) *val / 32767.0f);
+            }
+            break;
+        }
+        case NV097_SET_VERTEX_DATA_ARRAY_FORMAT_TYPE_F:
+            memcpy(attr->inline_value, data, attr->size * attr->count);
+            break;
+        case NV097_SET_VERTEX_DATA_ARRAY_FORMAT_TYPE_S32K: {
+            const int16_t *val = (const int16_t *) data;
+            for (uint32_t i = 0; i < attr->count; ++i, ++val) {
+                attr->inline_value[i] = (float)*val;
+            }
+            break;
+        }
+        case NV097_SET_VERTEX_DATA_ARRAY_FORMAT_TYPE_CMP: {
+            /* 3 signed, normalized components packed in 32-bits. (11,11,10) */
+            const int32_t val = *(const int32_t *)data;
+            int32_t x = val & 0x7FF;
+            if (x & 0x400) {
+                x |= 0xFFFFF800;
+            }
+            int32_t y = (val >> 11) & 0x7FF;
+            if (y & 0x400) {
+                y |= 0xFFFFF800;
+            }
+            int32_t z = (val >> 22) & 0x7FF;
+            if (z & 0x200) {
+                z |= 0xFFFFFC00;
+            }
+
+            attr->inline_value[0] = MAX(-1.0f, (float)x / 1023.0f);
+            attr->inline_value[1] = MAX(-1.0f, (float)y / 1023.0f);
+            attr->inline_value[2] = MAX(-1.0f, (float)z / 511.0f);
+            break;
+        }
+    default:
+        fprintf(stderr, "Unknown vertex attribute type: 0x%x for format 0x%x\n",
+                attr->gl_type, attr->format);
+        assert(!"Unsupported attribute type");
+        break;
+    }
+}
+
 static void pgraph_bind_vertex_attributes(NV2AState *d,
                                           unsigned int min_element,
                                           unsigned int max_element,
                                           bool inline_data,
-                                          unsigned int inline_stride)
+                                          unsigned int inline_stride,
+                                          unsigned int provoking_element)
 {
     PGRAPHState *pg = &d->pgraph;
     bool updated_memory_buffer = false;
@@ -6574,6 +6643,7 @@ static void pgraph_bind_vertex_attributes(NV2AState *d,
             pg->compressed_attrs |= (1 << i);
         }
 
+        hwaddr start = 0;
         if (inline_data) {
             glBindBuffer(GL_ARRAY_BUFFER, pg->gl_inline_array_buffer);
             attrib_data_addr = attr->inline_array_offset;
@@ -6586,7 +6656,7 @@ static void pgraph_bind_vertex_attributes(NV2AState *d,
             assert(attr->offset < dma_len);
             attrib_data_addr = attr_data + attr->offset - d->vram_ptr;
             stride = attr->stride;
-            hwaddr start = attrib_data_addr + min_element * stride;
+            start = attrib_data_addr + min_element * stride;
             pgraph_update_memory_buffer(d, start, num_elements * stride,
                                         updated_memory_buffer);
             updated_memory_buffer = true;
@@ -6602,6 +6672,24 @@ static void pgraph_bind_vertex_attributes(NV2AState *d,
         }
 
         glEnableVertexAttribArray(i);
+
+        uint32_t provoking_element_index = provoking_element - min_element;
+        size_t element_size = attr->size * attr->count;
+        assert(element_size <= sizeof(attr->inline_value));
+        const uint8_t *last_entry;
+
+        if (inline_data) {
+            last_entry = (uint8_t*)pg->inline_array + attr->inline_array_offset;
+        } else {
+            last_entry = d->vram_ptr + start;
+        }
+        if (stride) {
+            last_entry += stride * provoking_element_index;
+        } else {
+            last_entry += element_size * provoking_element_index;
+        }
+
+        pgraph_update_inline_value(attr, last_entry);
     }
 
     NV2A_GL_DGROUP_END();
@@ -6638,7 +6726,8 @@ static unsigned int pgraph_bind_inline_array(NV2AState *d)
     glBufferData(GL_ARRAY_BUFFER, NV2A_MAX_BATCH_LENGTH * sizeof(uint32_t),
                  NULL, GL_STREAM_DRAW);
     glBufferSubData(GL_ARRAY_BUFFER, 0, pg->inline_array_length*4, pg->inline_array);
-    pgraph_bind_vertex_attributes(d, 0, index_count-1, true, vertex_size);
+    pgraph_bind_vertex_attributes(d, 0, index_count-1, true, vertex_size,
+                                  index_count-1);
 
     return index_count;
 }


Shoegzer:
With this PR, JSRF crashes immediately prior to the main menu (after the loading screen), with the following message in term:

`qemu-system-i386: ../hw/xbox/nv2a/pgraph.c:6273: pgraph_bind_vertex_attributes: Assertion attr->gl_type == GL_FLOAT && "Need to convert non-float normal."' failed.
`

abaire:
I'm hoping that I'm being overly cautious with that assert so it can probably be removed (though I'm interested in seeing what is actually being used if it's not a float).

Unfortunately it seems like I can no longer get past the loading screen even with a clean build from `master` so I'm putting this back into draft mode until I can verify that it's safe to remove the assert.

abaire:
Looks like it uses unsigned byte for normals after the loading screen. Removed the assert and fixed the handling to support arbitrary types instead of assuming they'll always be floats..I also had an error in how I was grabbing the last element, so as an added bonus the coloring is more correct now as well.

Thanks for testing @Shoegzer !

hw/xbox/nv2a/pgraph.c
```diff
@@ -6433,6 +6434,26 @@ static void pgraph_bind_vertex_attributes(NV2AState *d,
...
+            // executing the fixed function pipeline. This is emulated by
+            // fetching the last value in the array and storing it as the
+            // inline_value.
+
+            // FIXME: Support inline_data as well?
+            if (!inline_data) {
+                uint32_t element_size = attr->size * attr->count;
+                const uint8_t* last_entry = d->vram_ptr + start;
+                if (stride) {
+                    last_entry += stride * (num_elements - 2);
```
abaire: Double checking on these. Is it possible that `num_elements` doesn't actually need the +1 at the top of the function (I'd expect that these should be `num_elements - 1`, but that definitely reads past the last element)? 

Not sure if there's an off-by-one error or just an error in my assumptions based on the naming.
- In my pgraph test I'm sending a single triangle (3 elements). 
- `count` is decremented by one before being sent to DRAW_ARRAYS as per spec, but that is handled [here](https://github.com/mborgerson/xemu/blob/01c2f0fa74d1f67ec135eaeed81f029058ef25a0/hw/xbox/nv2a/pgraph.c#L2982) so pg->draw_arrays_max_count ends up being correct (3)
- When the BEGIN_END is ended, the vertex attributes are bound w/ `3` as `max_element`. The naming and the +1 makes it sound like `max_element` is expected to be the maximum valid element (`2` in my test case) rather than the max count (`3`) that's being passed in.

Shoegzer:
Tested and confirmed the assert/crashing is gone, and the bottom-left panel looks good.  I haven't merged your other fix for the map itself, but with this PR alone it now shows:

pic

Nice work!

abaire:
Just happened to find that this crashes Gun Valkyrie, where a call is made with `num_elements` == 1

Shoegzer:
That's too bad.  Not sure what that means in terms of getting this PR merged, though from what you say here:

> I also had an error in how I was grabbing the last element, so as an added bonus the coloring is more correct now as well.

Would the panel color still be mostly correct if you avoided the condition for `num_elements == 1`?  If so, would it make any sense to move that part to another PR and merge this one?  Happy to help test any workarounds either way.

hw/xbox/nv2a/pgraph.c
```diff
@@ -2502,7 +2502,7 @@ DEF_METHOD(NV097, SET_BEGIN_END)
             assert(pg->inline_elements_length == 0);
 
             pgraph_bind_vertex_attributes(d, pg->draw_arrays_min_start,
-                                          pg->draw_arrays_max_count, false, 0);
+                                          pg->draw_arrays_max_count - 1, false, 0);
```
abaire: Based on the usage in `pgraph_bind_vertex_attributes`, it looks like the expectation is that this should be the last used element rather than the count.

Setting this to count requires a -2 to access the last used element in JSRF (and causes crashes in cases that use inline elements, such as Gun Valkyrie). As far as I can see, providing `max_element` in this path still works correctly in the games I've tested.

hw/xbox/nv2a/pgraph.c
abaire: Testing with Gun Valkyrie I feel more confident that the `- 2` is wrong and the `DRAW_ARRAYS` path should be passing `max_element` rather than `count`. Added a fix for that in this PR.

abaire:
> That's too bad. Not sure what that means in terms of getting this PR merged, though from what you say here:
> 
> > I also had an error in how I was grabbing the last element, so as an added bonus the coloring is more correct now as well.
> 
> Would the panel color still be mostly correct if you avoided the condition for `num_elements == 1`? If so, would it make any sense to move that part to another PR and merge this one? Happy to help test any workarounds either way.

I think the crash was due to me working around what I think is a bug elsewhere. To fix the crash I fixed the other bug, so I think this should still be merge-able. It'd be great if you could verify that it still works for you with this latest push and try out as many other games as you can to see if it causes any weird glitches. The underlying fix will affect many more things than lighting but ideally you would not see any actual difference.

Shoegzer:
You bet.  Tested JSRF and it still works fine.  I also tested Halo:CE, Halo 2 and those work as well as before at least.  Star Wars KOTOR regressed, however, as eyes are missing from the character select screen (though this issue occurred before until very recently):

PR:
prpic

Master:
masterpic

Side-note since you asked: I have many other discs but I don't want to dump them to non-archival quality (xiso) images for various reasons as mentioned in #238 or I'd test those for you too.  Hopefully xemu can support proper isos in the future (though I'm patient and respect priorities).

abaire:
Ah, that'd almost certainly mean that I'm wrong about the count versus max somehow, or that there's some interesting dependency on the draw mode that needs to be accounted for. I own KOTOR so I'll do some debugging in a few days when I'm back with my hardware.

abaire:
Well, the good news is that this seems to have been a simple case of the fix branch being older than the fix for KOTOR. I rebased to the latest master and the eyes now look correct. @Shoegzer would you mind re-verifying with the latest push?

The bad news is that I wrote a [much more in-depth test](https://github.com/abaire/nxdk_pgraph_tests/blob/321b73ce7de19735bd1f21c4c80329ac1e3edb48/src/tests/attribute_carryover_tests.cpp#L1) in anticipation of tracking this down and found that the behavior in the case of inline elements is slightly wrong (it should always take the last _rendered_ vertex, the code I wrote takes the last buffered vertex). I also verified that all the other attributes  was able to test are supposed to carry over in the same way the normals do (diffuse, specular, tex coords, almost certainly the rest but I can't easily test them with nxdk's Cg-based shaders).

@mborgerson would you be OK to review and merge this PR as is and I'll do a followup? 
The handling of the inline elements and additional attributes will add some complexity so I'm inclined to get this part that actually fixes a game in and then expand on it (I don't know if any other games suffer similar problems due to missing carryover).

mborgerson:
> The bad news is that I wrote a [much more in-depth test](https://github.com/abaire/nxdk_pgraph_tests/blob/321b73ce7de19735bd1f21c4c80329ac1e3edb48/src/tests/attribute_carryover_tests.cpp#L1) in anticipation of tracking this down and found that the behavior in the case of inline elements is slightly wrong (it should always take the last _rendered_ vertex, the code I wrote takes the last buffered vertex). I also verified that all the other attributes was able to test are supposed to carry over in the same way the normals do (diffuse, specular, tex coords, almost certainly the rest but I can't easily test them with nxdk's Cg-based shaders).

This is what I had expected. Thanks for creating a test case, it is actually good news.

> @mborgerson would you be OK to review and merge this PR as is and I'll do a followup? The handling of the inline elements and additional attributes will add some complexity so I'm inclined to get this part that actually fixes a game in and then expand on it (I don't know if any other games suffer similar problems due to missing carryover).

Let's just fix it generally, it should not be too difficult.

By the way, the arrays implementation is flawed. I have a patch for it that notably fixes some sports games, but it needs a couple of remaining improvements. You are welcome to work on it, or I'll return to it shortly

abaire:
> > The bad news is that I wrote a [much more in-depth test](https://github.com/abaire/nxdk_pgraph_tests/blob/321b73ce7de19735bd1f21c4c80329ac1e3edb48/src/tests/attribute_carryover_tests.cpp#L1) in anticipation of tracking this down and found that the behavior in the case of inline elements is slightly wrong (it should always take the last _rendered_ vertex, the code I wrote takes the last buffered vertex). I also verified that all the other attributes was able to test are supposed to carry over in the same way the normals do (diffuse, specular, tex coords, almost certainly the rest but I can't easily test them with nxdk's Cg-based shaders).
> 
> This is what I had expected. Thanks for creating a test case, it is actually good news.
> 
> > @mborgerson would you be OK to review and merge this PR as is and I'll do a followup? The handling of the inline elements and additional attributes will add some complexity so I'm inclined to get this part that actually fixes a game in and then expand on it (I don't know if any other games suffer similar problems due to missing carryover).
> 
> Let's just fix it generally, it should not be too difficult.
> 
> By the way, the arrays implementation is flawed. I have a patch for it that notably fixes some sports games, but it needs a couple of remaining improvements. You are welcome to work on it, or I'll return to it shortly

abaire:
Accidentally closed, sorry! This should be ready for review now.

> Let's just fix it generally, it should not be too difficult.

Sounds good. PR is updated, all testable cases now work correctly. I can't fully test the inline buffers case because xemu doesn't yet handle all of the attribute setter functions (e.g., `SET_DIFFUSE_COLOR` w/ 4 floats which my test program uses because I haven't gotten around to making use of floats versus DWORDs conditional).

> By the way, the arrays implementation is flawed. I have a patch for it that notably fixes some sports games, but it needs a couple of remaining improvements. You are welcome to work on it, or I'll return to it shortly

Cool, I can take a look, I assume it's [fix/arrays](https://github.com/mborgerson/xemu/tree/fix/arrays)? Do you know offhand which games it fixes?

While writing this PR I noticed that there's an off-by one in the call to `pgraph_bind_vertex_attributes` when handling `DRAW_ARRAYS` but I've stomped on my comments a couple times now. Mentioning it here in case you haven't read the full conversation history.

Triticum0:
It is fix/array and the game it fixes are on this issue #329.

Shoegzer:
> Well, the good news is that this seems to have been a simple case of the fix branch being older than the fix for KOTOR. I rebased to the latest master and the eyes now look correct. @Shoegzer would you mind re-verifying with the latest push?

I was wondering if this was a rebase issue since the eyes were fixed so recently in master, glad to know it was that simple.  In any event, KOTOR eyes are fine now with your PR, and of course JSRF lower-left panel looks good too.  Nice.

hw/xbox/nv2a/pgraph.c
```diff
@@ -6429,6 +6437,26 @@ static void pgraph_bind_vertex_attributes(NV2AState *d,
...
+        } else {
+            last_entry = (uint8_t*)pg->inline_array;
+            last_entry += attr->inline_array_offset;
+        }
+        if (stride) {
+            last_entry += stride * provoking_element_index;
+        } else {
+            last_entry += element_size * provoking_element_index;
+        }
+        memcpy(attr->inline_value, last_entry, element_size);
```
mborgerson: We store `inline_value` as `float` type. This can fail if attribute type is non-float

hw/xbox/nv2a/pgraph.c
```diff
@@ -2496,7 +2496,9 @@ DEF_METHOD(NV097, SET_BEGIN_END)
             assert(pg->inline_elements_length == 0);
 
             pgraph_bind_vertex_attributes(d, pg->draw_arrays_min_start,
-                                          pg->draw_arrays_max_count, false, 0);
+                                          pg->draw_arrays_max_count - 1,
```
mborgerson, line 2680: We should rename `draw_arrays_max_count` while we are at it

abaire: Do you mean we should change it to something like `draw_arrays_max_element` and remove the increment in the `DRAW_ARRAYS` handler? 
At the moment I believe it's tracking the count rather than the index of the last element so the current name seems to match the behavior.

Or do you mean we should change the suffix to `_length` to align with the other lengths?

hw/xbox/nv2a/pgraph.c
abaire: Added explicit handling of all the cases from `SET_VERTEX_DATA_ARRAY_FORMAT` below.

hw/xbox/nv2a/pgraph.c
mborgerson: > Do you mean we should change it to something like `draw_arrays_max_element` and remove the increment in the `DRAW_ARRAYS` handler? At the moment I believe it's tracking the count rather than the index of the last element so the current name seems to match the behavior.

`draw_arrays_max_count` tracks maximal (start + count) [max_element + 1]
 
> Or do you mean we should change the suffix to `_length` to align with the other lengths?

`draw_arrays_length` is already taken (number of arrays)

Actually nevermind, I'll just fix this in a follow up

hw/xbox/nv2a/pgraph.c
```diff
@@ -6538,11 +6544,60 @@ static void pgraph_update_memory_buffer(NV2AState *d, hwaddr addr, hwaddr size,
     }
 }
 
+static void pgraph_update_inline_value(VertexAttribute *attr,
+                                       const uint8_t *data)
+{
+    if (attr->gl_type == GL_FLOAT) {
```
mborgerson: Let's use a `switch` and test against `attr->format` and panic on `default` case like SET_VERTEX_DATA_ARRAY_FORMAT

abaire: Testing against `attr->format` feels like it introduces some duplication to me. 

E.g., if for some reason `FORMAT_TYPE_CMP` was changed to be a `GL_FOOBAR` instead of a `GL_INT` it'd silently do the wrong thing here. The intended behavior here is dependent on the combination of the `gl_type` and normalize/conversion flags and to me it doesn't seem like it necessarily needs to know or care about what the HW wants the format to be.

Changed to a switch with a panic, let me know if you feel strongly about the check against `format` (I realize my objection is a bit pedantic since I doubt there's a scenario where the gl types would actually change), but going that route feels like it'd also need an assert against `gl_type` which seems like it's effectively equivalent.

mborgerson: The input to this function is of the nv2a defined type, not the gl type. If cmp was changed to some other GL type it would be irrelevant for this matter, because we are always converting to float and we should know that the input is compressed from the nv2a type. Moreover, testing against the nv2a type keeps the code more generic, less tied to GL.

abaire: Makes sense, I'd erroneously assumed that there was more nv2a -> GL conversion happening to the raw data buffer than there is in practice - updated.

hw/xbox/nv2a/pgraph.c
```diff
@@ -6538,11 +6544,61 @@ static void pgraph_update_memory_buffer(NV2AState *d, hwaddr addr, hwaddr size,
...
+            const int16_t *val = (const int16_t *) data;
+            for (uint32_t i = 0; i < attr->count; ++i, ++val) {
+                attr->inline_value[i] = (float)*val;
+            }
+            break;
+        }
+        case NV097_SET_VERTEX_DATA_ARRAY_FORMAT_TYPE_CMP: {
+            /* 3 signed, normalized components packed in 32-bits. (11,11,10) */
+            typedef struct {
+                signed int x : 11;
```
mborgerson: I agree this looks nice, but let's do a shift and mask instead of depending on C struct bitfields; since the compiler can choose how to order them, we cannot depend on it to be correct

hw/xbox/nv2a/pgraph.c
```diff
@@ -6602,6 +6659,27 @@ static void pgraph_bind_vertex_attributes(NV2AState *d,
         }
 
         glEnableVertexAttribArray(i);
+
+        // Carry over the attribute value from the provoking vertex to mimic
+        // hardware behavior.
+        uint32_t provoking_element_index = provoking_element - min_element;
+        uint32_t element_size = attr->size * attr->count;
```
mborgerson: Let's use `size_t` for this variable

hw/xbox/nv2a/pgraph.c
```diff
@@ -6602,6 +6659,27 @@ static void pgraph_bind_vertex_attributes(NV2AState *d,
...
+        assert(element_size <= sizeof(attr->inline_value));
+        const uint8_t *last_entry;
+
+        if (!inline_data) {
+            last_entry = d->vram_ptr + start;
+        } else {
+            last_entry = (uint8_t*)pg->inline_array;
+            last_entry += attr->inline_array_offset;
```
mborgerson: nit: This can be a single line wlor

hw/xbox/nv2a/pgraph.c

hw/xbox/nv2a/pgraph.c

hw/xbox/nv2a/pgraph.c
```diff
@@ -6538,11 +6544,68 @@ static void pgraph_update_memory_buffer(NV2AState *d, hwaddr addr, hwaddr size,
     }
 }
 
+static void pgraph_update_inline_value(VertexAttribute *attr,
+                                       const uint8_t *data)
+{
+    switch (attr->format) {
```
mborgerson, line 6556: * Let's add a check here to make sure `attr->count <= 4`
* We should probably initialize `attr->inline_value` to [0,0,0,1] for the case that `attr->count < 4`

hw/xbox/nv2a/pgraph.c
```diff
@@ -6602,6 +6666,26 @@ static void pgraph_bind_vertex_attributes(NV2AState *d,
         }
 
         glEnableVertexAttribArray(i);
+
+        // Carry over the attribute value from the provoking vertex to mimic
```
mborgerson: nit: The function below already states this, so this comment is redundant

hw/xbox/nv2a/pgraph.c
```diff
@@ -6602,6 +6666,26 @@ static void pgraph_bind_vertex_attributes(NV2AState *d,
...
+        // hardware behavior.
+        uint32_t provoking_element_index = provoking_element - min_element;
+        size_t element_size = attr->size * attr->count;
+        assert(element_size <= sizeof(attr->inline_value));
+        const uint8_t *last_entry;
+
+        if (!inline_data) {
```
mborgerson: nit: Let's remove the `!` and swap the blocks

hw/xbox/nv2a/pgraph.c

hw/xbox/nv2a/pgraph.c

hw/xbox/nv2a/pgraph.c
abaire: Done.

mborgerson:
Thanks!

PR #625, opened 2022-01-08
smbus: Add dummy encoders for Focus and Xcalibur

GXTX:
Adds dummy handlers for Focus and Xcalibur, and exposes them via the `machine` option, defaults to Conexant.

Works as expected with 1.0, 1.4, and 1.6 based kernels.

### DIFF
--- a/hw/xbox/xbox.c
+++ b/hw/xbox/xbox.c
@@ -310,8 +310,20 @@ void xbox_init_common(MachineState *machine,
 
     /* smbus devices */
     smbus_xbox_smc_init(smbus, 0x10);
-    smbus_cx25871_init(smbus, 0x45);
-    smbus_adm1032_init(smbus, 0x4c);
+
+    const char *video_encoder =
+        object_property_get_str(qdev_get_machine(), "video-encoder", NULL);
+
+    if (strcmp(video_encoder, "xcalibur") != 0) {
+        if (!strcmp(video_encoder, "conexant")) {
+            smbus_cx25871_init(smbus, 0x45);
+        } else if (!strcmp(video_encoder, "focus")) {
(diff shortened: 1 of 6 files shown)


hw/xbox/xbox.c
```diff
@@ -310,8 +310,20 @@ void xbox_init_common(MachineState *machine,
...
+    const char *video_encoder =
+        object_property_get_str(qdev_get_machine(), "video-encoder", NULL);
+
+    if (strcmp(video_encoder, "xcalibur") != 0) {
+        if (!strcmp(video_encoder, "conexant")) {
+            smbus_cx25871_init(smbus, 0x45);
+        } else if (!strcmp(video_encoder, "focus")) {
+            smbus_fs454_init(smbus, 0x6A);
+        }
+        smbus_adm1032_init(smbus, 0x4C);
```
mborgerson, line 323: No temp sensor or xcalibur boxes?

GXTX: Not at that address. It's built into the XCalibur chip. According to *the scene*, would need someone with a 1.6 to properly verify. See: <https://xboxdevwiki.net/SMBus>

mborgerson: TIL. Let's revisit later

mborgerson:
Thanks!

hw/xbox/xbox.c
GXTX: Confirmed for not having it.
image

mborgerson: lpcmod has some info about xcalibur temp

https://bitbucket.org/psyko_chewbacca/lpcmod_os/src/b2b9fbb4dd43216acc936bee250029dee2492376/drivers/pci/i2cio.c?at=master#lines-221

PR #627, opened 2022-01-08
Allow to overwrite version with environment variables

j8r:
The change is fairly minimal, and allows to easily overwrite `XEMU_VERSION` with environment variables.

It could be done by using the already configured `XEMU_VERSION` file, but for now it won't work because the condition is after the one that checks the `.git` directory.

In the case of Flatpak, it allows us to just define an environment variable instead of adding a custom script, which will then call the build system inside - not ideal.

I have also added the feature for `XEMU_DATE` for those wanting reproducible builds.

### DIFF
--- a/ui/xemu.c
+++ b/ui/xemu.c
@@ -835,7 +835,7 @@ static void sdl2_display_very_early_init(DisplayOptions *o)
         SDL_GL_CONTEXT_PROFILE_CORE);
     SDL_GL_SetAttribute(SDL_GL_DOUBLEBUFFER, 1);
 
-    char *title = g_strdup_printf("xemu | v%s"
+    char *title = g_strdup_printf("xemu | %s"
 #if XEMU_DEBUG_BUILD
                                   " Debug"
 #endif
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -121,7 +121,7 @@ jobs:
       run: |
         pushd src
         echo -e "\
-        xemu (1:$(cat XEMU_VERSION)-0) unstable; urgency=medium\n\
+        xemu (1:$(cat XEMU_VERSION | cut -c2-)-0) unstable; urgency=medium\n\
           Built from $(cat XEMU_VERSION)\n\
          -- Matt Borgerson <contact@mborgerson.com>  $(date -R)" > debian/changelog
         popd
@@ -333,7 +333,7 @@ jobs:
       run: |
         pushd src
         echo -e "\
-        xemu (1:$(cat XEMU_VERSION)-0) unstable; urgency=medium\n\
+        xemu (1:$(cat XEMU_VERSION | cut -c2-)-0) unstable; urgency=medium\n\
           Built from $(cat XEMU_VERSION)\n\
          -- Matt Borgerson <contact@mborgerson.com>  $(date -R)" > debian/changelog
         popd
--- a/scripts/archive-source.sh
+++ b/scripts/archive-source.sh
@@ -76,7 +76,7 @@ done
 
 git rev-parse HEAD 2>/dev/null | tr -d '\n' > XEMU_COMMIT
 git symbolic-ref --short HEAD > XEMU_BRANCH
-git describe --tags --match 'xemu-v*' | cut -c 7- | tr -d '\n' > XEMU_VERSION
+git describe --tags --match 'xemu-v*' | cut -c 6- | tr -d '\n' > XEMU_VERSION
 tar -r --file "$tar_file" XEMU_COMMIT XEMU_BRANCH XEMU_VERSION
 
 exit 0
--- a/scripts/xemu-version.sh
+++ b/scripts/xemu-version.sh
@@ -3,7 +3,7 @@
 set -eu
 
 dir="$1"
-XEMU_DATE=$(date -u)
+: ${XEMU_DATE=$(date -u)}
 XEMU_COMMIT=$( \
   cd "$dir"; \
   if test -e .git; then \
--- a/scripts/xemu-version.sh
+++ b/scripts/xemu-version.sh
@@ -21,13 +21,13 @@ XEMU_BRANCH=$( \
 : ${XEMU_VERSION=$( \
   cd "$dir"; \
   if test -e .git; then \
-    git describe --tags --match 'xemu-v*' | cut -c 7- | tr -d '\n'; \
+    git describe --tags --match 'xemu-v*' | cut -c 6- | tr -d '\n'; \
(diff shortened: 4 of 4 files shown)


j8r:
@mborgerson have you the time to review it?

ui/xemu.c
```diff
@@ -835,7 +835,7 @@ static void sdl2_display_very_early_init(DisplayOptions *o)
         SDL_GL_CONTEXT_PROFILE_CORE);
     SDL_GL_SetAttribute(SDL_GL_DOUBLEBUFFER, 1);
 
-    char *title = g_strdup_printf("xemu | v%s"
+    char *title = g_strdup_printf("xemu | %s"
```
mborgerson, line 838: The `v` isn't extra. Are you passing in an extra `v` through the environment variable?

mborgerson:
Can you help me understand why this is necessary for flatpak builds?

If you need a .git free version for flatpak you can use the ppa-snapshot branch, which has an embedded version file

ui/xemu.c
j8r: It will become yes, otherwise the tittle will look like `vbuild-202201250122`

j8r:
It is not _necessary_, it will look better. For now the window tittle when launching `app.xemu.xemu` is: `xemu | v`.

The `build-` tags looks better, then this tag can be used as a version which is referenced in the Flatpak appstream metadata (so in https://flathub.org too)

ui/xemu.c
mborgerson: It's maybe a bit confusing, but there are two different version numbers here: one is the semantic version number `v0.6.2-55-gcaf84884c4` and the other is the build date `build-202201250122`. xemu displays the semantic version number at the top and in the about window and it is prefixed with a `v` here

j8r: Shouldn't the prefix added in `scripts/xemu-version.sh`, or before, rather than here? One can build xemu with any version (s)he wish, but this will keep appending a `v` for any value, even if not present.

j8r: How can we get this version number, in order to use in in the Flatpak build?

mborgerson: Why can you not use the provided build scripts, which automatically determine the build version number?

If you must build from some snapshot, you can clone the ppa-snapshot branch which will have all submodules embedded and correct version information

j8r: The Flatpak builder supports natively different build systems. It is generally preferable to avoid using the `simple` one (scripts) and e.g. `autotools`, `cmake` etc.

I guess the `ppa-snapshot` branch can be used then (even though I find the name not ideal).

Anyway, the points in this PR about being able to modify the date and version still stand: it should be possible to customize them freely.

j8r: The Flatpak builder supports natively different build systems. It is generally preferable to avoid using the `simple` one (scripts) and use e.g. `autotools`, `cmake` etc.

I guess the `ppa-snapshot` branch can be used then (even though I find the name not ideal).

Anyway, the points in this PR about being able to modify the date and version still stand: it should be possible to customize them freely.

j8r: I also think it will be more consistent to either prepend with `v` in the first place, and have it everywhere, or not prefixing it at all anywhere - I find it weird to have only this `v` prefix added  there.

mborgerson: > The Flatpak builder supports natively different build systems. It is generally preferable to avoid using the `simple` one (scripts) and use e.g. `autotools`, `cmake` etc.

Scripts are part of a project's build architecture, and a packager should use them.

> I guess the `ppa-snapshot` branch can be used then (even though I find the name not ideal).

Great. The ppa-snapshot branch is used to solve a problem of source releases when building Debian packages. If you have access to Git and egress for downloads though, it is not critical.

mborgerson:
> It is not _necessary_, it will look better. For now the window tittle when launching `app.xemu.xemu` is: `xemu | v`.
> 
> The `build-` tags looks better, then this tag can be used as a version which is referenced in the Flatpak appstream metadata (so in https://flathub.org too)

My biggest concern with this patch is not in its substance, but its motivation. Every official release (Windows, macOS, and Ubuntu PPA) and unofficial release (AUR) of xemu displays the semantic version (e.g. `v0.6.2-55-gcaf84884c4`) in the title bar and in the About dialog, and this is the version listed on the [xemu website](https://xemu.app).

Regardless of whether it _looks better_ or not, displaying the build date/tag as the version string is not consistent with other xemu releases and will inevitably lead to user confusion. Please make it so the flatpak build is consistent with other xemu releases.

j8r:
Ok then, agreed with that - will do!

So the `git describe --match 'xemu-v*' | cut -c 7- | tr -d '\n'` command is incorrect for `XEMU_VERSION`, I will change it too `git describe --match 'xemu-v*' | cut -c 7- | tr -d '\n'` to take the `v`.

j8r:
Ok then, agreed with that - will do!

So the `git describe --match 'xemu-v*' | cut -c 7- | tr -d '\n'` command is incorrect for `XEMU_VERSION`, I will change it too `git describe --match 'xemu-v*' | cut -c 6- | tr -d '\n'` to take the `v`.

j8r:
Ok then, agreed with that - will do!

So the `git describe --match 'xemu-v*' | cut -c 7- | tr -d '\n'` command is incorrect for `XEMU_VERSION`, I will change it to `git describe --match 'xemu-v*' | cut -c 6- | tr -d '\n'` to take the `v`.

j8r:
One thing remains that I haven't understood: what is the purpose of `build-202201250122`? Why not have a la Maven, `v0.6.2-55-gcaf84884c4-SNAPSHOT` or something like that?

j8r:
I'm working on the modification upstream, and the submodules in the `ppa-snapshot` are broken.
`Failed to download sources: module xemu: Not a gitlink tree: roms/seabios`

The submodules paths specified in the `.gitmodules` are no longer present in the branch.

Solutions I see:
- add `--shallow-submodules --recursive` to `.github/workflows/build.yml` in the `Integrate Debian packaging` step
- or, remove/modify the `.gitmodules` in this branch (through `.github/workflows/build.yml`)

j8r:
Also, why there is a `debian` branch instead of having a `debian` repository? What's more common in other projects is to have this files in `master` directly.

j8r:
Also, why there is a `debian` branch instead of having a `debian` directory in `master`? This is a more common approach in other projects.

Then an orphan branch can be checked out from master and then pushed from GitHub Actions.

j8r:
No longer needed

PR #660, opened 2022-01-19
nv2a: Match HW behavior when dealing with infinite fog coord

### LINKED ISSUE #653: Mechassault: Fog code causes intro videos to be rendered as a blank greyish screen
### Title

https://xemu.app/titles/4d530017/#MechAssault

### Bug Description

There appears to be a bug in the fog code that causes the MechAssault Microsoft Studios intro to be obscured by a blank grey screen. 

Commenting out the [`SET_FOG_ENABLE` handler](https://github.com/mborgerson/xemu/blob/3fd15218c76cedd89496db637c7ef5369ef35a63/hw/xbox/nv2a/pgraph.c#L1470) will allow the videos to be shown (but obviously will then prevent fog from working elsewhere and is just an indicator that there's something wrong w/ the fog rendering).

### Expected Behavior

The videos and menus should be fully visible.

### xemu Version

Latest master

Version: 0.6.2-51-g3fd15218c7
Branch: master
Commit: 3fd1521
Date: Thu Jan 13 04:44:48 PM UTC 2022

### System Information

CPU: Core i7-6700K @ 4GHz
GPU: NVIDIA GeForce GTX 1070

(Also seen on M1 Max - Big Sur)

### Additional Context

_No response_

abaire:
[Test shader](https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/shaders/fog_infinite_fogc_test.vs.cg)
[Test results](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-Fog_inf_coord)

Fixes #653 

Also removes some `FIXME`s on lines whose behavior is both described by nearby comments and matches observed HW (the `f =` lines describe the actual function implemented in the HW and documented by Microsoft, the `fogParam` comments match the observed behavior from Direct3D-generated binaries, and algebra explains the need to remove the bias added by D3D, so I'd argue there's nothing to be fixed).

### DIFF
--- a/hw/xbox/nv2a/shaders.c
+++ b/hw/xbox/nv2a/shaders.c
@@ -793,14 +793,25 @@ STRUCT_VERTEX_DATA);
         case FOG_MODE_LINEAR_ABS:
 
             /* f = (end - d) / (end - start)
-             *    fogParam[1] = 1 / (end - start)
-             *    fogParam[0] = 1 + end * fogParam[1];
+             *    fogParam[1] = -1 / (end - start)
+             *    fogParam[0] = 1 - end * fogParam[1];
              */
 
+            mstring_append(body,
+                "  if (isinf(fogDistance)) {\n"
+                "    fogDistance = 0.0;\n"
+                "  }\n"
+            );
             mstring_append(body, "  float fogFactor = fogParam[0] + fogDistance * fogParam[1];\n");
-            mstring_append(body, "  fogFactor -= 1.0;\n"); /* FIXME: WHHYYY?!! */
+            mstring_append(body, "  fogFactor -= 1.0;\n");
             break;
         case FOG_MODE_EXP:
+          mstring_append(body,
+                         "  if (isinf(fogDistance)) {\n"
+                         "    fogDistance = 0.0;\n"
+                         "  }\n"
+          );
+          /* fallthru */
         case FOG_MODE_EXP_ABS:
 
             /* f = 1 / (e^(d * density))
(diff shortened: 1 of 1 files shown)


hw/xbox/nv2a/shaders.c
```diff
@@ -793,14 +793,27 @@ STRUCT_VERTEX_DATA);
...
+                "  if (isinf(fogDistance)) {\n"
+                "    fogDistance = 0.0;\n"
+                "  }\n"
+            );
             mstring_append(body, "  float fogFactor = fogParam[0] + fogDistance * fogParam[1];\n");
-            mstring_append(body, "  fogFactor -= 1.0;\n"); /* FIXME: WHHYYY?!! */
+            mstring_append(body, "  fogFactor -= 1.0;\n");
             break;
         case FOG_MODE_EXP:
+          // Inf special casing should only be applied to the EXP case.
```
mborgerson: Comment doesn't match code? It's also applied for linear, above. If you mean it should only not be applied to EXP_ABS and other cases below, then that is implicit in the code fall-thru and I think the comment can be removed.

hw/xbox/nv2a/shaders.c
```diff
@@ -793,14 +793,27 @@ STRUCT_VERTEX_DATA);
...
+          mstring_append(body,
+                         "  if (isinf(fogDistance)) {\n"
+                         "    fogDistance = 0.0;\n"
+                         "  }\n"
+          );
+          // Intentional fallthrough, the rest of the handling is the same
+          // between EXP and EXP_ABS.
```
mborgerson: nit: Description is implied, we can simply write `/* fallthru */` to convey this and pass compiler accidental fall-thru checks

hw/xbox/nv2a/shaders.c
abaire: The latter, it is intended to be applied to both types of LINEAR and EXP (but not EXP_ABS). Comment removed.

hw/xbox/nv2a/shaders.c
abaire: Done.

mborgerson:
Thanks!

PR #664, opened 2022-01-19
nv2a: Make nv2a_dprintf toggleable

abaire:
Halo really does become intolerably slow w/ pgraph logging enabled. Added a trivial toggle in the UI so it can be enabled only when it's actually providing useful data. This is still guarded by the debug flag so it won't have any performance impact in normal builds.


abaire:
I think this is mostly obviated by the move to qemu tracing, which allows runtime toggling via the monitor https://www.qemu.org/docs/master/devel/tracing.html

PR #674, opened 2022-01-24
nv2a: Improve RenderDoc triggering

abaire:
Adds improved frame boundary detection for RenderDoc on Linux. 

I don't develop on Windows so I haven't implemented there, but it should be straightforward (https://renderdoc.org/docs/in_application_api.html just the `GetModuleHandleA` bit that needs to be brought in).

RenderDoc isn't actually supported on macOS so I haven't actually tried to test there yet.

### DIFF
--- a/hw/xbox/nv2a/debug.c
+++ b/hw/xbox/nv2a/debug.c
@@ -27,6 +27,14 @@
 #include <stdarg.h>
 #include <assert.h>
 
+#ifdef ENABLE_RENDERDOC
+#include <renderdoc_app.h>
+#include <dlfcn.h>
+
+static RENDERDOC_API_1_1_2 *rdoc_api = NULL;
+static int32_t renderdoc_capture_frames = 0;
+#endif
+
 static bool has_GL_GREMEDY_frame_terminator = false;
 static bool has_GL_KHR_debug = false;
 
@@ -52,6 +60,17 @@ void gl_debug_initialize(void)
        assert(glGetError() == GL_NO_ERROR);
 #endif
     }
+
+#ifdef ENABLE_RENDERDOC
+    void* renderdoc = dlopen("librenderdoc.so", RTLD_NOW | RTLD_NOLOAD);
+    if (renderdoc) {
+        pRENDERDOC_GetAPI RENDERDOC_GetAPI = (pRENDERDOC_GetAPI)dlsym(
+            renderdoc, "RENDERDOC_GetAPI");
+        int ret = RENDERDOC_GetAPI(eRENDERDOC_API_Version_1_1_2,
+                                   (void **)&rdoc_api);
+        assert(ret == 1 && "Failed to retrieve RenderDoc API.");
+    }
+#endif
 }
 
 void gl_debug_message(bool cc, const char *fmt, ...)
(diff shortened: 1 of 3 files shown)


hw/xbox/nv2a/debug.c
```diff
@@ -121,15 +140,40 @@ void gl_debug_label(GLenum target, GLuint name, const char *fmt, ...)
     va_end(ap);
 
     glObjectLabel(target, name, n, buffer);
+
+    GLenum err = glGetError();
+    assert(err != GL_INVALID_ENUM);
+    assert(err != GL_INVALID_OPERATION);
+    assert(err != GL_INVALID_VALUE);
+    assert(err == GL_NO_ERROR);
```
mborgerson, line 145: Redundant assertions

hw/xbox/nv2a/debug.h
```diff
@@ -40,6 +40,11 @@
 // #define DEBUG_NV2A_GL
 #ifdef DEBUG_NV2A_GL
 
+#if XEMU_DEBUG_BUILD == 1
```
mborgerson: We may not want to build in debug mode while enabling renderdoc. Let's just have the commented-out define, without the DEBUG_BUILD check

ui/xemu-hud.cc
```diff
@@ -2058,6 +2058,12 @@ static void process_keyboard_shortcuts(void)
     if (is_key_pressed(SDL_SCANCODE_GRAVE)) {
         monitor_window.toggle_open();
     }
+
+#ifdef ENABLE_RENDERDOC
+    if (is_key_pressed(SDL_SCANCODE_F10)) {
```
mborgerson, line 2067: Let's simply have a menu item for this (I will forget the key binding)

hw/xbox/nv2a/debug.h
```diff
@@ -149,6 +154,10 @@ extern NV2AStats g_nv2a_stats;
 const char *nv2a_profile_get_counter_name(unsigned int cnt);
 int nv2a_profile_get_counter_value(unsigned int cnt);
 
+#ifdef ENABLE_RENDERDOC
+void nv2a_dbg_renderdoc_capture_frames(uint32_t num_frames);
+#endif // ENABLE_RENDERDOC
```
mborgerson: Redundant comment (good when defines are complicated or spaced very far apart, but just noise here)

mborgerson:
In its current state the patch requires building in debug mode (for -ldl) to enable render doc tracing. I don't think we need to enforce such a requirement.

The preferred way to express the libdl requirement is through meson config; this is done for plugins for example. We can add an option in `configure` to enable renderdoc, then add libdl to executable dependencies and test against `CONFIG_RENDERDOC`.

I'm also fine to punt proper build integration to a follow-up PR; but take a look at provided feedback and let's drop the changes to build.sh

hw/xbox/nv2a/debug.c
abaire: Oops, this was me trying to track down the specific reason for a crash. Switched to just a single assert since it's otherwise going to assert far away.

hw/xbox/nv2a/debug.h
abaire: Done.

hw/xbox/nv2a/debug.h
abaire: Removed, but for context, this also helps when managing pending merges since git otherwise likes to erroneously flag conflicts on `#endif`s.

ui/xemu-hud.cc
abaire: Mind if we do both? Now that I'm used to the key binding it's way faster than having a menu item and has the benefit of not adding the menu rendering commands to the capture :)

Could also change it to be `F11` so it's the same as nsight (assuming it's not possible/useful to attach both)

abaire (commented):
sgtm. I've not looked at meson yet so will have to dig into how to pipe things through. I assume it's some combo of modifying `configure` and `meson_options.txt`/`meson.build`?

ui/xemu-hud.cc
mborgerson: I'm okay to have both, no preference on key

abaire: sg. I'm inclined to keep it F10 then, just in case I'm wrong and there actually is some compelling reason to support both.

mborgerson:
Thanks!

PR #672, opened 2022-01-24
nv2a: Implement vertex attribute setters

abaire:
Fixes #671 

[Tests here](https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/tests/attribute_carryover_tests.cpp)
Expected HW behavior is [here](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-Attrib_carryover#t-d00_00_10_10-dapng) but note that the bottom line / right hand triangle will not render correctly without #623.

Issue #671, opened 2022-01-24
pgraph: Several inline buffer vertex attribute setters are not implemented


abaire:
### Bug Description

Edit: See https://github.com/mborgerson/xemu/pull/672/files#r791195300 for additional setters that should be implemented.

While writing a test for #623 I found that xemu doesn't implement a handful of the vertex attribute setters [used around here](https://github.com/abaire/nxdk_pgraph_tests/blob/4e2776729ed6a3ea86b4eb15247bc9009ed7cff9/src/test_host.cpp#L493).

Specifically:
```
diffuse w/ 4 floats
nv2a: pgraph method (0): 0x97 -> 0x1550 (0x3f800000)
nv2a:     unhandled  (0x97 0x00001550)

normal w/ 3 floats
nv2a: pgraph method (0): 0x97 -> 0x1530 (0x0)
nv2a:     unhandled  (0x97 0x00001530)

specular w/ 4 floats
nv2a: pgraph method (0): 0x97 -> 0x1570 (0x2d002c)
nv2a:     unhandled  (0x97 0x00001570)

texcoord0 w/ 2 floats (and tc1, tc2, tc3)
nv2a: pgraph method (0): 0x97 -> 0x1590 (0x2e0036)
nv2a:     unhandled  (0x97 0x00001590)

```

### Expected Behavior

xemu should implement these setters in case some game makes use of them

### xemu Version

Latest master w/ #623 patched in, but that change does not remove any support.

Version: 0.6.2-54-g10b29fc45c
Branch:  updates_normal_inline_value
Commit:  10b29fc45c4f62f51fc44531349b61e5a152cc14
Date:    Mon Jan 24 08:47:18 PM UTC 2022

### System Information

OS: Kubuntu 21.04 - 5.11.0-38-generic
CPU: Core i7-6700K @ 4GHz
GPU: NVIDIA GeForce GTX 1070
GPU Driver: 4.6.0 NVIDIA 470.74

PR #675, opened 2022-01-25
nv2a: Fix slots for vertex attribute setters

abaire:
The values my test was using were insufficient to detect that some of the attributes were not being set correctly in the PR for #671.

The [new testbed that will cover the integer setters](https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/tests/attribute_explicit_setter_tests.cpp) has a better selection, this fixes the setters from #672 .

### DIFF
--- a/hw/xbox/nv2a/pgraph.c
+++ b/hw/xbox/nv2a/pgraph.c
@@ -2241,47 +2241,47 @@ DEF_METHOD_INC(NV097, SET_VERTEX4F)
     }
 }
 
-#define SET_VERTEX_ATTRIBUTE(command, attr_index, num_params) \
-    int slot = (method - (command)) / (num_params); \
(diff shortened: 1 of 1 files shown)


hw/xbox/nv2a/pgraph.c
```diff
@@ -2241,47 +2241,45 @@ DEF_METHOD_INC(NV097, SET_VERTEX4F)
     }
 }
 
-#define SET_VERTEX_ATTRIBUTE(command, attr_index, num_params) \
-    int slot = (method - (command)) / (num_params); \
+#define SET_VERTEX_ATTRIBUTE(command, attr_index) \
+    int slot = (method - (command)) / 4; \
```
mborgerson: Oops!

Also can you wrap this macro like this:
```c
do { \
... \
} while (0)
```

abaire: Definitely, done.

PR #676, opened 2022-01-25
nv2a: Add missing vertex attribute setters

### LINKED ISSUE #671: pgraph: Several inline buffer vertex attribute setters are not implemented

abaire:
I think this covers the rest of the attribute setters from Noveau.

[Tests](https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/tests/attribute_explicit_setter_tests.cpp)

[Golden results](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-Attrib_setter)

Fixes #671


hw/xbox/nv2a/nv2a_regs.h
```diff
@@ -1041,12 +1041,29 @@
 #   define NV097_SET_LIGHT_LOCAL_ATTENUATION                  0x00001068
 #   define NV097_SET_VERTEX4F                                 0x00001518
 #   define NV097_SET_NORMAL                                   0x00001530
```
mborgerson: We should use the type suffix `3F` here to be consistent with the other methods

hw/xbox/nv2a/pgraph.c
```diff
@@ -2241,7 +2241,118 @@ DEF_METHOD_INC(NV097, SET_VERTEX4F)
...
+#define FLOAT_SIGN(val) ((val) < 0 ? -1.0f : ((val) > 0 ? 1.0f : 0.0f))
+
+#define SET_VERTEX_ATTRIBUTE_SIGN(command, attr_index)                     \
+    do {                                                                   \
+        int slot = (method - (command)) / 4;                               \
+        unsigned int part = slot % 2;                                      \
+        VertexAttribute *attribute = &pg->vertex_attributes[(attr_index)]; \
+        pgraph_allocate_inline_buffer_vertices(pg, (attr_index));          \
+        int16_t val = parameter & 0xFFFF;                                  \
+        attribute->inline_value[part * 2 + 0] = FLOAT_SIGN(val);           \
```
mborgerson: Are you sure this is correct? It seems odd that the 16b value is clamped to {-1, 0, 1}. I expect that this is like SET_VERTEX_DATA2S and the signed 16b integer value should be converted to a float.

hw/xbox/nv2a/nv2a_regs.h
```diff
@@ -1041,12 +1041,29 @@
...
+#   define NV097_SET_NORMAL3S                                 0x00001540
 #   define NV097_SET_DIFFUSE_COLOR4F                          0x00001550
+#   define NV097_SET_DIFFUSE_COLOR3F                          0x00001560
+#   define NV097_SET_DIFFUSE_COLOR4I                          0x0000156C
```
mborgerson: The conventional type suffix for this should be `UB` (repeated below for other `I`s)

hw/xbox/nv2a/nv2a_regs.h
abaire: Done.

hw/xbox/nv2a/nv2a_regs.h
abaire: Done.

hw/xbox/nv2a/pgraph.c
abaire: Definitely wouldn't say I'm entirely sure, but it is aligned with what I see [out of the hardware](https://raw.githubusercontent.com/abaire/nxdk_pgraph_tests_golden_results/main/Attrib_setter/Setters-alpha.png)

Take a look at the result for `T02S`, the UV values for that test are 
```
        host_.SetTexCoord0i(0x4FFF, 0);
        host_.SetTexCoord0i(0, 0xFFFF);
        host_.SetTexCoord0i(1, 0x8000);
```

Your comment made me think about whether it could actually be a 16-bit float or a truncated 32-bit float stored as a uint16, so I tried both of those as well but I still see the signed behavior; setting any non-0 positive value (1-0x7FFF) seems to result in a 1.0.

It's entirely possible that I'm doing something wrong, but `NORMAL3S` acts dramatically differently for the same values and I can't seem to find any values that produce something that looks like fractional results.

abaire: Putting into draft mode until I have a chance to update the test to verify that it's not just passing raw values through to the shader for int16 tex coords, which would also explain the saturation behavior.

abaire: It was indeed an artifact of my test setup. Updated that to properly handle inputs outside of 0..1 and fixed the handling in xemu.

hw/xbox/nv2a/pgraph.c
```diff
@@ -2241,50 +2241,214 @@ DEF_METHOD_INC(NV097, SET_VERTEX4F)
...
+
+#undef SET_VERTEX_ATRIBUTE_TEX_2S
+
+#define SET_VERTEX_ATTRIBUTE_3F(command, attr_index, val_four)             \
     do {                                                                   \
         int slot = (method - (command)) / 4;                               \
         VertexAttribute *attribute = &pg->vertex_attributes[(attr_index)]; \
         pgraph_allocate_inline_buffer_vertices(pg, (attr_index));          \
         attribute->inline_value[slot] = *(float*)&parameter;               \
+        attribute->inline_value[3] = val_four;                             \
```
mborgerson: `val_four` is always `1.0` so we do not need it as a macro argument

abaire: Removed and renamed the macro to make it clearly non-generic.

hw/xbox/nv2a/pgraph.c
```diff
@@ -2241,50 +2241,212 @@ DEF_METHOD_INC(NV097, SET_VERTEX4F)
...
-#define SET_VERTEX_ATTRIBUTE(command, attr_index) \
+DEF_METHOD_INC(NV097, SET_NORMAL3S)
+{
+    int slot = (method - NV097_SET_NORMAL3S) / 4;
+    unsigned int part = slot % 2;
+    VertexAttribute *attribute =
+        &pg->vertex_attributes[NV2A_VERTEX_ATTR_NORMAL];
+    pgraph_allocate_inline_buffer_vertices(pg, NV2A_VERTEX_ATTR_NORMAL);
+    int16_t val = parameter & 0xFFFF;
+    attribute->inline_value[part * 2 + 0] = MAX(-1.0f, (float)val / 32767.0f);
```
mborgerson, line 2252: Are you sure this is correct? I would expect this is not different from the other 16b signed variants

abaire: I won't say I'm completely certain, but the evidence seems relatively compelling. 

You can see from the [test code](https://github.com/abaire/nxdk_pgraph_tests/blob/f09ecef58670fd74cb813a8c9a306e46078ef754/src/tests/attribute_explicit_setter_tests.cpp#L140) that I'm sending int16 values into the shader, then treating them as -1..1 values when mapping to colors (the final two parameters are the bias and multiplier).

Contrast that with the [tests for the texture coordinates](https://github.com/abaire/nxdk_pgraph_tests/blob/f09ecef58670fd74cb813a8c9a306e46078ef754/src/tests/attribute_explicit_setter_tests.cpp#L299) which treat the results as -32768..32767.

If the hardware wasn't normalizing the normal I'd expect [the results](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-Attrib_setter) to be very clearly wrong (in this case I believe the values would roughly be equivalent to clamp(0..1)). 

Just to be as sure as possible I changed the bias and multiplier to match the tex coord behavior and the resultant triangle is about 50% grey which makes sense if the values are normalized before hitting the shader ((32767 + 1) / 65535 is still going to be ~0.5)

mborgerson: Do you have normal normalization enabled in your test? I'm not sure if hardware will normalize or not in programmable path with it on. I think not, but worth checking.

abaire: I wasn't modifying it explicitly in my test, but setting it explicitly to false produces the exact same output.

This behavior makes sense to me; normalizing (1, -1, 0) should give ~(0.7, -0.7, 0) which would ~~clamp to (0.7, 0, 0) producing a red~~(note below) vertex whereas the resulting color is ~(.98, .01, .5) which feels about right given interpolation with the other two vertices.

*Edit: The bias/mult code would convert (0.7, -0.7, 0) -> (.85, 0.15, 0.5) in the shader so it wouldn't be red, but it'd still be a ways off from the ~(1, 0, 0.5) result*

To double check though https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/tests/attribute_explicit_setter_tests.cpp#L92 is the change I made - if there is another normalization setting that I should be looking at, please let me know, I have very little in the way of expertise in nv2a commands :)

mborgerson: I'll merge now; let's revisit in the future

mborgerson:
Thanks

Issue #567, opened 2021-11-23
Palettized volume texture format conversion is unimplemented


Triticum0:
### Title

https://xemu.app/titles/4a410006/#Room-Zoom-Race-for-Impact

### Bug Description

When trying to go in-game the game crashes with this assertion:
https://github.com/mborgerson/xemu/blob/5faa2419e7cfca1e03a8e3113de2df6453caf2e5/hw/xbox/nv2a/pgraph.c#L6368

### Expected Behavior

Should be able to make it in-game

### xemu Version

Version: 0.6.2
Branch: master
Commit: 5faa241
Date: Sat Nov 20 08:53:11 UTC 2021
Last-known Working: Unknown

### System Information

Field | Value
-- | --
OS | Ubuntu 20.04.3 LTS
CPU | Intel(R) Pentium(R) CPU 3825U @ 1.90GHz
Graphics Device | Mesa Intel(R) HD Graphics (BDW GT1)
Graphics Driver | Drivers: 4.6 (Core Profile) Mesa 21.0.3

### Additional Context
Also NV097_SET_TEXTURE_FORMAT_COLOR_LC_IMAGE_CR8YB8CB8YA8,YB8CR8YA8CB8 and NV097_SET_TEXTURE_FORMAT_COLOR_SZ_R6G5B5) should properly also be fixed even through there no known use case for these assertions.
Edit: Tested on lasted build still crashes

Ernegien:
[This](https://github.com/Ernegien/nxdk/blob/9dd778137f03dd81826db31b3097eebde48f11a8/samples/texture-formats/main.c#L77) would be a good starting point for unit-testing, which I may do later :)

dracc:
I haven't inspected code, issue is probably still present, but please consider updating your xemu - 0.5.0 is quite old by now.

Triticum0:
@dracc I tested on masters, still crashes with the same assertion

abaire:
Parameters when it asserts:

```
s.color_format == NV097_SET_TEXTURE_FORMAT_COLOR_SZ_I8_A8R8G8B8

width = 128
height = 128
depth = 32
row_pitch = 128
slice_pitch = 16384

palette data looks like it's at least pointing at something structured.
```

PR #684, opened 2022-01-28
nv2a: Fix swizzle_box for volumetric textures

abaire:
I don't think xemu actually uses this method in a way that exercises the bug, but anybody leveraging this to create volumetric testers in their xbe code will end up with corrupted results for slices beyond the first (and the first slice will be the last source slice).

I'm [currently leveraging this code](https://github.com/abaire/nxdk_pgraph_tests/blob/volumetric_texture/src/tests/volume_texture_tests.cpp) to create a repro case for #567 and can confirm that with this change the resultant (non-palettized) volumetric textures display correctly in xemu and on hw.


mborgerson:
Doh! Thanks

PR #693, opened 2022-02-01
nv2a: Add --enable-renderdoc build config

abaire:
Followup from #674

### DIFF
--- a/hw/xbox/nv2a/debug.h
+++ b/hw/xbox/nv2a/debug.h
@@ -40,11 +40,9 @@
 // #define DEBUG_NV2A_GL
 #ifdef DEBUG_NV2A_GL
 
-// Improve frame capture boundaries with RenderDoc.
-// #define ENABLE_RENDERDOC
-
 #include <stdbool.h>
 #include "gl/gloffscreen.h"
+#include "config-host.h"
 
 void gl_debug_initialize(void);
 void gl_debug_message(bool cc, const char *fmt, ...);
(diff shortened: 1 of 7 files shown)


hw/xbox/nv2a/debug.h
```diff
@@ -37,14 +37,12 @@
 #endif
 
 // #define DEBUG_NV2A_REG
-// #define DEBUG_NV2A_GL
+ #define DEBUG_NV2A_GL
```
mborgerson: Should not be enabled by default

abaire: Missed that after testing, sorry!

meson.build
```diff
@@ -1224,6 +1224,22 @@ endif
...
+if have_system and get_option('renderdoc').enabled()
+  if not cc.has_header('renderdoc_app.h')
+    error('renderdoc required, but renderdoc_app.h was not found')
+  else
+    if targetos == 'linux'
+       if not libdl.found()
+         error('renderdoc required, but libdl was not found')
+         renderdoc = not_found
+       else
+         renderdoc = declare_dependency(link_args: '-ldl')
```
mborgerson: We already did the libdl detection above, so to conditionally link libdl it's something like:

`specific_ss.add(when: 'CONFIG_RENDERDOC', if_true: [libdl])`

abaire: Done (I think)

mborgerson:
Thanks

PR #699, opened 2022-02-02
configure: Move optimization spec from build.sh

mborgerson:
And use -Og for debug builds.

### DIFF
--- a/configure
+++ b/configure
@@ -5290,7 +5290,7 @@ if test "$skip_meson" = no; then
         -Ddocdir="$docdir" \
         -Dqemu_firmwarepath="$firmwarepath" \
         -Dqemu_suffix="$qemu_suffix" \
-        -Doptimization=$(if test "$debug" = yes; then echo 0; else echo 2; fi) \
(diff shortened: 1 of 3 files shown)


configure
```diff
@@ -5290,7 +5290,7 @@ if test "$skip_meson" = no; then
         -Ddocdir="$docdir" \
         -Dqemu_firmwarepath="$firmwarepath" \
         -Dqemu_suffix="$qemu_suffix" \
-        -Doptimization=$(if test "$debug" = yes; then echo 0; else echo 2; fi) \
+        -Doptimization=$(if test "$debug" = yes; then echo g; else echo 3; fi) \
```
GXTX, line 5293: Perhaps `O2` would be best?

mborgerson: > Perhaps `O2` would be best?

Why would `O2` be best?

GXTX: Upstream default & more widely used as a sane release default.

mborgerson: xemu has been using -O3 for a long time now.

If you can provide a good technical reason why we should prefer O2 over O3, I'll consider it

PR #717, opened 2022-02-03
nv2a: Handle LSB mux flag

### LINKED ISSUE #716: Handling of combiner control mux LSB/MSB flag is incorrect
### Bug Description

The pixel shader doesn't actually use `PS_COMBINERCOUNT_MUX_MSB` when emitting the muxing code and produces incorrect results for shaders using muxers. It is doing a simple compare against 0.5 in both configurations rather than checking bits and respecting the value of the flag.

[Test program](https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/tests/combiner_tests.cpp)

Screenshot_20220202_233621

### Expected Behavior

Output should match the [HW results here](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-Combiner#muxpng)

In my test I'm leveraging the c0 constant to explicitly set alpha as an 8-bit value. In this configuration, the MSB mux is only enabled if the high bit is enabled, and the LSB mux is only enabled if the low bit is enabled, as expected.

### xemu Version

Version: 0.6.2-64-g758e36d398
Branch:  master
Commit:  758e36d3983f3acde38017276f31bb75ff32a92a
Date:    Thu Feb  3 06:52:47 AM UTC 2022

### System Information

CPU: Core i7-6700K @ 4GHz
GPU: NVIDIA GeForce GTX 1070

### Additional Context

_No response_

abaire:
Fixes #716 

May fix some games as well; I'm not sure how often the muxer mode is set to LSB, 

In Halo I see it used in the opening cutscene for The Silent Cartographer, but it's rendering something extremely far in the distance so I didn't notice any effect (it looks like it uses a shield texture or something)..

### DIFF
--- a/hw/xbox/nv2a/psh.c
+++ b/hw/xbox/nv2a/psh.c
@@ -469,8 +469,13 @@ static void add_stage_code(struct PixelShader *ps,
     if (output.muxsum_op == PS_COMBINEROUTPUT_AB_CD_SUM) {
         sum = mstring_from_fmt("(%s + %s)", mstring_get_str(ab), mstring_get_str(cd));
     } else {
-        sum = mstring_from_fmt("((r0.a >= 0.5) ? %s(%s) : %s(%s))",
-                               caster, mstring_get_str(cd), caster, mstring_get_str(ab));
+        if (ps->flags & PS_COMBINERCOUNT_MUX_MSB) {
+            sum = mstring_from_fmt("((r0.a >= 0.5) ? %s(%s) : %s(%s))",
+                                   caster, mstring_get_str(cd), caster, mstring_get_str(ab));
+        } else {
+            sum = mstring_from_fmt("(((uint(r0.a * 255.0) & 1u) == 1u) ? %s(%s) : %s(%s))",
+                                   caster, mstring_get_str(cd), caster, mstring_get_str(ab));
+        }
     }
 
     MString *sum_mapping = get_output(sum, output.mapping);


hw/xbox/nv2a/psh.c
```diff
@@ -469,8 +469,13 @@ static void add_stage_code(struct PixelShader *ps,
     if (output.muxsum_op == PS_COMBINEROUTPUT_AB_CD_SUM) {
         sum = mstring_from_fmt("(%s + %s)", mstring_get_str(ab), mstring_get_str(cd));
     } else {
-        sum = mstring_from_fmt("((r0.a >= 0.5) ? %s(%s) : %s(%s))",
-                               caster, mstring_get_str(cd), caster, mstring_get_str(ab));
+        if (ps->flags & PS_COMBINERCOUNT_MUX_MSB) {
```
mborgerson, line 472: nit: let's simplify a bit by moving the condition in

```c
        sum = mstring_from_fmt("((%s) ? %s(%s) : %s(%s))",
                               (ps->flags & PS_COMBINERCOUNT_MUX_MSB) ?
                                   "r0.a >= 0.5" :
                                   "(uint(r0.a * 255.0) & 1u) == 1u",
                               caster, mstring_get_str(cd), caster,
                               mstring_get_str(ab));
```

mborgerson: Nevermind, I'll just do this on a follow up. There's other stuff I want to change here too

Issue #286, opened 2021-05-19
Graphical issues in 'Grabbed by the Ghoulies'


Phat-Joe:
I believe this to be a lighting issue. cutscenes don't dsiplay properly. The game presents the story in a comic book fashion, and the pages are unreadable due to them being black. When ever the page turns, you can catch a glimpse of the panels, thats why I believe its the lighting of the game engine

theboy181:
Do you happen to know what part of the emulation codes is causing this issue? If not I would share this information with the compatibility section in the emulator instead.

mborgerson:
Thanks for reporting. Please update this issue to include screenshots of what happens in xemu vs on a real Xbox and include the version that you've tested with. You can update this issue or file a new one with the recently added templates.

PR #718, opened 2022-02-03
nv2a: Fix black pixels on compressed texture type

Metalsasi:
Compressed textures might have a height smaller than the minimum height required by opengl. The unused texture pixels are black
and so it blackens many materials. It seems like that the original hardware either ignores the black pixels, or fills them up with the available ones. This fix uses the latter. Only the case "2 pixel height" is handled this way because that was the only one I encountered. There may be other heights in other titles, but i am not sure if the behavior is the same.

Fixes black lighting: https://xemu.app/titles/4d530053/

### DIFF
--- a/hw/xbox/nv2a/pgraph.c
+++ b/hw/xbox/nv2a/pgraph.c
@@ -6596,6 +6596,20 @@ static uint8_t* convert_texture_data(const TextureShape s,
     }
 }
 
+static uint8_t *corrected_black_pixels(const uint8_t *data, GLsizei data_size)
+{
+    uint8_t *replacement_buffer = g_malloc(data_size);
+    memcpy(replacement_buffer, data, data_size);
+
+    for (int i = 0; i < data_size; i += 8)
+    {
+        replacement_buffer[i + 6] = replacement_buffer[i + 4];
+        replacement_buffer[i + 7] = replacement_buffer[i + 5];
+    }
+
+    return replacement_buffer;
+}
+
 static void upload_gl_texture(GLenum gl_target,
                               const TextureShape s,
                               const uint8_t *texture_data,
@@ -6643,6 +6657,11 @@ static void upload_gl_texture(GLenum gl_target,
         for (level = 0; level < s.levels; level++) {
             if (f.gl_format == 0) { /* compressed */
 
+                bool use_black_pixel_correction = false;
+                if (height == 2)
+                {
+                    use_black_pixel_correction = true;
+                }
                 width = MAX(width, 4); height = MAX(height, 4);
 
                 unsigned int block_size;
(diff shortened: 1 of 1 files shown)


hw/xbox/nv2a/pgraph.c
```diff
@@ -6596,6 +6596,20 @@ static uint8_t* convert_texture_data(const TextureShape s,
     }
 }
 
+static uint8_t *corrected_black_pixels(const uint8_t *data, GLsizei data_size)
+{
+    uint8_t *replacement_buffer = g_malloc(data_size);
+    memcpy(replacement_buffer, data, data_size);
+
+    for (int i = 0; i < data_size; i += 8)
+    {
```
abaire, line 6605: Nit: Move brace to previous line to align with xemu style (here and below, but not on function definitions).

hw/xbox/nv2a/pgraph.c
```diff
@@ -6596,6 +6596,20 @@ static uint8_t* convert_texture_data(const TextureShape s,
     }
 }
 
+static uint8_t *corrected_black_pixels(const uint8_t *data, GLsizei data_size)
```
abaire, line 6599: Nit: Defer to @mborgerson obviously, but this feels like it should be a verb like `correct_black_pixels`. Even then it seems like it should describe what it's doing rather than the end effect. Maybe `pad_compressed_texture_data`?

It'd also be useful to create a test (e.g., using the nxdk) to validate the behavior; your description mentions that it either ignores or reuses pixels, do you know if that's dependent on some other setting like the texture addressing mode?

Metalsasi: I chose the naming because it describes what problem it solves, but yes maybe your naming would be more technically correct, so i guess i can rename it to whatever seems the most accurate.

I was thinking about writing tests, but I am not yet quite familiar with nxdk, so if it necessary. it will take a while. I didn't know about texture addressing modes, so I will look into that as well.

abaire: > I was thinking about writing tests, but I am not yet quite familiar with nxdk, so if it necessary. it will take a while. I didn't know about texture addressing modes, so I will look into that as well.

No problem, you can use https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/tests/texture_format_tests.cpp as a foundation if you'd like. Feel free to ping me with questions if you decide to build tests and need any help.

Metalsasi:
I could confirm the similar behavior on hardware. pgraph test is on the way.

Metalsasi:
I tested the hardware with 1, 2 and 3 pixel heights for this particular texture format.
It seems like the case of 1 and 2 px heights, the hw ignores the remaining pixels.
Oddly enough if the texture has 3px height, the hw behaves like the 1px scenario, meaning it ignores the 2nd 3rd and 4th pixel.
Also xemu as it current state asserts and fails for 1px height, because 1D textures does not seem to be supported, but the hw does not appear to be crashing. 

I will fix the current PR to make the changes more closely resemble the "ignoring" behaviour.

LeJARB:
Hi there ! I've compile a Windows executable with that code and the game is really more beautiful !
I wanted to ask when this will be add to the master branch ?

Triticum0:
Close #286 when merged

LeJARB:
It will be merged soon ?

abaire:
> It will be merged soon ?

It's waiting for @Metalsasi to make additional changes to better match HW behavior; the PR is not in a mergeable state right now, which is why it has the "pending author response" tag.

Metalsasi:
Hi, life got a little busy to me in the previous months, so I didn't have time to finish the changes. I am not around my dev computer around this time, so the soonest I can push a change is the end of May when I get back to my place.

LeJARB:
Ok, I'm impatient to see the results (^_^)
And also do not have to compile manually

mborgerson:
Thank you for the investigation and work on this! I have made some improvements in #962 that I plan to merge in favor of the approach in this patch.

PR #720, opened 2022-02-05
Fix combiner independence

### LINKED ISSUE #51: Some Halo:CE fog effects broken
Fog on certain levels (Halo, Silent Cartographer) does not appear.

Xbox:

xemu:

Screenshots courtesy ngrst183.

Title: 4d530004

### LINKED ISSUE #53: Halo:CE laser effect missing parts
There is apparently blue colored tube object surrounding the beam which is not rendering correctly.

Xbox:

xemu:

Screenshots courtesy ngrst183.

Title: 4d530004

abaire:
Makes combiner stage output independent when assigning multiple results. (e.g., R1 can be set with the previous value of R0 in the same stage as R0 is assigned with the previous value of R1 without the one affecting the other).

[Test](https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/tests/combiner_tests.cpp#L148)
[Results](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-Combiner#independencepng)

Fixes #51
(Apparently, based on reports in xemu#general) fixes #53 

Screenshot_20220204_192824

Screenshot_20220204_192835

### DIFF
--- a/hw/xbox/nv2a/psh.c
+++ b/hw/xbox/nv2a/psh.c
@@ -448,10 +448,14 @@ static void add_stage_code(struct PixelShader *ps,
     MString *cd_dest = get_var(ps, output.cd, true);
     MString *muxsum_dest = get_var(ps, output.muxsum, true);
 
+    bool assign_ab = false;
+    bool assign_cd = false;
+    bool assign_muxsum = false;
+
     if (mstring_get_length(ab_dest)) {
-        mstring_append_fmt(ps->code, "%s.%s = clamp(%s(%s), -1.0, 1.0);\n",
-                           mstring_get_str(ab_dest), write_mask, caster,
-                           mstring_get_str(ab_mapping));
+        mstring_append_fmt(ps->code, "ab.%s = clamp(%s(%s), -1.0, 1.0);\n",
+                           write_mask, caster, mstring_get_str(ab_mapping));
+        assign_ab = true;
     } else {
         mstring_unref(ab_dest);
         mstring_ref(ab_mapping);
@@ -459,22 +463,13 @@ static void add_stage_code(struct PixelShader *ps,
     }
 
     if (mstring_get_length(cd_dest)) {
-        mstring_append_fmt(ps->code, "%s.%s = clamp(%s(%s), -1.0, 1.0);\n",
-                           mstring_get_str(cd_dest), write_mask, caster,
-                           mstring_get_str(cd_mapping));
+        mstring_append_fmt(ps->code, "cd.%s = clamp(%s(%s), -1.0, 1.0);\n",
+                           write_mask, caster, mstring_get_str(cd_mapping));
+        assign_cd = true;
     } else {
         mstring_unref(cd_dest);
         mstring_ref(cd_mapping);
         cd_dest = cd_mapping;
-    }
-
-    if (!is_alpha && output.flags & PS_COMBINEROUTPUT_AB_BLUE_TO_ALPHA) {
-        mstring_append_fmt(ps->code, "%s.a = %s.b;\n", mstring_get_str(ab_dest),
-                           mstring_get_str(ab_dest));
-    }
-    if (!is_alpha && output.flags & PS_COMBINEROUTPUT_CD_BLUE_TO_ALPHA) {
-        mstring_append_fmt(ps->code, "%s.a = %s.b;\n", mstring_get_str(cd_dest),
-                           mstring_get_str(cd_dest));
     }
 
     MString *muxsum;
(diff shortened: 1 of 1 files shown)


hw/xbox/nv2a/psh.c
```diff
@@ -438,45 +438,71 @@ static void add_stage_code(struct PixelShader *ps,
...
+        assign_ab = true;
     } else {
         mstring_unref(ab_dest);
         mstring_ref(ab_mapping);
         ab_dest = ab_mapping;
     }
 
     if (mstring_get_length(cd_dest)) {
-        mstring_append_fmt(ps->code, "%s.%s = clamp(%s(%s), -1.0, 1.0);\n",
-                           mstring_get_str(cd_dest), write_mask, caster, mstring_get_str(cd_mapping));
```
MasonT8198: Why was this changed to cd.%s

hw/xbox/nv2a/psh.c
```diff
@@ -438,45 +438,71 @@ static void add_stage_code(struct PixelShader *ps,
...
     }
     if (!is_alpha && output.flags & PS_COMBINEROUTPUT_CD_BLUE_TO_ALPHA) {
-        mstring_append_fmt(ps->code, "%s.a = %s.b;\n",
-                           mstring_get_str(cd_dest), mstring_get_str(cd_dest));
+        mstring_append_fmt(ps->code, "cd.a = cd.b;\n");
     }
 
     MString *sum;
     if (output.muxsum_op == PS_COMBINEROUTPUT_AB_CD_SUM) {
         sum = mstring_from_fmt("(%s + %s)", mstring_get_str(ab), mstring_get_str(cd));
```
PatrickvL: Shouldn't all 3 mux/sum paths read from the new `ab` and `cd` variables now?

hw/xbox/nv2a/psh.c
PatrickvL: Because he introduced two intermediate storage variables, specifically to fix this issue with

PatrickvL (commented):
Xemu isn't big on describing *why* code is written like it is, but this change is a good example of something that a future reader might see and think of "let's optimise this into a single assignment statement", overlooking the fact this specific construct is here to avoid sequential variable access (while hardware turns out to calculate in parallel). So IMHO even a single comment line on that would help tremendously here

PatrickvL:
The pull request description should also mention that it includes handling of the `PS_COMBINERCOUNT_MUX_MSB` flag

abaire:
> The pull request description should also mention that it includes handling of the `PS_COMBINERCOUNT_MUX_MSB` flag

It doesn't intentionally. This PR is just dependent on several other of my PRs that need to be merged first. As far as I'm aware the xemu convention is one-fix-one-commit so any PR with multiple commits is generally squashed pre-review.

abaire:
> Xemu isn't big on describing _why_ code is written like it is, but this change is a good example of something that a future reader might see and think of "let's optimise this into a single assignment statement", overlooking the fact this specific construct is here to avoid sequential variable access (while hardware turns out to calculate in parallel). So IMHO even a single comment line on that would help tremendously here

I don't disagree, but @mborgerson has asked me to remove comments a number of times so I'm defaulting to letting him ask for explanation where necessary.

hw/xbox/nv2a/psh.c
abaire: 'ab' and 'cd' in the C code actually contain the raw calculations (see ~line 419). They are only conditionally assigned to the glsl variables based on whether the results are referenced as part of the combiner output, so it would be unsafe to blindly use them here without adding code to blindly do the calculation and assignment to intermediates above.

hw/xbox/nv2a/psh.c
abaire: > Because he introduced two intermediate storage variables, specifically to fix this issue with

Yeah, the fix is indeed to use intermediate vars so an output that sets a variable dependent on another variable in the output is not then dependent on the ordering they were specified as outputs (matching HW behavior). 

E.g., if the output of the last stage had R0 set to (1.0, 0.0, 1.0) and the output of this stage is defined as `R1 = R0 * T0` and `R0 = T1 * C0` the final output should be: `R1 = (1.0, 0.0, 1.0) * T0`, `R0 = T1 * C0` whereas the current xemu code makes the value of R1 depend on whether it was calculated in AB (it'll be correct) or CD (it'll be wrong, because R0 gets updated before R1 is calculated and it erroneously uses the new value).

hw/xbox/nv2a/psh.c
```diff
@@ -438,45 +438,71 @@ static void add_stage_code(struct PixelShader *ps,
...
     } else {
         mstring_unref(cd_dest);
         mstring_ref(cd_mapping);
         cd_dest = cd_mapping;
     }
 
     if (!is_alpha && output.flags & PS_COMBINEROUTPUT_AB_BLUE_TO_ALPHA) {
-        mstring_append_fmt(ps->code, "%s.a = %s.b;\n",
-                           mstring_get_str(ab_dest), mstring_get_str(ab_dest));
+        mstring_append(ps->code, "ab.a = ab.b;\n");
```
PatrickvL: Admittedly an already existing issue might be that this flag is set, and thus ab is read, while ab isn't assigned to.

The same remark for cd below

hw/xbox/nv2a/psh.c
PatrickvL: Still, this currently results in duplicate expressions, which puts additional burden on the shader compiler, which could be avoided and reduced into a single assignment, when the ab and cd variables are used in these mux/sum paths. Indeed, that does require assignment to ab and/or cd, but that could also be evaluated upfront similar to the new assign_* booleans.
Such "require_*" booleans could also help fixing the potential "read without assignment" issue mentioned above.

hw/xbox/nv2a/psh.c
abaire: In practice I'm not sure this matters; if the flag is set but the combiner is discarding the calculated AB value, it doesn't matter if this produces incorrect data. That said gating behind whether the values are to be used or not makes complete sense and is straightforward. This also caused me to add an additional test for the b2a behavior that pointed out the fact that I was introducing a bug where the alpha was never actually used in practice (fixed now) - thanks!

hw/xbox/nv2a/psh.c
abaire: Definitely agree with you, I think Matt's preference in general has been to put in the basic fix, then optimize, so I'm inclined to do cleanup in a second pass since I don't think this PR changes the duplication in the case where the AB/CD results are assigned along with the muxsum.

ko81e24wy:
this PR fix lighting problem of NBA LIVE 2005
master
2022-02-06 20-33-38 的屏幕截图
PR
2022-02-08 01-42-13 的屏幕截图

mborgerson:
Awesome, thank you!

PR #727, opened 2022-02-07
Fix submodules checkout in GitHub Actions

j8r:
Hopefully it can fix the missing - but not necessary - submodules in the `ppa-snapshot` branch, which was making `flatpak-builder` return an error when running its `git` commands to initialize the build.

If this doesn't work, another option is to remove those from the `.gitmodules` file, or maybe do the commands ourselves without using https://github.com/peaceiris/actions-gh-pages.

### DIFF
--- a/scripts/archive-source.sh
+++ b/scripts/archive-source.sh
@@ -77,6 +77,9 @@ done
 git rev-parse HEAD 2>/dev/null | tr -d '\n' > XEMU_COMMIT
 git symbolic-ref --short HEAD > XEMU_BRANCH
 git describe --tags --match 'xemu-v*' | cut -c 7- | tr -d '\n' > XEMU_VERSION
-tar -r --file "$tar_file" XEMU_COMMIT XEMU_BRANCH XEMU_VERSION
+# Not all submodules are cloned, which can produce errors.
+# It is also no longer needed
+tar --delete --file "$tar_file" .gitmodules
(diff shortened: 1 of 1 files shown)


mborgerson:
> Hopefully it can fix the missing - but not necessary - submodules in the `ppa-snapshot` branch, which was making `flatpak-builder` return an error when running its `git` commands to initialize the build.

Submodule source should be there. e.g. https://github.com/mborgerson/xemu/tree/ppa-snapshot/ui/imgui

j8r:
There are some submodules missing from https://github.com/mborgerson/xemu/blob/ppa-snapshot/.gitmodules in the `ppa-snapshot` branch. For instance, lots of `roms/` submodules, like `roms/seabios` not in https://github.com/mborgerson/xemu/tree/ppa-snapshot/roms.

mborgerson:
Try dropping the .gitmodules file before your build

j8r:
Ho good idea I didn't think about this, it does the trick.

mborgerson:
I'll close this now because we want to avoid pulling unnecessary submodules.

Thanks for your continued help making xemu available to people!

j8r:
Thank you! Does it mean you'll work another solution to fix this issue, like modifying the `.gitmodules` directly in `master`?
My last commit was removing it totally from the source archive, thus also in the `ppa-snapshot` branch.

PR #732, opened 2022-02-11
nv2a: adjust display resolution in 1080i mode

### LINKED ISSUE #73: HDTV 1080i issues, and observations
There are a few games that have the ability to played with the 1080i option in xemu. One game that I have tested has an issue when trying to boot with 1080i is ``Dragons Lair 3D`` (remove videos to get it to boot into game)

This game will load properly and work in 1080i mode, but only after several reboots (ctr-r), eventually the game catches the output signal and will display correctly. 

First boot
image

after several resets
image

Game is set to 1080i and Scaling is set to center in my testing.

antangelo:
Detects 1080i mode by resolution and sets the height to 1080 in its place. This code was a suggestion made by @abaire to fix XBMC issues at 1080i that @LoveMHz was experiencing.

I've looked into `vga_get_resolution` (and `vga_ioport_write` where the relevant registers are set) to see if the height can be set to 1080 when the resolution is set by the guest program. The height as stored by the vga registers is capped at 1024, however, making this infeasible.

Looking at the datasheet for the Conexant encoder, specifically table E-2 in the appendix, it's possible to determine the video mode from the encoder registers. However, this approach seems a bit excessive, and checking the resolution in pgraph is functionally identical. Supporting Xcalibur using this method will require reverse engineering as well.

Resolves #73 . Tested on UnleashX running at 1080i.

### DIFF
--- a/hw/xbox/nv2a/pgraph.c
+++ b/hw/xbox/nv2a/pgraph.c
@@ -4907,6 +4907,11 @@ static void pgraph_render_display(NV2AState *d, SurfaceBinding *surface)
     d->vga.get_offsets(&d->vga, &pline_offset, &pstart_addr, &pline_compare);
     int line_offset = surface->pitch / pline_offset;
 
+    /* Adjust viewport height for interlaced mode, used only in 1080i */
+    if (d->vga.cr[NV_PRMCIO_INTERLACE_MODE] != NV_PRMCIO_INTERLACE_MODE_DISABLED) {
+        height *= 2;
+    }
+
     pgraph_apply_scaling_factor(pg, &width, &height);
 
     glBindFramebuffer(GL_FRAMEBUFFER, d->pgraph.disp_rndr.fbo);
--- a/hw/xbox/nv2a/nv2a_regs.h
+++ b/hw/xbox/nv2a/nv2a_regs.h
@@ -619,6 +619,10 @@
 #define NV_PCRTC_RASTER                                  0x00000808
 
 
+#define NV_PRMCIO_INTERLACE_MODE                         0x00000039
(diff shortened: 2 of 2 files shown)


hw/xbox/nv2a/pgraph.c
```diff
@@ -4907,6 +4907,10 @@ static void pgraph_render_display(NV2AState *d, SurfaceBinding *surface)
     d->vga.get_offsets(&d->vga, &pline_offset, &pstart_addr, &pline_compare);
     int line_offset = surface->pitch / pline_offset;
 
+    if (width == 1920 && height == 540) {
```
abaire: I wonder if we should use a symbolic constant so it's clearer that 540 is 1080i?

abaire: Also, do we need to/should we do anything for `480i`?

antangelo: Going off what LoveMHz said in the discord chat, the Xbox seems to run the GPU at 480p resolutions in 480i mode. If I add similar code to detect 480i (`width == 640 && height == 320`), and run a 480i game, it doesn't execute.

I'll let Matt chime in on whether he wants a symbolic constant here. I think a comment might be more appropriate given it'll be used only here. It's also possible he prefers an alternate approach to this one in any case.

GXTX: Perhaps a comment above this, skimming the code it's not immediately obvious

antangelo:
@LoveMHz found that the VGA register `0x39` stores the interlaced video state, so I've updated this change to use that register to detect 1080i mode. Qemu's VGA code doesn't seem to care about that register, so rather than defining the constants there, I defined them in `nv2a_regs.h`.

mborgerson:
Thanks!

Issue #577, opened 2021-12-01
some games shadows are broken or missing


Triticum0:
### Missing
- [x] https://xemu.app/titles/564e0006/#Baldur-s-Gate-Dark-Alliance (disappears at certain angles)
- [x] https://xemu.app/titles/4950000a/#Baldur-s-Gate-Dark-Alliance-II (disappears at certain angles/Untested)
- [x] https://xemu.app/titles/4b420005/#Chicago-Enforcer (missing)
- [x] Conan (missing)
- [x] Constantine (missing)
- [x] https://xemu.app/titles/534300f9/#Commandos-Strike-Force  (missing sometimes)
- [ ] https://xemu.app/titles/54510026/#Destroy-All-Humans (Missing)
- [ ] https://xemu.app/titles/54510106/#Destroy-All-Humans-2 (Missing)
- [ ] https://xemu.app/titles/45410075/#Madden-NFL-06 (Referee)
- [ ] https://xemu.app/titles/4541009f#Madden-NFL-07 (Referee)
- [ ] https://xemu.app/titles/454100ac#Madden-NFL-08 (Referee)
- [ ] https://xemu.app/titles/4541023b#Madden-NFL-09 (Referee)
- [x] https://xemu.app/titles/41560014/#Minority-Report-Everybody-Runs
- [x] Project-Gotham-Racing (missing/clipping)
- [x] Reign-of-Fire (missing)
- [x] https://xemu.app/titles/454100a4/#Tiger-Woods-PGA-Tour-07 (missing)
- [x] https://xemu.app/titles/4541007e/#Tiger-Woods-PGA-Tour-06 (missing)
- [x] https://xemu.app/titles/45410054/#Tiger-Woods-PGA-Tour-2005 (missing)
- [x] https://xemu.app/titles/5454009b/#The-Warriors (Shadows missing/lighting broken)
- [ ] (missing/broken lighting)
- [x] https://xemu.app/titles/4b4e0034/#Wallace-Gromit-The-Curse-of-the-Were-Rabbit (Missing)

### Rounding offset
- [ ] [Advent Rising](https://xemu.app/titles/4d4a0009/#Advent-Rising)
- [ ] [Aeon Flux](https://xemu.app/titles/4d4a0014#Aeon-Flux)
- [ ] [Beat Down](https://xemu.app/titles/43430015#Beat-Down-Fists-of-Vengeance)
[...]

Issue #576, opened 2021-11-30
Rendered text glyphs apparently offset in various titles When Upscaling


Triticum0:
### Title

https://xemu.app/titles/56550005/#Buffy-the-Vampire-Slayer-Chaos-Bleeds
https://xemu.app/titles/4d570005/#Area-51
https://xemu.app/titles/4c410004/#Star-Wars-The-Clone-Wars
https://xemu.app/titles/4343000b/#Lakemasters-Bass-Fishing-Game
https://xemu.app/titles/4b4f0003/#Dynasty-Warriors-3
https://xemu.app/titles/4b4f0004/#Dynasty-Warriors-4
https://xemu.app/titles/4d570012/#FireBlade
https://xemu.app/titles/4d53000f/#RalliSport-Challenge

### Bug Description

Some games have letters slightly cutoff as the letter seems to be one pixel to 2 pixels offset causing some artefacts in the typeface in menus. Depending on the typeface present In-game also affects the in-game font.
Area 51 xemu:
![Screenshot from 2021-11-30 16-15-19](https://user-images.githubusercontent.com/71127352/144098855-dbc7b12a-d715-4102-b45c-faeb46deabde.png)

xbox:
![Screenshot from 2021-11-30 16-31-45](https://user-images.githubusercontent.com/71127352/144098822-8af3c2d4-db38-4658-aab0-584f26eda5cc.png)
Buffy the vampire slayer: chaos bleeds xemu(Look at the letter T, t, r  and g): 
![Screenshot from 2021-11-30 17-15-53](https://user-images.githubusercontent.com/71127352/144099040-32e9214b-b0e3-4939-9a94-3ca5de3aeebc.png)

![Screenshot from 2021-11-30 17-16-07](https://user-images.githubusercontent.com/71127352/144099068-2ac2fa92-8f8c-49fb-995d-86e64052a7e8.png)

![Screenshot from 2021-11-30 17-16-17](https://user-images.githubusercontent.com/71127352/144099595-45d8c5c3-0f95-4518-bab4-2548b77861d5.png)

![Screenshot from 2021-11-30 17-16-22](https://user-images.githubusercontent.com/71127352/144100755-2ea51dd6-8feb-4065-8ef7-e8450ebb1f2d.png)

xbox:
![Screenshot from 2021-11-30 17-45-24](https://user-images.githubusercontent.com/71127352/144100261-b15025ab-f752-48e6-bf67-bba98f8d5815.png)
[...]

Issue #544, opened 2021-11-10
Pirates of the Caribbean menus Artifacting


Triticum0:
### Title

https://xemu.app/titles/42530004/#Pirates-of-the-Caribbean

### Bug Description

Menu are broken compared to original Xbox
![Screenshot from 2021-11-10 16-42-55](https://user-images.githubusercontent.com/71127352/141195703-fc7921dd-11ea-41c6-8bf2-160a1699c28d.png)
![Screenshot from 2021-11-10 16-44-46](https://user-images.githubusercontent.com/71127352/141195745-7d17c11b-00bd-41b1-9596-f279c73213e5.png)
![Screenshot from 2021-11-10 16-44-40](https://user-images.githubusercontent.com/71127352/141195793-166cf54b-9aa8-4f5b-98b2-eddda24bf613.png)

### Expected Behavior

Menu should look like this
![Screenshot from 2021-11-10 21-17-33](https://user-images.githubusercontent.com/71127352/141195016-ef339ab0-4d42-4f56-942b-ed66c130a968.png)
![Screenshot from 2021-11-10 21-26-13](https://user-images.githubusercontent.com/71127352/141197907-2b5c3ed7-c79c-41d4-a231-5369919ebd9f.png)

### xemu Version

Version: 0.6.2
Branch: master
Commit: bbcd576
Date: Sat Oct 30 23:09:11 UTC 2021
Last-known Working: Unknown

### System Information

Field | Value
-- | --
OS | Ubuntu 20.04.3 LTS
CPU | Intel(R) Pentium(R) CPU 3825U @ 1.90GHz
Graphics Device | Mesa Intel(R) HD Graphics (BDW GT1)
Graphics Driver | Drivers: 4.6 (Core Profile) Mesa 21.0.3

### Additional Context

Only video of xbox gameplay on youtube:https://www.youtube.com/watch?v=JmF-oh2C2SE

Triticum0:
Fixed on lasted build render correctly at all resolutions
![Image](https://github.com/user-attachments/assets/23b77342-b840-4046-8c9e-f6c0481deb00)

PR #735, opened 2022-02-15
nv2a: Fix pixel center offset problem

### LINKED ISSUE #57: Visual artifacts along screen margins
### Title

https://xemu.app/titles/45530005#Deus-Ex-Invisible-War
https://xemu.app/titles/4d53000d/#Fable
https://xemu.app/titles/464c0002/#Crusty-Demons
https://xemu.app/titles/4541005c/#Catwoman
https://xemu.app/titles/534300f9/#Commandos-Strike-Force
https://xemu.app/titles/56550016/#Bruce-Lee-Quest-of-the-Dragon
https://xemu.app/titles/4d530004/#Halo-Combat-Evolved
https://xemu.app/titles/55530038/#Tom-Clancy-s-Splinter-Cell-Chaos-Theory
https://xemu.app/titles/4d530036/#Counter-Strike
https://xemu.app/titles/55530013/#Tom-Clancy-s-Rainbow-Six-3

Some titles display incorrectly rendered lines at margins of the screen. Indicates possible off-by-one/0.5 issue. Issue is exacerbated by screen effects like zoom and bloom.

image

image

### LINKED ISSUE #725: The Elder Scrolls III: Morrowind: GOTY Edition: water is bugged, shows up in a grid pattern
### Title

https://xemu.app/titles/42530005/#The-Elder-Scrolls-III-Morrowind-Game-of-the-Year-Edition

### Bug Description

Any water in the game (lakes, ponds, etc.) is not rendered properly. The water is visible but broken up into a grid pattern. I assume the water was programmed this way but the grid itself wasn't supposed to be visible. Halo CE's water was programmed as a repeating grid, for example.

Steps to reproduce:
Any source of water has this issue but here is how to get to water the quickest
1. Start a new game
2. Skip the cutscene with the `Start` button
3. Name your character
4. Follow the guard to leave the ship
5. You are now surrounded by water

morrowind-bugged-water

### Expected Behavior

I don't have my real Xbox to get a screenshot so I used my PC version. The water should look the same except without the grid. The ~~shimmering effect~~ (shimmering works, sorry) and shadows are also missing on xemu but I do not remember if the effect was also missing on real Xbox hardware. It could have been cut for performance reasons. I would need someone else to test this or find real Xbox footage. The video below is from my PC version.

EDIT: This YouTube video (not mine) shows the same water for a few seconds on a real Xbox https://youtu.be/nhP49PlN6HY?t=102 more water is visible for longer at 19:24

https://user-images.githubusercontent.com/29314595/152713903-e79970e4-3a0c-4cd7-804b-175cd51c9eb4.mp4

### xemu Version

0.6.2-65-g517e4b3414

### System Information

OS: Windows 10
CPU: AMD Ryzen 5 3600X
GPU: NVIDIA GeForce GTX 1070
GPU Driver: 511.23

### Additional Context

_No response_

abaire:
This addresses the difference in pixel center between OpenGL and the DirectX 8
approach used by the hardware.

Note that the resultant behavior still does not fully match the hardware, where
screen-space pixel coordinates below (x + 0.5625) => x, but it is close enough
to fix issues with surface-target rendering.
[Test](https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/tests/vertex_shader_rounding_tests.cpp)
[HW results](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-Vertex_shader_rounding_tests)
With this patch, xemu will produce correct output for the render-target but will not fully match the geometry-based test output.

Fixes #725 
Fixes #57

### DIFF
--- a/hw/xbox/nv2a/pgraph.c
+++ b/hw/xbox/nv2a/pgraph.c
@@ -4269,6 +4269,13 @@ static void pgraph_shader_update_constants(PGRAPHState *pg,
         glUniform2f(binding->clip_range_loc, zclip_min, zclip_max);
     }
 
+    if (binding->gl_viewport_size_loc != -1) {
+        unsigned int vp_width = pg->surface_binding_dim.width,
+                vp_height = pg->surface_binding_dim.height;
+        pgraph_apply_scaling_factor(pg, &vp_width, &vp_height);
+        glUniform2f(binding->gl_viewport_size_loc, vp_width, vp_height);
+    }
+
     /* Clipping regions */
     unsigned int max_gl_width = pg->surface_binding_dim.width;
     unsigned int max_gl_height = pg->surface_binding_dim.height;
--- a/hw/xbox/nv2a/pgraph.c
+++ b/hw/xbox/nv2a/pgraph.c
@@ -4269,13 +4269,6 @@ static void pgraph_shader_update_constants(PGRAPHState *pg,
         glUniform2f(binding->clip_range_loc, zclip_min, zclip_max);
     }
 
-    if (binding->gl_viewport_size_loc != -1) {
-        unsigned int vp_width = pg->surface_binding_dim.width,
-                vp_height = pg->surface_binding_dim.height;
-        pgraph_apply_scaling_factor(pg, &vp_width, &vp_height);
-        glUniform2f(binding->gl_viewport_size_loc, vp_width, vp_height);
-    }
-
     /* Clipping regions */
     unsigned int max_gl_width = pg->surface_binding_dim.width;
     unsigned int max_gl_height = pg->surface_binding_dim.height;
--- a/hw/xbox/nv2a/shaders.c
+++ b/hw/xbox/nv2a/shaders.c
@@ -724,9 +724,13 @@ GLSL_DEFINE(materialEmissionColor, GLSL_LTCTXA(NV_IGRAPH_XF_LTCTXA_CM_COL) ".xyz
         mstring_append(body, "  tPosition = position;\n");
     }
 
-    mstring_append(body,
-    "   oPos = invViewport * (tPosition * compositeMat);\n"
-    "   oPos.z = oPos.z * 2.0 - oPos.w;\n");
+    mstring_append(
+        body,
+        "  oPos = tPosition * compositeMat;\n"
+        "  oPos.xy = fix_pixel_rounding(oPos.xy, oPos.w);\n"
+        "  oPos = invViewport * oPos;\n"
+        "  oPos.z = oPos.z * 2.0 - oPos.w;\n"
+    );
 
     /* FIXME: Testing */
     if (state->point_params_enable) {
@@ -753,6 +757,58 @@ GLSL_DEFINE(materialEmissionColor, GLSL_LTCTXA(NV_IGRAPH_XF_LTCTXA_CM_COL) ".xyz
                    "  vtx_inv_w_flat = vtx_inv_w;\n");
 }
 
+/*
+ * Generates a GLSL function that emulates hardware rounding at 9/16 pixel.
+ *
+ * FIXME: This approach only works at low render scale.
+ * At higher scales it needs an increasingly large bias as small amounts
+ * become full intermediate pixels. This seems like it should be by design
+ * (render scale increases the range of intermediate values that should map
+ * to raster pixels), so it's possible that this whole approach is flawed
+ * and should focus on fixing the first row/column offset by shifting
+ * everything left,up in NDC instead of messing with subpixels in nv2a
+ * screen space like this.
+ */
+static void append_fix_pixel_rounding(MString *header)
+{
+    // A bias is used to throw values on either side of 0.5 in order to
+    // avoid numerical precision issues (e.g., with a surfaceSize of
+    // 640,480, a 0.5625 value will map to (0.5000..71, 0.49999..82) when
+    // going to NDC and back to screen coords instead of the correct (0.5,
+    // 0.5)). at 1x, 0.002 is enough, 2x and beyond need substantially
+    // higher biases
+    static const float bias = 0.05f;
+    static const float scale = 0.5f - bias;
+    static const float offset = 0.5f + bias;
+
+    mstring_append_fmt(
+        header,
+        "\n"
+        "vec2 fix_pixel_rounding(vec2 screen_coords, float w) {\n"
+        "  if (w == 0.0 || isinf(w)) {\n"
+        "    w = 1.0;\n"
+        "  }\n"
+        "  screen_coords /= w;\n"
+        "  vec2 floored_pos = floor(screen_coords);\n"
+        "  vec2 subpixel = screen_coords - floored_pos;\n"
+        // Calculate what percentage of a full pixel should be added by shifting
+        // the 9/16 rounding point to 0.5.
+        //
+        // (0, 0.5625) => (0,0.5)
+        "  vec2 lt = subpixel / 0.5625 * %f;\n"
+        // (.5625,1) => (0.5,1)
+        "  vec2 gte = %f + (subpixel - 0.5625) / 0.4375 * %f;\n"
+        // Select between the two mappings based on whether the subpixel is
+        // above or below the rounding point.
+        "  vec2 round_down = vec2(lessThan(subpixel, vec2(0.5625)));\n"
+        "  subpixel = mix(gte, lt, round_down);\n"
+
+        "  return (floored_pos + subpixel) * w;\n"
+        "}\n",
+        scale, offset, scale
+    );
+}
+
 static MString *generate_vertex_shader(const ShaderState *state,
                                        bool prefix_outputs)
 {
@@ -762,7 +818,6 @@ static MString *generate_vertex_shader(const ShaderState *state,
 "\n"
 "uniform vec2 clipRange;\n"
 "uniform vec2 surfaceSize;\n"
-"uniform vec2 glViewportSize;\n"
 "\n"
 /* All constants in 1 array declaration */
 "uniform vec4 c[" stringify(NV2A_VERTEXSHADER_CONSTANTS) "];\n"
@@ -828,6 +883,8 @@ GLSL_DEFINE(texMat3, GLSL_C_MAT4(NV_IGRAPH_XF_XFCTX_T3MAT))
                                i, i);
         }
     }
+
+    append_fix_pixel_rounding(header);
     mstring_append(header, "\n");
 
     MString *body = mstring_from_str("void main() {\n");
@@ -1119,7 +1176,6 @@ ShaderBinding *generate_shaders(const ShaderState *state)
     }
     ret->surface_size_loc = glGetUniformLocation(program, "surfaceSize");
     ret->clip_range_loc = glGetUniformLocation(program, "clipRange");
-    ret->gl_viewport_size_loc = glGetUniformLocation(program, "glViewportSize");
     ret->fog_color_loc = glGetUniformLocation(program, "fogColor");
     ret->fog_param_loc[0] = glGetUniformLocation(program, "fogParam[0]");
     ret->fog_param_loc[1] = glGetUniformLocation(program, "fogParam[1]");
--- a/hw/xbox/nv2a/shaders.c
+++ b/hw/xbox/nv2a/shaders.c
@@ -762,6 +762,7 @@ static MString *generate_vertex_shader(const ShaderState *state,
 "\n"
 "uniform vec2 clipRange;\n"
 "uniform vec2 surfaceSize;\n"
+"uniform vec2 glViewportSize;\n"
 "\n"
 /* All constants in 1 array declaration */
 "uniform vec4 c[" stringify(NV2A_VERTEXSHADER_CONSTANTS) "];\n"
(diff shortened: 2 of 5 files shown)


antangelo:
This appears to resolve shadow issues in 5451001a, but only at 1x scaling. On 2x or above, the behavior is the same as master.
bfbb fix_pixel_rounding demo

Debated between posting here or in the issue, but this seems to be related to the code change so I opted to put it here.

dracc:
This removes the edge bloom (top and left) present in Fable: The Lost Chapters. :+1:

Fabxx:
Progression: artifacts on borders have disappeared for splinter cell double agent and chaos theory with this PR on native scaling:
image
image

revix-0:
conker
@abaire  on conker there's that weird bar on the screen, posting it here since it might still be related to the offset

abaire:
> on conker there's that weird bar on the screen, posting it here since it might still be related to the offset

@revix-0 is that screenshot taken with or without this PR? Does this change introduce the bar on the right hand side?

revix-0:
with the PR

abaire:
> with the PR, as for the side, i have to check without the PR

Please do and let us know if it's caused by this change (and if you're using a scale other than 1x, please test at 1x too if you can and report back :))

revix-0:
master:
image

both 1x scaling

Fabxx:
lights

Splinter cell 1 has been fixed, no more greenish bar on the right side

abaire:
Still need to fix this at higher render scales, and I just found that it doesn't quite work correctly on M1 (at 0.5 there is already a red bar along the left side, though not along the top) so there may be some device/GL dependency to work through as well.

LeRutY:
Also fixes Counter Strike
Fix

abaire:
The current state of this PR also causes some corruption in certain games like Splinter Cell:Double Agent when combined with #879 (which unmasks the issue by allowing compounding errors).

In the pgraph test for #879, I see that there's a yellow column along the right hand side of the texture after it is drawn the second time. I suspect that this happens in games as well, leading textures that go through multiple phases of compositing to be smeared (this is exactly what @revix-0 has reported in SC:DA w/ this fix).

Here is a screenshot from xemu, 
Screenshot_20220505_070346

and this is the same screenshot zoomed in to look at the right column
Screenshot_20220505_071114

Removing this PR and running #879 alone results in correct output.

abaire:
It looks like there is differing behavior between cases where we are rendering to the framebuffer versus to a texture. 

When rendering direct to the framebuffer, it appears to be correct that we should adjust the coordinate by flooring the (screenspace) output of the xbox vertex shader + 0.5625 to match hardware rounding behavior.

When rendering to a texture, it seems like even without this PR (and with any offset values I've tried) there is texture smearing, which makes me think that it's related to the way xemu handles render targets (it looks like it's being filtered rather than faithfully copied).

abaire:
Here is an illustration of the problem. This screen is the result of compositing the left hand side of a texture on top of itself (leaving the right hand side as the initial value, a hand-defined checkerboard pattern that I've verified is correct by dumping the memory passed to glTexImage2D in `pgraph_upload_surface_data`).

Whenever xemu does compositing, it does a partial pixel shift, despite the s2t_rndr being set up with correct coordinates and even if I force GL_NEAREST filtering. I'm not yet sure where the failure is coming from, but this is probably the root cause of the texture corruption that can be seen in combination with #879.

Screenshot_20220506_211642

abaire:
A note on debugging this: Unfortunately renderdoc seems to fail to capture interim states for some of the textures. If I take the hand-initialized texture and just render it to the standard color surface I can look at the input texture in renderdoc and confirm that it is as expected (a crisp checkerboard).

If I add a self-modifying step (input texture renders to a surface pointing at the same memory address), renderdoc ends up showing the final modification as the input stage, so it becomes very confusing as to whether the input texture was corrupted somehow (doesn't seem likely given the above check) or if renderdoc ends up updating its view of the texture after xemu updates it post-render. This step also corrupts the input (in renderdoc) for the test draw that is not self modifying it.

E.g.,

1. Create texture manually
2. Draw texture to normal color buffer. If you stop here and run in renderdoc, the input texture will be exactly as expected
3. Draw texture to a surface at the same address as the texture (on HW this would literally render into itself). If you examine the input texture for this phase in renderdoc, it'll show the final (corrupted) output. If you examine the input texture for step 2 in renderdoc, it'll also show the corrupted output, indicating a problem with renderdoc's resolution (it seems to only correctly track the final state).

abaire:
With nsight it's a bit more clear:
1. When creating the initial texture from the surface, the result is correct (no smearing).
2. When doing the first draw (rendering into the surface target) the smearing is introduced.

So I think we can rule out the s2t_rndr, but I'm still left wondering why we don't get consistently shifted results (e.g., if I don't render back into the surface at the same hwaddr as the texture, I get a pixel-aligned image).

abaire:
It looks like the way xemu handles the WINDOW_CLIP commands may be incorrect, which also contributes to this problem (and may explain uninitialized data along the outer edges of surfaces)

abaire:
`WINDOW_CLIP` differences do not explain the behavior.

I have a fix that gets a pixel perfect match in my test case, but still breaks SC:DA (it does a bunch of downscaling + jitter to do lighting effects and with this fix the resultant overlay is misaligned compared to the plain framebuffer).

ryzendew:
can you rebase this with master please so i can do more testing

abaire:
@RyzenDew - it should've already been automatically mergable, but rebased anyway

ryzendew:
image
fable is looking good but npc's weird outline is still a bit off just glad the white bits are gone

abaire:
Note: The current state of this PR breaks the highlight texture generation in NFS MW (#890) where the low res version of the color buffer appears to suffer from a pixel offset issue. Applying this PR causes the output to be completely black.

Triticum0:
Fixes Deus Ex Invisible War

Triticum0:
@abaire This break the tiger wood 2005 as well as avatar the last air bender

https://user-images.githubusercontent.com/71127352/171655676-757e8c37-12df-4a16-87e1-74535386030e.mp4

Triticum0:
Also Fixes C.A.T: Cyber Attack Team and Nobunaga no Yabou Ranseiki #598
xemu-2022-06-02-15-50-28
xemu-2022-06-02-16-08-52

Triticum0:
This cause major z-fighting issues on most of the games I tested

Triticum0:
Fixed Commandos Strike Force
Before
xemu-2022-06-02-18-10-18

After
xemu-2022-06-02-18-10-57

Triticum0:
Also fixes Peter Jackson's King Kong The Official Game of the Movie
### Before 
xemu-2022-06-02-19-05-58
### After
xemu-2022-06-02-18-50-40
 
And the Character Shadow
### Before 
xemu-2022-06-02-19-11-31

### After
xemu-2022-06-02-19-15-44

Triticum0:
Also fixes the Character Shadow in beat down
### Before
xemu-2022-06-02-21-18-53


### After 
xemu-2022-06-02-21-16-06

Triticum0:
Also Fixes Curse: The Eye of Isis Character Shadow
### Before
xemu-2022-06-02-21-22-35
### After
xemu-2022-06-02-21-21-54

Triticum0:
Also Fixes Aeon Flux Character Shadow and screen border, but causes terrible z-fighting
### Before
xemu-2022-06-03-20-50-10

### After
xemu-2022-06-03-20-50-53

Triticum0:
Fixes Crime Life: Gang Wars Character Shadow
### Before 
xemu-2022-06-03-22-11-01

### After
xemu-2022-06-03-22-13-13

Triticum0:
Fixes the text in AirForce Delta Storm
### Before
xemu-2022-06-03-21-43-45

### After
xemu-2022-06-03-21-43-45

Triticum0:
Fixes Strike Force Bowling Character Shadow
Before
xemu-2022-06-03-23-46-43
After
xemu-2022-06-03-23-47-32

Triticum0:
Fixes Without Warning Shadows
### Before
xemu-2022-06-04-14-36-56

### After
xemu-2022-06-04-14-43-18

Triticum0:
Fixes Harry Potter and the Goblet of Fire Shadows and border artifacting
### Before
xemu-2022-06-04-14-51-19

### After
xemu-2022-06-04-14-56-37

### Before
xemu-2022-06-04-14-53-50

### After
xemu-2022-06-04-14-52-31

Triticum0:
Fixes Pirates of the Caribbean Menu Issues #544
xemu-2022-06-05-01-03-50

WitchfinderMH:
Fixes shadows in Manhunt. 

Before: 

After:

WitchfinderMH:
And this. 

Before: 

After:

WitchfinderMH:
> This appears to resolve shadow issues in 5451001a, but only at 1x scaling. On 2x or above, the behavior is the same as master. 
> 
> Debated between posting here or in the issue, but this seems to be related to the code change so I opted to put it here.

It all becoming broken again when upsacling is being used, yeah. Same with Manhunt but it's not emulation issue i guess.

Triticum0:
FYI: This might be useful for the SpongeBob issue https://dolphin-emu.org/blog/2021/09/07/dolphin-progress-report-august-2021/

abaire:
> FYI: This might be useful for the SpongeBob issue https://dolphin-emu.org/blog/2021/09/07/dolphin-progress-report-august-2021/

Interesting find; vertex rounding is exactly what this PR does, and is what causes the z-fighting and shimmer seen in some games with this PR while fixing others.

abaire:
Confirmed that the latest version of the fix passes all of the pgraph tests (there is very slight ghosting in the multi-compositor tests at high render-scales, but it is similar to the differences I see in the raw texture rendering tests so I think it's acceptable).

@antangelo FYI - I also confirmed the shadows remain fixed for Spongebob since I happened to have a save state right at your screenshot :)

Triticum0:
The z-fighting isn't fixed. also there a wobble affect when upscaling

WitchfinderMH:
> The z-fighting isn't fixed. also there a wobble affect when upscaling

Wobble effect has nothing to do with upscaling, at least on my side and games on which i did testing.

dracc:
xemu-2022-06-10-20-38-55
Fable TLC is (still) looking good with this (7351aa1) change. <3

Triticum0:
@5d-0 Your right, my bad

abaire:
Surprisingly it looks like Tiger Woods 2005 uses a mix of fixed function and programmable pipelines any this fix is currently only applied in the programmable path, which may explain why there are gaps in the terrain (looking specifically at one of the seams, it is between two pieces of geometry using different paths). I'll look at Aeon Flux to see if the z-fighting is a similar situation.

UPDATE: Actually the version of the PR I had up during that testing had an issue where it created a big discontinuity between the values < 0.5625 and >= (each side would be compressed very close to its rounding target). I've since taken a slightly different approach that scales each half space such that there is no 1st order discontinuity, which may help things.

abaire:
Interestingly the problems that usually manifest when I bump up the render scale with the current version of this PR don't happen when testing on Monterey+M1, only on my Linux+GTX1070 machine.

Triticum0:
@abaire Placed all the games with this issue into #577. Note this game doesn't seem to fix all the shadow offset issues case in point advent rising, cold fear and Bloodrayne 2

Triticum0:
This seem to cause a regression on Hummer Badlands xemu-2022-07-26-12-57-58

Triticum0:
You Fixed the Hello Kitty: Roller Rescue Texture Sampling Misalignment issue
xemu-2022-08-05-21-09-37

The only issue left to be fixed is the slight misalignment on the save screen bar then the game will be perfect. Well done

Triticum0:
Just to make this a bit easier to parse through, I've made a list for simplicity to combat the comment bloat
Games Fixed 
AirForce Delta Storm #576
Aeon Flux (Shadows)
Alter Echo #979
Area 51 #576
Black Stone: Magic & Steel #979
Beat Down: Fists of Vengeance (Shadows)
Buffy the Vampire Slayer Chaos Bleeds #576
C.A.T: Cyber Attack Team #979
Capcom Fighting Jam #979
Cold war
Counter Strike
Commandos Strike Force https://github.com/mborgerson/xemu/issues/57
Corvette (Border) https://github.com/mborgerson/xemu/issues/57
Curse: The Eye (Shadows)
Crime Life: Gang (Shadows)
Dead or Alive 3 https://github.com/mborgerson/xemu/issues/57
Dynasty Warriors 3 #576 
Dynasty Warriors 4 #576
F1 Career Challenge (Shadows)
Fable https://github.com/mborgerson/xemu/issues/57
FireBlade #576
Harry Potter and the Goblet of Fire (Shadows)
Hello Kitty: Roller Rescue
Jaws: Unleashed
LMA 2003 #979
Peter Jackson's King Kong (Shadows)
ProStroke Golf: World Tour 2007
Manhunt 
Nickelodeon SpongeBob SquarePants in: Battle for Bikini Bottom
Nobunaga no Yabou Ranseik #979
Secret Weapons over Normandy
Shenmue 2 #979
Star Wars The Clone Wars #576
Strike Force Bowling (Shadows)
Without Warning (Shadows)


Partial fixes
The King of Fighters: Neowave #576 (Still slightly visible on [Kyo Kusanagi](https://www.giantbomb.com/kyo-kusanagi/3005-197/) legs) 
Tron 2.0: Killer App #979 (Still slightly visible)
Dynasty Warriors 3 #979 (Still slightly visible)
Shinchou Mahjong #979 (Still slightly visible)

LeRutY:
@Triticum0 Where´s Counter Strike?

Triticum0:
I never tested the game, only add games I tested that were fixed.

AhayriSG:
> Where´s Counter Strike?

a9b70a9
[JPCSP-dev - PSP XrossMediaBar (XMB) + UMD movie decoding test](https://www.youtube.com/watch?v=bGVnn_GzuzY "PR 735 CounterStrike test")

HunterAhlquist:
Has this potential fix been abandoned?

abaire:
Closing in favor of #2183

Issue #387, opened 2021-06-30
DreamWorks Madagascar: Crashes after Intro screen.


RageXbox:
### Title Information

* Titles: DreamWorks Madagascar AV-059 [4156003b]

* https://xemu.app/titles/4156003b/#DreamWorks-Madagascar

### Bug Description

After the Intro screen, it crashes with these:
Assertion failed!
D:/a/xemu/hw/xbox/nv2a/pgraph.c
Line 717: Expression: glGetError() = = GL_NO_ERROR
Line 4584: Expression: surface != d->pgraph.color_binding
Line 4831: Expression: glCheckFramebufferStatus(GL_FRAMEBUFFER = = GL_FRAMEBUFFER_COMPLETE
https://github.com/mborgerson/xemu/blob/7cde5a222cd77864b3bea5424970f6e86c85c28d/hw/xbox/nv2a/pgraph.c#L717
https://github.com/mborgerson/xemu/blob/7cde5a222cd77864b3bea5424970f6e86c85c28d/hw/xbox/nv2a/pgraph.c#L4584
https://github.com/mborgerson/xemu/blob/7cde5a222cd77864b3bea5424970f6e86c85c28d/hw/xbox/nv2a/pgraph.c#L4831
Steps to reproduce this issue:
1. Insert disc
2. Game crashes at Intro screen

### Expected Behavior

Goes properly to the menus without issues.

### Version

* Version: v0.5.4-9-g7cde5a222c
* Last-known Working: Unknown

### System Information

| Field | Value |
|-------|-------|
| OS			| (Windows 10 (64-bit) 
| CPU			| (Intel(R) Xeon(R) CPU E5-2678 v3 @2.50 GHz) 
| Graphics Device	| (NVIDIA Quadro P5000) 
| Graphics Driver	| (NVIDIA 457.30)

### Additional Context

I tested this game at February 20th, 2021 and Line changed from 4072 to 4584 for some reason. DSP is disabled. I kept pressing "Ignore" until it goes for another assertion.

Triticum0:
This issue is link to #288, #116 and #373

RageXbox:
Fixed

Triticum0:
@RageXbox Please Add new info to the compatibly list

Issue #288, opened 2021-05-21
Halo 2 v1.0: Assertion failed! 


theboy181:
https://xemu.app/titles/4d530064/#Halo-2

### Bug Description
Assertion failed! 

Error will occur when you zoom in  looking through glass railing in the first level at the spot shown in the screen capture.
![image](https://user-images.githubusercontent.com/9299743/119065605-c831c200-b992-11eb-9dfb-0341143ecc50.png)

Video
https://user-images.githubusercontent.com/9299743/119066318-607c7680-b994-11eb-86ec-5155c20dc7bb.mp4

Version: 0.5.2-3-g68b8043af9
Last-known Working: Unknown

System Specs
https://valid.x86.fr/r74im1

Triticum0:
This also affect half life 2, Full spectrum Warrior: 10 Hammers and Midnight Club 3: DUB Edition https://github.com/mborgerson/xemu/issues/349.
https://xemu.app/titles/45410091/#Half-Life-2
https://xemu.app/titles/545100f8/#Full-Spectrum-Warrior-Ten-Hammers (I Tested it independently so it is not on compact report)
https://xemu.app/titles/54540079/#Midnight-Club-3-DUB-Edition
https://xemu.app/titles/55530018/#Beyond-Good-Evil
https://xemu.app/titles/5451001a/#Nickelodeon-SpongeBob-SquarePants-in-Battle-for-Bikini-Bottom

tmob03:
Still happens in Half life 2 as of 1/10/21

alejandrodf95:
When i start xemu crash after one minute with this message.

Triticum0:
@w4rexx Can you retest

Triticum0:
@theboy181 Can you retest

Triticum0:
@theboy181 Can you retest the issue been fixed

theboy181:
When was it fixed. It's been a while since I tested this game.

Triticum0:
@theboy181 It was merged 17 days ago #989 and the assertion probably changed

theboy181:
Not tested myself, but have been told the issue has been resolved.

antangelo:
> It was merged 17 days ago #989 and the assertion probably changed

Are you sure this is the same issue? This assert message is very generic as compared to the one in #635 which is what that PR closed.

Goonslaper:
[...]

PR #738, opened 2022-02-17
nv2a: Ignore color/depth mask in CLEAR_SURFACE

### LINKED ISSUE #730: Half-Life 2: Assertion failed!
### Title

https://xemu.app/titles/45410091/#Half-Life-2

EA-145

### Bug Description

Game triggers assert on scenes when certain in game events are active.

1. Launch Game
2. Start New game
3. Play Point Insertion until you are at where the Combine asks you to pick up the can.

xemu_2022-02-08_22-30-13

### Expected Behavior

Ability to progress the game as it plays on the actual system.

### xemu Version

0.6.2-64-g758e36d398

### System Information

OS: Windows 11
CPU: Intel(R) Core(TM) i7-6700 CPU @ 3.40GHz
GPU: Nvidia GeForce GTX 1050

### Additional Context

Same assertion continues on into the game  (tracked it as far as Red Letter Day) and ony seems to happen when certain NPC's are active. Seems to not trigger when just kleiner and alyx are in the lab but as soon as barney walks into room it immediately trigers.

abaire:
Fixes #730 and matches observed HW behavior where NV097_SET_COLOR_MASK/NV097_SET_DEPTH_MASK masking is not respected by `CLEAR_SURFACE` call (where masking is set via a parameter).

[Test](https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/tests/clear_tests.cpp)
[Golden results](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-Clear)

### DIFF
--- a/hw/xbox/nv2a/pgraph.c
+++ b/hw/xbox/nv2a/pgraph.c
@@ -5918,9 +5918,9 @@ static void pgraph_update_surface(NV2AState *d, bool upload,
     pg->surface_shape.z_format = GET_MASK(pg->regs[NV_PGRAPH_SETUPRASTER],
                                           NV_PGRAPH_SETUPRASTER_Z_FORMAT);
 
-    /* FIXME: Does this apply to CLEARs too? */
-    color_write = color_write && pgraph_color_write_enabled(pg);
-    zeta_write = zeta_write && pgraph_zeta_write_enabled(pg);
+    color_write = color_write &&
+            (pg->clearing || pgraph_color_write_enabled(pg));
+    zeta_write = zeta_write && (pg->clearing || pgraph_zeta_write_enabled(pg));
 
     if (upload) {
         bool fb_dirty = pgraph_framebuffer_dirty(pg);


hw/xbox/nv2a/pgraph.c
```diff
@@ -5911,16 +5911,15 @@ static void pgraph_unbind_surface(NV2AState *d, bool color)
...
 {
     PGRAPHState *pg = &d->pgraph;
 
     pg->surface_shape.z_format = GET_MASK(pg->regs[NV_PGRAPH_SETUPRASTER],
                                           NV_PGRAPH_SETUPRASTER_Z_FORMAT);
 
-    /* FIXME: Does this apply to CLEARs too? */
-    color_write = color_write && pgraph_color_write_enabled(pg);
-    zeta_write = zeta_write && pgraph_zeta_write_enabled(pg);
+    color_write = color_write && (is_clear || pgraph_color_write_enabled(pg));
```
antangelo: Does this also work if you use `pg->clearing` instead of `is_clear`?

abaire: Ha, I wish I'd noticed that before piping through a bunch of bools :) Thanks for the tip, update to follow shortly.

Triticum0:
I think this bug also fixes Full spectrum Warrior: 10 Hammers, Midnight Club 3: DUB Edition,Beyond Good & Evil, #140?(issue not documented well there over 10 different assertion with the same name), #387 and #288. All have the same assertion to my knowledge but don't know if the bugs are related. I not going to test so don't ask but games are probably fixed with this pr.

abaire:
Thanks for the pointers!

I think we'll need to farm this out to the discord to get people to re-test those games post-merge to verify. Unfortunately the `glGetError` assert doesn't specify what the actual error is, so some of those failures could be due to totally different root causes. Similarly, it's possible to get into a framebuffer != complete status a few different ways so it's possible that even the #387 failures are due to different problems.

antangelo:
> I think this bug also fixes Full spectrum Warrior: 10 Hammers, Midnight Club 3: DUB Edition,Beyond Good & Evil, #140?(issue not documented well there over 10 different assertion with the same name), #387 and #288. All have the same assertion to my knowledge but don't know if the bugs are related. I not going to test so don't ask but games are probably fixed with this pr.

#140 is not related to this. That issue stems from `SET_BEGIN_END` and isn't related to clearing.

mborgerson:
> I think we'll need to farm this out to the discord to get people to re-test those games post-merge to verify.

Preferably things should be tested __before__ being merged. Feel free to ask whomever filed the last compat report to help test a PR

abaire:
> > I think we'll need to farm this out to the discord to get people to re-test those games post-merge to verify.
> 
> Preferably things should be tested **before** being merged. Feel free to ask whomever filed the last compat report to help test a PR

Agreed in general, but in this particular case we have a fix for one particular cause of a generic error message that could be triggered many different ways. 

Re-testing every game that has been found to have a generic gl failure (which presumably could even be driver-related) seems like it's going to burn the good will of testers and I'd argue that it's better to get known issues fixed than delay in the hope that some other issues might've been fixed unknowingly.

Alternatively/in-addition, perhaps we can expand on the error messaging in gl error cases so this is a non-issue in the future? Perhaps we could have the debug build leverage `glDebugMessageCallback` and log the underlying error so we don't have to rely on a developer to try to repro to make the issue de-dupable (this is what I happen to do on my work branch)

mborgerson:
> 



> > > I think we'll need to farm this out to the discord to get people to re-test those games post-merge to verify.
> > 
> > 
> > Preferably things should be tested **before** being merged. Feel free to ask whomever filed the last compat report to help test a PR
> 
> Agreed in general, but in this particular case we have a fix for one particular cause of a generic error message that could be triggered many different ways.
> 
> Re-testing every game that has been found to have a generic gl failure (which presumably could even be driver-related) seems like it's going to burn the good will of testers and I'd argue that it's better to get known issues fixed than delay in the hope that some other issues might've been fixed unknowingly.
> 
> Alternatively/in-addition, perhaps we can expand on the error messaging in gl error cases so this is a non-issue in the future? Perhaps we could have the debug build leverage `glDebugMessageCallback` and log the underlying error so we don't have to rely on a developer to try to repro to make the issue de-dupable (this is what I happen to do on my work branch)

If the problem is too generic, we should do the work to isolate what the actual failures are and begin cataloging and fixing them individually. `glDebugMessageCallback` is an option although sometimes it is not quite as helpful as you'd expect/hope.

abaire:
> If the problem is too generic, we should do the work to isolate what the actual failures are and begin cataloging and fixing them individually. `glDebugMessageCallback` is an option although sometimes it is not quite as helpful as you'd expect/hope.

Totally agree. I'm happy to file an issue to replace/augment any `glGetError` assertions with something that either provides sufficient information to determine the underlying issue or at least makes it clear to non-developers that more info is required to be actionable. 

If the debug messaging is insufficient I don't personally have enough of a GL background to know what alternative would be viable (other than checking for errors after every GL call and logging the actual error value along with the assert, which presumably would have enough of a perf impact to be non-viable for the release build).

As an aside, I'd suggest a similar thing for uses of `assert(false)`; if we `assert(!"Some explanation of the issue")` the assertion message becomes better future-proofed and more immediately actionable. At the moment things like pgraph change frequently enough that the line numbers go out of date fast and it'd reduce the need to switch back to old release branches to understand GH issues.

Triticum0:
#745  also need to be retested with this commit

Triticum0:
[Full-Spectrum-Warrior-Ten-Hammers](https://xemu.app/titles/545100f8/#Full-Spectrum-Warrior-Ten-Hammers) is fixed with this pull

PR #742, opened 2022-02-21
nv2a/mcpx: fix freezing on loadvm failure

### LINKED ISSUE #741: nv2a/mcpx: Loading nonexistent save state freezes the guest
### Bug Description

Loading a save state using `loadvm` that does not exist will cause the guest to freeze. Unpausing and resetting will not return to a normal state, only closing and restarting the emulator will do so.

### Expected Behavior

The emulator should not be frozen after `loadvm` fails. Either the emulator should be paused, or should continue as if the command had not been executed.

### xemu Version

xemu_version: 0.6.2-75-g8919bae311
xemu_branch: master
xemu_commit: 8919bae31156a8534d529be0012670020fcd5704

### System Information

OS: Arch Linux (5.16.10-arch1-1)
CPU: Intel i5-10400
GL_VENDOR: NVIDIA Corporation
GL_RENDERER: NVIDIA GeForce RTX 3060 Ti/PCIe/SSE2
GL_VERSION: 4.0.0 NVIDIA 510.54

### Additional Context

This issue is caused by the fact that when the VM state is set to `RUN_STATE_RESTORE_VM`, the pfifo is halted. The pfifo is normally resumed when the load completes, but since `loadvm` has failed, this never happens.

If the above is resolved, then the MCPX APU is reset on this same state change, so the emulator will resume without audio.

antangelo:
Resolves #741 

Introduces a flag in the nv2a state that tracks whether we are in the process of loading a VM or not. This flag is set to true on state change to `RUN_STATE_RESTORE_VM`, and false once the load completes. If the state is changed to running with this flag enabled, then the pfifo is resumed, as the load must have failed.

In addition, move the MCPX APU reset to the pre-load hook. On failed loads, the APU will no longer be reset.

With the behavior of this PR, the emulator will be paused after a failed load. Changing this to allow the emulator to continue instead is trivial.

### DIFF
--- a/hw/xbox/nv2a/nv2a.c
+++ b/hw/xbox/nv2a/nv2a.c
@@ -375,13 +375,16 @@ static void nv2a_vm_state_change(void *opaque, bool running, RunState state)
         nv2a_lock_fifo(d);
         qatomic_set(&d->pfifo.halt, true);
         nv2a_unlock_fifo(d);
+    } else if (state == RUN_STATE_RUNNING) {
+        nv2a_lock_fifo(d);
+        qatomic_set(&d->pfifo.halt, false);
+        nv2a_unlock_fifo(d);
     }
 }
 
 static int nv2a_post_save(void *opaque)
 {
     NV2AState *d = opaque;
-    qatomic_set(&d->pfifo.halt, false);
     nv2a_unlock_fifo(d);
     return 0;
 }
@@ -396,7 +399,6 @@ static int nv2a_pre_load(void *opaque)
 static int nv2a_post_load(void *opaque, int version_id)
 {
     NV2AState *d = opaque;
-    qatomic_set(&d->pfifo.halt, false);
     qatomic_set(&d->pgraph.flush_pending, true);
     nv2a_unlock_fifo(d);
     return 0;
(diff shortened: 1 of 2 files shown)


hw/xbox/nv2a/nv2a.c
```diff
@@ -374,6 +374,12 @@ static void nv2a_vm_state_change(void *opaque, bool running, RunState state)
     } else if (state == RUN_STATE_RESTORE_VM) {
         nv2a_lock_fifo(d);
         qatomic_set(&d->pfifo.halt, true);
+        qatomic_set(&d->loadingState, true);
+        nv2a_unlock_fifo(d);
+    } else if (state == RUN_STATE_RUNNING && qatomic_read(&d->loadingState)) {
+        nv2a_lock_fifo(d);
+        qatomic_set(&d->pfifo.halt, false);
+        qatomic_set(&d->loadingState, false);
```
mborgerson: I do not think it's necessary to track the `loadingState` variable and can just un-halt fifo if in the run state.

`post_save` and `post_load` will be called before entering run state so let's remove the `qatomic_set(&d->pfifo.halt, false);` lines from `nv2a_post_save` and `nv2a_post_load` and just do it here to make sure fifo is running whenever we finally enter the run state.

antangelo: Sounds good. I didn't know if there was a case where the fifo is supposed to be halted after loading a save state, so I introduced an additional variable for safety. Removed the variable since it's not needed.

mborgerson:
Thanks!

Issue #498, opened 2021-10-14
Serial broken on Windows


HoRnEyDvL:
### Bug Description

Xemu freezes (Not Responding) when WinDbg is open & listening to COM ports

Unable to debug any Kernels

Steps:
Install Virtual Com Ports I used (https://www.virtual-serial-port.org/) 14 day trial

Open VSPD  & create a new pair in my case COM1 & COM 2

Open WinDbg & Connect to COM Port 2

-  Microsoft (R) Windows Debugger Version 6.12.0002.633 X86
- Copyright (c) Microsoft Corporation. All rights reserved.
- Opened \\.\com2
- Waiting to reconnect...

Launch Xemu Using D:\Xemu\xemu.exe -device lpc47m157 -serial COM1

Pop up appears for COM port settings but Xemu locks up.

![image](https://user-images.githubusercontent.com/73101201/137322828-1ba29fac-bee9-40cd-841d-49f05cf2b047.png)

### Expected Behavior

Xemu Should output Kernel debug to WindDbg Via Com Port specified

![image](https://user-images.githubusercontent.com/73101201/137322524-4937f2cf-0f33-4a2a-b380-06a6cdc40802.png)

### xemu Version

Version: 0.6.1-22-g5622af4981
Branch:  master
Commit:  5622af49811cdde72e45b01ef80c5a76de3ee390
Date:    Wed Oct 13 23:54:07 UTC 2021

### System Information

Win 10
AMD 5950X
GTX 980

HoRnEyDvL:
Just did further testing & can confirm that this also causes crash when using Putty.

This eliminates  WinDbg been the issue.

GXTX:
This is a known issue. See: #86. I believe it's an issue with QEMU itself.

You should keep this issue open for documentation. And rename to "Serial broken on Windows".

HoRnEyDvL:
> This is a known issue. See: #86. I believe it's an issue with QEMU itself.
> 
> You should keep this issue open for documentation. And rename to "Serial broken on Windows".

Renamed.

Thanks

mborgerson:
Does it hang when you use a physical COM port? Have you tried other chardev backends, tcp/stdio for instance? Is it only when you're using this virtual COM port that xemu hangs? Does QEMU work with your virtual COM port?

Can you help investigate the issue

GXTX:
[...]

PR #751, opened 2022-02-28
Fix serial-based Xbox kernel debugging on Windows

gaasedelen:
This PR contains two commits.

* The first commit explicitly 'fixes' the QEMU lockup that occurs on Windows when attempting to connect to the serial port for kernel debugging with WinDbg. (Issue #498) 
* The second commit sets default parameters for the connection so that you don't have to click/close the COM config dialog every launch (it gets annoying, I promise).

Without going into excessive detail, the crux of the issue is that QEMU does not provide ample opportunity for the emulated system (in this case, the Xbox kernel) to process data received on the serial port.

When the 'received data' buffer (which is LITERALLY 1 byte) is 'full' and QEMU sees that there is more data available on the COM port (almost a certainty!) it will return early from the main QEMU / emulation loop -- meaning the Xbox kernel will never execute, thus the single received byte is never processed. This is our impasse.

While the fix is technically probably not the 'correct' long-term solution for QEMU at large (note their XXX comment, a proper set of async polling/buffering infrastructure should be built out instead), it works for our purposes.

I have successfully been using WinDbg and WinDbg Preview (requests much more data, so can be a lot slower) using these fixes for the past month or so.

xemu

### DIFF
--- a/chardev/char-win.c
+++ b/chardev/char-win.c
@@ -110,7 +110,8 @@ int win_chr_serial_init(Chardev *chr, const char *filename, Error **errp)
     size = sizeof(COMMCONFIG);
     GetDefaultCommConfig(filename, &comcfg, &size);
     comcfg.dcb.DCBlength = sizeof(DCB);
-    CommConfigDialog(filename, NULL, &comcfg);
+    comcfg.dcb.BaudRate = CBR_115200;
+    //CommConfigDialog(filename, NULL, &comcfg);
 
     if (!SetCommState(s->file, &comcfg.dcb)) {
         error_setg(errp, "Failed SetCommState");
--- a/chardev/char-win.c
+++ b/chardev/char-win.c
@@ -111,7 +111,6 @@ int win_chr_serial_init(Chardev *chr, const char *filename, Error **errp)
     GetDefaultCommConfig(filename, &comcfg, &size);
     comcfg.dcb.DCBlength = sizeof(DCB);
     comcfg.dcb.BaudRate = CBR_115200;
-    //CommConfigDialog(filename, NULL, &comcfg);
 
     if (!SetCommState(s->file, &comcfg.dcb)) {
         error_setg(errp, "Failed SetCommState");
--- a/util/main-loop.c
+++ b/util/main-loop.c
@@ -490,16 +490,6 @@ static int os_host_main_loop_wait(int64_t timeout)
 
     g_main_context_acquire(context);
 
-    /* XXX: need to suppress polling by better using win32 events */
-    ret = 0;
-    for (pe = first_polling_entry; pe != NULL; pe = pe->next) {
-        ret |= pe->func(pe->opaque);
-    }
-    if (ret != 0) {
-        g_main_context_release(context);
-        return ret;
-    }
-
     FD_ZERO(&rfds);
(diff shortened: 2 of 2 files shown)


HBelusca:
Quite based for using WinDbg 4 as test <3

GXTX:
Using a Digi EdgePort/4 going from port to port it's incredibly slow but still functional. Going from motherboard port to one of the EdgePort's COM's it's quicker but still quite slow with noticeable hiccups when transmitting or receiving.

My EdgePort has a couple of port modifiers for low latency and such, also attempted those.
image

~~Side note: It's interesting that the kernel thinks the FSB is 16MHz, CPU is 91MHz and MCLK is 16MHz. I wonder if that effects anything.~~ FSB and such is wrong because xemu is missing the device so it gets reported as `0xFFFFFFFF` so w/e calculations they're using is wrong.

gaasedelen:
Yes, debugging/communication is slow. And there is notable slowdown of XEMU when using serial debugging with this. But it does work :-P

I am almost certain it's due to the insane amount of overhead with QEMU trying to bring data in over the COM port. It will only peek & read 1 byte at a time which can involve several calls to underlying Windows APIs. Then, this byte may sit untouched for several cycles through the QEMU main loop until the scheduler/thread in the emulated system finally runs through the Kernel Debug stub to process the read byte. 

And even when the emulated system processes a single byte, it may require another byte... which means it'll have to wait till QEMU breaks from emulation and reads in another byte (again taking millions of instructions for polling / reading 1 byte, etc).

This is why they have an XXX comment stating they need async polling and buffering of data. Which would be the more performant and long term solution for QEMU.

chardev/char-win.c
```diff
@@ -110,7 +110,8 @@ int win_chr_serial_init(Chardev *chr, const char *filename, Error **errp)
     size = sizeof(COMMCONFIG);
     GetDefaultCommConfig(filename, &comcfg, &size);
     comcfg.dcb.DCBlength = sizeof(DCB);
-    CommConfigDialog(filename, NULL, &comcfg);
+    comcfg.dcb.BaudRate = CBR_115200;
+    //CommConfigDialog(filename, NULL, &comcfg);
```
abaire: Nit: why leave the code commented out versus remove it entirely?

util/main-loop.c
```diff
@@ -562,6 +552,16 @@ static int os_host_main_loop_wait(int64_t timeout)
         g_main_context_dispatch(context);
     }
 
+    /* XXX: need to suppress polling by better using win32 events */
+    ret = 0;
+    for (pe = first_polling_entry; pe != NULL; pe = pe->next) {
+        ret |= pe->func(pe->opaque);
+    }
+    if (ret != 0) {
+        g_main_context_release(context);
```
mborgerson: Redundant release with line below

chardev/char-win.c
gaasedelen: It's a good question. I think I left it in for reference, or maybe to provide some context when pulling / de-conflicting edits against QEMU upstream.

I went ahead and removed it.

util/main-loop.c
gaasedelen: Fixed per your request

mborgerson:
Thanks

PR #753, opened 2022-03-02
xemu: Add wireframe mode shortcuts

GXTX:
Adds support to use CTRL+F5 to enable wireframe.

Should probably be edited to invalidate all of the surfaces. Also doesn't seem to work on all surfaces.


GXTX:
Doesn't work as expected in most games. Will only work if *at boot* it's enabled. Maybe it'll require moving into some other already existing function?

PR #757, opened 2022-03-05
xbox: Add a dummy device in place of memory controller

GXTX:
This is required for the kernel to correctly calculate clock speeds. This should be fixed in a later commit to add the correct device.

See https://github.com/mborgerson/xemu/issues/638 for missing devices.

Before:
before

After:
after

Assumption is that the kernel was trying to read the FSB register that's present at 0x6C on this device and improperly calculating clocks based on the fail response of `0xFFFFFFFF`. I don't know if this has any real implications, if the kernel is doing it's own time keeping based off of these numbers, etc.

### DIFF
--- a/hw/xbox/xbox.c
+++ b/hw/xbox/xbox.c
@@ -353,6 +353,9 @@ void xbox_init_common(MachineState *machine,
     /* GPU! */
     nv2a_init(agp_bus, PCI_DEVFN(0, 0), ram_memory);
 
+    /* FIXME: Stub the memory controller */
+    pci_create_simple(pci_bus, PCI_DEVFN(0, 3), "pci-testdev");
+
     if (pci_bus_out) {
         *pci_bus_out = pci_bus;
     }


hw/xbox/xbox.c
```diff
@@ -353,6 +353,9 @@ void xbox_init_common(MachineState *machine,
     /* GPU! */
     nv2a_init(agp_bus, PCI_DEVFN(0, 0), ram_memory);
 
+    /* FIXME: Kernel incorrectly calculates clock speeds if we don't have a device at 0:0:3 */
```
mborgerson: * What clock speed? What is the impact? Does having this dummy device resolve the issue?
* Which part should be fixed? Just stubbing out the device properly? Is the device just being present enough to resolve the issue or is more required?

GXTX: > What clock speed?

Front-side bus.

> What is the impact?

There may be hidden uses currently unknown.

> Does having this dummy device resolve the issue?

It resolves the issue of the kernel and xshell improperly calculating speeds. See images.

> Just stubbing out the device properly? Is the device just being present enough to resolve the issue or is more required?

Having *a* device present here is good enough. The only thing that is going on is the xcodes are writing to the register 0x6C on this device and then using that for what the FSB is should be rated at. Doing testing with my [XboxOverclock](https://github.com/GXTX/XboxOverclock) utility this is indeed what this particular register is doing.

```
xc_pci_write 0x8000036C, 0x00230801         ; FSB
xc_pci_write 0x8000036C, 0x01230801         ; FSB
```

When I say "fixme" I mean to emulate the correct name with a stubbed device, I don't know the correct use case is for the `pci-testdev` device but it's likely not to have just a random PCI device present. I haven't tested to see if any other part of the system is reading or writing to this device.

mborgerson: >  When I say "fixme" I mean to emulate the correct name with a stubbed device, I don't know the correct use case is for the pci-testdev device but it's likely not to have just a random PCI device present.

Can you change the FIXME to state this pls

PR #768, opened 2022-03-14
nv2a: Cache shaders to disk

antangelo:
Shaders now use an LRU cache eviction strategy. On shader compilation, the shader programs will be saved to disk. On shutdown, a list of shaders currently in the LRU cache is saved to disk as well. On startup, this list (along with all shaders in it) are then loaded, effectively restoring the shader cache to the state it was when the emulator was last closed.

I have tested this on Linux, Windows, and macOS. Unfortunately, my Intel MacBook Pro doesn't support saving and loading program binaries, so I have only been able to verify that case on macOS.

### DIFF
--- a/hw/xbox/nv2a/nv2a_int.h
+++ b/hw/xbox/nv2a/nv2a_int.h
@@ -305,12 +305,15 @@ typedef struct PGRAPHState {
 
     hwaddr dma_a, dma_b;
     Lru texture_cache;
-    struct TextureLruNode *texture_cache_entries;
+    TextureLruNode *texture_cache_entries;
     bool texture_dirty[NV2A_MAX_TEXTURES];
     TextureBinding *texture_binding[NV2A_MAX_TEXTURES];
 
-    GHashTable *shader_cache;
+    Lru shader_cache;
+    ShaderLruNode *shader_cache_entries;
     ShaderBinding *shader_binding;
+    QemuMutex shader_cache_lock;
+    QemuThread shader_disk_thread;
 
     bool texture_matrix_enable[NV2A_MAX_TEXTURES];
 
@@ -370,7 +373,7 @@ typedef struct PGRAPHState {
     uint16_t compressed_attrs;
 
     Lru element_cache;
-    struct VertexLruNode *element_cache_entries;
+    VertexLruNode *element_cache_entries;
 
     unsigned int inline_array_length;
     uint32_t inline_array[NV2A_MAX_BATCH_LENGTH];
@@ -403,10 +406,12 @@ typedef struct PGRAPHState {
     bool download_dirty_surfaces_pending;
     bool flush_pending;
     bool gl_sync_pending;
+    bool shader_cache_writeback_pending;
     QemuEvent downloads_complete;
     QemuEvent dirty_surfaces_download_complete;
     QemuEvent flush_complete;
     QemuEvent gl_sync_complete;
+    QemuEvent shader_cache_writeback_complete;
 
     unsigned int surface_scale_factor;
     uint8_t *scale_buf;
--- a/hw/xbox/nv2a/pgraph.c
+++ b/hw/xbox/nv2a/pgraph.c
@@ -451,8 +451,6 @@ static bool vertex_cache_entry_compare(Lru *lru, LruNode *node, void *key)
 
 static void pgraph_mark_textures_possibly_dirty(NV2AState *d, hwaddr addr, hwaddr size);
 static bool pgraph_check_texture_dirty(NV2AState *d, hwaddr addr, hwaddr size);
-static guint shader_hash(gconstpointer key);
-static gboolean shader_equal(gconstpointer a, gconstpointer b);
 static unsigned int kelvin_map_stencil_op(uint32_t parameter);
 static unsigned int kelvin_map_polygon_mode(uint32_t parameter);
 static unsigned int kelvin_map_texgen(uint32_t parameter, unsigned int channel);
@@ -3958,10 +3956,12 @@ void pgraph_init(NV2AState *d)
     pg->downloads_pending = false;
 
     qemu_mutex_init(&pg->lock);
+    qemu_mutex_init(&pg->shader_cache_lock);
     qemu_event_init(&pg->gl_sync_complete, false);
     qemu_event_init(&pg->downloads_complete, false);
     qemu_event_init(&pg->dirty_surfaces_download_complete, false);
     qemu_event_init(&pg->flush_complete, false);
+    qemu_event_init(&pg->shader_cache_writeback_complete, false);
 
     /* fire up opengl */
     glo_set_current(g_nv2a_context_render);
@@ -4018,7 +4018,7 @@ void pgraph_init(NV2AState *d)
     pg->element_cache.init_node = vertex_cache_entry_init;
     pg->element_cache.compare_nodes = vertex_cache_entry_compare;
 
-    pg->shader_cache = g_hash_table_new(shader_hash, shader_equal);
+    shader_cache_init(pg);
 
     pg->material_alpha = 0.0f;
     SET_MASK(pg->regs[NV_PGRAPH_CONTROL_3], NV_PGRAPH_CONTROL_3_SHADEMODE,
@@ -4053,6 +4053,7 @@ void pgraph_init(NV2AState *d)
 void pgraph_destroy(PGRAPHState *pg)
 {
     qemu_mutex_destroy(&pg->lock);
+    qemu_mutex_destroy(&pg->shader_cache_lock);
 
     glo_set_current(g_nv2a_context_render);
 
@@ -4060,7 +4061,9 @@ void pgraph_destroy(PGRAPHState *pg)
 
     glDeleteFramebuffers(1, &pg->gl_framebuffer);
 
-    // TODO: clear out shader cached
+    // Clear out shader cache
+    shader_write_cache_reload_list(pg);
+    free(pg->shader_cache_entries);
 
     // Clear out texture cache
     lru_flush(&pg->texture_cache);
@@ -4640,19 +4643,24 @@ static void pgraph_bind_shaders(PGRAPHState *pg)
         state.psh.conv_tex[i] = kernel;
     }
 
-    ShaderBinding* cached_shader = (ShaderBinding*)g_hash_table_lookup(pg->shader_cache, &state);
-    if (cached_shader) {
-        pg->shader_binding = cached_shader;
+    uint64_t shader_state_hash = fast_hash((uint8_t*) &state, sizeof(ShaderState));
+    qemu_mutex_lock(&pg->shader_cache_lock);
+    LruNode *node = lru_lookup(&pg->shader_cache, shader_state_hash, &state);
+    ShaderLruNode *snode = container_of(node, ShaderLruNode, node);
+    if (snode->binding || shader_load_from_memory(snode)) {
+        pg->shader_binding = snode->binding;
     } else {
         pg->shader_binding = generate_shaders(&state);
         nv2a_profile_inc_counter(NV2A_PROF_SHADER_GEN);
 
         /* cache it */
-        ShaderState *cache_state = (ShaderState *)g_malloc(sizeof(*cache_state));
-        memcpy(cache_state, &state, sizeof(*cache_state));
-        g_hash_table_insert(pg->shader_cache, cache_state,
-                            (gpointer)pg->shader_binding);
+        snode->binding = pg->shader_binding;
+        if (g_config.perf.cache_shaders) {
+            shader_cache_to_disk(snode);
+        }
     }
+
+    qemu_mutex_unlock(&pg->shader_cache_lock);
 
     binding_changed = (pg->shader_binding != old_binding);
     if (binding_changed) {
@@ -7647,17 +7655,6 @@ static bool texture_cache_entry_compare(Lru *lru, LruNode *node, void *key)
 {
     TextureLruNode *tnode = container_of(node, TextureLruNode, node);
     return memcmp(&tnode->key, key, sizeof(TextureKey));
-}
-
-/* hash and equality for shader cache hash table */
-static guint shader_hash(gconstpointer key)
-{
-    return fast_hash((const uint8_t *)key, sizeof(ShaderState));
-}
-static gboolean shader_equal(gconstpointer a, gconstpointer b)
-{
-    const ShaderState *as = (const ShaderState *)a, *bs = (const ShaderState *)b;
-    return memcmp(as, bs, sizeof(ShaderState)) == 0;
 }
 
 static unsigned int kelvin_map_stencil_op(uint32_t parameter)
--- a/hw/xbox/nv2a/shaders.c
+++ b/hw/xbox/nv2a/shaders.c
@@ -25,6 +25,9 @@
 
 #include "shaders_common.h"
 #include "shaders.h"
+#include "nv2a_int.h"
+#include "ui/xemu-settings.h"
+#include "xemu-version.h"
 
 void mstring_append_fmt(MString *qstring, const char *fmt, ...)
 {
@@ -71,6 +74,37 @@ void mstring_append_va(MString *qstring, const char *fmt, va_list va)
     g_free(buf);
 }
 
+GLenum get_gl_primitive_mode(enum ShaderPolygonMode polygon_mode, enum ShaderPrimitiveMode primitive_mode)
+{
+    if (polygon_mode == POLY_MODE_POINT) {
+        return GL_POINTS;
+    }
+
+    switch (primitive_mode) {
+    case PRIM_TYPE_POINTS: return GL_POINTS;
+    case PRIM_TYPE_LINES: return GL_LINES;
+    case PRIM_TYPE_LINE_LOOP: return GL_LINE_LOOP;
+    case PRIM_TYPE_LINE_STRIP: return GL_LINE_STRIP;
+    case PRIM_TYPE_TRIANGLES: return GL_TRIANGLES;
+    case PRIM_TYPE_TRIANGLE_STRIP: return GL_TRIANGLE_STRIP;
+    case PRIM_TYPE_TRIANGLE_FAN: return GL_TRIANGLE_FAN;
+    case PRIM_TYPE_QUADS: return GL_LINES_ADJACENCY;
+    case PRIM_TYPE_QUAD_STRIP: return GL_LINE_STRIP_ADJACENCY;
+    case PRIM_TYPE_POLYGON:
+        if (polygon_mode == POLY_MODE_LINE) {
+            return GL_LINE_LOOP;
+        } else if (polygon_mode == POLY_MODE_FILL) {
+            return GL_TRIANGLE_FAN;
+        }
+
+        assert(!"PRIM_TYPE_POLYGON with invalid polygon_mode");
+        return 0;
+    default:
+        assert(!"Invalid primitive_mode");
+        return 0;
+    }
+}
+
 static MString* generate_geometry_shader(
                                       enum ShaderPolygonMode polygon_front_mode,
                                       enum ShaderPolygonMode polygon_back_mode,
@@ -78,14 +112,14 @@ static MString* generate_geometry_shader(
                                       GLenum *gl_primitive_mode,
                                       bool smooth_shading)
 {
-
     /* FIXME: Missing support for 2-sided-poly mode */
     assert(polygon_front_mode == polygon_back_mode);
     enum ShaderPolygonMode polygon_mode = polygon_front_mode;
 
+    *gl_primitive_mode = get_gl_primitive_mode(polygon_mode, primitive_mode);
+
     /* POINT mode shouldn't require any special work */
     if (polygon_mode == POLY_MODE_POINT) {
-        *gl_primitive_mode = GL_POINTS;
         return NULL;
     }
 
@@ -94,12 +128,11 @@ static MString* generate_geometry_shader(
     const char *layout_out = NULL;
     const char *body = NULL;
     switch (primitive_mode) {
-    case PRIM_TYPE_POINTS: *gl_primitive_mode = GL_POINTS; return NULL;
-    case PRIM_TYPE_LINES: *gl_primitive_mode = GL_LINES; return NULL;
-    case PRIM_TYPE_LINE_LOOP: *gl_primitive_mode = GL_LINE_LOOP; return NULL;
-    case PRIM_TYPE_LINE_STRIP: *gl_primitive_mode = GL_LINE_STRIP; return NULL;
+    case PRIM_TYPE_POINTS: return NULL;
+    case PRIM_TYPE_LINES: return NULL;
+    case PRIM_TYPE_LINE_LOOP: return NULL;
+    case PRIM_TYPE_LINE_STRIP: return NULL;
     case PRIM_TYPE_TRIANGLES:
-        *gl_primitive_mode = GL_TRIANGLES;
         if (polygon_mode == POLY_MODE_FILL) { return NULL; }
         assert(polygon_mode == POLY_MODE_LINE);
         layout_in = "layout(triangles) in;\n";
@@ -111,7 +144,6 @@ static MString* generate_geometry_shader(
                "  EndPrimitive();\n";
         break;
     case PRIM_TYPE_TRIANGLE_STRIP:
-        *gl_primitive_mode = GL_TRIANGLE_STRIP;
         if (polygon_mode == POLY_MODE_FILL) { return NULL; }
         assert(polygon_mode == POLY_MODE_LINE);
         layout_in = "layout(triangles) in;\n";
@@ -133,7 +165,6 @@ static MString* generate_geometry_shader(
                "  EndPrimitive();\n";
         break;
     case PRIM_TYPE_TRIANGLE_FAN:
-        *gl_primitive_mode = GL_TRIANGLE_FAN;
         if (polygon_mode == POLY_MODE_FILL) { return NULL; }
         assert(polygon_mode == POLY_MODE_LINE);
         layout_in = "layout(triangles) in;\n";
@@ -147,7 +178,6 @@ static MString* generate_geometry_shader(
                "  EndPrimitive();\n";
         break;
     case PRIM_TYPE_QUADS:
-        *gl_primitive_mode = GL_LINES_ADJACENCY;
         layout_in = "layout(lines_adjacency) in;\n";
         if (polygon_mode == POLY_MODE_LINE) {
             layout_out = "layout(line_strip, max_vertices = 5) out;\n";
@@ -170,7 +200,6 @@ static MString* generate_geometry_shader(
         }
         break;
     case PRIM_TYPE_QUAD_STRIP:
-        *gl_primitive_mode = GL_LINE_STRIP_ADJACENCY;
         layout_in = "layout(lines_adjacency) in;\n";
         if (polygon_mode == POLY_MODE_LINE) {
             layout_out = "layout(line_strip, max_vertices = 5) out;\n";
@@ -198,11 +227,9 @@ static MString* generate_geometry_shader(
         break;
     case PRIM_TYPE_POLYGON:
         if (polygon_mode == POLY_MODE_LINE) {
-            *gl_primitive_mode = GL_LINE_LOOP;
             return NULL;
         }
         if (polygon_mode == POLY_MODE_FILL) {
-            *gl_primitive_mode = GL_TRIANGLE_FAN;
             if (smooth_shading) {
                 return NULL;
             }
@@ -291,7 +318,6 @@ static void append_skinning_code(MString* str, bool mix,
                                  const char* output, const char* input,
                                  const char* matrix, const char* swizzle)
 {
-
     if (count == 0) {
         mstring_append_fmt(str, "%s %s = (%s * %s0).%s;\n",
                            type, output, input, matrix, swizzle);
@@ -999,11 +1025,107 @@ static GLuint create_gl_shader(GLenum gl_shader_type,
     return shader;
 }
 
-ShaderBinding *generate_shaders(const ShaderState *state)
+void update_shader_constant_locations(ShaderBinding *binding, const ShaderState *state)
 {
     int i, j;
     char tmp[64];
 
+    /* set texture samplers */
+    for (i = 0; i < NV2A_MAX_TEXTURES; i++) {
+        char samplerName[16];
+        snprintf(samplerName, sizeof(samplerName), "texSamp%d", i);
+        GLint texSampLoc = glGetUniformLocation(binding->gl_program, samplerName);
+        if (texSampLoc >= 0) {
+            glUniform1i(texSampLoc, i);
+        }
+    }
+
+    /* validate the program */
+    glValidateProgram(binding->gl_program);
+    GLint valid = 0;
+    glGetProgramiv(binding->gl_program, GL_VALIDATE_STATUS, &valid);
+    if (!valid) {
+        GLchar log[1024];
+        glGetProgramInfoLog(binding->gl_program, 1024, NULL, log);
+        fprintf(stderr, "nv2a: shader validation failed: %s\n", log);
+        abort();
+    }
+
+    /* lookup fragment shader uniforms */
+    for (i = 0; i < 9; i++) {
+        for (j = 0; j < 2; j++) {
+            snprintf(tmp, sizeof(tmp), "c%d_%d", j, i);
+            binding->psh_constant_loc[i][j] = glGetUniformLocation(binding->gl_program, tmp);
+        }
+    }
+    binding->alpha_ref_loc = glGetUniformLocation(binding->gl_program, "alphaRef");
+    for (i = 1; i < NV2A_MAX_TEXTURES; i++) {
+        snprintf(tmp, sizeof(tmp), "bumpMat%d", i);
+        binding->bump_mat_loc[i] = glGetUniformLocation(binding->gl_program, tmp);
+        snprintf(tmp, sizeof(tmp), "bumpScale%d", i);
+        binding->bump_scale_loc[i] = glGetUniformLocation(binding->gl_program, tmp);
+        snprintf(tmp, sizeof(tmp), "bumpOffset%d", i);
+        binding->bump_offset_loc[i] = glGetUniformLocation(binding->gl_program, tmp);
+    }
+
+    for (int i = 0; i < NV2A_MAX_TEXTURES; i++) {
+        snprintf(tmp, sizeof(tmp), "texScale%d", i);
+        binding->tex_scale_loc[i] = glGetUniformLocation(binding->gl_program, tmp);
+    }
+
+    /* lookup vertex shader uniforms */
+    for(i = 0; i < NV2A_VERTEXSHADER_CONSTANTS; i++) {
+        snprintf(tmp, sizeof(tmp), "c[%d]", i);
+        binding->vsh_constant_loc[i] = glGetUniformLocation(binding->gl_program, tmp);
+    }
+    binding->surface_size_loc = glGetUniformLocation(binding->gl_program, "surfaceSize");
+    binding->clip_range_loc = glGetUniformLocation(binding->gl_program, "clipRange");
+    binding->fog_color_loc = glGetUniformLocation(binding->gl_program, "fogColor");
+    binding->fog_param_loc[0] = glGetUniformLocation(binding->gl_program, "fogParam[0]");
+    binding->fog_param_loc[1] = glGetUniformLocation(binding->gl_program, "fogParam[1]");
+
+    binding->inv_viewport_loc = glGetUniformLocation(binding->gl_program, "invViewport");
+    for (i = 0; i < NV2A_LTCTXA_COUNT; i++) {
+        snprintf(tmp, sizeof(tmp), "ltctxa[%d]", i);
+        binding->ltctxa_loc[i] = glGetUniformLocation(binding->gl_program, tmp);
+    }
+    for (i = 0; i < NV2A_LTCTXB_COUNT; i++) {
+        snprintf(tmp, sizeof(tmp), "ltctxb[%d]", i);
+        binding->ltctxb_loc[i] = glGetUniformLocation(binding->gl_program, tmp);
+    }
+    for (i = 0; i < NV2A_LTC1_COUNT; i++) {
+        snprintf(tmp, sizeof(tmp), "ltc1[%d]", i);
+        binding->ltc1_loc[i] = glGetUniformLocation(binding->gl_program, tmp);
+    }
+    for (i = 0; i < NV2A_MAX_LIGHTS; i++) {
+        snprintf(tmp, sizeof(tmp), "lightInfiniteHalfVector%d", i);
+        binding->light_infinite_half_vector_loc[i] =
+            glGetUniformLocation(binding->gl_program, tmp);
+        snprintf(tmp, sizeof(tmp), "lightInfiniteDirection%d", i);
+        binding->light_infinite_direction_loc[i] =
+            glGetUniformLocation(binding->gl_program, tmp);
+
+        snprintf(tmp, sizeof(tmp), "lightLocalPosition%d", i);
+        binding->light_local_position_loc[i] = glGetUniformLocation(binding->gl_program, tmp);
+        snprintf(tmp, sizeof(tmp), "lightLocalAttenuation%d", i);
+        binding->light_local_attenuation_loc[i] =
+            glGetUniformLocation(binding->gl_program, tmp);
+    }
+    for (i = 0; i < 8; i++) {
+        snprintf(tmp, sizeof(tmp), "clipRegion[%d]", i);
+        binding->clip_region_loc[i] = glGetUniformLocation(binding->gl_program, tmp);
+    }
+
+    if (state->fixed_function) {
+        binding->material_alpha_loc =
+            glGetUniformLocation(binding->gl_program, "material_alpha");
+    } else {
+        binding->material_alpha_loc = -1;
+    }
+}
+
+ShaderBinding *generate_shaders(const ShaderState *state)
+{
     char *previous_numeric_locale = setlocale(LC_NUMERIC, NULL);
     if (previous_numeric_locale) {
         previous_numeric_locale = g_strdup(previous_numeric_locale);
@@ -1011,10 +1133,9 @@ ShaderBinding *generate_shaders(const ShaderState *state)
 
     /* Ensure numeric values are printed with '.' radix, no grouping */
     setlocale(LC_NUMERIC, "C");
-
     GLuint program = glCreateProgram();
 
-    /* Create an option geometry shader and find primitive type */
+    /* Create an optional geometry shader and find primitive type */
     GLenum gl_primitive_mode;
     MString* geometry_shader_code =
         generate_geometry_shader(state->polygon_front_mode,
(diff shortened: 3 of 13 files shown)


hw/xbox/nv2a/pgraph.c
```diff
@@ -7176,15 +7219,269 @@ static bool texture_cache_entry_compare(Lru *lru, LruNode *node, void *key)
...
+
+    char name[20];
+    snprintf(name, 20, "%llu", (unsigned long long) snode->node.hash);
+    qemu_thread_create(&snode->save_thread, name, shader_cache_write_to_disk, snode, QEMU_THREAD_JOINABLE);
+}
+
+static void shader_write_lru_list_entry_to_disk(Lru *lru, LruNode *node, void *opaque)
+{
+    FILE *lru_list_file = (FILE*) opaque;
+    fwrite(&node->hash, sizeof(uint64_t), 1, lru_list_file);
```
abaire: Check to see if the write was successful

hw/xbox/nv2a/pgraph.c
```diff
@@ -649,6 +657,20 @@ void pgraph_flush(NV2AState *d)
     qemu_event_set(&d->pgraph.flush_complete);
 }
 
+void pgraph_cache_shaders_to_disk(NV2AState *d)
+{
+    char *shader_lru_path = shader_get_lru_cache_path();
+    FILE *lru_list = qemu_fopen(shader_lru_path, "wb");
```
abaire: Should we write to a temp file first and then move only once the write is completed?

My worry is that a power outage or untimely hard kill could leave a partially written cache file. Doing a move would minimize the failure window.

hw/xbox/nv2a/pgraph.c
```diff
@@ -3829,7 +3852,20 @@ void pgraph_init(NV2AState *d)
     pg->element_cache.init_node = vertex_cache_entry_init;
     pg->element_cache.compare_nodes = vertex_cache_entry_compare;
 
-    pg->shader_cache = g_hash_table_new(shader_hash, shader_equal);
+    shader_create_cache_folder();
+    const size_t shader_cache_size = 50*1024;
```
abaire: Out of curiosity, how did you decide on 50k?

hw/xbox/nv2a/pgraph.c
```diff
@@ -649,6 +657,20 @@ void pgraph_flush(NV2AState *d)
     qemu_event_set(&d->pgraph.flush_complete);
 }
 
+void pgraph_cache_shaders_to_disk(NV2AState *d)
```
abaire: Double checking as I'm not super familiar with the qemu life cycle: is it possible to get into a state where the `shader_disk_thread` is still running and the LRU cache is still being mutated when this is called?

If that's guaranteed by the life cycle, maybe we should add a comment or assert to clarify that this should only be called during the shutdown state?

If not, should we lock the cache while it's being populated and here while it's being saved?

hw/xbox/nv2a/pgraph.c
```diff
@@ -7176,15 +7219,269 @@ static bool texture_cache_entry_compare(Lru *lru, LruNode *node, void *key)
...
+    char *shader_lru_path = shader_get_lru_cache_path();
+
+    FILE *lru_shaders_list = qemu_fopen(shader_lru_path, "rb");
+    g_free(shader_lru_path);
+    if (!lru_shaders_list) {
+        rcu_unregister_thread();
+        return NULL;
+    }
+
+    fseek(lru_shaders_list, 0, SEEK_END);
```
abaire: Nit: While I think it's probably fine on all of the platforms xemu supports, I believe `SEEK_END` behavior is not actually guaranteed by the standard.

hw/xbox/nv2a/pgraph.c
```diff
@@ -7176,15 +7219,269 @@ static bool texture_cache_entry_compare(Lru *lru, LruNode *node, void *key)
...
+    size_t shader_list_size = ftell(lru_shaders_list);
+    fseek(lru_shaders_list, 0, SEEK_SET);
+
+    for (size_t read = 0; read < shader_list_size; read += sizeof(uint64_t)) {
```
abaire: Should we bomb out early if `shader_list_size` is not evenly divisible by `sizeof(uint64_t)` (presumably that'd indicate that the cache is corrupted or truncated)?

hw/xbox/nv2a/pgraph.c
```diff
@@ -7176,15 +7219,269 @@ static bool texture_cache_entry_compare(Lru *lru, LruNode *node, void *key)
...
+{
+    assert(glGetError() == GL_NO_ERROR);
+
+    if (!snode->program) {
+        return false;
+    }
+
+    GLuint gl_program = glCreateProgram();
+    glProgramBinary(gl_program, snode->program_format, snode->program, snode->program_size);
+    if (glGetError() != GL_NO_ERROR) {
```
abaire: This seems potentially trace-worthy (w/ the actual error preserved).

hw/xbox/nv2a/shaders.c
```diff
@@ -70,6 +70,37 @@ void mstring_append_va(MString *qstring, const char *fmt, va_list va)
...
+    case PRIM_TYPE_QUADS: return GL_LINES_ADJACENCY;
+    case PRIM_TYPE_QUAD_STRIP:
+    case PRIM_TYPE_POLYGON:
+        if (polygon_mode == POLY_MODE_LINE) {
+            return GL_LINE_LOOP;
+        } else if (polygon_mode == POLY_MODE_FILL) {
+            return GL_TRIANGLE_FAN;
+        }
+
+        assert(false);
```
abaire: Can we use `assert(!"Some meaningful message")` so that user reports are less prone to expire? 

Trying to look at year+ old "It crashes with assert(false); at line <something_long_since_moved_around>" makes it more painful to try to address problems than it needs to be.

hw/xbox/nv2a/shaders.c
```diff
@@ -70,6 +70,37 @@ void mstring_append_va(MString *qstring, const char *fmt, va_list va)
     g_free(buf);
 }
 
+GLenum get_gl_primitive_mode(enum ShaderPolygonMode polygon_mode, enum ShaderPrimitiveMode primitive_mode)
```
abaire, line 77: Is this refactoring intentionally part of this PR?

hw/xbox/nv2a/shaders.c
```diff
@@ -936,156 +957,162 @@ static GLuint create_gl_shader(GLenum gl_shader_type,
...
+    }
+}
+
+ShaderBinding* generate_shaders(const ShaderState state)
+{
+
+    char vtx_prefix;
+    GLuint program = glCreateProgram();
+
+    /* Create an option geometry shader and find primitive type */
```
abaire: Supernit: Since you're in here, I believe the intent was to say "optional" not "option"

ui/xemu-settings.c
```diff
@@ -227,6 +227,18 @@ void xemu_settings_set_path(const char *path)
 	fprintf(stderr, "%s: config path: %s\n", __func__, settings_path);
 }
 
+char *xemu_get_config_dir_path(void)
+{
+	char *base = xemu_settings_detect_portable_mode()
```
abaire: Why not take the same approach as `xemu_settings_get_path` and permacache this so the caller doesn't have to remember to free it?

ui/xemu-settings.c
```diff
@@ -227,6 +227,18 @@ void xemu_settings_set_path(const char *path)
...
+	             ? SDL_GetBasePath()
+	             : SDL_GetPrefPath("xemu", "xemu");
+	assert(base != NULL);
+	char *path = g_malloc(strlen(base) + 1);
```
abaire: `g_strdup`?

ui/xemu-settings.c
```diff
@@ -227,6 +227,18 @@ void xemu_settings_set_path(const char *path)
...
+	strcpy(path, base);
+	SDL_free(base);
+	return path;
+}
+
 const char *xemu_settings_get_path(void)
```
abaire: GH won't let me comment below, but should we update `xemu_settings_get_path` to use `xemu_get_config_dir_path` to reduce code duplication? I assume these should never drift.

hw/xbox/nv2a/pgraph.c
```diff
@@ -7176,15 +7219,269 @@ static bool texture_cache_entry_compare(Lru *lru, LruNode *node, void *key)
...
+    }
+
+    GLint program_size;
+    glGetProgramiv(snode->binding->gl_program, GL_PROGRAM_BINARY_LENGTH, &program_size);
+    
+    if (snode->program) {
+        g_free(snode->program);
+    }
+
+    /* program_size might be zero on some systems, if no binary formats are supported */
```
abaire: Should we differentiate between these two cases then? If `program_size` is 0, bail out silently. If it's not zero but `g_malloc` failed, blow up (or at least yell in the log)?

hw/xbox/nv2a/pgraph.c
```diff
@@ -7176,15 +7219,269 @@ static bool texture_cache_entry_compare(Lru *lru, LruNode *node, void *key)
...
+
+cleanup:
+    unlink(shader_path);
+    g_free(shader_path);
+    g_free(snode->program);
+    snode->program = NULL;
+    return NULL;
+}
+
+static void shader_cache_to_disk(ShaderLruNode *snode)
```
abaire: (I don't know enough about GL shader binding performance) Is the performance of this new cache writing comparable to the old trivial hash table insertion?

hw/xbox/nv2a/pgraph.c
```diff
@@ -7176,15 +7219,269 @@ static bool texture_cache_entry_compare(Lru *lru, LruNode *node, void *key)
...
+
+    GLsizei program_size_copied;
+    glGetProgramBinary(snode->binding->gl_program, program_size, &program_size_copied, &snode->program_format, snode->program);
+    assert(glGetError() == GL_NO_ERROR);
+
+    snode->program_size = program_size_copied;
+    snode->cached = true;
+
+    char name[20];
+    snprintf(name, 20, "%llu", (unsigned long long) snode->node.hash);
```
abaire: Nit: maybe "scache-%llu" or something so it's clear what threads with 20 character hashes are really for (and in case we find a need to cache other things using background threads in the future)

hw/xbox/nv2a/pgraph.c
```diff
@@ -7176,15 +7219,269 @@ static bool texture_cache_entry_compare(Lru *lru, LruNode *node, void *key)
...
+    ShaderLruNode *snode = (ShaderLruNode*) arg;
+
+    char *cfg_path = xemu_get_config_dir_path();
+    size_t shader_path_len = strlen(cfg_path) + 30;
+    char *shader_path = (char*) g_malloc(shader_path_len);
+    snprintf(shader_path, shader_path_len, "%s/shaders/%llu", cfg_path, (unsigned long long) snode->node.hash);
+    g_free(cfg_path);
+
+    FILE *shader_file = qemu_fopen(shader_path, "wb");
+    if (!shader_file) {
```
abaire: Should we warn on any of these failures?

If a user complained about performance issues and it turned out that the shader cache was never being updated for some reason I think it'd be useful to know when trying to diagnose the problem.

hw/xbox/nv2a/pgraph.c
```diff
@@ -7176,15 +7219,269 @@ static bool texture_cache_entry_compare(Lru *lru, LruNode *node, void *key)
...
+    }
+
+    g_free(shader_path);
+    g_free(snode->program);
+    snode->program = NULL;
+
+    rcu_unregister_thread();
+    return NULL;
+
+cleanup:
```
abaire: Nit, maybe "error_cleanup" or "fail_cleanup" or something that indicates that this is only meant to be jumped to in a failure case?

hw/xbox/nv2a/pgraph.c
```diff
@@ -7176,15 +7219,269 @@ static bool texture_cache_entry_compare(Lru *lru, LruNode *node, void *key)
...
+        goto cleanup;
+    }
+
+    written = fwrite(&snode->state, sizeof(ShaderState), 1, shader_file);
+    if (written != 1) {
+        fclose(shader_file);
+        goto cleanup;
+    }
+
+    written = fwrite(snode->program, snode->program_size, 1, shader_file);
```
abaire: I'd suggest we write the program size explicitly as well. With the current code I think the size of the program is inferred by subtracting sizeof GLEnum + SS from the total size, which means a file that is truncated for some reason (e.g., xemu crashing during this write) would be treated as a valid entry. If we write the size explicitly and we can verify that the total cache file size matches expectations as a proxy for validating that the program data is correct.

hw/xbox/nv2a/pgraph.c
```diff
@@ -7176,15 +7219,269 @@ static bool texture_cache_entry_compare(Lru *lru, LruNode *node, void *key)
...
+    g_free(snode->program);
+    snode->program = NULL;
+
+    update_shader_constant_locations(binding, &snode->state);
+
+    return true;
 }
-static gboolean shader_equal(gconstpointer a, gconstpointer b)
+
+static void shader_load_from_disk(PGRAPHState *pg, uint64_t hash)
```
abaire: Nit: Maybe `shader_cache_load_from_disk` or `shader_cache_read_from_disk` to be aligned with naming used below (`shader_cache_write_to_disk` specifically). Or vice versa, rename below to be aligned with this.

hw/xbox/nv2a/pgraph.c
```diff
@@ -7176,15 +7219,269 @@ static bool texture_cache_entry_compare(Lru *lru, LruNode *node, void *key)
...
+{
+    char *cfg_dir = xemu_get_config_dir_path();
+    size_t shader_path_len = strlen(cfg_dir) + 30;
+    char *shader_path = g_malloc(shader_path_len);
+    snprintf(shader_path, shader_path_len, "%s/shaders/%llu", cfg_dir, (unsigned long long) hash);
```
abaire: Maybe make the format string a constant or create a helper function to get the shader cache path for a hash, since it's important that the read here and the write function below to use the same path?

hw/xbox/nv2a/pgraph.c
```diff
@@ -7176,15 +7219,269 @@ static bool texture_cache_entry_compare(Lru *lru, LruNode *node, void *key)
...
+    g_free(cfg_dir);
+
+    FILE *shader_file = qemu_fopen(shader_path, "rb");
+    if (!shader_file) {
+        goto cleanup;
+    }
+
+    fseek(shader_file, 0, SEEK_END);
+    size_t shader_size = ftell(shader_file) - sizeof(GLenum) - sizeof(ShaderState);
```
abaire: I added a comment in the write path below, but I'm worried that this would happily accept truncated files (as long as they're at least 1 byte larger than the enum + state sizes).

hw/xbox/nv2a/pgraph.c
```diff
@@ -7176,15 +7219,269 @@ static bool texture_cache_entry_compare(Lru *lru, LruNode *node, void *key)
...
+
+    /* If we happened to regenerate this shader already, then we may as well use the new one */
+    if(snode->binding) {
+        qemu_mutex_unlock(&pg->shader_cache_lock);
+        return;
+    }
+
+    snode->program_format = program_binary_format;
+    snode->program_size = shader_size;
+    snode->program = program_buffer;
```
abaire: If you corrupt the program data in the cache file, what happens when it's used out of the cache? Do we have a path to regen? If not, should we also add a hash or CRC of the program so we can be confident that we're not adding a vector for HDD corruption to introduce inexplicable crashes?

hw/xbox/nv2a/pgraph.c
antangelo: Good catch, I can call `qemu_thread_join` on it before visiting the LRU entries.

hw/xbox/nv2a/pgraph.c
antangelo: It's not critical that the LRU list is written properly: the load routine will fail quietly if it cannot load part or all of the list, and then regenerate those shaders that it was unable to load when they are next encountered.

I suppose this would result in losing part of the cache and having to regenerate it, which isn't ideal.

hw/xbox/nv2a/pgraph.c
antangelo: Copied it from the element cache, mostly. I tested a game that compiled shaders every time the camera was moved, so I don't think this value is unreasonable (it might be low, even).

hw/xbox/nv2a/pgraph.c
antangelo: Take a look at the spec: https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glProgramBinary.xhtml
It only fails if `gl_program` is invalid, or if the program binary format is unsupported (which can happen due to driver updates). I didn't think it was necessary to log these, as they result in the shader being regenerated anyway.

hw/xbox/nv2a/pgraph.c
antangelo: Good point, will fix.

hw/xbox/nv2a/pgraph.c
antangelo: Will fix.

hw/xbox/nv2a/pgraph.c
antangelo: I had assumed that `glProgramBinary` will validate that the binary is valid but this doesn't seem to be the case. I'll fix this by using `glValidateProgram` after loading.

hw/xbox/nv2a/pgraph.c
antangelo: I don't know if there's another good cross-platform way to do this. I can move it into the platform specific utils in case we introduce a new platform that doesn't support it, but I don't know if that's necessary at this point.

hw/xbox/nv2a/pgraph.c
antangelo: I don't think it makes a difference, if it's not a multiple of `sizeof(uint64_t)` then the `fread` will fail and the loop will (effectively) end. `shader_load_from_disk` will also quietly fail if the read hash is invalid, in the case that the shader list is corrupted.

hw/xbox/nv2a/pgraph.c
antangelo: I can print a warning message ~and add a flag to ensure it's only printed once~. On second thought, I don't think a flag is necessary at this point, as the only possible failures have to do with writing the file.

hw/xbox/nv2a/pgraph.c
antangelo: Good catch. I'll write the size and validate the program on load to mitigate this.

hw/xbox/nv2a/pgraph.c
antangelo: Will fix

hw/xbox/nv2a/shaders.c
PatrickvL: Besides below, it's also called from shader_load_from_memory(), so it seems intentional

hw/xbox/nv2a/pgraph.c
antangelo: Theoretically it should be slower (LRU is backed by `QTAILQ`), but I didn't notice a performance impact on my system. However, it could just be that my system is fast and I don't have enough shaders in the cache to notice the difference.

hw/xbox/nv2a/pgraph.c
antangelo: If `g_malloc` itself has failed then I would imagine things have already blown up, but I can make this check explicit.

hw/xbox/nv2a/pgraph.c
antangelo: Will fix

hw/xbox/nv2a/pgraph.c
antangelo: What would you suggest the failure behavior be? I can print a log message but that may be noisy if subsequent writes also fail. I can't think of a way to recover otherwise from this point.

hw/xbox/nv2a/shaders.c
antangelo: Yes it's intentional, I need it when I create the `ShaderBinding` when loading from disk (and likewise for updating the constant locations).

hw/xbox/nv2a/shaders.c
antangelo: Will fix

ui/xemu-settings.c
antangelo: Good idea, will fix

ui/xemu-settings.c
antangelo: Will fix

ui/xemu-settings.c
antangelo: Will fix

hw/xbox/nv2a/pgraph.c
abaire: sg, having to regenerate a part of the cache seems no worse than having to regenerate the whole thing (which would be the case if we did a write-then-move and had it fail) so your approach seems better.

hw/xbox/nv2a/pgraph.c
abaire: Making sure I understand the flow: the only way we'd anticipate hitting those would be if the shader loaded from disk has become invalid (either due to it being corrupted or the driver being updated). I think if you switch to doing explicit validation on the reload case it should no longer be possible for this to be hit?

If that's the case, then should we just assert instead of handling it silently-but-gracefully (and potentially hiding some failure mode)?

hw/xbox/nv2a/pgraph.c
abaire: Yeah, I don't think it's necessary to change, just something to note in case we hear about weird behavior on some esoteric platform in the future.

hw/xbox/nv2a/pgraph.c
abaire: sg

hw/xbox/nv2a/pgraph.c
abaire: I'd say we should probably try to catch the case where shader bin formats aren't supported before even bothering to create the file, no?

My assumption is that once we're in this path we'd really expect to be able to write to the filesystem without any error, and failure to do basic writes would either mean there's some misconfiguration (e.g., read-only target) or some very unexpected low level error, both of which seem like they'd be good to know when debugging.

hw/xbox/nv2a/pgraph.c
abaire: Devil's advocate against my original suggestion - if you're going to validate the shaders on load, maybe we don't need to also save the size? I guess it'd let you bomb out early if the file is truncated, but presumably that'd be a rare case and if we're going to do full validation anyway I'm dubious that it'd save much time in practice.

hw/xbox/nv2a/pgraph.c
abaire: I don't think it'd be unreasonable to log it, even if it's spammy, since it should essentially never happen in typical scenarios (I guess an out of HDD space case is the most likely culprit). If log spam is the biggest concern I think we could have opaque be a struct that has the `FILE *` and a boolean and just && the bool w/ the result, then log failure in `pgraph_cache_shaders_to_disk`.

That said, since it seems likely that opening the file would fail if the issue is an unwritable target (which strikes me as the most likely fail scenario), I guess it's also reasonable to go without a check here.

hw/xbox/nv2a/pgraph.c
antangelo: Writing the size removes the need for the `fseek` to file end, so it has other benefits that make it an attractive solution. Perhaps we don't need to validate the shader after loading it in memory, but I think it might be good practice to do so anyway.

hw/xbox/nv2a/pgraph.c
antangelo: My bad on this one, the case where shader bin formats are unsupported is detected beforehand, before this function is called. I'll add a print to stderr (which shouldn't have a performance impact since this is running in its own thread).

hw/xbox/nv2a/pgraph.c
antangelo: That's correct. This path is happening on load, and I don't know if there's another way to check if the binary format is supported beforehand (or if it's worth doing, since the GL implementation will do it here as well).

hw/xbox/nv2a/shaders.c
antangelo: Missed this last time, but will fix

hw/xbox/nv2a/pgraph.c
antangelo: Since it's unlikely that the write will fail if open didn't fail, I'll add a log there for good measure. I'll also add a log if open failed, since I missed that initially.

antangelo:
Pushed some changes to reflect the review comments, but I have only tested them on Linux so far.

hw/xbox/nv2a/pgraph.c
mborgerson: > I tested a game that compiled shaders every time the camera was moved, so I don't think this value is unreasonable

@antangelo which game was this?

antangelo: Battle for Bikini Bottom (5451001a). It seems to be somewhat level dependent. I haven't looked into what part of the shader state is changing so frequently.


https://user-images.githubusercontent.com/25046221/159141456-112253e4-6d62-44e2-b60c-cf2e23542890.mp4

hw/xbox/nv2a/shaders.c
```diff
@@ -936,156 +957,162 @@ static GLuint create_gl_shader(GLenum gl_shader_type,
...
+        binding->material_alpha_loc = glGetUniformLocation(binding->gl_program, "material_alpha");
     } else {
-        ret->material_alpha_loc = -1;
+        binding->material_alpha_loc = -1;
+    }
+}
+
+ShaderBinding* generate_shaders(const ShaderState state)
+{
+
```
mborgerson, line 1133: nit: blank line

ui/xemu-os-utils-windows.c
```diff
@@ -29,3 +29,13 @@ void xemu_open_web_browser(const char *url)
 {
 	ShellExecute(0, "open", url, 0, 0 , SW_SHOW);
 }
+
+int xemu_mkdir(const char *path)
+{
+	BOOL dirResult = CreateDirectoryA(path, 0);
```
mborgerson: Should use unicode/wide version and convert `path` from utf-8 to utf-16

hw/xbox/nv2a/pgraph.c
```diff
@@ -3826,7 +3860,22 @@ void pgraph_init(NV2AState *d)
     pg->element_cache.init_node = vertex_cache_entry_init;
     pg->element_cache.compare_nodes = vertex_cache_entry_compare;
 
-    pg->shader_cache = g_hash_table_new(shader_hash, shader_equal);
+    shader_create_cache_folder();
+    //const size_t shader_cache_size = 50*1024;
+    const size_t shader_cache_size = 5;
```
mborgerson: 5 is pretty low. Left over from debug?

hw/xbox/nv2a/pgraph.c
```diff
@@ -7176,15 +7231,307 @@ static bool texture_cache_entry_compare(Lru *lru, LruNode *node, void *key)
...
+static void shader_load_from_disk(PGRAPHState *pg, uint64_t hash)
+{
+    char *shader_path = shader_get_binary_path(hash);
+
+    FILE *shader_file = qemu_fopen(shader_path, "rb");
+    if (!shader_file) {
+        goto error;
+    }
+
+    GLenum program_binary_format;
```
mborgerson: Let's also add a couple extra fields to avoid bugs:
* Let's add a check that strictly enforces we only load shaders generated by the exact same build of xemu
* We can also add a check for GL vendor/driver, which can change if someone is using a system with hybrid graphics for instance. If the current vendor doesn't match the vendor of the loaded shader we can skip uploading + validating the shader for it to just eventually fail.

hw/xbox/nv2a/pgraph.c
```diff
@@ -7176,15 +7231,307 @@ static bool texture_cache_entry_compare(Lru *lru, LruNode *node, void *key)
...
+    }
+
+    g_free(shader_path);
+
+    qemu_mutex_lock(&pg->shader_cache_lock);
+    LruNode *node = lru_lookup(&pg->shader_cache, hash, &state);
+    ShaderLruNode *snode = container_of(node, ShaderLruNode, node);
+
+    /* If we happened to regenerate this shader already, then we may as well use the new one */
+    if(snode->binding) {
```
mborgerson: Let's do this check earlier, before opening/reading the shader program file, etc

hw/xbox/nv2a/pgraph.c
```diff
@@ -7176,15 +7231,307 @@ static bool texture_cache_entry_compare(Lru *lru, LruNode *node, void *key)
...
+    rcu_unregister_thread();
+    return NULL;
 }
-static gboolean shader_equal(gconstpointer a, gconstpointer b)
+
+static void *shader_write_to_disk(void *arg)
 {
-    const ShaderState *as = (const ShaderState *)a, *bs = (const ShaderState *)b;
-    return memcmp(as, bs, sizeof(ShaderState)) == 0;
+    rcu_register_thread();
```
mborgerson: This thread doesn't use rcu so we can drop the registration

hw/xbox/nv2a/pgraph.c
```diff
@@ -7176,15 +7231,307 @@ static bool texture_cache_entry_compare(Lru *lru, LruNode *node, void *key)
...
+error:
+    /* Delete the shader so it won't be loaded again */
+    unlink(shader_path);
+    g_free(shader_path);
+}
+
+static void *shader_reload_lru_from_disk(void *arg)
 {
-    return fast_hash((const uint8_t *)key, sizeof(ShaderState));
+    rcu_register_thread();
```
mborgerson: This thread isn't using rcu, so we can drop this registration

hw/xbox/nv2a/pgraph.c
```diff
@@ -3826,7 +3860,22 @@ void pgraph_init(NV2AState *d)
...
+    assert(pg->shader_cache_entries != NULL);
+    for (i = 0; i < shader_cache_size; i++) {
+        lru_add_free(&pg->shader_cache, &pg->shader_cache_entries[i].node);
+    }
+
+    pg->shader_cache.init_node = shader_cache_entry_init;
+    pg->shader_cache.compare_nodes = shader_cache_entry_compare;
+    pg->shader_cache.post_node_evict = shader_cache_entry_post_evict;
+
+    qemu_thread_create(&pg->shader_disk_thread, "pgraph.shader_cache", shader_reload_lru_from_disk, pg, QEMU_THREAD_JOINABLE);
```
mborgerson: Let's add a config item that controls whether or not we cache shaders. I added a stub item in config_spec.yml for this

hw/xbox/nv2a/pgraph.c
```diff
@@ -7176,15 +7231,307 @@ static bool texture_cache_entry_compare(Lru *lru, LruNode *node, void *key)
...
+
+    return true;
+}
+
+static char *shader_get_binary_path(uint64_t hash)
+{
+    const char *cfg_dir = xemu_settings_get_base_path();
+    size_t shader_path_len = strlen(cfg_dir) + 30;
+    char *shader_path = g_malloc(shader_path_len);
+    snprintf(shader_path, shader_path_len, "%s/shaders/%llu", cfg_dir, (unsigned long long) hash);
```
mborgerson: * We simplify the `strlen`+`malloc`+`sprintf` with `g_strdup_printf`
* Let's change the filename from decimal to hex and pad it to 16 digits (eg `"%016"PRIx64`)
* Let's bin the shaders into subdirectories based on some number of hash bits

hw/xbox/nv2a/pgraph.c
```diff
@@ -7176,15 +7231,307 @@ static bool texture_cache_entry_compare(Lru *lru, LruNode *node, void *key)
...
+}
+
+void nv2a_delete_shader_cache(void)
+{
+    NV2AState *d = g_nv2a;
+
+    qatomic_set(&d->pgraph.shader_cache_save_list, false);
+
+    char *path = shader_get_lru_cache_path();
+    unlink(path);
```
mborgerson: * `unlink` (and other references here to it) should be using `qemu_unlink`
* `qemu_unlink` should be updated to use `_wunlink` on Windows, and convert the utf-8 path to utf-16 accordingly

mborgerson (changes requested):
Thanks for the patch! Sorry about the review delay. Left some change requests

Also this needs a rebase but there are some conflicts. Can you please fix them up

hw/xbox/nv2a/pgraph.c
antangelo: My bad, this is left over from testing. As an aside, do you think the original `50 * 1024` is adequate, or should that number be increased/decreased in your opinion?

hw/xbox/nv2a/pgraph.c
antangelo: The earliest this can be done is after the state is read from disk (as the state struct is necessary to lookup the LRU node), at which point the only remaining item to be read is the shader program. If I did put a check for this there, we'd still need to check after the shader program is loaded from disk (the lock would be released before continuing to load from disk, and the shader might be regenerated between then and acquiring the lock again). I'm not sure if there's much utility in doing that, especially given that this runs once at startup and then never again.

hw/xbox/nv2a/shaders.c
antangelo: Fixed

ui/xemu-os-utils-windows.c
antangelo: Done

hw/xbox/nv2a/pgraph.c
antangelo: Done, using xemu version string and GL vendor

hw/xbox/nv2a/pgraph.c
antangelo: Done

hw/xbox/nv2a/pgraph.c
antangelo: Done

hw/xbox/nv2a/pgraph.c
antangelo: Done

hw/xbox/nv2a/pgraph.c
antangelo: Done, using first 16 bits for the bin and remaining 48 bits for the filename.

hw/xbox/nv2a/pgraph.c
antangelo: Done

hw/xbox/nv2a/pgraph.c
```diff
@@ -7194,15 +7259,348 @@ static bool texture_cache_entry_compare(Lru *lru, LruNode *node, void *key)
     return memcmp(&tnode->key, key, sizeof(TextureKey));
 }
 
-/* hash and equality for shader cache hash table */
-static guint shader_hash(gconstpointer key)
+static void shader_cache_entry_init(Lru *lru, LruNode *node, void *state)
```
mborgerson: Lots of new code added here. Can we break this out to shaders.c

hw/xbox/nv2a/pgraph.c
mborgerson: Hash collision should be rare but in the event it does happen, it would conflict on disk anyway. We can just check the cache and see if we already have a generated shader before checking the filesystem. It can be that the shader is generated very early in startup for instance while we are loading a potentially large cache of shaders, so we can add an early out check without too much cost and skip loading if its already generated

hw/xbox/nv2a/pgraph.c
```diff
@@ -7194,15 +7259,348 @@ static bool texture_cache_entry_compare(Lru *lru, LruNode *node, void *key)
...
+static void shader_load_from_disk(PGRAPHState *pg, uint64_t hash)
+{
+    char *shader_path = shader_get_binary_path(hash);
+
+    FILE *shader_file = qemu_fopen(shader_path, "rb");
+    if (!shader_file) {
+        goto error;
+    }
+
+    uint64_t xemu_cache_version_len;
```
mborgerson: Can we refactor this read size / read data pattern to make this function smaller

hw/xbox/nv2a/nv2a_int.h
```diff
@@ -391,10 +394,13 @@ typedef struct PGRAPHState {
     bool download_dirty_surfaces_pending;
     bool flush_pending;
     bool gl_sync_pending;
+    bool shader_cache_pending;
+    bool shader_cache_save_list;
     QemuEvent downloads_complete;
     QemuEvent dirty_surfaces_download_complete;
     QemuEvent flush_complete;
     QemuEvent gl_sync_complete;
+    QemuEvent shader_cache_complete;
```
mborgerson: The event names are kind of ambiguous about what they signify. I think `shader_cache_disk_writeback_complete` or something like that might be more clear, for instance

hw/xbox/nv2a/pgraph.c
```diff
@@ -7194,15 +7259,348 @@ static bool texture_cache_entry_compare(Lru *lru, LruNode *node, void *key)
...
+
+    FILE *shader_file = qemu_fopen(shader_path, "wb");
+    if (!shader_file) {
+        goto error;
+    }
+
+    uint64_t xemu_version_len = (uint64_t) (strlen(xemu_version) + 1);
+    size_t written = fwrite(&xemu_version_len, sizeof(xemu_version_len), 1, shader_file);
+    if (written != 1) {
+        fclose(shader_file);
```
mborgerson: Let's refactor this 8x repeated pattern

ui/xemu-os-utils-linux.c
```diff
@@ -95,3 +98,13 @@ void xemu_open_web_browser(const char *url)
 
 	free(cmd);
 }
+
+int xemu_mkdir(const char *path)
```
mborgerson: The preferred place for these kinds of functions is osdep.{h,c} with other functions like qemu_{open, close, ...}

ui/xemu-settings.cc
```diff
@@ -68,18 +69,31 @@ void xemu_settings_set_path(const char *path)
     fprintf(stderr, "%s: config path: %s\n", __func__, settings_path);
 }
 
-const char *xemu_settings_get_path(void)
+const char *xemu_settings_get_base_path(void)
 {
-    if (settings_path != NULL) {
-        return settings_path;
+    if (base_path != NULL) {
```
mborgerson, line 75: Let's move `base_path` scope into this function

Adamillo:
@antangelo what's the state of this PR?

antangelo:
> @antangelo what's the state of this PR?

Needs a few changes, I'll probably have time to get back to it in the next week or so.

hw/xbox/nv2a/pgraph.c
antangelo: Done

hw/xbox/nv2a/pgraph.c
antangelo: Done, I added an early check using the hash only.

hw/xbox/nv2a/pgraph.c
antangelo: Done, added a similar macro to writes

hw/xbox/nv2a/nv2a_int.h
antangelo: Done

hw/xbox/nv2a/pgraph.c
antangelo: Done, moved into macro

ui/xemu-os-utils-linux.c
antangelo: Done

ui/xemu-settings.cc
antangelo: Done

hw/xbox/nv2a/shaders.c
```diff
@@ -994,98 +1108,394 @@ ShaderBinding* generate_shaders(const ShaderState state)
...
+{
+    return g_strdup_printf("%s/shader_cache_list", xemu_settings_get_base_path());
+}
+
+static void shader_write_lru_list_entry_to_disk(Lru *lru, LruNode *node, void *opaque)
+{
+    FILE *lru_list_file = (FILE*) opaque;
+    size_t written = fwrite(&node->hash, sizeof(uint64_t), 1, lru_list_file);
+    if (written != 1) {
+        fprintf(stderr, "nv2a: Failed to write shader list entry %llu to disk\n",
```
mborgerson: Hashes are identified by 64b hex number so let's printf `%lx` here instead

hw/xbox/nv2a/shaders.c
```diff
@@ -994,98 +1108,394 @@ ShaderBinding* generate_shaders(const ShaderState state)
...
+
+static char *shader_get_binary_path(uint64_t hash)
+{
+    const char *cfg_dir = xemu_settings_get_base_path();
+    uint64_t bin_mask = 0xffffUL << 48;
+    char *shader_bin_dir = g_strdup_printf("%s/shaders/%04lx",
+                                           cfg_dir, (hash & bin_mask) >> 48);
+    char *shader_bin_path = g_strdup_printf("%s/%012lx", shader_bin_dir,
+                                            hash & (~bin_mask));
+    qemu_mkdir(shader_bin_dir);
```
mborgerson: I think we should not *create* directories in this function, rather only create the directory when writing the shaders out to disk if it doesn't exist yet

hw/xbox/nv2a/shaders.c
```diff
@@ -994,98 +1108,394 @@ ShaderBinding* generate_shaders(const ShaderState state)
...
+{
+    if (!snode->binding || snode->cached) {
+        return;
+    }
+
+    GLint program_size;
+    glGetProgramiv(snode->binding->gl_program, GL_PROGRAM_BINARY_LENGTH, &program_size);
+    
+    if (snode->program) {
+        g_free(snode->program);
```
mborgerson, line 1574: We should set `snode->program = NULL;` here?

hw/xbox/nv2a/shaders.c
```diff
@@ -994,98 +1108,394 @@ ShaderBinding* generate_shaders(const ShaderState state)
...
+    GLsizei program_size_copied;
+    glGetProgramBinary(snode->binding->gl_program, program_size, &program_size_copied,
+                       &snode->program_format, snode->program);
+    assert(glGetError() == GL_NO_ERROR);
+
+    snode->program_size = program_size_copied;
+    snode->cached = true;
+
+    char name[27];
+    snprintf(name, 27, "scache-%llu", (unsigned long long) snode->node.hash);
```
mborgerson: * Possible truncation (name should be 28 to fit maximal hash value and null byte)
* With hex representation it would be 24

hw/xbox/nv2a/shaders.c
```diff
@@ -994,98 +1108,394 @@ ShaderBinding* generate_shaders(const ShaderState state)
...
-    for (i = 1; i < NV2A_MAX_TEXTURES; i++) {
-        snprintf(tmp, sizeof(tmp), "bumpMat%d", i);
-        ret->bump_mat_loc[i] = glGetUniformLocation(program, tmp);
-        snprintf(tmp, sizeof(tmp), "bumpScale%d", i);
-        ret->bump_scale_loc[i] = glGetUniformLocation(program, tmp);
-        snprintf(tmp, sizeof(tmp), "bumpOffset%d", i);
-        ret->bump_offset_loc[i] = glGetUniformLocation(program, tmp);
+    qemu_mutex_unlock(&pg->shader_cache_lock);
+
+    FILE *shader_file = qemu_fopen(shader_path, "rb");
```
mborgerson, line 1336: Another latency optimization for this system would be combining all of the shaders into one big file to reduce startup disk I/O overhead. Could wait for a follow-up, just noting it here for future iteration.

hw/xbox/nv2a/shaders.c
```diff
@@ -994,98 +1108,394 @@ ShaderBinding* generate_shaders(const ShaderState state)
...
+    lru_visit_active(&pg->shader_cache, shader_write_lru_list_entry_to_disk, lru_list);
+    fclose(lru_list);
+
+    lru_flush(&pg->shader_cache);
+
+    qatomic_set(&pg->shader_cache_writeback_pending, false);
+    qemu_event_set(&pg->shader_cache_writeback_complete);
+}
+
+bool shader_load_from_memory(ShaderLruNode *snode)
```
mborgerson, line 1253: I took a brief look at the performance impact of loading and validating shaders at runtime in the way this implementation does by instrumenting this function to print out some experimental measurements. Playing around in Halo 2 with a 'warm' cache, the per-frame latency added here if the shaders for that frame are pre-loaded from disk but not yet created in GL are: min=284us median=1242us, max=29797us. Even with the savings of having the shader program 'binary' in memory, it looks like loading and validating can still add some significant latency.

I augmented the patch to create all the shader programs up front and eliminate this runtime cost. About 13ms to load ~615 shaders from disk, and 333ms to load them into GL state. I'm not certain what the impact will be after long-term practical user play, but scaling this figure out to the current shader cache size of `50*1024` suggests a possible startup cost of something like `(346/615)*(50*1024)=28805ms`. Intermittent frame latency is not great, but nearly 30s of startup latency is also not a good user experience. The cache size could be reduced to limit the up-front loading cost, but even at a tenth of the size, 2.8 seconds would be noticeable and annoying startup delay.

A compromise between the full-lazy and full-upfront approaches would be loading these shaders at startup, but in a separate context; that way we would not block startup, and be able to reduce some latency compared to the full-lazy approach.

Anyway, I think the approach in this patch is good as it is, and we can defer this proposed improvement to a follow-up.

hw/xbox/nv2a/shaders.c
```diff
@@ -994,98 +1108,394 @@ ShaderBinding* generate_shaders(const ShaderState state)
...
+}
+
+void shader_cache_init(PGRAPHState *pg)
+{
+    if (!shader_gl_vendor) {
+        shader_gl_vendor = (const char *) glGetString(GL_VENDOR);
+    }
+
+    shader_create_cache_folder();
+    const size_t shader_cache_size = 50*1024;
```
mborgerson, line 1482: I don't really know what the right number would be for this. We may want to make this configurable in the future. We can leave it hard-coded for now.

hw/xbox/nv2a/shaders.c
```diff
@@ -87,17 +120,18 @@ static MString* generate_geometry_shader(
         return NULL;
     }
 
+    *gl_primitive_mode = get_gl_primitive_mode(polygon_mode, primitive_mode);
```
mborgerson: Let's move this function up so we only set `gl_primitive_mode` the one time

hw/xbox/nv2a/shaders.c
antangelo: Done

hw/xbox/nv2a/shaders.c
antangelo: Done, moved to before the shader file is opened.

hw/xbox/nv2a/shaders.c
antangelo: Yes, thanks for the catch. I believe this can only be an issue in the edge case below where the platform doesn't support shader binaries (otherwise we set it almost immediately afterwards anyway).

hw/xbox/nv2a/shaders.c
antangelo: Done, switched to hex and adjusted size accordingly

hw/xbox/nv2a/shaders.c
antangelo: Unless I'm mistaken, this would necessitate jumping through the file to determine the individual shader offsets (as the binary sizes are not constant), before the loading work can be parallelized. Would be interesting to see if this diminishes the IO performance benefits

hw/xbox/nv2a/shaders.c
antangelo: Gotcha, I've added a fixme comment above to indicate this

hw/xbox/nv2a/shaders.c
antangelo: Done

Adamillo:
@mborgerson are there any other changes required in order for this PR to get merged?

Spidy123222:
From testing it seems I have some issues with burnout 3 with this PR. 

before restart you can see it has a black blob on the streets on the texturing.
https://cdn.discordapp.com/attachments/680488810429153315/1017332510566404136/Screenshot_300.png

After a xemu restart it is much more clearly green on the texture ahead.
https://cdn.discordapp.com/attachments/680488810429153315/1017332528065019914/Screenshot_301.png

Master branch has not had this issue as i have tested multiple versions of master branch and here is example versions 0.7.67, 0.7.66, 0.7.64, 0.6.2-90.
https://cdn.discordapp.com/attachments/680488810429153315/1017332560491196456/Screenshot_302.png

Hope this info is helpful to you

mborgerson:
@Spidy123222 

> From testing it seems I have some issues with burnout 3 with this PR.

I recall these issues being fixed on master branch previously. I've rebased this PR now, can you test again and confirm whether the issues are still present or resolved?

Spidy123222:
> @Spidy123222
> 
> > From testing it seems I have some issues with burnout 3 with this PR.
> 
> I recall these issues being fixed on master branch previously. I've rebased this PR now, can you test again and confirm whether the issues are still present or resolved?

I don’t see any problems after rebasing. It helps some performance on some games and load times.

mborgerson:
Nice job!

PR #770, opened 2022-03-15
nv2a: Implement v1r0_sum flags

### LINKED ISSUE #765: Handling of final combiner flags is silently unimplemented
### Bug Description

While investigating the shimmer effect in #270 I noticed that Ninja Gaiden Black sets the `PS_FINALCOMBINERSETTING_CLAMP_SUM` flag in the bugged render pass (it doesn't seem to actually use the V1R0 sum so I don't think it's directly causing the issue here).

Looking at the pixel shader generator code, I see that the [flags are set](https://github.com/mborgerson/xemu/blob/master/hw/xbox/nv2a/psh.c#L1020), but they don't seem to actually be applied anywhere (there are no references to any of those flags other than this initial assignment in `psh_translate`:

```
        ps.final_input.clamp_sum = flags & PS_FINALCOMBINERSETTING_CLAMP_SUM;
        ps.final_input.inv_v1 = flags & PS_FINALCOMBINERSETTING_COMPLEMENT_V1;
        ps.final_input.inv_r0 = flags & PS_FINALCOMBINERSETTING_COMPLEMENT_R0;
```

### Expected Behavior

The flags should be implemented to match HW behavior.

Ninja Gaiden Black does make use of the `PS_REGISTER_V1R0_SUM` register for several other draws in that same frame (which was captured when entering the magma caverns in chapter 12).

The clamp flag also set during the bootup sequence, though I didn't capture the pgraph logs to see if the sum register is actually used in the pixel shader.

### xemu Version

My near head development version (I'm confident that I have made no changes that would've removed acting on the flags)

Version: 0.6.2-109-ga16eea7d86
Branch:  work
Commit:  a16eea7d86a1cbe5f6c9cfb6b076839abd64ea7d
Date:    Sat Mar 12 06:38:59 AM UTC 2022

### System Information

  "cpu": "Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz",
  "gl_renderer": "NVIDIA GeForce GTX 1070/PCIe/SSE2",
  "gl_shading_language_version": "4.00 NVIDIA via Cg compiler",
  "gl_vendor": "NVIDIA Corporation",
  "gl_version": "4.0.0 NVIDIA 470.103.01",
  "os_platform": "Linux",
  "os_version": "Ubuntu 21.10",

### Additional Context

_No response_

abaire:
Fixes #765

### DIFF
--- a/hw/xbox/nv2a/psh.c
+++ b/hw/xbox/nv2a/psh.c
@@ -290,7 +290,17 @@ static MString* get_var(struct PixelShader *ps, int reg, bool is_dest)
         return mstring_from_str("r1");
     case PS_REGISTER_V1R0_SUM:
         add_var_ref(ps, "r0");
-        return mstring_from_str("vec4(v1.rgb + r0.rgb, 0.0)");
+        if (ps->final_input.clamp_sum) {
+            return mstring_from_fmt(
+                    "clamp(vec4(%s.rgb + %s.rgb, 0.0), 0.0, 1.0)",
+                    ps->final_input.inv_v1 ? "(1.0 - v1)" : "v1",
+                    ps->final_input.inv_r0 ? "(1.0 - r0)" : "r0");
+        } else {
+            return mstring_from_fmt(
+                    "vec4(%s.rgb + %s.rgb, 0.0)",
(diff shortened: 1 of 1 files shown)


mborgerson:
Thanks!

PR #773, opened 2022-03-15
ui: Save screen size on resize

GXTX:
#437 appears stalled and has design issues imo so I went ahead and rewrote it.

I tried to cherry-pick the base commit on his branch but couldn't do it cleanly. Hopefully @7oxicshadow doesn't mind i've added him as a co-author.

Thinking, now that users have the ability to resize and save it, would we be willing to have the base resolution be 640x480 which is the 'normal' Xbox size?

### DIFF
--- a/ui/xemu.c
+++ b/ui/xemu.c
@@ -591,6 +591,11 @@ static void handle_windowevent(SDL_Event *ev)
             info.width = ev->window.data1;
             info.height = ev->window.data2;
             dpy_set_ui_info(scon->dcl.con, &info);
+
+            if (!gui_fullscreen) {
+                xemu_settings_set_int(XEMU_SETTINGS_DISPLAY_WINDOW_WIDTH, ev->window.data1);
+                xemu_settings_set_int(XEMU_SETTINGS_DISPLAY_WINDOW_HEIGHT, ev->window.data2);
+            }
         }
         sdl2_redraw(scon);
         break;
@@ -843,7 +848,7 @@ static void sdl2_display_very_early_init(DisplayOptions *o)
 
     // Create main window
     m_window = SDL_CreateWindow(
-        title, SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED, 1280, 960,
+        title, SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED, 640, 480,
         SDL_WINDOW_OPENGL | SDL_WINDOW_RESIZABLE | SDL_WINDOW_ALLOW_HIGHDPI);
     if (m_window == NULL) {
         fprintf(stderr, "Failed to create main window\n");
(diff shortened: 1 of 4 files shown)


ui/xemu.c
```diff
@@ -591,6 +591,12 @@ static void handle_windowevent(SDL_Event *ev)
             info.width = ev->window.data1;
             info.height = ev->window.data2;
             dpy_set_ui_info(scon->dcl.con, &info);
+
+            if (!gui_fullscreen) {
+                xemu_settings_set_int(XEMU_SETTINGS_DISPLAY_WINDOW_WIDTH, ev->window.data1);
+                xemu_settings_set_int(XEMU_SETTINGS_DISPLAY_WINDOW_HEIGHT, ev->window.data2);
+                xemu_settings_save();
```
mborgerson: Saving writes the settings out to disk, and during a resize this can be called many times as the user drags the window to the preferred size. How about we register an exit handler with `atexit` to simply save the config when xemu exits

ui/xemu.c
```diff
@@ -841,9 +847,20 @@ static void sdl2_display_very_early_init(DisplayOptions *o)
...
+    int win_w;
+    int win_h;
+
+    xemu_settings_get_int(XEMU_SETTINGS_DISPLAY_WINDOW_WIDTH, &win_w);
+    xemu_settings_get_int(XEMU_SETTINGS_DISPLAY_WINDOW_HEIGHT, &win_h);
+
+    if (win_w <= 0 || win_h <= 0) {
+        win_w = 1280;
+        win_h = 960;
+    }
```
mborgerson: While we are here, can we add a check that queries screen size to make sure we do not create a window larger than the screen size

ui/xemu.c
GXTX: Fixed

ui/xemu.c
GXTX: We can't get the current display params without knowing the display index. To do this we need to create the window first. The solution I have creates the window then resizes if appropriate. It's possible there's a SDL env variable to get the default display. Pushed the changes.

ui/xemu.c
```diff
@@ -852,6 +857,22 @@ static void sdl2_display_very_early_init(DisplayOptions *o)
     }
     g_free(title);
 
+    SDL_DisplayMode disp_mode;
+    if(SDL_GetCurrentDisplayMode(SDL_GetWindowDisplayIndex(m_window), &disp_mode)) {
+        fprintf(stderr, "Failed to query display size properties\n");
+        SDL_Quit();
```
mborgerson: I don't think we need to exit in this case

ui/xemu.c
```diff
@@ -852,6 +857,22 @@ static void sdl2_display_very_early_init(DisplayOptions *o)
...
+        exit(1);
+    }
+
+    int win_w, win_h;
+    xemu_settings_get_int(XEMU_SETTINGS_DISPLAY_WINDOW_WIDTH, &win_w);
+    xemu_settings_get_int(XEMU_SETTINGS_DISPLAY_WINDOW_HEIGHT, &win_h);
+
+    if (disp_mode.w >= win_w || disp_mode.h >= win_h) {
```
mborgerson: This can create a window too large for the screen. Let's check that the desired window size is smaller than the screen in both dimensions

ui/xemu.c
```diff
@@ -852,6 +857,22 @@ static void sdl2_display_very_early_init(DisplayOptions *o)
...
 
+    SDL_DisplayMode disp_mode;
+    if(SDL_GetCurrentDisplayMode(SDL_GetWindowDisplayIndex(m_window), &disp_mode)) {
+        fprintf(stderr, "Failed to query display size properties\n");
+        SDL_Quit();
+        exit(1);
+    }
+
+    int win_w, win_h;
+    xemu_settings_get_int(XEMU_SETTINGS_DISPLAY_WINDOW_WIDTH, &win_w);
```
mborgerson, line 864: Should check values are >0

ui/xemu.c
GXTX: 👍

ui/xemu.c
GXTX: 👍

ui/xemu.c
GXTX: 👍

mborgerson:
Nice, thanks!

PR #782, opened 2022-03-17
ui: Add snapshot management UI

antangelo:
Enables the snapshots menu in the new UI, complete with thumbnails, game title information, searching, and the ability to bind snapshots to F5-F8 keys for ease of use.

For anyone who wishes to test this before merge, keep in mind that snapshots created or saved with this branch are not backwards compatible with previous xemu versions. That is, those snapshots will not work on any other version of xemu. Snapshots created on previous or current xemu versions will work fine with this branch, however.

https://user-images.githubusercontent.com/25046221/168232353-fe025b84-39c4-4621-bf7b-60ed884a9025.mp4

### DIFF
--- a/block/file-win32.c
+++ b/block/file-win32.c
@@ -36,6 +36,7 @@
 #include "qapi/qmp/qstring.h"
 #include <windows.h>
 #include <winioctl.h>
+#include <assert.h>
 
 #define FTYPE_FILE 0
 #define FTYPE_CD     1
@@ -341,6 +342,9 @@ static int raw_open(BlockDriverState *bs, QDict *options, int flags,
     bool use_aio;
     OnOffAuto locking;
     int ret;
+#ifdef XBOX
+    int sharing_flags;
+#endif
 
     s->type = FTYPE_FILE;
 
@@ -396,9 +400,21 @@ static int raw_open(BlockDriverState *bs, QDict *options, int flags,
     if (!filename) {
         goto fail;
     }
+
+#ifdef XBOX
+    sharing_flags = FILE_SHARE_READ;
+    if (flags & BDRV_O_RO_WRITE_SHARE) {
+        assert(access_flags == GENERIC_READ);
+        sharing_flags = FILE_SHARE_READ | FILE_SHARE_WRITE;
+    }
+    s->hfile = CreateFileW(wfilename, access_flags,
+                          sharing_flags, NULL,
+                          OPEN_EXISTING, overlapped, NULL);
+#else
     s->hfile = CreateFileW(wfilename, access_flags,
                           FILE_SHARE_READ, NULL,
                           OPEN_EXISTING, overlapped, NULL);
+#endif
     g_free(wfilename);
     if (s->hfile == INVALID_HANDLE_VALUE) {
         int err = GetLastError();
--- a/ui/xemu-snapshots.c
+++ b/ui/xemu-snapshots.c
@@ -1,0 +1,319 @@
+/*
+ * xemu User Interface
+ *
+ * Copyright (C) 2020-2022 Matt Borgerson
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "xemu-snapshots.h"
+#include "xemu-settings.h"
+#include "xemu-xbe.h"
+
+#include <SDL2/SDL.h>
+#include <epoxy/gl.h>
+
+#include "block/aio.h"
+#include "block/block_int.h"
+#include "block/qapi.h"
+#include "block/qdict.h"
+#include "migration/qemu-file.h"
+#include "migration/snapshot.h"
+#include "qapi/error.h"
+#include "sysemu/runstate.h"
+#include "qemu-common.h"
+
+#include "ui/console.h"
+#include "ui/input.h"
+
+static QEMUSnapshotInfo *xemu_snapshots_metadata = NULL;
+static XemuSnapshotData *xemu_snapshots_extra_data = NULL;
+static int xemu_snapshots_len = 0;
+static bool xemu_snapshots_dirty = true;
+
+const char **g_snapshot_shortcut_index_key_map[] = {
+    &g_config.general.snapshots.shortcuts.f5,
+    &g_config.general.snapshots.shortcuts.f6,
+    &g_config.general.snapshots.shortcuts.f7,
+    &g_config.general.snapshots.shortcuts.f8,
+};
+
+static bool xemu_snapshots_load_thumbnail(BlockDriverState *bs_ro,
+                                          XemuSnapshotData *data,
+                                          int64_t *offset)
+{
+    int res = bdrv_load_vmstate(bs_ro, (uint8_t *)&data->thumbnail, *offset,
+                                sizeof(TextureBuffer) -
+                                    sizeof(data->thumbnail.buffer));
+    if (res != sizeof(TextureBuffer) - sizeof(data->thumbnail.buffer))
+        return false;
+    *offset += res;
+
+    data->thumbnail.buffer = g_malloc(data->thumbnail.size);
+
+    res = bdrv_load_vmstate(bs_ro, (uint8_t *)data->thumbnail.buffer, *offset,
+                            data->thumbnail.size);
+    if (res != data->thumbnail.size) {
+        return false;
+    }
+    *offset += res;
+
+    return true;
+}
+
+static void xemu_snapshots_load_data(BlockDriverState *bs_ro,
+                                     QEMUSnapshotInfo *info,
+                                     XemuSnapshotData *data, Error **err)
+{
+    int res;
+    XemuSnapshotHeader header;
+    int64_t offset = 0;
+
+    data->xbe_title_present = false;
+    data->thumbnail_present = false;
+    res = bdrv_snapshot_load_tmp(bs_ro, info->id_str, info->name, err);
+    if (res < 0)
+        return;
+
+    res = bdrv_load_vmstate(bs_ro, (uint8_t *)&header, offset, sizeof(header));
+    if (res != sizeof(header))
+        goto error;
+    offset += res;
+
+    if (header.magic != XEMU_SNAPSHOT_DATA_MAGIC)
+        goto error;
+
+    res = bdrv_load_vmstate(bs_ro, (uint8_t *)&data->xbe_title_len, offset,
+                            sizeof(data->xbe_title_len));
+    if (res != sizeof(data->xbe_title_len))
+        goto error;
+    offset += res;
+
+    data->xbe_title = (char *)g_malloc(data->xbe_title_len);
+
+    res = bdrv_load_vmstate(bs_ro, (uint8_t *)data->xbe_title, offset,
+                            data->xbe_title_len);
+    if (res != data->xbe_title_len)
+        goto error;
+    offset += res;
+
+    data->xbe_title_present = (offset <= sizeof(header) + header.size);
+
+    if (offset == sizeof(header) + header.size)
+        return;
+
+    if (!xemu_snapshots_load_thumbnail(bs_ro, data, &offset)) {
+        goto error;
+    }
+
+    data->thumbnail_present = (offset <= sizeof(header) + header.size);
+
+    if (data->thumbnail_present) {
+        glGenTextures(1, &data->gl_thumbnail);
+        xemu_snapshots_render_thumbnail(data->gl_thumbnail, &data->thumbnail);
+    }
+    return;
+
+error:
+    g_free(data->xbe_title);
+    data->xbe_title_present = false;
+
+    g_free(data->thumbnail.buffer);
+    data->thumbnail_present = false;
+}
+
+static void xemu_snapshots_all_load_data(QEMUSnapshotInfo **info,
+                                         XemuSnapshotData **data,
+                                         int snapshots_len, Error **err)
+{
+    BlockDriverState *bs_ro;
+    QDict *opts = qdict_new();
+
+    assert(info && data);
+
+    if (*data) {
+        for (int i = 0; i < xemu_snapshots_len; ++i) {
+            if ((*data)[i].xbe_title_present) {
+                g_free((*data)[i].xbe_title);
+            }
+
+            if ((*data)[i].thumbnail_present) {
+                g_free((*data)[i].thumbnail.buffer);
+                glDeleteTextures(1, &((*data)[i].gl_thumbnail));
+            }
+        }
+        g_free(*data);
+    }
+
+    *data =
+        (XemuSnapshotData *)g_malloc(sizeof(XemuSnapshotData) * snapshots_len);
+    memset(*data, 0, sizeof(XemuSnapshotData) * snapshots_len);
+
+    qdict_put_bool(opts, BDRV_OPT_READ_ONLY, true);
+    bs_ro = bdrv_open(g_config.sys.files.hdd_path, NULL, opts,
+                      BDRV_O_RO_WRITE_SHARE | BDRV_O_AUTO_RDONLY, err);
+    if (!bs_ro) {
+        return;
+    }
+
+    for (int i = 0; i < snapshots_len; ++i) {
+        xemu_snapshots_load_data(bs_ro, (*info) + i, (*data) + i, err);
+        if (*err) {
+            break;
+        }
+    }
+
+    bdrv_flush(bs_ro);
+    bdrv_drain(bs_ro);
+    bdrv_unref(bs_ro);
+    assert(bs_ro->refcnt == 0);
+    if (!(*err))
+        xemu_snapshots_dirty = false;
+}
+
+int xemu_snapshots_list(QEMUSnapshotInfo **info, XemuSnapshotData **extra_data,
+                        Error **err)
+{
+    BlockDriverState *bs;
+    AioContext *aio_context;
+    int snapshots_len;
+    assert(err);
+
+    if (!xemu_snapshots_dirty && xemu_snapshots_extra_data &&
+        xemu_snapshots_metadata) {
+        goto done;
+    }
+
+    if (xemu_snapshots_metadata)
+        g_free(xemu_snapshots_metadata);
+
+    bs = bdrv_all_find_vmstate_bs(NULL, false, NULL, err);
+    if (!bs) {
+        return -1;
+    }
+
+    aio_context = bdrv_get_aio_context(bs);
+
+    aio_context_acquire(aio_context);
+    snapshots_len = bdrv_snapshot_list(bs, &xemu_snapshots_metadata);
+    aio_context_release(aio_context);
+    xemu_snapshots_all_load_data(&xemu_snapshots_metadata,
+                                 &xemu_snapshots_extra_data, snapshots_len,
+                                 err);
+    if (*err) {
+        return -1;
+    }
+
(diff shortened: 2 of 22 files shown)


ui/xemu-hud.cc
```diff
@@ -2063,6 +2071,30 @@ static void process_keyboard_shortcuts(void)
         monitor_window.toggle_open();
     }
 
+    for (int fkey = 0; fkey < 12; fkey++) {
```
abaire: Can we reserve F9-F12 for dev tools? Renderdoc uses F12, nSight uses F11, and the xemu deeper renderdoc integration (just below this block) uses F10.

antangelo: Good point, completely forgot about those. I'll reserve them in the next iteration.

ui/xui/menubar.cc
```diff
@@ -101,6 +101,7 @@ void ShowMainMenu()
             if (ImGui::MenuItem(" Display")) g_main_menu.ShowDisplay();
             if (ImGui::MenuItem(" Audio")) g_main_menu.ShowAudio();
             if (ImGui::MenuItem(" Network")) g_main_menu.ShowNetwork();
+            if (ImGui::MenuItem(" Snapshots")) g_main_menu.ShowSnapshots();
```
mborgerson, line 147: * We should have a menu item here to take a snapshot, I think below Screenshot
* Similarly, we need two new entries in the popup menu for `Save Snapshot` (ICON_FA_DOWNLOAD), and `Snapshots...` (ICON_FA_CLOCK_ROTATE_LEFT)

ui/xemu-snapshots.c
```diff
@@ -0,0 +1,342 @@
...
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAX_LEVEL, 0);
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_LINEAR);
+
+    glTexImage2D(GL_TEXTURE_2D, 0, GL_RGBA, thumbnail->width,
+                 thumbnail->height, 0, thumbnail->format, thumbnail->type,
+                 thumbnail->buffer);
+    glBindTexture(GL_TEXTURE_2D, tex);
+}
+
+static TextureBuffer *xemu_snapshots_make_thumbnail(void)
```
mborgerson: A few issues:
* `glGetTextureImage` is only available on GL 4.5+, but xemu targets GL 4.0 today, so unfortunately we cannot use this function (some drivers like macOS just don't support it, so this will cause a crash in those cases).
* This does not resize the framebuffer, nor is any image compression used. At a baseline resolution of 640x480x4 this is 1,200KiB. At typical 3x surface scaling this little thumbnail will be a chonky ~10.5MiB.
* Does not handle aspect ratio / gamma correction / alpha channel also so it may not look like what you expect to see in all circumstances

The good news is that I have already solved these problems for the screenshot feature and it will just need a minor edit to be usable for this purpose too. Let's update the common code to support both use cases and avoid duplication on it

block/file-win32.c
```diff
@@ -396,9 +400,21 @@ static int raw_open(BlockDriverState *bs, QDict *options, int flags,
     if (!filename) {
         goto fail;
     }
+
+#ifdef XBOX
+    sharing_flags = FILE_SHARE_READ;
+    if (flags & BDRV_O_RO_WRITE_SHARE) {
+        assert(access_flags == GENERIC_READ);
+        sharing_flags = FILE_SHARE_READ | FILE_SHARE_WRITE;
```
mborgerson, line 408: Do we need to permit write access for subsequent processes? I'm concerned this will allow multiple instances of xemu to corrupt a single HDD image

ui/xui/main-menu.cc
```diff
@@ -634,115 +636,349 @@ void MainMenuNetworkView::DrawUdpOptions(bool appearing)
...
+}
+
+void MainMenuSnapshotsView::Draw()
+{
+    Load();
+    SectionTitle("Snapshots");
+    ImGui::Checkbox("Filter by Current Game", filter_current_game);
+
+    ImGui::SameLine();
+    ImGui::InputText("Search", search_buf, 40, ImGuiInputTextFlags_CallbackEdit, &MainMenuSnapshotsViewUpdateSearchBox, this);
```
mborgerson: Some usability thoughts from my experience trying this version out.

On filtering + creating snapshot:
* It's not clear that this is a text box that both filters snapshots by name, and is also the name entry for new snapshots. Maybe we can use the `InputTextWithHint` widget instead and show a <kbd>Filter by name...</kbd> hint, or something
* It's not clear that you need to enter some text in the name filter box AND have a non-exact name match for the 'Create Snapshot' button appear.
* When 'Create Snapshot' button does appear, it is at the end of the list. This would be fine but if the name is too generic and many snapshots are shown, the create button will not be visible unless the user scrolls all the way down to reveal it. Moreover, the distance from the input box to the create button is big which is a bit uncomfortable.
* Maybe the solution for named snapshots is to have a separate input box (from filtering) for new snapshot name, and have the create button always visible.

On ordering:
* I think it might be preferable to have snapshots displayed in reverse-chronological order, with most recent snapshots at top and oldest snapshots at bottom, as opposed to currently chronological order. Personally I access recently saved snapshots more frequently, so having them in the viewport by default (at top) makes sense to me.
* Then maybe the create button at the top would also make sense, maybe as a placeholder item in the snapshot list with the name entry box where the name would be displayed for instance

On key binding:
* The function key binding stuff is not apparent. We should provide some hint to users that they can bind to a function key by holding the key and selecting the snapshot. Once they press the key, we should instruct them to click on a snapshot item to bind it.
* Maybe a subtle <kbd>F5</kbd> (with outline around the text) or <kbd>Not Bound</kbd> would be preferable to the "Bound to F5" text
* There's also no way to know that mod+fn key overwrites a snapshot.

ui/xui/main-menu.hh
```diff
@@ -102,9 +103,28 @@ public:
...
 {
+protected:
+    QEMUSnapshotInfo *snapshots;
+    XemuSnapshotData *extra_data;
+    int snapshots_len, prev_snapshots_len;
+    GLuint *thumbnails;
+    bool *filter_current_game;
+    uint32_t current_title_id;
+    char *current_title_name;
+    char search_buf[40];
```
mborgerson: This could be made to use std::string instead of a hardcoded buffer size

ui/xui/main-menu.cc
```diff
@@ -634,115 +636,349 @@ void MainMenuNetworkView::DrawUdpOptions(bool appearing)
...
+            }
+        }
+
+        search_buf_equal |= g_strcmp0(search_buf, snapshots[i].name) == 0;
+
+        ImGui::PushID(i);
+        GLuint thumbnail = 0;
+        if (extra_data[i].thumbnail_present) {
+            thumbnail = thumbnails[i];
+            xemu_snapshots_render_thumbnail(thumbnail, &extra_data[i].thumbnail);
```
mborgerson: This is uploading every thumbnail on every frame with potentially very big thumbnails, but snapshot thumbnails should not change on every frame so we can just upload them once and refresh when necessary.

e.g. you can easily kill xemu performance by setting higher resolution scale, creating some snapshots, watch performance suffer, then delete the snapshots and see it shoot back up

ui/xui/main-menu.hh
```diff
@@ -102,9 +103,28 @@ public:
 
 class MainMenuSnapshotsView : public virtual MainMenuTabView
 {
+protected:
+    QEMUSnapshotInfo *snapshots;
+    XemuSnapshotData *extra_data;
+    int snapshots_len, prev_snapshots_len;
```
mborgerson: * I don't think `prev_snapshots_len` needs to be a member variable
* Maybe it would be better to refactor these data into a vector of objects if they are closely related
* Member variable naming convention in this file is `m_` prefix

ui/xui/main-menu.hh
```diff
@@ -102,9 +103,28 @@ public:
...
+    GLuint *thumbnails;
+    bool *filter_current_game;
```
mborgerson: Don't need an extra pointer if you're just setting a config variable with it, same for `snapshot_binding` below

ui/xui/main-menu.cc
```diff
@@ -634,115 +636,349 @@ void MainMenuNetworkView::DrawUdpOptions(bool appearing)
...
+
+    // Snapshot thumbnail
+    ImGui::SetItemAllowOverlap();
+    ImGui::SameLine();
+    ImGui::SetCursorPosX(pos.x + thumbnail_pos.x);
+    ImGui::SetCursorPosY(pos.y + thumbnail_pos.y);
+    if (screenshot > 0) {
+        ImGui::Image((ImTextureID)(uint64_t)screenshot, thumbnail_size, ImVec2(0,0), ImVec2(1,1));
+    } else {
+        ImGui::Dummy(thumbnail_size);
```
mborgerson: A placeholder image would be preferable

ui/xemu-snapshots.c
```diff
@@ -0,0 +1,342 @@
...
+{
+    /* 
+     * Avoids crashing if a snapshot is made on a thread with no GL context
```
mborgerson: We should be able to dispatch a request for the snapshot thumbnail without needing a context on this thread to support this case

ui/xui/main-menu.cc
```diff
@@ -634,115 +636,349 @@ void MainMenuNetworkView::DrawUdpOptions(bool appearing)
...
+        xemu_queue_error_message(error_get_pretty(err));
+        error_free(err);
+    }
+}
+
+void MainMenuSnapshotsView::Draw()
+{
+    Load();
+    SectionTitle("Snapshots");
+    ImGui::Checkbox("Filter by Current Game", filter_current_game);
```
mborgerson: In Xbox parlance, games are called 'titles', and the sentence should be all lower-case except for the first letter. This is the verbiage also used on the compatibility reporter

ui/xui/menubar.cc
antangelo: Done

ui/xemu-snapshots.c
antangelo: Done

ui/xui/main-menu.cc
antangelo: - Separated filtering and snapshot creation into separate textboxes
- Snapshots are now ordered in reverse-chronological order
- I overhauled the key binding system.
  - There is now an explicit keybinding button on each snapshot for selecting which key to bind to (or unbind)
  - There are corresponding entries for each key in the menubar (similar to how other emulators will have "Save state n" and "Load state n" menu buttons), these also have hints for which keyboard keys perform the same actions

ui/xui/main-menu.hh
antangelo: Done

ui/xui/main-menu.cc
antangelo: Done, will now refresh only when the snapshot list changes

ui/xui/main-menu.hh
antangelo: Done

ui/xui/main-menu.hh
antangelo: Done

ui/xui/main-menu.cc
antangelo: Done, using xemu logo as the placeholder

ui/xui/main-menu.cc

ui/xemu-snapshots.c
antangelo: I haven't come across a method that actually exercises this case in practice (External QAPI maybe?). I put this here mostly in the event that someone does manage to trigger snapshot creation on a thread with no GL context, since crashing here will result in the loss of both the old snapshot (in case it is being overwritten) as well as the new.

However, in practice the vast majority (if not all) of snapshots will be created by the UI thread (either through the new management UI, keybinds, or through the monitor), so I'm not sure whether the added complexity of being able to dispatch a request to the UI thread for this is worth it at this time.

block/file-win32.c
antangelo: The behavior here is a bit unintuitive. The snapshot UI must create a new read-only file handle in order to be able to open snapshots for reading (read-only is a hard requirement within QEMU block drivers, so the current file handle cannot be reused). The RO handle must coexist with the RW handle that is already open, and so it must have write sharing permissions to do so. The RW handle already has read sharing permissions, satisfying the sharing permissions on that side.

I've verified that you cannot open a second xemu instance on Windows, both before, after, and while the snapshots window is open. The first RW handle, given that it only allows read sharing, will not permit any subsequent RW handles to be open (and it already allows RO handles to be opened on the current master branch).

ui/xemu-snapshots.c
```diff
@@ -0,0 +1,317 @@
...
+{
+    BlockDriverState *bs_ro;
+    QDict *opts = qdict_new();
+
+    assert(info && data);
+
+    if (*data) {
+        for (int i = 0; i < xemu_snapshots_len; ++i) {
+            if ((*data)[i].xbe_title_present) {
+                g_free((*data)[i].xbe_title);
```
mborgerson, line 148: Possible double free if an error occurs in `xemu_snapshots_load_data`

ui/xemu-snapshots.c
```diff
@@ -0,0 +1,317 @@
...
+    data->thumbnail_present = (offset <= sizeof(header) + header.size);
+
+    if (data->thumbnail_present) {
+        glGenTextures(1, &data->gl_thumbnail);
+        xemu_snapshots_render_thumbnail(data->gl_thumbnail, &data->thumbnail);
+    }
+    return;
+
+error:
+    g_free(data->xbe_title);
```
mborgerson, line 129: It's possible that `xbe_title` is `free`'d here, but `xbe_title_present` remains `true` which will crash when we render the title name

ui/xui/main-menu.cc
```diff
@@ -639,115 +641,357 @@ void MainMenuNetworkView::DrawUdpOptions(bool appearing)
...
+            continue;
+        }
+
+        if (m_search_regex) {
+            GMatchInfo *match;
+            bool keep_entry = false;
+        
+            g_regex_match(m_search_regex, m_snapshots[i].name, (GRegexMatchFlags)0, &match);
+            keep_entry |= g_match_info_matches(match);
+            g_free(match);
```
mborgerson: I think we are supposed to call `g_match_info_free` instead of `g_free`

ui/xemu-thumbnail.cc
```diff
@@ -0,0 +1,81 @@
...
+     * here.
+     * FIXME: Allow for dispatching a thumbnail request to the UI thread to
+     * remove this altogether.
+     */
+    if (!SDL_GL_GetCurrentContext() || display_tex == 0) {
+        return NULL;
+    }
+
+    std::vector<uint8_t> png;
+    if (!ExtractFramebufferPixels(display_tex, display_flip, png)) {
```
mborgerson: Should we resize the thumbnail?

ui/xui/main-menu.cc
```diff
@@ -639,115 +641,357 @@ void MainMenuNetworkView::DrawUdpOptions(bool appearing)
...
+    if (ImGui::Button(snapshot_with_create_name_exists ? "Save" : "Create") && !m_create_buf.empty()) {
+        xemu_snapshots_save(m_create_buf.c_str(), NULL);
+        m_create_buf.clear();
+    }
+
+    if (snapshot_with_create_name_exists && ImGui::IsItemHovered()) {
+        ImGui::SetTooltip("A snapshot with the name \"%s\" already exists. This button will overwrite the existing snapshot.", m_create_buf.c_str());
+    }
+
+    ImGui::InputTextWithHint("##search", "Filter by name...", &m_search_buf, ImGuiInputTextFlags_CallbackEdit,
```
mborgerson: Right now `Filter by current title` and `Filter by name` are separated by `Create new snapshot`. Let's put the filter options together

ui/xemu-snapshots.c
antangelo: I don't follow the condition that would lead to that. After this loop runs, `*data` is freed and then replaced with a new allocation. I've moved the `memset` out of `xemu_snapshots_load_data` and into this function so that the new region is zeroed immediately in case there's garbage data in the newly allocated buffer, but I don't see how an error in `xemu_snapshots_load_data` would otherwise cause a double free.

ui/xemu-snapshots.c
antangelo: Good catch, fixed

ui/xui/main-menu.cc
antangelo: Fixed

ui/xemu-thumbnail.cc
antangelo: I've added a scale-down factor that scales the image down by 2 and undoes the internal resolution scaling

ui/xemu-snapshots.c
mborgerson: This related to the case I mentioned below. If an error occurs in `xemu_snapshots_load_data` after `data->xbe_title_present` is set to `true`, at the end it will `g_free(data->xbe_title)`. Then when `xemu_snapshots_all_load_data` is run a second time it will free `data->xbe_title` here again

antangelo: I see, I assumed they were separate. The issue with `xemu_snapshots_load_data` should be fixed as below now.

ui/xemu-thumbnail.cc
```diff
@@ -0,0 +1,88 @@
...
+     * FIXME: Allow for dispatching a thumbnail request to the UI thread to
+     * remove this altogether.
+     */
+    if (!SDL_GL_GetCurrentContext() || display_tex == 0) {
+        return NULL;
+    }
+
+    std::vector<uint8_t> png;
+    uint32_t scale_factor = 2 * nv2a_get_surface_scale_factor();
+    if (!ExtractFramebufferPixels(display_tex, display_flip, png, scale_factor)) {
```
mborgerson: What do you think about making this so it just uses a fixed thumbnail width? We only render the thumbnail as 160px or 320px wide anyway

antangelo: I've modified the code to render at the thumbnail size at 2x UI scaling. My initial rationale behind having it be dynamic based on the display width/height was that it preserves additional quality and aspect ratio in the event that we want to change the way the thumbnail is shown in the future.
However, it's probably unlikely, and there's nothing wrong with upscaling the smaller thumbnail in that case until the snapshot is overwritten, so having it be static makes sense.

mborgerson: > preserves additional quality and aspect ratio

I do think we should preserve the aspect ratio in the thumbnail

> change the way the thumbnail is shown in the future

Yeah in the future if we want the thumbnails to have more detail (maybe for large background rendering, or something) we can just increase the thumbnail resolution. For now, smaller, fixed-size thumbnails makes most sense to me

antangelo: Make sense. I've updated the height to be computed from the aspect ratio and thumbnail width, instead of being fixed.

ui/xemu-thumbnail.cc
```diff
@@ -0,0 +1,93 @@
...
+    glBindTexture(GL_TEXTURE_2D, display_tex);
+    int thumbnail_width = XEMU_SNAPSHOT_WIDTH * 2;
+    int tex_width;
+    int tex_height;
+    glGetTexLevelParameteriv(GL_TEXTURE_2D, 0, GL_TEXTURE_WIDTH, &tex_width);
+    glGetTexLevelParameteriv(GL_TEXTURE_2D, 0, GL_TEXTURE_HEIGHT, &tex_height);
+    int thumbnail_height = (int)(((float)tex_height / (float)tex_width) * (float)thumbnail_width);
+
+    std::vector<uint8_t> png;
+    if (!ExtractFramebufferPixels(display_tex, display_flip, png, thumbnail_width, thumbnail_height)) {
```
mborgerson, line 81: A couple issues remain around resolution scaling and display modes. If resolution scaling is >1 and display mode is set to 'center', the thumbnail will be of a zoomed in portion in the center of the image. If display mode is 'Widescreen 16:9', the thumbnail is distorted.

antangelo: This seems to have been related to how `RenderFramebuffer` and `ExtractFramebufferPixels` scales the width/height depending on UI fit. I updated these functions for this case where if the width/height of the PNG are manually specified, there is no adjustment to the scaling, and it seems to work well.

antangelo:
Superseded by #1447

PR #794, opened 2022-03-19
nv2a: Rename sum to muxsum, simplify mux string build

mborgerson:


abaire (commented):
Was going to add a comment on the non-matching indentation but you already fixed it. lgtm, +1 for more consistent naming.

PR #819, opened 2022-03-21
nv2a: Move snapshot surface download into savevm hook 

### LINKED ISSUE #818: nv2a: Creating save states while the VM is paused does not preserve the paused frame
### Bug Description

Steps to reproduce:

1. Pause the VM (Ctrl + P)
2. Create a save state using `savevm` in the monitor
3. Load the same save state while still paused
4. Observe that the frame that is displayed is not the same as the frame that was shown when the save state was made

### Expected Behavior

Loading the VM should not result in the frame changing (or being incorrect)

### xemu Version

xemu_branch: master
xemu_commit: 6e1969001eda4cdedd9e880602bfbff03710914b

### System Information

OS: Arch Linux (5.16.15-arch1-1)
CPU: Intel Core i5-10400
GL_VENDOR: NVIDIA Corporation
GL_RENDERER: NVIDIA GeForce RTX 3060 Ti/PCIe/SSE2
GL_VERSION: 4.0.0 NVIDIA 510.54

### Additional Context

_No response_

antangelo:
When the vm is paused, the `nv2a_vm_state_change` function is not called when saving or loading the VM. As such, the surface download that is normally triggered here never happens. This change will perform the download when the state changes to paused to mitigate this.

Resolves #818

### DIFF
--- a/hw/xbox/nv2a/nv2a.c
+++ b/hw/xbox/nv2a/nv2a.c
@@ -21,6 +21,8 @@
 
 #include "hw/xbox/nv2a/nv2a_int.h"
 
+static NV2AState *g_nv2a;
+
 void nv2a_update_irq(NV2AState *d)
 {
     /* PFIFO */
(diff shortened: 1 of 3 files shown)


mborgerson:
Hmm..I don't think we should flush surfaces on a pause. Instead, we should just make sure they are flushed before the VM save event

antangelo:
> Hmm..I don't think we should flush surfaces on a pause. Instead, we should just make sure they are flushed before the VM save event

I moved the state handler code into its own function that gets called in savevm regardless of whether or not the normal state change handler is called. While not the cleanest solution, I think this is better than modifying qemu's state change handler behavior (and potentially introducing bugs in qemu code that expects it to be the way it is).

PR #831, opened 2022-04-06
nv2a: Prevent GPU download after blit operation

abaire:
Fixes #788 

[Test](https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/tests/texture_framebuffer_blit_tests.cpp)
[Results](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-Texture_Framebuffer_Blit)

PR #835, opened 2022-04-08
nv2a: Upload blitted surfaces used as textures

### LINKED ISSUE #786: Splinter cell double agent: steam heat waves effect not handled correctly without goggles
### Title

### Bug Description

RE-tested on master:

https://user-images.githubusercontent.com/30447649/159268322-1639ca0b-3b90-4500-859a-4273d7265235.mp4

WHere is happens:
Level name: Iceland, first part, second part
kinshasa, second part beginning
sea of okosth, first part oging through a fire camp

### Expected Behavior

Steam waves should not cause a blue screen

### xemu Version

### System Information

Manjaro Linux 
i7-10700 4.80GHz
GTX 970 STRIX 4GB VRAM

abaire:
Fixes #786 (may require #831 as well to have the full effect)

[Test](https://github.com/abaire/nxdk_pgraph_tests/blob/b64de98e2c87d8b979991d95eddb2d67e4625860/src/tests/texture_framebuffer_blit_tests.cpp#L214)
[HW Results](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-Texture_Framebuffer_Blit#fbtooldrendertargetpng)

### DIFF
--- a/hw/xbox/nv2a/pgraph.c
+++ b/hw/xbox/nv2a/pgraph.c
@@ -6327,7 +6327,11 @@ static void pgraph_bind_textures(NV2AState *d)
         SurfaceBinding *surface = pgraph_surface_get(d, texture_vram_offset);
         if (surface != NULL) {
             surf_to_tex = pgraph_check_surface_to_texture_compatibility(
-                surface, &state);
+                    surface, &state);
+
+            if (surf_to_tex && surface->upload_pending) {
+                pgraph_upload_surface_data(d, surface, false);
+            }
         }
 
         if (!surf_to_tex) {


Fabxx:
tested:
image

removes the blue screen

hw/xbox/nv2a/pgraph.c
```diff
@@ -6317,8 +6317,14 @@ static void pgraph_bind_textures(NV2AState *d)
         bool surf_to_tex = false;
         SurfaceBinding *surface = pgraph_surface_get(d, texture_vram_offset);
         if (surface != NULL) {
-            surf_to_tex = pgraph_check_surface_to_texture_compatibility(
-                surface, &state);
+            if (surface->upload_pending) {
+                // Surfaces marked as dirty due to other paths (e.g., image
+                // blits) should be treated the same as render targets.
+                pgraph_upload_surface_data(d, surface, false);
```
mborgerson, line 6333: This forces a surface upload now AND skips the surface-to-texture check, resulting in two uploads of the same data here, one for the surface, and another for the texture. I think either we should:
* Do the s2t check first, and if it is `true`, then do the surface upload to let the texture be created from s2t and save an upload assuming the surface is going to be used again, or
* Skip the check if an upload is pending, let the surface be updated later, and just update the texture from RAM

Probably the former is the better way to go

abaire: Changed.

To make sure I follow (sorry, still getting a handle on this part of the code):

This will now continue to do the RAM->GPU path for the surface, but will end up following the fast path to duplicate the surface GL binding to the texture GL binding around line 6413 below.

mborgerson: > This will now continue to do the RAM->GPU path for the surface, but will end up following the fast path to duplicate the surface GL binding to the texture GL binding around line 6413 below.

That's the intent anyway :)

mborgerson: An additional nit: can we drop the comment here? Surfaces are render targets

abaire: Done.

abaire: Looks like this change breaks SC:DA (though it still passes my pgraph tests)

abaire: Specifically, allowing `surf_to_tex` to be true prevents the game from working correctly

Issue #580, opened 2021-12-01
Lighting problems


RageXbox:
### Bug Description

For some reason the startup animation has a lighting problem where the background is darker than hardware and at the end the flash is brighter than hardware. Halo 2 has one where the Battle Rifle number screen is not shining. Possibly all the other games has one too.
https://xemu.app/titles/4d530064/#Halo-2
### Expected Behavior

Should light correctly as real hardware.

### xemu Version

v0.6.2-9-g69ceec4446

### System Information

Windows 10 (64-bit) 
(Intel(R) Xeon(R) CPU E5-2678 v3 @2.50 GHz) 
(NVIDIA Quadro P5000)
(NVIDIA 472.47)

### Additional Context

![093C27BD-2124-4D3D-9DB5-7E1F67568FFA](https://user-images.githubusercontent.com/76413188/144318185-6c574540-1615-435a-ace7-a7a53925e838.png)
![D9FD24C2-4399-48AB-8766-0CDF652C5FD3](https://user-images.githubusercontent.com/76413188/144318191-90f0015a-fcec-41dd-bf4a-5b6891a5aa07.png)
Xemu
![Screenshot_2021-11-30_202809](https://user-images.githubusercontent.com/76413188/144319345-456cc49f-d74d-4739-bb72-4208b34e26d2.png)
Hardware
![Screenshot_2021-11-30_203053](https://user-images.githubusercontent.com/76413188/144319378-96c6c4ca-d697-4833-b7f7-16b4c0945efc.png)

abaire:
@RageXbox can you update this with the results from the latest master? Are there still remaining issues that should be tracked here or are they tracked elsewhere already?

Triticum0:
Xbox boot screen fixed

PR #840, opened 2022-04-13
nv2a: Implement shadow samplers

### LINKED ISSUE #712: nv2a 0x97 -> 0x1e6c (3D_TEX_RCOMP?) unhandled
### Title

https://xemu.app/titles/53430007
Boot animation (EvoX m8plus 1.6)

### Bug Description

According to the debug output, 0x97 -> 0x1e6c is unhandled.
In the [NV20 nouveau sources](https://gitlab.freedesktop.org/mesa/mesa/-/blob/184a690fca700aa9ef2ae2e58b0f4bd86562aa6d/src/mesa/drivers/dri/nouveau/nv20_3d.xml.h#L1140), this instruction is `3D_TEX_RCOMP`.

The parameter of the call has so far been observed as one of the following:
`0x0`, `0x4`, `0x7`

### Expected Behavior

The NV2A instruction 0x1e6c should be handled.

### xemu Version

0.6.2-55-gcaf84884c4

### System Information

N/A

### Additional Context

_No response_

### LINKED ISSUE #781: Ninja Gaiden Black shadow problem
### Title

https://xemu.app/titles/5443000d/#Ninja-Gaiden-Black

### Bug Description

Ryu's shadow missing or incorrect

enemy shadows are blobs, xemu just renders it fine,Ryu's are dynamic, so it's probably stencil related maybe
shadows

### Expected Behavior

expect to display correctly

### xemu Version

Version: 0.6.2-82-ghttps://github.com/mborgerson/xemu/commit/2ff5f23235387a296fc9a5f11943b4dd29bd2837235
Branch: master
Commit: 2ff5f23
Date: Thu Mar 17 04:24:58 UTC 2022

### System Information

  "cpu": "Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz",
  "gl_renderer": "NVIDIA GeForce GTX 1080/PCIe/SSE2",
  "gl_shading_language_version": "4.00 NVIDIA via Cg compiler",
  "gl_vendor": "NVIDIA Corporation",
  "gl_version": "4.0.0 NVIDIA 515.36",
  "os_platform": "Windows",
  "os_version": "Windows 11"

### Additional Context

_No response_

### LINKED ISSUE #787: Splinter cell Double Agent: surface lights being rendered in distance only
### Title

### Bug Description

https://user-images.githubusercontent.com/30447649/159268426-56095736-f2ef-4a5a-a26c-ec254534f490.mp4

light xemu

Happens in all levels

### Expected Behavior

lights should remain rendered on surface while getting closer to them.

hardware:
hardware

### xemu Version

### System Information

Manjaro Linux
i7-10700 4.80GHz
GTX 970 STRIX 4GB VRAM

### Additional Context

### LINKED ISSUE #807: Splinter cell double agent: Dynamic lights not being rendered on surfaces
### Title

### Bug Description

Happens in iceland level, first part, but also in other levels using the same lights.

https://user-images.githubusercontent.com/30447649/159269038-ddbc1c9e-501c-49a9-b767-57a6d753bcbf.mp4

image

### Expected Behavior

image

https://www.youtube.com/watch?v=QDKhIzr6Slg

### xemu Version

### System Information
Manjaro Linux
i7-10700 4.80GHz
GTX 970 STRIX 4GB VRAM

### Additional Context

_No response_

### LINKED ISSUE #808: Splinter cell chaos theory versus: torch light not rendered
### Title

https://xemu.app/titles/55530038/#Tom-Clancy-s-Splinter-Cell-Chaos-Theory

### Bug Description

tochlight

### Expected Behavior

image

### xemu Version

### System Information

Manjaro Linux
i7-10700 4.80GHz
GTX 970 STRIX 4GB VRAM

### Additional Context

_No response_

### LINKED ISSUE #833: Splinter cell Double Agent: Shadows not being rendered on surfaces/objects
### Title

### Bug Description

the shadows are not being projected on surfaces/objects like in the screens:
shadow xemu
shadow xemu 2
xemu shadow 3

Happens in all levels, the most noticable one is the JBA HQ part 1/2

### Expected Behavior

hardware
shadow3
shadow hardware

### xemu Version

0.6.2-90-g6f507c80af

### System Information

Manjaro Linux
i7-10700 4.70GHz
GTX 970 STRIX
NVIDIA driver: 510.54
CUDA version: 11.6
RAM: DDR4 2933MHz HyperX Predator

### Additional Context

_No response_

abaire:
Implements shadow map functionality for 16-bit and 24-bit fixed point depth buffer modes.

This PR does not address 16-bit float mode (planning to do as a followup as it will require a different approach) or 24-bit float mode (which wouldn't be possible due to #322). 

This PR does not exactly match HW behavior, so some of the pgraph tests do not pass, however it does seem to work for a number of games so I'm taking it out of draft mode to encourage further testing. I am strongly suspicious that there will be edge cases where shadows will still be incorrect (e.g., anything using the fixed function pipeline with a non-default clipping range) but will need some real life test cases to rule out issues with my tests as a cause (i.e., there's some chance a failure to initialize something coupled with a difference in xemu and HW default states causes the difference).

Some of these need other recent PRs to fully resolve the problems, I'm going to assume this PR will be merged after the older fixes and mark them as fixed here:
Fixes #807 

Fixes #712 
Fixes #781 
Fixes #787 
Fixes #808 
Fixes #833 

[Test](https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/tests/texture_shadow_comparator_tests.cpp)
[HW results](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-Texture_shadow_comparator)


abaire:
@revix-0 can you re-test #787 with this so I can mark it as fixed by this PR if it is indeed fixed? Might be good to re-test all of the SplinterCell bugs that relate to lighting, shadows, or weird texture artifacts with this PR to see how many go away.

RageXbox:
This partially fixes #580 since the boot animation has got it's real effect. Combining with #720 would give it the full effect.
Screenshot 2022-04-12 223213
Screenshot 2022-04-12 223224
Screenshot 2022-04-12 223231
Screenshot 2022-04-12 223240

abaire:
> This partially fixes #580 since the boot animation has got it's real effect.

Good to hear and thanks for testing! This makes me optimistic that the implementation in this PR may be correct and the fact that the HW lets me do things that can't be replicated in GL without jumping through hoops may be irrelevant in practice, though I need to investigate more to really feel confident.

Fabxx:
image
now there are light reflections on the character where the lights are missing in some spots

on the fan as well:

image

image

the lights on surfaces are still missing tho

Fabxx:
UPDATE: those screens are from the fix branch with this fix individually, but combined with the other previous fixes for this game, the lights are rendered

image
image
image
image

fixed https://github.com/mborgerson/xemu/issues/792 as well
image

UPDATE 2: some shadow objects are still not rendered for some reason:

xemu:

obj shadow

hardware:
missing shadow

migh be same problem here, fisher shadow:

hardware:
fisher shadow hardware

xemu:
fisher shadow

LEVEL: JBA HQ Part 1 (third level)

xemu (video and screen):
https://user-images.githubusercontent.com/30447649/163219672-fdf98192-d626-487e-b200-0d67acf822d4.mp4
xemu



hardware:
hardware shadows

ko81e24wy:
cause some light problem in NGB
#840
NGBBB
#835 
NGBBBBBB

Fabxx:
fixes https://github.com/mborgerson/xemu/issues/808 with the other PRs for the lighting merged, both pandora tomorrow and chaos theory versus

image

dracc:
This adds "shadows" back into Fable TLC. :+1:
fable_840_shadows

(Lighting is somehow inverted)

abaire:
> UPDATE 2: some shadow objects are still not rendered for some reason:
> 
> xemu:
> 
> obj shadow
> 
> hardware: missing shadow
> 
> migh be same problem here, fisher shadow:
> 
> hardware: fisher shadow hardware
> 
> xemu: fisher shadow

@revix-0 can you provide instructions with how to get to those parts of the game? I've not played past the area that used to have the blue screen steam pipe bug so I'm not sure where I have to go to reproduce.

abaire:
> cause some light problem in NGB #840 

@ko81e24wy I'm failing to see the problem from those screenshots (might just be because of glare on my screen). Can you add a circle or an arrow or something so I can see exactly what you mean?
Also can you tell me where that screenshot was taken (relative to the nearest save point would be best) so I can repo?

abaire:
> (Lighting is somehow inverted)

Thanks @dracc, this one is very helpful as it indicates that there's still some interesting bug in the way the comparisons are handled. Any chance you could look at the pgraph log or the nv2a state and tell me if this is using a d24s8 depth buffer? So far I've only tested d16 (both in Splinter Cell and in the pgraph tester) so I wonder if that's part of the problem. 

Also possible that this is a signed/unsigned issue or a float/fixed issue, so if you could turn on https://github.com/mborgerson/xemu/blob/master/hw/xbox/nv2a/debug.h#L92 to see if it's setting signed components and check to see if the depth is using floating point (https://github.com/mborgerson/xemu/blob/master/hw/xbox/nv2a/pgraph.c#L1408) that'd be extremely helpful in narrowing this down. I own TLC so but am away for a bit and won't be able to test until I'm back.

Triticum0:
@abaire Look at the rock on the screenshot there is no gradient on the darkening of the rock face with the diagonal line going through the whole of the rock face where on the side much darker that the other

ko81e24wy:
> > cause some light problem in NGB #840
> 
> @ko81e24wy I'm failing to see the problem from those screenshots (might just be because of glare on my screen). Can you add a circle or an arrow or something so I can see exactly what you mean? Also can you tell me where that screenshot was taken (relative to the nearest save point would be best) so I can repo?

OK，I will make a vid after work,its before the boss of chapter 2

ko81e24wy:
some test in NGB 
https://www.youtube.com/watch?v=insPpGKnHUk

RageXbox:
There is a shadow bug on the boot animation.
Screenshot 2022-04-14 005553
Screenshot 2022-04-15 145908

Fabxx:
fixes crash twinsanity shadows:
image

abaire:
> some test in NGB https://www.youtube.com/watch?v=insPpGKnHUk

Now I see it, thanks! It looks similar to Fable, the shadow effect is inverted into a lighting effect. I'll hopefully have some time to do a deeper investigation this weekend, I'm still suspicious that the way that OpenGL normalizes depth textures is not directly compatible with the way the nv2a hardware uses them, may just need to move away from GL depth components and do comparisons manually in the shader.

Fabxx:
These PRs fix the issue in GTA 3/VC about the red tail lights getting over the car
image
image

abaire:
It looks like the problem might actually be with the vertex shader:

https://github.com/mborgerson/xemu/blob/master/hw/xbox/nv2a/vsh.c#L775

```
        /* Map the clip range into clip space so z is clipped correctly.
         * Note this makes the values in the depth buffer wrong. This should be
         * handled with gl_ClipDistance instead, but that has performance issues
         * on OS X.
         */
```

which makes me surprised this works as well as it does.

Playing around with renderdoc I definitely see inexplicable Z values in my test. I have a depth map that is cleared to 0xFFFF and a number of quads rendered inside of it that do not reach all the way to the edge. One of those quads has a depth that is also projected to 0xFFFF (confirmed to be correct in HW by dumping the depth texture). However in xemu, instead of mapping to 1.0 like the cleared area of the depth texture, this quad maps to `0.0039063`. All of the actual rendered depth values are compressed into very small values.
I don't think it's possible to re-map these values since there's no way in the pixel shader to know that certain areas of the depth texture were not touched by the vertex shader and thus have correct values whereas other areas would need work.

I suspect fixing this PR will require the vertex shader to be fixed first such that the depth values are correct.


UPDATE: It's a bit more complicated than that, the fixed function shader path also produces results that don't match HW. This is presumably because the GL vertex shader outputs clip coordinates whereas the nv2a shader outputs screen coordinates. The fixed function shader applies an inverse of the viewport matrix to the transformed coordinate and adjusts it to be in NDC:
```
  oPos = invViewport * (tPosition * compositeMat);
  oPos.z = oPos.z * 2.0 - oPos.w;
```

Saving the raw transformed position produces values are at least very close to what I see when I dump the depth buffer from hardware. Even if we blindly undo this transform (or also transform the interpolated z coord in the fragment shader back to clip space) I think we run afoul of the fact that xemu is clearing the depth buffer to what would likely be a screen space value so any pixels untouched by rendering will have an untransformed value. We may be able to adjust the clear code to also work in clip space, but it is quickly turning into a large chain of changes with a lot of potential to introduce errors. The inverse transform also squashes values into a very small portion of the 0..1 range so there's a good chance that precision issues will crop up as well.

abaire:
After diving into getting renderdoc working, I can confirm that, for the programmable pipeline path, it looks like the values in the depth buffer are actually reasonable (at least very close to expectations for my test case) and that the OpenGL depth texture also looks correct. The test is still failing, but I expect this will be a mistake or misunderstanding in my code rather than a major underlying issue. 

That said, both the programmable and fixed function pipelines use some estimates to try to map values back into clip space that will only work in specific situations that may or may not hold true in practice, so even once I get this working there may be additional effort needed to find a better general solution to faithfully reproducing the original depths such that shadow comparators are valid.

abaire:
Getting closer, this mostly passes the tests for 16bit fixed depth buffers and 24-bit fixed depth buffers (at least on my M1 mac). There are likely some rounding issues.

16 bit float buffers are not yet handled correctly, possibly because of the hack used to map nv2a screenspace coords back into clip space.

@dracc I also discovered that I had misread the comparator functions, shadow maps apparently swap the behavior as compared to depth/stencil maps so the value that usually maps to "GL_LESS" actually should map to "GL_GREATER". It would be extremely helpful if you could see if this changes behavior for Fable, as inverting the comparison could certainly invert where shadows are applied.

vanderleivieirainacio:
Shadows in Ninja Gaiden Black are much better now! Ryu Hayabusa's shadows flicker a bit but anyway it's much better than before:

Captura de tela 2022-04-28 014426

Captura de tela 2022-04-28 013838

Captura de tela 2022-04-28 013809

dracc:
I'm honestly not familiar enough with graphics debugging to provide useful technical data at this time, but this is what Fable TLC looks like right now with this PR applied:
image
Brighter overall, which is correct. But no shadows as far as the eyes can see.

abaire:
@dracc thanks for checking! I've got a renderdoc capture from SC:DA that exhibits a clear problem with this current draft and I know that there are situations where it's definitely not matching HW; will ping you again to test Fable once I get things resolved.

One piece of info that would be extremely helpful would be to know what depth buffer format Fable is using (I own the game and will eventually check but am traveling for another week). The easiest way to find this is to set a breakpoint someplace like https://github.com/mborgerson/xemu/blob/master/hw/xbox/nv2a/pgraph.c#L3322 and check the values of `pg->surface_shape.zeta_format` and `pg->surface_shape.z_format`. 
Alternatively, the nv2a debugger under the debug menu in my `work` branch will print this out in the `nv2a last BeginEnd` box.

dracc:
> @dracc thanks for checking! I've got a renderdoc capture from SC:DA that exhibits a clear problem with this current draft and I know that there are situations where it's definitely not matching HW; will ping you again to test Fable once I get things resolved.
> 
> One piece of info that would be extremely helpful would be to know what depth buffer format Fable is using (I own the game and will eventually check but am traveling for another week). The easiest way to find this is to set a breakpoint someplace like https://github.com/mborgerson/xemu/blob/master/hw/xbox/nv2a/pgraph.c#L3322 and check the values of `pg->surface_shape.zeta_format` and `pg->surface_shape.z_format`. Alternatively, the nv2a debugger under the debug menu in my `work` branch will print this out in the `nv2a last BeginEnd` box.

```
Thread 35 "xemu" hit Breakpoint 1, pgraph_NV097_CLEAR_SURFACE_handler (
    d=0x7fff7cacd010, pg=0x7fff7caefcc8, subchannel=<optimized out>, 
    method=7572, parameter=243, parameters=0x7fff2980f254, 
    num_words_available=1, num_words_consumed=0x7fff15ff0688, inc=true)
    at ../hw/xbox/nv2a/pgraph.c:3343
3343	        switch(pg->surface_shape.zeta_format) {
(gdb) print pg->surface_shape.zeta_format
$1 = 2
(gdb) print pg->surface_shape.z_format
$2 = 0
```
`zeta_format` is indeed 24-bit

abaire:
@dracc perfect, thanks!

I was able to track down an error w/ @revix-0 's help on SC:DA and have just pushed another fix. Things still don't quite match up w/ HW behavior but they're pretty close, at least when not using the fixed function pipeline and having a z clip that closely matches the viewport projection. When you have a moment, could you re-test Fable and see if it brings back shadows (ideally flipped the right way round this time)?

vanderleivieirainacio:
Testing here again Ninja Gaiden Black I saw that Ryu Hayabusa's shadows flicker when I change the resolution in the middle of the game and if I don't change it in the middle of the game, the shadows work smoothly. Everything else works perfectly.

abaire:
The current draft still has an issue in SplinterCell DA where some shadows have disappeared. After doing some investigation, it turns out that SC:DA uses different Z clip ranges when rendering the shadow map (from the perspective of the light) versus when it actually applies the map. This means we can't simply use the current clip range to try to convert the texture reference value into the [0,1] interval that OpenGL uses and would need to reliably track the range used to create the depth buffer in the first place and use that to convert the texture coord's z component.

This is easy to do as a proof of concept in renderdoc by manually editing the pixel shader:

Using the clip range at the time of the shadow map application [0.00, 1.67772E+07]:
Screen Shot 2022-04-28 at 21 33 47

Forcibly editing the shader to use [0,65535] (the value observed during creation of the depth texture for this frame):
Screen Shot 2022-04-28 at 21 32 06

Note the lack of shadows behind the box on the wall on the left and on the floor behind the left wall in the first image.

I think fixing this while continuing to use depth textures as-is will be very tricky, and the clip range mechanism is already trivial to break (see #843, for instance). We may be able to use the depth texture as a type other than a depth component and do shadow comparisons in the shader. I'm not sure of the performance impact of doing this (if there is any dedicated hardware to do the shadow comparison this would bypass it) and I know that this will produce different results based on filtering mode.

Fabxx:
i can confirm what abaire said, just tested this a minute ago: 
image

Triticum0:
@abaire #577 will also be fixed.

abaire:
> @abaire #577 will also be fixed.

@Triticum0 just checking - are you sure all of those cases will be fixed (that's a long list :))? There are some other approaches to dynamic shadows so I could imagine that not all games follow the same pattern. In particular, games that use actual dynamic lighting and the stencil buffer to do shadows would probably not be affected by this fix.
There are also still some differences in the behavior compared to HW with the way I'm implementing this; I'm not sure if I can get it to match perfectly in every case, though I'm getting very close now.

@revix-0 @RageXbox @dracc @vanderleivieirainacio I've just pushed a substantial change to the way this fix is implemented (there were cases in SplinterCell that were going to be extremely difficult to resolve using OpenGL depth texture comparison functions). Would it be possible for you to re-test things to verify that it still works as expected?

vanderleivieirainacio:
Ninja Gaiden Black with shadow graphics bugs and error window now:

Captura de tela 2022-04-29 113512

Captura de tela 2022-04-29 113735

Fabxx:
@vanderleivieirainacio please retry with the recent push abaire did, same thing happened on first splinter cell but now the crash  is fixed

vanderleivieirainacio:
> @vanderleivieirainacio please retry with the recent push abaire did, same thing happened on first splinter cell but now the crash is fixed

I tested it here. the error window no longer appears but the shadow bugs continue:

Captura de tela 2022-04-29 121845

abaire:
@vanderleivieirainacio thanks for testing! Do you know if this is easy to repro someplace near the start of the game by any chance? I don't have my saves on my current computer.

Triticum0:
@abaire  I Think it all game depends. I would need to use debug setting to check what causing the bugs. it will at least fix half of the issues. That is not even the full list because of vague issue reports and not reporting all the issues to Github. Though the games that have missing shadows will most likely be fixed.

abaire:
> @abaire I Think it all game depends. I would need to use debug setting to check what causing the bugs. it will at least fix half of the issues. That is not even the full list because of vague issue reports and not reporting all the issues to Github. Though the games that have missing shadows will most likely be fixed.

Yeah, that was my worry, I think we'll just have to ask people to re-test all of those games after this gets merged so we can remove any with shadows/lights that are fixed by this PR and leave the rest to be fixed in future PRs.

Fabxx:
@vanderleivieirainacio just to be sure, have you tested with merged https://github.com/mborgerson/xemu/pull/835 and https://github.com/mborgerson/xemu/pull/831 with this? those PRs are linked with this draft, merge these two and retest, see how it behaves

vanderleivieirainacio:
> @vanderleivieirainacio thanks for testing! Do you know if this is easy to repro someplace near the start of the game by any chance? I don't have my saves on my current computer.

This shadow bug in NGB happens throughout the game from the demo to any selected stage.

vanderleivieirainacio:
> @vanderleivieirainacio just to be sure, have you tested with merged #835 and #831 with this? those PRs are linked with this draft, merge these two and retest, see how it behaves

I haven't tried it. Unfortunately I don't know how to blend and still the shadows were working correctly in NGB. Maybe someone else would have to do the merge to test...

abaire:
> 

I would not expect either of those PRs to have significant influence on the NGB issue, and given that the shadows worked better before it seems clear that the problem is related to the new approach that fixes SC:DA.

There are some significant issues with the threshold value that comes out of the vertex shader in 24-bit depth mode (which we can tell is used by NGB due to the assert that came up when 24-bit was briefly broken). I'll work on trying to find a way to resolve this and hope that it is fixable.

@vanderleivieirainacio if it's possible to point me at exactly some area in the first level where this new problem is clear (e.g., from a save point or ideally from the starting point) it'll make my life much easier in trying to reproduce and fix the problem. (I haven't had a chance to actually test the game yet, so apologies if it's immediately obvious right when starting a new game, but even if it's something silly like "walk forward 10 steps and look at the wall on the left" that is enormously useful)

dracc:
With the latest commit (d9dcc8b), behavior in Fable TLC is basically identical with current master.
image

vanderleivieirainacio:
> > 
> 
> I would not expect either of those PRs to have significant influence on the NGB issue, and given that the shadows worked better before it seems clear that the problem is related to the new approach that fixes SC:DA.
> 
> There are some significant issues with the threshold value that comes out of the vertex shader in 24-bit depth mode (which we can tell is used by NGB due to the assert that came up when 24-bit was briefly broken). I'll work on trying to find a way to resolve this and hope that it is fixable.
> 
> @vanderleivieirainacio if it's possible to point me at exactly some area in the first level where this new problem is clear (e.g., from a save point or ideally from the starting point) it'll make my life much easier in trying to reproduce and fix the problem. (I haven't had a chance to actually test the game yet, so apologies if it's immediately obvious right when starting a new game, but even if it's something silly like "walk forward 10 steps and look at the wall on the left" that is enormously useful)

In the first phase, as soon as Ryu jumps from the waterfall to kill the enemy, the bug starts, but if you start from the enemy's corpse at the beginning of the game, you take a few steps and look at the wall on the left, the shadow graphics problems are already noticeable . I hope this could have helped you. I can print other images of the game with the graphical problems if you want...

Captura de tela 2022-04-29 133526

Fabxx:
image
@vanderleivieirainacio  i had the other two PRs, and the difference i can see is from the character shadow which is visible

vanderleivieirainacio:
@revix-0 Yes, your image and my image show that the shadow is inside a light square with a big dark part around it. Only Ryu's shadow should appear.

Fabxx:
@vanderleivieirainacio  what i meant was that yes, we both get the square ecc, but in your screen compared to mine, i don't see the ninja shadow at all, and i'm in the same spot

abaire:
> @revix-0 Yes, your image and my image show that the shadow is inside a light square with a big dark part around it. Only Ryu's shadow should appear.

I'm surprised the other PRs make a difference but I guess it gives some insight as to how NGB is actually building up the depth texture. As @vanderleivieirainacio points out, it's clear that something is very wrong in both captures. I also wonder if using those PRs as well would fix shadows in Fable, but I assume @dracc was testing just this branch when they were "working" but inverted.

I have an idea for how to address the 24-bit reference being wrong, will ping you all when I confirm that it works and update this PR.

dracc:
> I also wonder if using those PRs as well would fix shadows in Fable, but I assume @dracc was testing just this branch when they were "working" but inverted.

Yes, I avoid merging multiple branches when testing new stuff as far as possible.

RageXbox:
Now it is properly rendering shadows without graphical glitches for boot animation
https://user-images.githubusercontent.com/76413188/166069801-486aaaa8-a548-4507-a3e6-cf0826662408.mp4

vanderleivieirainacio:
@abaire improved the shadows a bit (it's not Ryu's shadow in a light square anymore) but there's still a lot of malformed shadow in the environment. There are also moments where the shadow flickers throughout the game, fading and coming back:

Captura de tela 2022-05-01 170708

Captura de tela 2022-05-01 173912

Captura de tela 2022-05-01 170952

Captura de tela 2022-05-01 171310

abaire:
@RageXbox @dracc @vanderleivieirainacio I just pushed another attempt that is worth re-testing. 

I've tested the attract mode in NGB and it looks correct now (before I could easily repro the issue @vanderleivieirainacio reported). @revix-0 also verified SC:DA works as expected.

dracc:
Shadows in Forza Motorsport are now working, before only the blurred shadows below the car were visible.
image
Shadows in Fable TLC looks perfect!
image

vanderleivieirainacio:
> @RageXbox @dracc @vanderleivieirainacio I just pushed another attempt that is worth re-testing.
> 
> I've tested the attract mode in NGB and it looks correct now (before I could easily repro the issue @vanderleivieirainacio reported). @revix-0 also verified SC:DA works as expected.

Thanks for the hard work! Ryu's shadows and environment in NGB are fixed again! Sorry for the stress.

Captura de tela 2022-05-02 153009

Captura de tela 2022-05-02 153218

dracc:
Just Cause now has working shadows too!
image
Current master:
image

hw/xbox/nv2a/nv2a_regs.h
```diff
@@ -1218,6 +1218,15 @@
 #   define NV097_SET_COMBINER_COLOR_OCW                       0x00001E40
 #   define NV097_SET_COMBINER_CONTROL                         0x00001E60
 #   define NV097_SET_SHADOW_ZSLOPE_THRESHOLD                  0x00001E68
+#   define NV097_SET_SHADOW_COMPARE_FUNC                      0x00001E6C
+#       define NV097_SET_SHADOW_COMPARE_FUNC_NEVER                0x00000000
+#       define NV097_SET_SHADOW_COMPARE_FUNC_GREATER              0x00000001
```
mborgerson: I think the order might be flipped here. In other instances (here in this file) the order is:

```
#       define ..._NEVER               0
#       define ..._LESS                1
#       define ..._EQUAL               2
#       define ..._LEQUAL              3
#       define ..._GREATER             4
#       define ..._NOTEQUAL            5
#       define ..._GEQUAL              6
#       define ..._ALWAYS              7
```

hw/xbox/nv2a/nv2a_regs.h
```diff
@@ -1218,6 +1218,15 @@
 #   define NV097_SET_COMBINER_COLOR_OCW                       0x00001E40
 #   define NV097_SET_COMBINER_CONTROL                         0x00001E60
 #   define NV097_SET_SHADOW_ZSLOPE_THRESHOLD                  0x00001E68
+#   define NV097_SET_SHADOW_COMPARE_FUNC                      0x00001E6C
```
mborgerson: I think this should be `SET_SHADOW_DEPTH_FUNC` with other conventions here

hw/xbox/nv2a/psh.c
```diff
@@ -731,12 +780,40 @@ static MString* psh_convert(struct PixelShader *ps)
...
-            sampler_type = "sampler3D";
-            mstring_append_fmt(vars, "vec4 t%d = textureProj(texSamp%d, pT%d.xyzw);\n",
-                               i, i, i);
+            if (ps->state.shadow_map[i]) {
+                if (ps->state.shadow_compare_func == SHADOW_COMPARE_FUNC_NEVER) {
+                    mstring_append_fmt(vars, "vec4 t%d = vec4(0.0);\n", i);
+                } else if (ps->state.shadow_compare_func == SHADOW_COMPARE_FUNC_ALWAYS) {
+                    mstring_append_fmt(vars, "vec4 t%d = vec4(1.0);\n", i);
+                } else {
+                    mstring_append_fmt(vars, "pT%d.xy *= texScale%d;\n", i, i);
```
mborgerson: Let's collapse this to one call to `mstring_append_fmt`

hw/xbox/nv2a/psh.c
```diff
@@ -711,7 +760,7 @@ static MString* psh_convert(struct PixelShader *ps)
                                i);
             break;
         case PS_TEXTUREMODES_PROJECT2D: {
-            sampler_type = ps->state.rect_tex[i] ? "sampler2DRect" : "sampler2D";
+            assert(!ps->state.shadow_map[i] && "Shadow handling not implemented for Project2D");
```
mborgerson: We only handle it for one texture mode, can we drop all these asserts and just have one?

hw/xbox/nv2a/pgraph.c
```diff
@@ -300,14 +301,19 @@ static const ColorFormatInfo kelvin_color_format_map[66] = {
...
+         {GL_RED, GL_ZERO, GL_ZERO, GL_ZERO}, true},
+    [NV097_SET_TEXTURE_FORMAT_COLOR_LU_IMAGE_DEPTH_X8_Y24_FIXED] =
+        {4, true, GL_DEPTH_COMPONENT, GL_DEPTH_STENCIL, GL_UNSIGNED_INT_24_8,
+         {GL_RED, GL_ONE, GL_ZERO, GL_ZERO}, true},
     [NV097_SET_TEXTURE_FORMAT_COLOR_LU_IMAGE_DEPTH_Y16_FIXED] =
-        {2, true, GL_DEPTH_COMPONENT16, GL_DEPTH_COMPONENT, GL_UNSIGNED_SHORT},
+        {2, true, GL_DEPTH_COMPONENT16, GL_DEPTH_COMPONENT, GL_UNSIGNED_SHORT,
+         {GL_RED, GL_ZERO, GL_ZERO, GL_ZERO}, true},
     [NV097_SET_TEXTURE_FORMAT_COLOR_LU_IMAGE_DEPTH_Y16_FLOAT] =
         {2, true, GL_DEPTH_COMPONENT16, GL_DEPTH_COMPONENT, GL_FLOAT},
```
mborgerson: Should `.depth=true` for `NV097_SET_TEXTURE_FORMAT_COLOR_LU_IMAGE_DEPTH_Y16_FLOAT`?

hw/xbox/nv2a/psh.c
```diff
@@ -731,12 +780,40 @@ static MString* psh_convert(struct PixelShader *ps)
...
+                    mstring_append_fmt(vars, "vec4 t%d = vec4(0.0);\n", i);
+                } else if (ps->state.shadow_compare_func == SHADOW_COMPARE_FUNC_ALWAYS) {
+                    mstring_append_fmt(vars, "vec4 t%d = vec4(1.0);\n", i);
+                } else {
+                    mstring_append_fmt(vars, "pT%d.xy *= texScale%d;\n", i, i);
+                    mstring_append_fmt(vars, "vec4 t%d_depth = textureProj(texSamp%d, pT%d.xyw);\n",
+                                       i, i, i);
+                    mstring_append_fmt(vars,
+                                       "float t%d_max_depth;\n"
+                                       "if (t%d_depth.y > 0) {\n"
```
mborgerson, line 801: Ternary op would be more succinct

hw/xbox/nv2a/nv2a_regs.h
abaire: Done.

hw/xbox/nv2a/nv2a_regs.h
abaire: Yeah, I initially thought the same (in fact I had a bug where it inverted shadows because I blindly followed the pattern of other calls)

These values are from nouveau (linked from #712):  https://gitlab.freedesktop.org/mesa/mesa/-/blob/184a690fca700aa9ef2ae2e58b0f4bd86562aa6d/src/mesa/drivers/dri/nouveau/nv20_3d.xml.h#L1140

hw/xbox/nv2a/pgraph.c
abaire: I left it out because this PR doesn't actually implement the float path. Changed since I don't think it would do anything useful in either way and this will be one less thing to miss when the float path is implemented.

hw/xbox/nv2a/psh.c
abaire: Done, though I feel the number of "i"'s that have to be passed makes it quite difficult to read which is why I split it initially.

hw/xbox/nv2a/psh.c
abaire: Agreed, but when float support gets implemented it'll outgrow a ternary.

hw/xbox/nv2a/psh.c
abaire: The problem with using one assert is that we lose the ability to know what the actual value was. I don't know if any games use any of these invalid modes (I assume `PROJECT2D` could be used for a 1D shadow map, but I don't know if that exists in practice).

I'm hoping that by having explicit asserts we can actually get useful info on what is supported in the wild without having to either buy the game or teach a random user how to build xemu (which seems to often times include teaching *nix command line).

mborgerson: How about we just log the value with printf in this error case?

hw/xbox/nv2a/nv2a_regs.h
mborgerson: Let's follow the pattern we already have to be consistent; if you swap comparison operands it will be the same logic. I think the nouveau definitions may be swapped in this case

hw/xbox/nv2a/nv2a_regs.h
```diff
@@ -1218,6 +1218,15 @@
 #   define NV097_SET_COMBINER_COLOR_OCW                       0x00001E40
 #   define NV097_SET_COMBINER_CONTROL                         0x00001E60
 #   define NV097_SET_SHADOW_ZSLOPE_THRESHOLD                  0x00001E68
+#   define NV097_SET_SHADOW_DEPTH_FUNC                      0x00001E6C
```
mborgerson: nit: alignment

abaire: Sorry, fixed.

hw/xbox/nv2a/psh.c
abaire: Sounds good to me. 

I made a bit of a compromise: since there are actually quite a number of paths where it isn't necessarily an error to pass a shadow map, but it won't be used as one (e.g., the `NONE` path) and since I'm already asserting in get_sampler_type, I used the condensed switch there to do the asserts w/ an stderr log indicating the specific type so we can implement support where necessary based on actual use.

hw/xbox/nv2a/nv2a_regs.h
abaire: Done.

hw/xbox/nv2a/pgraph.c
```diff
@@ -180,12 +180,12 @@ static const GLenum pgraph_cull_face_map[] = {
 
 static const GLenum pgraph_depth_func_map[] = {
     GL_NEVER,
-    GL_LESS,
```
mborgerson: Did you mean to change this

hw/xbox/nv2a/psh.h
```diff
@@ -33,6 +33,17 @@ enum PshAlphaFunc {
     ALPHA_FUNC_ALWAYS,
 };
 
+enum PshShadowDepthFunc {
+    SHADOW_DEPTH_FUNC_NEVER,
+    SHADOW_DEPTH_FUNC_GREATER,
+    SHADOW_DEPTH_FUNC_EQUAL,
```
mborgerson: Let's keep ordering consistent with defs

hw/xbox/nv2a/psh.c
```diff
@@ -558,6 +558,64 @@ static void add_final_stage_code(struct PixelShader *ps, struct FCInputInfo fina
...
+        if (state->shadow_map[i]) {
+            fprintf(stderr, "Shadow map support not implemented for mode %d\n", mode);
+            assert(!"Shadow map support not implemented for this mode");
+        }
+        return sampler2D;
+    }
+}
+
+static const char *shadow_comparison_map[] = {
+    [SHADOW_DEPTH_FUNC_GREATER] = ">",
```
mborgerson: nit: Let's change ordering to be consistent with def

hw/xbox/nv2a/pgraph.c
abaire: Nope.

hw/xbox/nv2a/psh.h
abaire: Swapped.

hw/xbox/nv2a/psh.c
abaire: Sorry, missed this comment, fixed now.

vanderleivieirainacio:
Shadows in Mortal Kombat Shaolin Monks:

Captura de tela 2022-05-10 212659

Captura de tela 2022-05-10 213258

abaire:
@vanderleivieirainacio what does that exact screen look like on the `master` branch? What does it look like on hardware?

We already know this PR doesn't exactly match HW as it doesn't pass all of my pgraph tests. I also know #843 triggers breaks, but don't know if any games follow that setup pattern, we'd need to know if that game uses the fixed function pipeline and whether it changes the depth clipping in an interesting way.

vanderleivieirainacio:
The only thing I can say that I understand is that the master branch has no shadows. Shadows were implemented in this latest shadow sampler implementation PR:

Last master branch:
Captura de tela 2022-05-10 221913

PR:
Captura de tela 2022-05-10 222856

mborgerson:
> This PR does not address 16-bit float mode (planning to do as a followup as it will require a different approach) or 24-bit float
> This PR does not exactly match HW behavior, so some of the pgraph tests do not pass

We can continue iterating on this. Thanks for the patch!

Triticum0:
@abaire When you added this pr it did fix Ed Edd n Eddy The Mis Edventures
Edit it was'nt this pr that broke it 
xemu-2022-06-02-00-12-08

Triticum0:
I tested z16 float shadow comparison it didn't fix it

PR #849, opened 2022-04-25
Add new configuration system

mborgerson:
The UI overhaul has prompted me to finally improve the config system: no more overly complex code changes to add/use simple config items.

* Replaces .ini with .toml
* Reduces complexity in adding new config items
* Simplifies code using config items

Things to note:
* Users will need to reconfigure their xemu because I did not add config migration.
* New build dep of pyyaml to read the config spec.
* It might be tempting to add new configuration items to xemu, but please wait until the new UI is merged so you don't add work I'll end up replacing.


config_spec.yml
```diff
@@ -0,0 +1,71 @@
...
+    default: nat
+  pcap:
+    netif: string
+  udp:
+    bind_addr:
+      type: string
+      default: 0.0.0.0:9368
+    remote_addr:
+      type: string
+      default: 1.2.3.4:9368
```
GXTX, line 54: Should this be some non-public address?

PR #850, opened 2022-04-25
UI: Display hardware info in Help > About window

### LINKED ISSUE #668: Add machine information to Help->About screen
### Feature Request

The Help->Report compatibility screen contains interesting information about the user's machine (OS, driver info) but this is buried under "Report Details".

In order to facilitate GitHub bug reports and easier `#help` conversations, it'd be nice to replicate this information in the Help->About screen, where the Version info is already displayed.

Ideally we would segment this in a way that makes it easy to cut and paste into the GitHub new issue form (i.e., put some whitespace between the version info and the system info, use the same line prefixes as in the template).

### Alternatives

Alternatively, or as an additional enhancement, we could add a mechanism to launch a browser window with the relevant info pre-populated into the form (including title info if relevant from the current running context).

### Additional Context

_No response_

Fabxx:
Fixes: https://github.com/mborgerson/xemu/issues/668

image

### DIFF
--- a/ui/xemu-hud.cc
+++ b/ui/xemu-hud.cc
@@ -838,11 +838,11 @@ private:
 
 public:
     AboutWindow()
-        {
-            snprintf(build_info_text, sizeof(build_info_text),
-                "Version: %s\nBranch:  %s\nCommit:  %s\nDate:    %s",
-                xemu_version, xemu_branch, xemu_commit, xemu_date);
-        }
+    {
+        snprintf(build_info_text, sizeof(build_info_text),
+            "Version: %s\nBranch:  %s\nCommit:  %s\nDate:    %s",
+            xemu_version, xemu_branch, xemu_commit, xemu_date);
+    }
 
     void Draw()
     {
--- a/ui/xemu-hud.cc
+++ b/ui/xemu-hud.cc
@@ -793,15 +793,15 @@ static const char *get_os_platform(void)
 {
     const char *platform_name;
 
-#if defined(__linux__)
+    #if defined(__linux__)
         platform_name = "Linux";
-#elif defined(_WIN32)
+    #elif defined(_WIN32)
        platform_name = "Windows";
-#elif defined(__APPLE__)
+    #elif defined(__APPLE__)
        platform_name = "macOS";
-#else
+    #else
         platform_name = "Unknown";
-#endif
+    #endif
         return platform_name;
 }
 
--- a/ui/xemu-hud.cc
+++ b/ui/xemu-hud.cc
@@ -827,29 +827,39 @@ static const char *get_cpu_info(void)
     return cpu_info;
 }
 
+static const char *get_gl_vendor(void) 
+{
+    const char *gl_vendor = (const char*)glGetString(GL_VENDOR); 
+    return gl_vendor;
+}
+
+static const char *get_gl_renderer(void) 
+{
+    const char *gl_renderer = (const char*)glGetString(GL_RENDERER);
+    return gl_renderer;
+}
+
+static const char *get_gl_version(void)
+{
+    const char *gl_version = (const char*)glGetString(GL_VERSION);
+    return gl_version;
+}
+
+static const char *get_gl_shader_version(void)
+{
+    const char *gl_shader_version = (const char*)glGetString(GL_SHADING_LANGUAGE_VERSION);
+    return gl_shader_version;
+}
+
 class AboutWindow
 {
 public:
     bool is_open;
 
 private:
-    char build_info_text[256];
+    char build_info_text[600];
 
 public:
-    AboutWindow()
-    {
-        snprintf(build_info_text, sizeof(build_info_text),
-            "Version: %s\n" "Branch:  %s\n" "Commit:  %s\n" "Date: %s\n"
-            "CPU: %s\n" "OS Platform: %s\n" "OS Version: %s\n",
-            xemu_version, xemu_branch, xemu_commit, xemu_date, get_cpu_info(), 
-            get_os_platform(), xemu_get_os_info());
-        // FIXME: Show driver
-        // FIXME: Show BIOS/BootROM hash
-    }
-
-    ~AboutWindow()
-    {
-    }
 
     void Draw()
     {
@@ -864,6 +874,17 @@ public:
 
         static uint32_t time_start = 0;
         if (ImGui::IsWindowAppearing()) {
+             
+        snprintf(build_info_text, sizeof(build_info_text),
+            "Version: %s\n" "Branch:  %s\n" "Commit:  %s\n" "Date: %s\n\n"
+            "----------------System Information-----------------------------\n\n"
+            "CPU: %s\n" "OS Platform: %s\n" "OS Version: %s\n" "Manufacturer: %s\n"
+            "GPU Model: %s\n" "Driver: %s\n" "Shading Language Version: %s\n",
+            xemu_version, xemu_branch, xemu_commit, xemu_date, get_cpu_info(), 
+            get_os_platform(), xemu_get_os_info(), get_gl_vendor(), get_gl_renderer(),
+            get_gl_version(), get_gl_shader_version());
+            // FIXME: Show BIOS/BootROM hash
+
             time_start = SDL_GetTicks();
         }
         uint32_t now = SDL_GetTicks() - time_start;
@@ -900,7 +921,7 @@ public:
         ImGui::Dummy(ImVec2(0,40*g_ui_scale));
 
         ImGui::PushFont(g_fixed_width_font);
-        ImGui::InputTextMultiline("##build_info", build_info_text, sizeof(build_info_text), ImVec2(-FLT_MIN, ImGui::GetTextLineHeight() * 6), ImGuiInputTextFlags_ReadOnly);
+        ImGui::InputTextMultiline("##build_info", build_info_text, sizeof(build_info_text), ImVec2(-FLT_MIN, ImGui::GetTextLineHeight() * 12), ImGuiInputTextFlags_ReadOnly);
         ImGui::PopFont();
 
         ImGui::End();
--- a/ui/xemu-hud.cc
+++ b/ui/xemu-hud.cc
@@ -864,7 +864,7 @@ public:
              
             snprintf(platform_info_text, sizeof(platform_info_text),
                 "CPU: %s\nOS Platform: %s\nOS Version: %s\nManufacturer: %s\n"
-                "GPU Model: %s\nDriver: %s\nShading Language Version: %s\n",
+                "GPU Model: %s\nDriver: %s\nShading Language Version: %s",
                  get_cpu_info(), get_os_platform(), xemu_get_os_info(), gl_renderer,
                  gl_vendor, gl_version, gl_shader_version);
             // FIXME: Show BIOS/BootROM hash
--- a/ui/xemu-hud.cc
+++ b/ui/xemu-hud.cc
@@ -793,16 +793,16 @@ static const char *get_os_platform(void)
 {
     const char *platform_name;
 
-    #if defined(__linux__)
-        platform_name = "Linux";
-    #elif defined(_WIN32)
-       platform_name = "Windows";
-    #elif defined(__APPLE__)
-       platform_name = "macOS";
-    #else
-        platform_name = "Unknown";
-    #endif
-        return platform_name;
+#if defined(__linux__)
+    platform_name = "Linux";
+#elif defined(_WIN32)
+    platform_name = "Windows";
+#elif defined(__APPLE__)
+    platform_name = "macOS";
+#else
+    platform_name = "Unknown";
+#endif
+    return platform_name;
 }
 
 #ifndef _WIN32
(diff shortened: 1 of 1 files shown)


ui/xemu-hud.cc
```diff
@@ -1225,6 +1146,7 @@ class CompatibilityReporter
         report.xemu_branch = xemu_branch;
         report.xemu_commit = xemu_commit;
         report.xemu_date = xemu_date;
+
```
abaire: Nit: remove whitespace changes (or explain why they're worth bundling with a functionality change)

ui/xemu-hud.cc
```diff
@@ -1422,6 +1345,100 @@ class CompatibilityReporter
...
 };
 
+class AboutWindow
+{
+public:
+    bool is_open;
+
+private:
+    char build_info_text[256];
+public:
```
abaire: Nit: Add back the newline that was removed here as well?

ui/xemu-hud.cc
```diff
@@ -1422,6 +1345,100 @@ class CompatibilityReporter
...
+    AboutWindow()
+    {
+        CompatibilityReport info;
```
abaire: It doesn't look like you actually need a `CompatabilityReport` instance here at all, since you're populating it in this constructor anyway. E.g., you can just call `get_cpu_info();` and pass it to your `sprintf`, etc...

ui/xemu-hud.cc
```diff
@@ -1422,6 +1345,100 @@ class CompatibilityReporter
     }
 };
 
+class AboutWindow
```
abaire: I'd probably move the helper functions up instead of moving the class down (this will also make it easier to review as GitHub should do a better job of showing the actual changed lines instead of marking everything as changed).

While you're at it, `get_cpu_info` can probably be made static, most like the other helper methods as well if they're not used outside of this file.

ui/xemu-hud.cc
```diff
@@ -1362,6 +1372,7 @@ class CompatibilityReporter
         if (ImGui::TreeNode("Report Details")) {
             ImGui::PushFont(g_fixed_width_font);
             if (dirty) {
+
```
abaire: Remove newline

ui/xemu-hud.cc
```diff
@@ -838,25 +838,62 @@ class SettingsWindow
...
     {
         snprintf(build_info_text, sizeof(build_info_text),
-            "Version: %s\n" "Branch:  %s\n" "Commit:  %s\n" "Date:    %s",
-            xemu_version,  xemu_branch,   xemu_commit,   xemu_date);
-        // FIXME: Show platform
-        // FIXME: Show driver
-        // FIXME: Show BIOS/BootROM hash
-    }
-
+            "Version: %s\n" "Branch: %s\n" "Commit: %s\n" "Date: %s\n"
```
abaire: It looks like spaces have been removed that would've previously aligned all of the values, is this intentional?

ui/xemu-hud.cc
```diff
@@ -838,25 +838,62 @@ class SettingsWindow
...
+       platform_name = "Windows";
+#elif defined(__APPLE__)
+       platform_name = "macOS";
+#else
+        platform_name = "Unknown";
+#endif
+        return platform_name;
+}
+
+#ifndef _WIN32
```
abaire, line 808: Why does this not work on win32? Where is the endif for this ifndef?

ui/xemu-hud.cc
Fabxx: yes the removed spaces are intentional to match original xemu space

ui/xemu-hud.cc
abaire: For posterity: this PR was initially diffbased against an older master, the cpuid include was intentionally removed for Windows.

ui/xemu-hud.cc
```diff
@@ -801,9 +839,10 @@ class AboutWindow
     AboutWindow()
     {
         snprintf(build_info_text, sizeof(build_info_text),
-            "Version: %s\n" "Branch:  %s\n" "Commit:  %s\n" "Date:    %s",
-            xemu_version,  xemu_branch,   xemu_commit,   xemu_date);
-        // FIXME: Show platform
+            "Version: %s\n" "Branch:  %s\n" "Commit:  %s\n" "Date: %s\n"
+            "CPU: %s\n" "OS Platform: %s\n" "OS Version: %s\n",
+            xemu_version,  xemu_branch,   xemu_commit,   xemu_date, get_cpu_info(), get_os_platform(), xemu_get_os_info()
```
abaire: Probably worth wrapping this long line. Also I think you can remove the extra spaces after the commas to save some space (the original code doesn't seem to follow the prevailing style)

ui/xemu-hud.cc
Fabxx: Confirmed that this PR was initially based on 0.6.2 version, now it is rebased on 0.6.3 and everything is as should be

ui/xemu-hud.cc
Fabxx: Taking care of it, thanks

ui/xemu-hud.cc
```diff
@@ -789,28 +789,77 @@ class SettingsWindow
...
+        __get_cpuid(0x80000003, brand+0x4, brand+0x5, brand+0x6, brand+0x7);
+        __get_cpuid(0x80000004, brand+0x8, brand+0x9, brand+0xa, brand+0xb);
+    }
+    cpu_info = (const char *)brand;
+#endif
+    // FIXME: Support other architectures (e.g. ARM)
+    return cpu_info;
+}
+
+static const char *get_gl_vendor(void)
```
abaire: I'd argue it's not worth creating functions for each of these since they're not doing any real work. Just do `(const char*)glGetString(GL_VENDOR)` where you need to get the vendor string.

ui/xemu-hud.cc
```diff
@@ -825,6 +886,14 @@ class AboutWindow
 
         static uint32_t time_start = 0;
         if (ImGui::IsWindowAppearing()) {
+
```
abaire, line 864: Nit: remove newline

ui/xemu-hud.cc
```diff
@@ -825,6 +886,14 @@ class AboutWindow
 
         static uint32_t time_start = 0;
         if (ImGui::IsWindowAppearing()) {
+             
+        snprintf(platform_info_text, sizeof(platform_info_text),
```
abaire: Indent another 4 spaces since this is in an if block.

ui/xemu-hud.cc
```diff
@@ -825,6 +886,14 @@ class AboutWindow
...
+            "CPU: %s\n" "OS Platform: %s\n" "OS Version: %s\n" "Manufacturer: %s\n"
```
abaire: Just make this one string, no need for multiple `"`'s. If you want to break it across two lines to keep it small then that'd be fine, but at most it should be two string literals.

ui/xemu-hud.cc
```diff
@@ -789,28 +789,89 @@ class SettingsWindow
...
+        {
+            snprintf(build_info_text, sizeof(build_info_text),
+            "Version: %s\n" "Branch:  %s\n" "Commit:  %s\n" "Date: %s\n\n",
+            xemu_version, xemu_branch, xemu_commit, xemu_date);
+        }
 
-    ~AboutWindow()
-    {
-    }
+        ~AboutWindow()
```
abaire: I'd just remove this altogether, but at least get rid of the blank line 873

ui/xemu-hud.cc
```diff
@@ -789,28 +789,89 @@ class SettingsWindow
...
-            "Version: %s\n" "Branch:  %s\n" "Commit:  %s\n" "Date:    %s",
-            xemu_version,  xemu_branch,   xemu_commit,   xemu_date);
-        // FIXME: Show platform
-        // FIXME: Show driver
-        // FIXME: Show BIOS/BootROM hash
-    }
+        AboutWindow()
+        {
+            snprintf(build_info_text, sizeof(build_info_text),
+            "Version: %s\n" "Branch:  %s\n" "Commit:  %s\n" "Date: %s\n\n",
```
abaire: Fix indentation.
Remove the double newline after the date.

I'd also make this a single string but not sure why it was formatted this way originally.

ui/xemu-hud.cc
```diff
@@ -789,28 +789,60 @@ class SettingsWindow
...
-        // FIXME: Show BIOS/BootROM hash
-    }
-
-    ~AboutWindow()
-    {
-    }
+        AboutWindow()
+        {
+            snprintf(build_info_text, sizeof(build_info_text),
+                "Version: %s\nBranch: %s\nCommit: %s\nDate: %s\n",
```
abaire: Are the changes to formatting here intentional? The values used to be aligned and there was no newline after the date value.

ui/xemu-hud.cc
```diff
@@ -789,28 +789,60 @@ class SettingsWindow
...
+}
+
 class AboutWindow
 {
 public:
     bool is_open;
 
 private:
-    char build_info_text[256];
+    char build_info_text[600];
```
abaire: Why the increase to 600?

ui/xemu-hud.cc
abaire: Some code got moved around since I added this comment, you can resolve it as it's no longer relevant.

ui/xemu-hud.cc
```diff
@@ -789,28 +789,60 @@ class SettingsWindow
     }
 };
 
+static const char *get_os_platform(void) 
+{
+    const char *platform_name;
+
+#if defined(__linux__)
+        platform_name = "Linux";
```
abaire: nit: Unindent these, they should all be aligned with the `const char` above. You can also just return them, there's no need to assign to a temporary variable.

ui/xemu-hud.cc
```diff
@@ -789,28 +789,60 @@ class SettingsWindow
...
-            xemu_version,  xemu_branch,   xemu_commit,   xemu_date);
-        // FIXME: Show platform
-        // FIXME: Show driver
-        // FIXME: Show BIOS/BootROM hash
-    }
-
-    ~AboutWindow()
-    {
-    }
+        AboutWindow()
```
abaire: Fix indent, this should align with the "char" on line 837

ui/xemu-hud.cc
```diff
@@ -789,28 +789,60 @@ class SettingsWindow
     }
 };
 
+static const char *get_os_platform(void) 
+{
+    const char *platform_name;
+
+    #if defined(__linux__)
```
abaire: (Unindent this too, prevailing style is that #if's are not indented)

ui/xemu-hud.cc
```diff
@@ -861,7 +905,8 @@ class AboutWindow
         ImGui::Dummy(ImVec2(0,40*g_ui_scale));
 
         ImGui::PushFont(g_fixed_width_font);
-        ImGui::InputTextMultiline("##build_info", build_info_text, sizeof(build_info_text), ImVec2(-FLT_MIN, ImGui::GetTextLineHeight() * 6), ImGuiInputTextFlags_ReadOnly);
+        ImGui::InputTextMultiline("##build_info", build_info_text, sizeof(build_info_text), ImVec2(-FLT_MIN, ImGui::GetTextLineHeight() * 5), ImGuiInputTextFlags_ReadOnly);
+        ImGui::InputTextMultiline("##plaftorm_info", platform_info_text, sizeof(platform_info_text), ImVec2(-FLT_MIN, ImGui::GetTextLineHeight() * 8), ImGuiInputTextFlags_ReadOnly);
```
abaire: Check the spelling of "##platform_info"

ui/xemu-hud.cc
Fabxx: done

ui/xemu-hud.cc
Fabxx: done

revix-0:
Current status of the about information:
image

abaire (approved):
lgtm

ui/xemu-hud.cc
```diff
@@ -824,7 +856,19 @@ class AboutWindow
...
         static uint32_t time_start = 0;
-        if (ImGui::IsWindowAppearing()) {
+        if (ImGui::IsWindowAppearing()) { 
+            const char *gl_shader_version = (const char*)glGetString(GL_SHADING_LANGUAGE_VERSION);
+            const char *gl_version = (const char*)glGetString(GL_VERSION);
+            const char *gl_renderer = (const char*)glGetString(GL_RENDERER);
+            const char *gl_vendor = (const char*)glGetString(GL_VENDOR);
+             
+            snprintf(platform_info_text, sizeof(platform_info_text),
+                "CPU: %s\nOS Platform: %s\nOS Version: %s\nManufacturer: %s\n"
```
mborgerson: nit: The fields in the above text box are aligned, but the fields in this box are not aligned. I think either they should both be aligned or not

Fabxx: Took note and updated in the first comment, thanks for pointing out!

mborgerson:
Thanks!

PR #853, opened 2022-04-25
nv2a: Fix renderdoc detection on macOS

abaire:
Renderdoc seems to work on M1 macs now, this fixes the dlopen so xemu can properly detect that it has been launched via renderdoc to enable the improved capture API.

### DIFF
--- a/hw/xbox/nv2a/debug.c
+++ b/hw/xbox/nv2a/debug.c
@@ -62,7 +62,13 @@ void gl_debug_initialize(void)
     }
 
 #ifdef CONFIG_RENDERDOC
-    void* renderdoc = dlopen("librenderdoc.so", RTLD_NOW | RTLD_NOLOAD);
+    const char *renderdoc_lib;
+#ifdef __APPLE__
+    renderdoc_lib = "librenderdoc.dylib";
+#else
+    renderdoc_lib = "librenderdoc.so";
+#endif
+    void* renderdoc = dlopen(renderdoc_lib, RTLD_NOW | RTLD_NOLOAD);
     if (renderdoc) {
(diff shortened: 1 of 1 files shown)


mborgerson:
Can you update this too https://github.com/mborgerson/xemu/blob/87919cfb13e401f381b5cfb7baff034aa02d39c0/meson.build#L1240

abaire:
As far as I know macOS doesn't require any libraries for dlopen so the libdl test isn't actually required (i.e., no meson changes should be needed to get it to work). I did a quick search to confirm my understanding and the top few answers seem to agree (it also builds and correctly enables renderdoc on my machine without any modification to meson).


Unfortunately I found that renderdoc causes a GL error that triggers an assert in xemu when capturing, but I am optimistic that'll be fixable upstream.

mborgerson:
> As far as I know macOS doesn't require any libraries for dlopen so the libdl test isn't actually required (i.e., no meson changes should be needed to get it to work). I did a quick search to confirm my understanding and the top few answers seem to agree (it also builds and correctly enables renderdoc on my machine without any modification to meson).

Good to know!

> Unfortunately I found that renderdoc causes a GL error that triggers an assert in xemu when capturing, but I am optimistic that'll be fixable upstream.

Bummer. Should we report the issue?

abaire:
> Bummer. Should we report the issue?

Unfortunately macOS isn't an officially supported platform so platform specific reports generally seem to get a standard "I'm happy to help educate you about renderdoc internals but will not spend time helping to debug issues". There are a couple folks actively contributing macOS fixes right now though, so there's a good chance it'll get fixed accidentally and if not I'll probably start debugging it myself in the nearish future. 

That said, I'll at least spend some time to figure out exactly what error is being introduced and whether it's reproducible with a trivial app.

PR #868, opened 2022-04-28
Fix Flatpak builds by updating tomlplusplus submodule

j8r:
This fixes the currently failing flatpak builds, 

The cause is the currently used tomlplusplus submodule commit  `0f6a856dc436ef72667830de993f027c132b0529` has itself a `external/dox` submodule entry referred in [.gitmodules](https://github.com/marzer/tomlplusplus/blob/0f6a856dc436ef72667830de993f027c132b0529/.gitmodules), but actually in the project tree no such directory exists at [external/dox](https://github.com/marzer/tomlplusplus/tree/0f6a856dc436ef72667830de993f027c132b0529/external).
This issue has since been resolved in master.


mborgerson:
Why is xemu building everywhere else, but failing on flatpak here? Maybe there is a configuration issue

I'll merge this to get things going.

j8r:
Thanks! That's because the Flatpak build process invoke several git commands. I have no knowledge of its intrinsics, they may be safe guards to ensure a perfectly "sane" git repository, for example for reproducible build by looking up commits for each submodule.

There are no issues with the xemy flatpak manifest itself, it is rather standard.

Other xemu build processes are likely less strict.

j8r:
Thanks! That's because the Flatpak build process invoke several git commands. I have no knowledge of its intrinsics, they may be safe guards to ensure a perfectly "sane" git repository, for example for reproducible build by looking up commits for each submodule.

There are no issues with the xemu flatpak manifest itself, it is rather standard.

Other xemu build processes are likely

j8r:
The build is still in error but now because of https://github.com/TartanLlama/optional, which is used by tomlplusplus.
A submodule use an already existing directory in the git tree, which is invalid. The Flatpak builder is very picky...

I have no idea to fix this since the project is inactive, I'll investigate how to go around that issue.

PR #877, opened 2022-05-01
tomlplusplus: Update submodule - again

j8r:
Continuation of #868 .
I assumed just updating tomlplusplus would have fixed all git-related issues, but another one appeared, related to a library used by this library as a submodule called `TartanLlama/optional`.

After [opening an issue about this library in tomlplusplus](https://github.com/marzer/tomlplusplus/issues/151), also that also it was no longer active and optional for this library, @marzer decided to drop all submodules from the project and [vendor those that are needed](https://github.com/marzer/tomlplusplus/commit/e47ef8cfb2dfa35e48dbc17d49c9a4ad739e7899).

After adding PyYAML to the build instructions, the Flatpak finally compiles :tada:


mborgerson:
Awesome. Thanks @j8r

PR #879, opened 2022-05-03
nv2a: Fix stale texture in surface input and output case

### LINKED ISSUE #790: Splinter cell double agent: Nightvision filter displaying artifacts
### Title

### Bug Description

https://user-images.githubusercontent.com/30447649/159268681-f7a22121-4efa-40bc-b96c-8ef09037aec1.mp4

image

### Expected Behavior

image

image

nightvision should be clean and not show altered visuals

### xemu Version

### System Information

Manjaro Linux
i7-10700 4.80GHz
GTX 970 STRIX 4GB VRAM

abaire:
Fixes #790

[Test](https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/tests/texture_render_update_in_place_tests.cpp)
[HW results](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-Texture_render_update_in_place)

### DIFF
--- a/hw/xbox/nv2a/pgraph.c
+++ b/hw/xbox/nv2a/pgraph.c
@@ -2787,6 +2787,14 @@ DEF_METHOD(NV097, SET_BEGIN_END)
             glEndQuery(GL_SAMPLES_PASSED);
         }
 
+        pg->draw_time++;
+        if (pg->color_binding && pgraph_color_write_enabled(pg)) {
+            pg->color_binding->draw_time = pg->draw_time;
+        }
+        if (pg->zeta_binding && pgraph_zeta_write_enabled(pg)) {
+            pg->zeta_binding->draw_time = pg->draw_time;
+        }
+
         NV2A_GL_DGROUP_END();
     } else {
         NV2A_GL_DGROUP_BEGIN("NV097_SET_BEGIN_END: 0x%x", parameter);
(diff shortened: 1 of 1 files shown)


hw/xbox/nv2a/pgraph.c
```diff
@@ -6251,6 +6245,16 @@ static void pgraph_bind_textures(NV2AState *d)
         palette_data += palette_offset;
         hwaddr palette_vram_offset = palette_data - d->vram_ptr;
 
+        SurfaceBinding *surface = pgraph_surface_get(d, texture_vram_offset);
```
abaire: As an alternative to checking `surface->draw_dirty` I also considered updating `pg->draw_time` in the _end_ handler (it currently only gets updated in the begin part of the BEGIN_END handler) and updating the color/zeta bindings as appropriate (e.g., if they're defined and writable by the draw).

I don't know what the full set of effects would be of doing that versus this (but can confirm that it also passes my pgraph test)

mborgerson: Surfaces remain dirty until written back to RAM, which means these timestamps would not be used effectively when a surface is rendered then used as a texture multiple times. Let's look at fixing the timestamps so a surface is not rendered to a texture more than once if it is not changed

abaire: Sounds good, added a commit with the alternative I've been considering. I still don't follow the intent behind the `draw_time`, specifically why it's incremented on upload at the beginning of the draw rather than after the draw, so I'm not entirely certain this is the right fix either.

Roughly paraphrasing from discord so it's captured here: the current approach to `draw_time` falls over the case where a texture and surface reference the same VRAM and then the texture is used in a subsequent draw with a different target. The surface's draw_time is incremented before the texture is created, then not incremented in the second draw (since a different surface is the target), so we lose sight of the fact that the surface was actually drawn to after the texture was created.

abaire:
EDIT: Removed my comment about this PR exposing an issue with pixel offsetting; I was testing on a branch that had #735 merged. If I remove that code, the pgraph test works.

hw/xbox/nv2a/pgraph.c
```diff
@@ -6251,6 +6253,15 @@ static void pgraph_bind_textures(NV2AState *d)
...
+        TextureBinding *tbind = pg->texture_binding[i];
```
mborgerson: Can we move this check up a bit and skip some extra attribute computation that's not required if this check passes

hw/xbox/nv2a/pgraph.c
```diff
@@ -6146,7 +6154,32 @@ static void pgraph_bind_textures(NV2AState *d)
...
+        if (dma_select) {
+            texture_data = (uint8_t*)nv_dma_map(d, pg->dma_b, &dma_len);
+        } else {
+            texture_data = (uint8_t*)nv_dma_map(d, pg->dma_a, &dma_len);
+        }
+        assert(offset < dma_len);
+        texture_data += offset;
+        hwaddr texture_vram_offset = texture_data - d->vram_ptr;
+
+        hwaddr palette_dma_len;
```
mborgerson, line 6168: We don't consider the palette in the below check, so we could skip mapping it here

abaire: Sorry, I totally missed this comment. I can fix in a followup.

Triticum0:
This pr cause performace regression in Carve

PR #889, opened 2022-05-06
ui: Redesign user interface

mborgerson:
Introduces a new user interface that looks much nicer, is easier to
navigate with controllers, provides more context to users, and is
scalable. Some additional features are included.

* Adds 'popup menu' with actions that can be used easily from controller
* Adds 'main menu', unifying other configuration dialogs
* Adds port-forwarding user interface
* Adds screenshot feature
* Adds volume control feature
* Adds gamepad auto-bind option
* Adds vsync configuration option
* Adds auto UI scaling
* Adds preferred window size selection
* Adds AV pack selection
* Exposes some existing config items in GUI


GXTX:
Size saving is broken, and AV pack should probably be allowed to be swapped while running, Xbox has a SMC interrupt thing to notify the OS and it's on the OS to decide what it wants to do. Looks good overall!

mborgerson:
> Size saving is broken, and AV pack should probably be allowed to be swapped while running, Xbox has a SMC interrupt thing to notify the OS and it's on the OS to decide what it wants to do. Looks good overall!

@GXTX Agree! I'll accept a PR for this. I added smc interrupts a while back so I imagine it's probably an easy add.

config_spec.yml
```diff
@@ -16,26 +18,57 @@ input:
...
     surface_scale:
       type: integer
       default: 1
   window:
-    last_width: integer
-    last_height: integer
+    fullscreen_on_startup: bool
+    startup_size:
+      type: enum
+      values: [last_used, 640x480, 1280x720, 1280x800, 1280x960, 1920x1080, 2560x1440, 2560x1600, 2560x1920, 3840x2560]
```
MasonT8198: I'd consider adding 3840x2160 to this, 2560 is an odd resolution

mborgerson: You are right! That should have been 2160

MasonT8198: Alright, thanks for the fix!

MasonT8198: I'd consider adding 1366x768 and 1600x900 as they are still pretty common laptop resolutions, even on some older machines with dgpus in them

PR #896, opened 2022-05-08
menubar.cc: Add F12 Shortcut to Screenshot function

MasonT8198:


MasonT8198:
I think F12 would be a nice shortcut method, however Print Screen works too!

Issue #800, opened 2022-03-19
Serious Sam: Incorrect display of the map flythrough during the loading screen


abaire:
### Title

https://xemu.app/titles/54540004/#Serious-Sam

### Bug Description

The small viewport rendering the flythrough of the upcoming level is rendered incorrectly, the top part is cut off and displayed under the bulk of the content.

![Screenshot_20220318_212924](https://user-images.githubusercontent.com/448413/159106847-ea49fb0e-3b18-40c7-ad5d-3b0038da77c6.png)

### Expected Behavior

The sky portion of the viewport in the screenshot above should be at the top of the image, not the bottom.

### xemu Version

Version: 0.6.2-82-ghttps://github.com/mborgerson/xemu/commit/https://github.com/mborgerson/xemu/commit/2ff5f23235387a296fc9a5f11943b4dd29bd2837235387a296fc9a5f11943b4dd29bd2837235
Branch: master
Commit: https://github.com/mborgerson/xemu/commit/2ff5f23235387a296fc9a5f11943b4dd29bd2837
Date: Sat Mar 19 03:47:53 AM UTC 2022

### System Information

"cpu": "Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz",
"gl_renderer": "NVIDIA GeForce GTX 1070/PCIe/SSE2",
"gl_shading_la844c-cc0d466e4729.png)
nguage_version": "4.00 NVIDIA via Cg compiler",
"gl_vendor": "NVIDIA Corporation",
"gl_version": "4.0.0 NVIDIA 470.103.01",
"os_platform": "Linux",
"os_version": "Ubuntu 21.10",

abaire:
This is a pvideo rendering.

The issue is that the pvideo shader fails to account for the flipped GL origin, so the rect it is using is calculated based on the bottom of the screen instead of the top (causing it to be shifted and also causing the video texture to wrap around)

PR #908, opened 2022-05-09
nv2a: Match HW behavior when setting window clip

### LINKED ISSUE #907: Behavior when setting `NV097_SET_WINDOW_CLIP_*` does not match HW
### Bug Description

While debugging #735 I kept running into issues where surfaces had uninitialized data along edges. I wrote a [test case](https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/tests/window_clip_tests.cpp) and found that xemu's behavior departs substantially from hardware for all the tests in the case that the test does not explicitly initialize all 8 clip regions. 

Compared to the [HW results](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-Window_clip), on xemu, all of the exclusive mode clips fail to fill any of the region and all of the inclusive mode clips fill the entire region.

NOTE: When all clip regions are explicitly initialized (setting the unused ones to 0,0,0,0) the xemu output matches HW.

### Expected Behavior

On hardware, any set of a clip window updates the clip value for all following clip windows.

E.g., when setting clip0, then clip1, 2-7 will all have the value of clip1. Similarly, if setting clip1 then clip0, all clip values will be set to clip0. (Confirmed by reading back the NV_PGRAPH_WINDOWCLIPX0 and NV_PGRAPH_WINDOWCLIPY0 registers and by inverting ordering of clip rects).

[HW results](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-Window_clip)

### xemu Version

Version:      0.6.6
Branch:       master
Commit:       c2e0067538fa5d24e57effce1adca172db6e8117
Date:         Mon May  9 02:38:25 AM UTC 2022

### System Information

OS Platform:  Linux
OS Version:   Ubuntu 21.10
GPU Model:    NVIDIA GeForce GTX 1070/PCIe/SSE2
Driver:       4.0.0 NVIDIA 470.103.01

### Additional Context

_No response_

abaire:
Fixes #907 

[Test](https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/tests/window_clip_tests.cpp)
[HW results](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-Window_clip)

### DIFF
--- a/hw/xbox/nv2a/pgraph.c
+++ b/hw/xbox/nv2a/pgraph.c
@@ -1507,13 +1507,17 @@ DEF_METHOD(NV097, SET_WINDOW_CLIP_TYPE)
 DEF_METHOD_INC(NV097, SET_WINDOW_CLIP_HORIZONTAL)
 {
     int slot = (method - NV097_SET_WINDOW_CLIP_HORIZONTAL) / 4;
-    pg->regs[NV_PGRAPH_WINDOWCLIPX0 + slot * 4] = parameter;
+    for (; slot < 8; ++slot) {
+        pg->regs[NV_PGRAPH_WINDOWCLIPX0 + slot * 4] = parameter;
+    }
 }
 
 DEF_METHOD_INC(NV097, SET_WINDOW_CLIP_VERTICAL)
 {
     int slot = (method - NV097_SET_WINDOW_CLIP_VERTICAL) / 4;
-    pg->regs[NV_PGRAPH_WINDOWCLIPY0 + slot * 4] = parameter;
+    for (; slot < 8; ++slot) {
+        pg->regs[NV_PGRAPH_WINDOWCLIPY0 + slot * 4] = parameter;
+    }
 }
 
 DEF_METHOD(NV097, SET_ALPHA_TEST_ENABLE)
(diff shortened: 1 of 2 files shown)


abaire:
Found an additional bug in xemu wrt. dealing with clip regions larger than the surface size (there's a signed/unsigned mismatch, likely something else as well). Pulling that fix into this PR so the improved tests pass

Example of the bug: (the top line ends up at the bottom of the image)
Screenshot_20220509_071809

Depending on the specifics, this may also explain #800 as it's very similar behavior.

abaire:
Looks like that wrapping bug is was the final display only (the surfaces emulating the guest are fine, the error is introduced in the last display blit) and has been fixed with the new UI.

hw/xbox/nv2a/pgraph.c
```diff
@@ -1507,13 +1507,17 @@ DEF_METHOD(NV097, SET_WINDOW_CLIP_TYPE)
 DEF_METHOD_INC(NV097, SET_WINDOW_CLIP_HORIZONTAL)
 {
     int slot = (method - NV097_SET_WINDOW_CLIP_HORIZONTAL) / 4;
-    pg->regs[NV_PGRAPH_WINDOWCLIPX0 + slot * 4] = parameter;
+    for (; slot < 8; ++slot) {
+        pg->regs[NV_PGRAPH_WINDOWCLIPX0 + slot * 4] = parameter;
```
mborgerson, line 1511: Nice discovery! I confirmed this with your test

mborgerson (changes requested):
Please remove the updated submodule and I'll merge

abaire:
> Please remove the updated submodule and I'll merge

Oops, fixed. My script to auto rebase my PR branches apparently doesn't account for submodule update properly.

Issue #801, opened 2022-03-19
Serious Sam: Skybox is broken


abaire:
### Title

https://xemu.app/titles/54540004/#Serious-Sam

### Bug Description

The skybox in the game is only half rendered when entering the first level.
![Screenshot_20220318_213310](https://user-images.githubusercontent.com/448413/159106903-90188d6a-5140-46cd-960b-86d3980595af.png)

Rotating the view will cause it to display a different, glitchy skybox (or all black, or half correct half not, it's unpredictable).

### Expected Behavior

The skybox should not glitch out.

### xemu Version

Version: 0.6.2-82-ghttps://github.com/mborgerson/xemu/commit/https://github.com/mborgerson/xemu/commit/2ff5f23235387a296fc9a5f11943b4dd29bd2837235387a296fc9a5f11943b4dd29bd2837235
Branch: master
Commit: https://github.com/mborgerson/xemu/commit/2ff5f23235387a296fc9a5f11943b4dd29bd2837
Date: Sat Mar 19 03:47:53 AM UTC 2022

### System Information

"cpu": "Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz",
"gl_renderer": "NVIDIA GeForce GTX 1070/PCIe/SSE2",
"gl_shading_la844c-cc0d466e4729.png)
nguage_version": "4.00 NVIDIA via Cg compiler",
"gl_vendor": "NVIDIA Corporation",
"gl_version": "4.0.0 NVIDIA 470.103.01",
"os_platform": "Linux",
"os_version": "Ubuntu 21.10",

abaire:
SSam uses the fixed function pipeline for rendering.

In renderdoc I see geometry being rendered clearly out of place up in the sky, and this appears to be what blanks out the skybox. For the first draw:



```
Position:
|x|y|z|
|--|--|--|
|-128|	-4|	-128|
|128|	-4|	-128|
|128|	0|	-128|
|-128|	0|	-128|
|-128|	-4|	-128|
|128|	0|	-128|

compositeMat:
c[0] 319.9999695, 0.00, -320.00, 0.00        float4
c[1] 0.00, -319.999939, -240.00, 0.00        float4
c[2] 0.00, 0.00, -1.6777220E+07, -419431.53125 float4
c[3] 0.00, 0.00, -1.00, 0.00                 float4


invViewport {0.003125, 0.00, 0.00, -0.9983398}
{0.00, -0.0041667, 0.00, 0.9977865}
[...]

PR #913, opened 2022-05-11
nv2a: Adjust NaN handling to be similar to HW

### LINKED ISSUE #365: Blackened Texture Issues and Documentation
### Title Information

* Titles: Otogi FS-002 [46530002]
* https://xemu.app/titles/46530002/

* Titles: Otogi: Hyakki Toubatsu Emaki FS-004 [46530004]
* https://xemu.app/titles/46530004/

### Bug Description

Otogi 1: The only major issue I have encountered for the first game is a failure to load the character/weapon textures properly. Instead Raikoh has some sort of shadow texture over their model.

1. Start the game.
2. Skip any cutscenes.
3. Start any stage and observe the main character. 

https://cdn.discordapp.com/attachments/680221390359888154/856388676602298368/Otogi_1_1.png
https://cdn.discordapp.com/attachments/680221390359888154/856388690561204244/Otogi_1_2.png

Otogi 2: The second game has more issues, starting with major slowdown in nearly every part of the game. The menus, saving, loading, all of it is locked at 5 fps. The only thing unaffected is the stage gameplay, which has normal slowdown similar to the first game.

The other major issue is some stages not loading environment textures properly, resulting in the entire level being covered in shadow textures. This can be observed in the starting stage...
https://cdn.discordapp.com/attachments/680221390359888154/856389461860810762/Otogi_2_1.png

However, some stages load properly...
https://cdn.discordapp.com/attachments/680221390359888154/856389890529558548/Otogi_2_2.png
So I will be doing further testing to see which stages load and which don't to find out if there is any pattern. Also note that the character/weapon texture load fail from the first game has persisted.

Steps to reproduce these issues:
1. Start a new game, you will have to go through game, sound, and controls settings, then save before beginning gameplay.
2. Note the slowdown to 5 fps during these menus and the save process.
3. Skip any cutscenes.
4. Observe the first stage.

### Expected Behavior

For Otogi 1, Raikoh and his weapon(s) should obviously have their textures.

For Otogi 2, there shouldn't be any slowdown in any of the menus. The menus in Otogi 1 worked fine, and I am fairly certain the games use the same engine, so I have no idea what could have been changed between games to be causing this problem. As for the missing textures, obviously the stages should all have their textures. I am not sure why some stages work and others don't, and plan to investigate further.

### Version

* Version: v0.5.4-7-ga7f3f8f212
* Last-known Working: Unknown

### System Information

| Field | Value |
|-------|-------|
| OS			| Windows 10 Pro			| 
| CPU			| AMD FX(tm)-8350 Eight-Core Processor			| 
| Graphics Device	| Radeon RX 580		| 
| Graphics Driver	| Radeon 21.4.1		|

abaire:
Fixes #365 by forcing treatment of NaN values generated in the vertex shader to follow the same pattern as HW.

Unfortunately this does not address the fact that on HW, `+/- NaN` `*` `0` (or very close to 0) should be `0`. This difference in behavior is also responsible for #1008.

This also does not address the same issue for `+/- inf` and the shader version currently used by xemu explicitly notes that operations on inf/NaN's are undefined (which may explain why Otogi textures are not blacked out on M1 mac).

[Tests](https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/tests/attribute_float_tests.cpp)
[HW results](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-Attrib_float)

### DIFF
--- a/hw/xbox/nv2a/shaders.c
+++ b/hw/xbox/nv2a/shaders.c
@@ -853,6 +853,23 @@ GLSL_DEFINE(texMat3, GLSL_C_MAT4(NV_IGRAPH_XF_XFCTX_T3MAT))
                                i, i);
         }
     }
+    mstring_append(header,
+                   "/* Converts the input to vec4, pads with last component */\n"
+                   "vec4 _in(float v) { return vec4(v); }\n"
+                   "vec4 _in(vec2 v) { return v.xyyy; }\n"
+                   "vec4 _in(vec3 v) { return v.xyzz; }\n"
+                   "vec4 _in(vec4 v) { return v.xyzw; }\n"
+                   "#define FixNaN(src, mask) _FixNaN(_in(src)).mask\n"
+                   "vec4 _FixNaN(vec4 src)\n"
+                   "{\n"
+                   "  bvec4 nans = isnan(src);\n"
+                   "  if (!any(nans)) {\n"
+                   "    return src;\n"
+                   "  }\n"
+                   "  ivec4 signs = floatBitsToInt(src);\n"
+                   "  vec4 negative = vec4(lessThan(signs, ivec4(0)));"
+                   "  return mix(src, mix(vec4(1.0), vec4(0.0), negative), vec4(nans));\n"
+                   "}\n");
     mstring_append(header, "\n");
 
     MString *body = mstring_from_str("void main() {\n");
@@ -964,15 +981,15 @@ GLSL_DEFINE(texMat3, GLSL_C_MAT4(NV_IGRAPH_XF_XFCTX_T3MAT))
     /* Set outputs */
     const char *shade_model_mult = state->smooth_shading ? "vtx_inv_w" : "vtx_inv_w_flat";
     mstring_append_fmt(body, "\n"
-                      "  vtxD0 = clamp(oD0, 0.0, 1.0) * %s;\n"
-                      "  vtxD1 = clamp(oD1, 0.0, 1.0) * %s;\n"
-                      "  vtxB0 = clamp(oB0, 0.0, 1.0) * %s;\n"
-                      "  vtxB1 = clamp(oB1, 0.0, 1.0) * %s;\n"
-                      "  vtxFog = oFog.x * vtx_inv_w;\n"
-                      "  vtxT0 = oT0 * vtx_inv_w;\n"
-                      "  vtxT1 = oT1 * vtx_inv_w;\n"
-                      "  vtxT2 = oT2 * vtx_inv_w;\n"
-                      "  vtxT3 = oT3 * vtx_inv_w;\n"
+                      "  vtxD0 = clamp(FixNaN(oD0, xyzw), 0.0, 1.0) * %s;\n"
+                      "  vtxD1 = clamp(FixNaN(oD1, xyzw), 0.0, 1.0) * %s;\n"
+                      "  vtxB0 = clamp(FixNaN(oB0, xyzw), 0.0, 1.0) * %s;\n"
+                      "  vtxB1 = clamp(FixNaN(oB1, xyzw), 0.0, 1.0) * %s;\n"
+                      "  vtxFog = FixNaN(oFog, x) * vtx_inv_w;\n"
+                      "  vtxT0 = FixNaN(oT0, xyzw) * vtx_inv_w;\n"
+                      "  vtxT1 = FixNaN(oT1, xyzw) * vtx_inv_w;\n"
+                      "  vtxT2 = FixNaN(oT2, xyzw) * vtx_inv_w;\n"
+                      "  vtxT3 = FixNaN(oT3, xyzw) * vtx_inv_w;\n"
                       "  gl_Position = oPos;\n"
                       "  gl_PointSize = oPts.x;\n"
                       "\n"
--- a/hw/xbox/nv2a/vsh.c
+++ b/hw/xbox/nv2a/vsh.c
@@ -614,17 +614,6 @@ static const char* vsh_header =
      * https://www.opengl.org/registry/specs/NV/vertex_program1_1.txt
      */
     "\n"
-//QQQ #ifdef NICE_CODE
-    "/* Converts the input to vec4, pads with last component */\n"
-    "vec4 _in(float v) { return vec4(v); }\n"
-    "vec4 _in(vec2 v) { return v.xyyy; }\n"
-    "vec4 _in(vec3 v) { return v.xyzz; }\n"
-    "vec4 _in(vec4 v) { return v.xyzw; }\n"
-//#else
-//    "/* Make sure input is always a vec4 */\n"
-//   "#define _in(v) vec4(v)\n"
-//#endif
-    "\n"
     "#define INFINITY (1.0 / 0.0)\n"
     "\n"
     "#define MOV(dest, mask, src) dest.mask = _MOV(_in(src)).mask\n"


abaire:
Hey @NZJenkins I wasn't sure if you were already looking at a fix, but I started to suspect this might be partly to blame for #801 so I threw together a partial fix to unblock my debugging there. 

This is mostly a proof of concept at the moment as it doesn't fully match the HW behavior and because some investigation will need to be done on the performance impact (and possibly some optimization)

hw/xbox/nv2a/shaders.c
```diff
@@ -769,6 +769,27 @@ STRUCT_VERTEX_DATA);
...
+                   "vec4 _in(vec4 v) { return v.xyzw; }\n"
+                   "#define FixNaN(dest, mask, src) dest.mask = _FixNaN(_in(src)).mask\n"
+                   "vec4 _FixNaN(vec4 src)\n"
+                   "{\n"
+                   "  bvec4 nans = isnan(src);\n"
+                   "  if (!any(nans)) {\n"
+                   "    return src;\n"
+                   "  }\n"
+                   "  ivec4 signs = floatBitsToInt(src);\n"
+                   "  vec4 dest;\n"
```
abaire: This stuff can be vectorized, assuming the logic will stay the same:

`"  return mix(src, mix(vec4(1.0), vec4(0.0), lessThan(signs, ivec4(0))), nans);\n"`

ko81e24wy:
glitch in NGB this PR
may relate to issue  https://github.com/mborgerson/xemu/issues/783
2022-05-11 11-48-33 mp4_20220511_115252406
same glitch in DOA2U
xemu_2022_05_11_12_52_47_379

abaire:
@ko81e24wy where is the nearest save point so I can reproduce that?

ko81e24wy:
> @ko81e24wy where is the nearest save point so I can reproduce that?

I think you just need to start a new game

abaire:
> > @ko81e24wy where is the nearest save point so I can reproduce that?
> 
> I think you just need to start a new game

There was a stupid mistake in my code that is now fixed, can you test again when you have a chance?

ko81e24wy:
I am just going to work and test with a combine version with this PR merged
that white glitch was gone,and I dont know if this PR relate to https://github.com/mborgerson/xemu/issues/783
xemu-2022-05-17-07-42-08

abaire:
@ko81e24wy cool, thanks for double checking!

I have not tried to debug 783 yet, so if this PR happens to fix the missing lighting that'd be a nice bonus. I'm not exactly sure what that screenshot should look like compared to HW though; are you saying it looks more correct with this PR than before?

ko81e24wy:
> @ko81e24wy cool, thanks for double checking!
> 
> I have not tried to debug 783 yet, so if this PR happens to fix the missing lighting that'd be a nice bonus. I'm not exactly sure what that screenshot should look like compared to HW though; are you saying it looks more correct with this PR than before?

that white issue remind me of that light effect,the last PR just as before with that effect missing.

mysteria25:
Was hoping this would fix the black graphics of Inside Pitch 2003 but guess that is a separate issue from this pr since that game still looks like this: 

inside pitch 2003

abaire:
> Was hoping this would fix the black graphics of Inside Pitch 2003 but guess that is a separate issue from this pr since that game still looks like this:

@mysteria25 as far as I can see from the compatibility page there is no bug tracking this problem, please file one and it'll get looked into eventually (PRs are definitely not the place to report untracked issues). Grabbed by the Ghoulies has a black texture problem related to texture compression; there are many ways for things to fail so it's entirely possible what you're seeing with that game is something novel that will need a specific fix.

abaire:
This doesn't entirely match HW behavior on my GTX 1070 but it is much closer and fixes #365 despite the remaining departure around infinite values and zeros.

hw/xbox/nv2a/shaders.c
```diff
@@ -769,6 +769,22 @@ STRUCT_VERTEX_DATA);
...
+                   "vec4 _in(vec4 v) { return v.xyzw; }\n"
+                   "#define FixNaN(src, mask) _FixNaN(_in(src)).mask\n"
+                   "vec4 _FixNaN(vec4 src)\n"
+                   "{\n"
+                   "  bvec4 nans = isnan(src);\n"
+                   "  if (!any(nans)) {\n"
+                   "    return src;\n"
+                   "  }\n"
+                   "  ivec4 signs = floatBitsToInt(src);\n"
+                   "  return mix(src, mix(vec4(1.0), vec4(0.0), lessThan(signs, ivec4(0))), nans);\n"
```
mborgerson: * The sign bit vectorization is neat, nice :)
* So "negative NaN" (msb set) gets 0 and "positive NaN" gets 1.0?
* Seems odd because you mentioned 0*NaN should == 0? ~~So under what conditions do you expect it to be 1?~~(nvm, I see [your comment](https://github.com/mborgerson/xemu/issues/365#issuecomment-1123121908) about it) I don't think this it makes sense to generalize this based on sign bit alone here. More about this thought below.
* It looks like `mix(genBType)` was added in GL4.5 so this might fail in a 4.0 context. Does it work on your M1 Mac? You could cast the boolean to a float

As far as I can see in the general case, I think this NaN handling is not more correct than just using 0.0 in the event we do get a NaN (which would force all platforms to behave the same at least, but might be unnecessary if we don't have issues cataloged for it). Is there another case in vsh output that this solves for, beyond fog?

To fix this fog issue for Otogi (and now at least 1 other game (Triggerman) that is affected in the same way that I've seen), I do think it's a fair compromise to have specialized fog NaN handling (like we did with Inf previously); and if this is only intended to fix the fog issue, I think we should just stick with the specialized handling for that case.

The fishing game is a different, interesting case that appears to be setting light colors to NaN, so for this second specialized case I think it makes sense to interpret NaN as 1.0 before performing lighting calculations, so lighting is calculated correctly. Perhaps it makes sense to generalize NaNs in the uniform spaces to 1.0 for such cases but I'd have to think more about it. Needs more investigation to understand how hardware is handling that case.

abaire: Partial response:

> As far as I can see in the general case, I think this NaN handling is not more correct than just using 0.0 in the event we do get a NaN (which would force all platforms to behave the same at least, but might be unnecessary if we don't have issues cataloged for it). Is there another case in vsh output that this solves for, beyond fog?

The tester itself is using the diffuse channel and the xemu output does not match HW, if that's what you're asking. I don't know of any game where this makes a difference offhand, but that doesn't mean there isn't one with graphics glitching attributable to this handling.

An example difference with all three cases in one place:

master:



This branch:



HW:


> To fix this fog issue for Otogi (and now at least 1 other game (Triggerman) that is affected in the same way that I've seen), I do think it's a fair compromise to have specialized fog NaN handling (like we did with Inf previously); and if this is only intended to fix the fog issue, I think we should just stick with the specialized handling for that case.

Otogi's fog is the only case I know of (I assume someone has already looked into Triggerman and determined that just applying the fix to the fog fixes that game? I don't own that one to test). I guess the issue is balancing the performance impact of doing this check against everything in an attempt to be closer to HW behavior versus the time a contributor might end up spending tracing some glitch back to a FIXME in the shader code.

> The fishing game is a different, interesting case that appears to be setting light colors to NaN, so for this second specialized case I think it makes sense to interpret NaN as 1.0 before performing lighting calculations, so lighting is calculated correctly. Perhaps it makes sense to generalize NaNs in the uniform spaces to 1.0 for such cases but I'd have to think more about it. Needs more investigation to understand how hardware is handling that case.

Hopefully @Triticum0 wouldn't mind splitting [Lakemasters](https://github.com/mborgerson/xemu/issues/365#issuecomment-974732862) into a new issue given that it sounds like it's a similar symptom due to a different bug?

abaire: Verified that the shader works on macOS 12.4 (M1)

Triticum0: @abaire Lakemaster is fixed with this PR
xemu-2022-06-02-20-51-16

Fabxx:
image
current handling causes this bug where the character is highlighted in proximity of lights

abaire:
>  current handling causes this bug where the character is highlighted in proximity of lights

Looks like limiting the NaN fix to fog will fix that regression and align with Matt's suggestion, will update and re-push in a bit.

hw/xbox/nv2a/shaders.c
abaire: #1008 is due to a similar cause, INF * 0 should == 0 but is instead NaN (at least on my machine and presumably other nvidia systems since the original report was on a 3060).

In this case the incorrect number causes a chain of errors and catching it at the very end where the outputs are assigned would also lead to incorrect behavior; in the test cases I examined for #1008 the output diffuse should be set to values between 0 and 1 so blindly detecting the NaN and setting to either would do the wrong thing.

Fabxx:
The highlight problem seems to be fixed with #1045

ko81e24wy:
I test the lastest build of this PR,and it crash when I run xemu.
[xemu.log](https://github.com/mborgerson/xemu/files/9009708/xemu.log)

abaire:
> I test the lastest build of this PR,and it crash when I run xemu. [xemu.log](https://github.com/mborgerson/xemu/files/9009708/xemu.log)

Bad merge, should be fixed now. Thanks for letting me know!

Triticum0:
@abaire could you rebase this pr with the work you did on Match inv_w qualifier to attribute qualifier, Thanks

Kohryujin:
Can we get another pull request or rebase or however that works so there is a working download available for this branch for testing? The one from July is expired and no longer available.

TheHollows:
Any update on when this will be merged?

mborgerson:
@TheHollows 

> Any update on when this will be merged?

When it's ready

hw/xbox/nv2a/shaders.c
```diff
@@ -853,6 +853,23 @@ GLSL_DEFINE(texMat3, GLSL_C_MAT4(NV_IGRAPH_XF_XFCTX_T3MAT))
...
+                   "#define FixNaN(src, mask) _FixNaN(_in(src)).mask\n"
+                   "vec4 _FixNaN(vec4 src)\n"
+                   "{\n"
+                   "  bvec4 nans = isnan(src);\n"
+                   "  if (!any(nans)) {\n"
+                   "    return src;\n"
+                   "  }\n"
+                   "  ivec4 signs = floatBitsToInt(src);\n"
+                   "  vec4 negative = vec4(lessThan(signs, ivec4(0)));"
+                   "  return mix(src, mix(vec4(1.0), vec4(0.0), negative), vec4(nans));\n"
```
antoniodesousa, line 871:
```suggestion
                   "  if (isnan(src.x)) {\n"
                   "    src.x = step(0.0, sign(src.x));\n"
                   "  }\n"
                   "  if (isnan(src.y)) {\n"
                   "    src.y = step(0.0, sign(src.y));\n"
                   "  }\n"
                   "  if (isnan(src.z)) {\n"
                   "    src.z = step(0.0, sign(src.z));\n"
                   "  }\n"
                   "  if (isnan(src.w)) {\n"
                   "    src.w = step(0.0, sign(src.w));\n"
                   "  }\n"
                   "  return src;\n"
```

Hi @abaire, may a suggest to change the code to what I wrote? I tried your current solution but it didn't fix anything on my PC. It behaves exactly as master. From my tests, my change behaves almost exactly as real hardware. Almost because there's a slight difference in the -NaNQ_NaNQ and -NaNS_NaNS tests cases in the N/A column, but I'm pretty sure that it's safe to ignore.

I tested a lot of games I couldn't find any regressions in any of them. Games that are fixed by this change:
- Otogi
- Otogi 2
- Pro Cast: Sports Fishing Game
- Trigger Man

## HW - PC (AMD RX 6600M )

Attrib_float_0_1
Attrib_float_0_8
Attrib_float_-1_1
Attrib_float_-8_1
Attrib_float_-INF_INF
Attrib_float_-Max_Max
Attrib_float_-MaxSN_MaxSN
Attrib_float_-Min_Min
Attrib_float_-MinN_MinN
Attrib_float_-NaNq_NaNq
Attrib_float_-NaNs_NaNs

abaire: You're very welcome to take this one over, I've got some other projects going at the moment and don't have much time for xemu. It'd be good to test on an nvidia card as well, I'm not sure if glsl guarantees NaN behavior so it may be manufacturer or device specific.

medievil1: I have a gtx 3060 and can test nvidia,  just need the build and the test program...

antoniodesousa: That's too bad to heard. I will need to open a new PR then but I will link it to this one. Once is merged we can close both. Thanks for your research and I hope that you are back to xemu in the future.

@medievil1 you can download the build once I open the new PR.

mborgerson:
Thank you for the PR! Since this work is continuing in #1780 with original author sign off, I'll close this PR now

PR #927, opened 2022-05-18
nv2a: Invalidate GL textures after surface blit

### LINKED ISSUE #786: Splinter cell double agent: steam heat waves effect not handled correctly without goggles

abaire:
Fixes #786 after a change to PR #835 prevented the fix from working (but allowed tests to pass). 

The original code in #835 worked because it over-aggressively forced the texture bindings to be regenerated, when this was scaled back in code review, the fix was left in a state where the stale texture binding was still being used, even though the surface binding (for the same hwaddr) was correctly updated.

### DIFF
--- a/hw/xbox/nv2a/pgraph.c
+++ b/hw/xbox/nv2a/pgraph.c
@@ -397,7 +397,7 @@ static void pgraph_download_surface_data_to_buffer(NV2AState *d,
                                                    uint8_t *pixels);
 static void pgraph_upload_surface_data(NV2AState *d, SurfaceBinding *surface, bool force);
 static bool pgraph_check_surface_compatibility(SurfaceBinding *s1, SurfaceBinding *s2, bool strict);
-static bool pgraph_check_surface_to_texture_compatibility(SurfaceBinding *surface, TextureShape *shape);
+static bool pgraph_check_surface_to_texture_compatibility(const SurfaceBinding *surface, const TextureShape *shape);
 static void pgraph_render_surface_to_texture(NV2AState *d, SurfaceBinding *surface, TextureBinding *texture, TextureShape *texture_shape, int texture_unit);
 static void pgraph_update_surface_part(NV2AState *d, bool upload, bool color);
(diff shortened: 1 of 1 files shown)


hw/xbox/nv2a/pgraph.c
```diff
@@ -397,7 +397,7 @@ static void pgraph_download_surface_data_to_buffer(NV2AState *d,
                                                    uint8_t *pixels);
 static void pgraph_upload_surface_data(NV2AState *d, SurfaceBinding *surface, bool force);
 static bool pgraph_check_surface_compatibility(SurfaceBinding *s1, SurfaceBinding *s2, bool strict);
-static bool pgraph_check_surface_to_texture_compatibility(SurfaceBinding *surface, TextureShape *shape);
+static bool pgraph_check_surface_to_texture_compatibility(const SurfaceBinding *surface, const TextureShape *shape);
```
abaire, line 400: I can split this into a separate PR if desired; having these mutable made me worry that the check might've been doing something interesting when I was trying to pin down the regression from the original PR.

mborgerson: No worries

hw/xbox/nv2a/pgraph.c
```diff
@@ -887,7 +887,9 @@ static void pgraph_image_blit(NV2AState *d)
             surf_dest->download_pending = false;
             surf_dest->draw_dirty = false;
         }
+        surf_dest->draw_time = pg->draw_time;
```
mborgerson: How about we set a surface's draw time in the upload handler instead

PR #942, opened 2022-05-20
nv2a: Use width to calculate swizzled surface pitch

mborgerson:
I forget the case that prompted this change. Would like a test case to confirm hardware behavior wrt pitch on swizzled surfaces

### DIFF
--- a/hw/xbox/nv2a/pgraph.c
+++ b/hw/xbox/nv2a/pgraph.c
@@ -5741,7 +5741,8 @@ static void pgraph_populate_surface_binding_entry_sized(NV2AState *d,
     entry->vram_addr = dma.address + surface->offset;
     entry->width = width;
     entry->height = height;
-    entry->pitch = surface->pitch;
+    entry->pitch = entry->swizzle ? (entry->width * entry->fmt.bytes_per_pixel)
+                                  : surface->pitch;
     entry->size = height * surface->pitch;
     entry->upload_pending = true;
     entry->download_pending = false;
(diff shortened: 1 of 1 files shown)


abaire:
I added a test case for this to the pgraph tester:
[Test](https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/tests/surface_pitch_tests.cpp)
[HW results](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-Surface_pitch)

At a high level, it sets up a green linear surface, then sets up a smaller swizzled surface at the same address and renders a checkerboard pattern into it. The original green surface is used to determine the in-memory layout of the smaller draw; the final output is rendered as a linear texture using the full surface size.

It does this with 4 permutations of `NV097_SET_SURFACE_FORMAT_WIDTH/HEIGHT` + `NV097_SET_SURFACE_PITCH` to see if the results differ, they do not.


Unfortunately with or without your fix, it crashes xemu (looks like there's an undersized malloc someplace, I haven't attempted to debug yet). The triggering case seems to be the [case where the surface is set to the outer texture size and the pitch is set to the inner texture size](https://github.com/abaire/nxdk_pgraph_tests/blob/2387ff17d0a3dced3e3fa9e08d5ddebd632850f6/src/tests/surface_pitch_tests.cpp#L147). 

On HW this case works fine and the others produce output that differs from xemu; HW appears to completely ignore pitch when dealing with swizzled surfaces.

abaire:
I forgot that I wrote the test for this PR and ended up creating #1153 to fix the undersized malloc. Once that's merged this should probably be retested to verify that it matches HW behavior.

Triticum0:
Update?

PR #963, opened 2022-05-24
nv2a: Implement shadow comparison for PROJECT2D mode

### LINKED ISSUE #958: Shadow map for PROJECT2D not implemented
### Title

https://xemu.app/titles/4d530005/#Amped-Freestyle-Snowboarding
https://xemu.app/titles/4c410005/#Star-Wars-Jedi-Starfighter

### Bug Description

Shadow mapping was recently introduced, but it is not complete yet, and some games using incomplete parts of this feature now assert, regressing from playable to not playable.

PROJECT2D should be an easy add, and prioritized to bring these games back to playable state.

### Expected Behavior

Implemented

### xemu Version

0.7.17

### System Information

N/A

### Additional Context

_No response_

abaire:
Fixes #958 

[Updated test](https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/tests/texture_shadow_comparator_tests.cpp)
[Updated HW results](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-Texture_shadow_comparator)

### DIFF
--- a/hw/xbox/nv2a/psh.c
+++ b/hw/xbox/nv2a/psh.c
@@ -571,6 +571,8 @@ static const char* get_sampler_type(enum PS_TEXTUREMODES mode, const PshState *s
         return NULL;
 
     case PS_TEXTUREMODES_PROJECT2D:
+        return state->rect_tex[i] ? sampler2DRect : sampler2D;
+
     case PS_TEXTUREMODES_BUMPENVMAP:
     case PS_TEXTUREMODES_BUMPENVMAP_LUM:
     case PS_TEXTUREMODES_DOT_ST:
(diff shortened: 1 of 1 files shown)


hw/xbox/nv2a/psh.c
```diff
@@ -771,51 +806,31 @@ static MString* psh_convert(struct PixelShader *ps)
...
-                                       i, i, i, i, i,
-                                       i, i, i, i, i,
-                                       i, i, i, i, i);
-
-                    const char *comparison = shadow_comparison_map[ps->state.shadow_depth_func];
-                    mstring_append_fmt(vars,
-                                       "vec4 t%d = vec4(t%d_depth.x %s pT%d.z ? 1.0 : 0.0);\n",
-                                       i, i, comparison, i);
-                }
+                psh_append_shadowmap2d(ps, i, vars);
```
mborgerson: I suspect 2D and 3D are not identical

abaire: You're right, the test results are different. Amped looked reasonable, I should've looked at the actual detailed results from HW.

abaire: Examining the results of the tests where I set the depth values explicitly, it looks like the PROJECT2D mode just compares against 0 (versus 3D where it compares against the texture Z coord).

Updated to reflect that, and the raw results now match hardware for both modes. The projected modes still differ more or less across the board, so there is still some nuance that needs to be handled.

Did a spot check in Amped and the shadow looks reasonable to me with this new implementation (but it looked reasonable before as well).

PR #965, opened 2022-05-24
Minimal change to expose the keyboard config in the toml file

7oxicshadow:
This change should be transparent to the end user but will expose the controller keyboard mapping in the .toml file for any advanced users who would like to have a non standard keyboard layout.

I have tired to keep this change as minimal as possible so that even if it doesn't get accepted into master, someone else may be able to use it as a base for when the UI is updated to allow configurable keyboard support.


config_spec.yml
```diff
@@ -23,6 +23,83 @@ input:
     type: bool
     default: true
   background_input_capture: bool
+  keyboard_mappings_dig:
+    KEY_CTRLR_DIG_A:
```
mborgerson: Config is already structured so

```
keyboard_mappings_dig:
    KEY_CTRL_DIG_A:
    ...
keyboard_mappings_ana:
    KEY_CTRL_ANA_LSTICK_UP
    ....
```

is more verbose than necessary. Suggest something like:

```
keyboard_controller_scancode_map:
  a:
    type: integer
    default: 4
    ...
```

ui/xemu-input.c
```diff
@@ -112,6 +114,23 @@ void xemu_input_init(void)
     new_con->name = "Keyboard";
     new_con->bound = -1;
 
+    //init the keyboard digital button map from the toml config
```
mborgerson: * Comment is not necessary
* Need bounds checking for eventual use in `kbd[]` below

ui/xemu-input.c
```diff
@@ -336,23 +338,23 @@ void xemu_input_update_sdl_kbd_controller_state(ControllerState *state)
     S     F
```
mborgerson: This comment and the one below are now possibly inaccurate

config_spec.yml
7oxicshadow: Done

ui/xemu-input.c
7oxicshadow: Comment removed. Bounds checking added with printf to inform the user of any rejected scancodes

ui/xemu-input.c
7oxicshadow: Comments removed from here and moved to the init function where a comment for each input has been added to show default bindings

ui/xemu-input.c
```diff
@@ -112,6 +116,48 @@ void xemu_input_init(void)
...
+    for (i=0; i<25; i++){
+        /* SDL allocates an array of size SDL_NUM_SCANCODES
+           that can be indexed to get the pressed states of
+           each key. We need to validate that none of the
+           assignments in the toml file are outside of this
+           range */
+
+        if( (sdl_kbd_scancode_map[i] < SDL_SCANCODE_UNKNOWN) ||
+            (sdl_kbd_scancode_map[i] >= SDL_NUM_SCANCODES) ) {
+            printf("WARNING: Keyboard Controller Scancode Out Of Range (%d) : Disabled\n", sdl_kbd_scancode_map[i]);
```
mborgerson: * Error messages should be logged to stderr
* Capitalization: `Keyboard Controller Scancode Out Of Range` => `Keyboard controller map scancode out of range`

ui/xemu-input.c
```diff
@@ -94,8 +94,12 @@ static const char **port_index_to_settings_key_map[] = {
     &g_config.input.bindings.port4,
 };
 
+static int sdl_kbd_scancode_map[25];
+
 void xemu_input_init(void)
 {
+    int i;
```
mborgerson: Let's just move the `i` definition down to the `for` loop its used in

ui/xemu-input.c
```diff
@@ -112,6 +116,48 @@ void xemu_input_init(void)
...
+    sdl_kbd_scancode_map[17] = g_config.input.keyboard_controller_scancode_map.lstick_right; //default key : f
+    sdl_kbd_scancode_map[18] = g_config.input.keyboard_controller_scancode_map.lstick_down;  //default key : d
+    sdl_kbd_scancode_map[19] = g_config.input.keyboard_controller_scancode_map.ltrigger;     //default key : w
+    sdl_kbd_scancode_map[20] = g_config.input.keyboard_controller_scancode_map.rstick_up;    //default key : i
+    sdl_kbd_scancode_map[21] = g_config.input.keyboard_controller_scancode_map.rstick_left;  //default key : j
+    sdl_kbd_scancode_map[22] = g_config.input.keyboard_controller_scancode_map.rstick_right; //default key : l
+    sdl_kbd_scancode_map[23] = g_config.input.keyboard_controller_scancode_map.rstick_down;  //default key : k
+    sdl_kbd_scancode_map[24] = g_config.input.keyboard_controller_scancode_map.rtrigger;     //default key : o
+
+    for (i=0; i<25; i++){
```
mborgerson: * nit: space between `)` and `{`
* nit: space around `=` and `<`

ui/xemu-input.c
```diff
@@ -112,6 +116,48 @@ void xemu_input_init(void)
     new_con->name = "Keyboard";
     new_con->bound = -1;
 
+    //see config_spec.yml for default scancode values
+    sdl_kbd_scancode_map[0] = g_config.input.keyboard_controller_scancode_map.a;             //default key : a
```
mborgerson: * We don't need the 'see config_spec.yml' comment (it's implied because it's using config)
* We don't need the default key comments here. You could comment about them in config_spec.yml though

ui/xemu-input.c
7oxicshadow: Changes implemented

ui/xemu-input.c
7oxicshadow: Changes implemented

ui/xemu-input.c
7oxicshadow: Changes implemented

ui/xemu-input.c
7oxicshadow: comments moved to config_spec.yml. can change the comments to a single block rather than line by line if prefered

ui/xemu-input.c
```diff
@@ -112,6 +114,47 @@ void xemu_input_init(void)
...
+        /* SDL allocates an array of size SDL_NUM_SCANCODES
+           that can be indexed to get the pressed states of
+           each key. We need to validate that none of the
+           assignments in the toml file are outside of this
+           range */
+
+        if( (sdl_kbd_scancode_map[i] < SDL_SCANCODE_UNKNOWN) ||
+            (sdl_kbd_scancode_map[i] >= SDL_NUM_SCANCODES) ) {
+            fprintf(stderr, "WARNING: Keyboard controller map scancode out of range (%d) : Disabled\n", sdl_kbd_scancode_map[i]);
+            // Default to an unknown code (0)
```
mborgerson: Don't need this comment, the code says this

ui/xemu-input.c
```diff
@@ -112,6 +114,47 @@ void xemu_input_init(void)
...
+    sdl_kbd_scancode_map[18] = g_config.input.keyboard_controller_scancode_map.lstick_down;
+    sdl_kbd_scancode_map[19] = g_config.input.keyboard_controller_scancode_map.ltrigger;
+    sdl_kbd_scancode_map[20] = g_config.input.keyboard_controller_scancode_map.rstick_up;
+    sdl_kbd_scancode_map[21] = g_config.input.keyboard_controller_scancode_map.rstick_left;
+    sdl_kbd_scancode_map[22] = g_config.input.keyboard_controller_scancode_map.rstick_right;
+    sdl_kbd_scancode_map[23] = g_config.input.keyboard_controller_scancode_map.rstick_down;
+    sdl_kbd_scancode_map[24] = g_config.input.keyboard_controller_scancode_map.rtrigger;
+
+    for (int i = 0; i < 25; i++) {
+        /* SDL allocates an array of size SDL_NUM_SCANCODES
```
mborgerson: Don't need this comment, the code says this

config_spec.yml
```diff
@@ -23,6 +23,108 @@ input:
     type: bool
     default: true
   background_input_capture: bool
+  keyboard_controller_scancode_map:
+    # Scancode reference : https://github.com/libsdl-org/SDL/blob/main/include/SDL_scancode.h
+    a:
+      type: integer
+      # default key : a (4)
```
mborgerson: * Don't need to state the value, it's already listed right here
* Move comment to same line as value

```yaml
a:
  type: integer
  default: 4 # a
```

7oxicshadow: all comments updated

PR #677, opened 2022-01-26
nv2a: Set oFog.x with most significant oFog mutation

abaire:
Sets `oFog.x` to the most significant component modified by the vertex program, mimicking HW behavior.

[Tests](https://github.com/abaire/nxdk_pgraph_tests/blob/3c23adb4fcb6a059f42a34b76d0e931d1fb9cd30/src/tests/fog_tests.cpp#L329)
[HW results](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-Fog_coord_vec4)

Fixes #673 

Appears to fix #564 as well (testing with Need For Speed: Underground)

PR #981, opened 2022-05-27
nv2a: Fix oFog write behavior with non-x destination masks.

### LINKED ISSUE #978: Vertex shader `_in` expansion leads to incorrect behavior in `MOV` w/ a mask that does not include `x`
### Bug Description

While looking into fog behavior, I discovered that `MOV(oFog,yw, c[108].yw);` does not result in the same behavior as `oFog.yw = c[108].yw` (which is the intent and HW behavior).

`vec4 _in(vec2 v) { return v.xyyy; }` causes the MOV to be interpreted as `oFog.yw = c[108].ww` since the `y` is erroneously moved to the `x` component of the expanded vector and the `w` is used for the rest of the values.

### Expected Behavior

The values need to be expanded in a manner consistent with the source mask. In this particular case, `yyww` or `yyyw` would've worked.

### xemu Version

Based on the [git blame](https://github.com/mborgerson/xemu/blame/0cdcdde70c13f6cf3ed09aca29185ed790b22485/hw/xbox/nv2a/vsh.c#L617), this affects all version of xemu since the code came over from the xqemu fork

### System Information

n/a this bug is system independent.

### Additional Context

_No response_

abaire:
Fixes #978 

Modifies the write masking behavior to fit the expansion done by `_in` and removes the need for the (slightly erroneous) fix made in #677.

The original tests were insufficient to detect this case since it wasn't straightforward to get Cg to produce the necessary code, the tests now use nv2a assembly directly.

[Expanded tests, the vec4 ones are the ones of interest](https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/tests/fog_tests.cpp)
[Updated HW results](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-Fog_coord_vec4)

### DIFF
--- a/hw/xbox/nv2a/vsh.c
+++ b/hw/xbox/nv2a/vsh.c
@@ -202,6 +202,28 @@ static const char* mask_str[] = {
     ",xyzw" // 1111 xyzw
 };
 
+/* Writes to the oFog register apply the most significant masked component to
+ * `x`. The remaining values are assigned arbitrarily to fit the 4-component
+ * function behavior. */
+static const char* fog_mask_str[] = {
+    ",",
+    ",x",
+    ",x",
+    ",xy",
+    ",x",
+    ",xy",
+    ",xy",
+    ",xyz",
+    ",x",
+    ",xy",
+    ",xy",
+    ",xyz",
+    ",xy",
+    ",xyz",
+    ",xyz",
+    ",xyzw"
+};
+
 /* Note: OpenGL seems to be case-sensitive, and requires upper-case opcodes! */
 static const char* mac_opcode[] = {
     "NOP",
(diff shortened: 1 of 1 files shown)


hw/xbox/nv2a/vsh.c
```diff
@@ -436,24 +456,10 @@ static MString* decode_opcode(const uint32_t *shader_token,
         }
 
         int write_mask = vsh_get_field(shader_token, FLD_OUT_O_MASK);
-        mstring_append(ret, mask_str[write_mask]);
+        const char *write_mask_str = write_fog_register ? fog_mask_str[write_mask] : mask_str[write_mask];
```
PatrickvL, line 459: If a slightly different ordering of the rest after X is acceptable, this could also be done without the fog table by indexing the regular mask table with mask_index, but for fog first assert that zero won't appear in mask_index and shift it left while the X bit is not set.

abaire: Not sure I'm following. There are a lot of ways that this could be done, a lookup table (that's tiny and running on modern machines with tons of RAM) seems preferable to doing any sort of looping.

PR #982, opened 2022-05-27
nv2a: Do not force oFog to 1.0

### LINKED ISSUE #876: Need for Speed: Most Wanted - Map doesn't render
### Title

https://xemu.app/titles/4541007b/#Need-for-Speed-Most-Wanted
EA-123 [4541007b]

### Bug Description

Streets do not render on the Map on the radar or on the map view.

Need for Speed Most Wanted No Map

Need for Speed Most Wanted No Map on radar

### Expected Behavior

Streets should be visible so that you can see the best routes to use. This is crucial during police chases. The map objects do render so at least you can get a general sense of which direction to go. 

### xemu Version

0.6.5 commit a809d8557d48005f9e88145d398b87851b76e8fa

### System Information

OS: Windows 10
CPU: AMD Ryzen 9 5950X 16-Core Processor
GPU: NVIDIA GeForce RTX 3070/PCIe/SSE2
GPU Driver: 4.0.0 NVIDIA 512.15

### Additional Context

Created QEMU launch parameters: C:\Users\dbill\LaunchBox\Emulators\xemu-win-release\xemu.exe -machine xbox,bootrom=C:\Users\dbill\LaunchBox\Emulators\xemu-win-release\Xbox BIOS\mcpx_1.0.bin,kernel-irqchip=off -device smbus-storage,file=C:\Users\dbill\AppData\Roaming\xemu\xemu\eeprom.bin -bios C:\Users\dbill\LaunchBox\Emulators\xemu-win-release\Xbox BIOS\Complex_4627v1.03.bin -m 64 -drive index=0,media=disk,file=C:\Users\dbill\LaunchBox\Emulators\xemu-win-release\xbox_hdd.qcow2\xbox_hdd.qcow2,locked=on -drive index=1,media=cdrom,file=C:\ROMS\Microsoft Xbox\Need for Speed - Most Wanted (USA).iso -display xemu -audiodev none,id=snd0 
xemu_version: 0.6.5
xemu_branch: master
xemu_commit: a809d8557d48005f9e88145d398b87851b76e8fa
xemu_date: Sat Apr 30 05:06:34 UTC 2022
xemu_settings_get_path: config path: C:\Users\dbill\AppData\Roaming\xemu\xemu\xemu.toml
CPU: AMD Ryzen 9 5950X 16-Core Processor            
OS_Version: Windows
GL_VENDOR: NVIDIA Corporation
GL_RENDERER: NVIDIA GeForce RTX 3070/PCIe/SSE2
GL_VERSION: 4.0.0 NVIDIA 512.15
audio: Device ac97: audiodev default parameter is deprecated, please specify audiodev=snd0

abaire:
Fixes #876

WARNING: This does not exactly follow HW behavior and it's possible that it will cause regressions in any games that happen to set the oFog to something high before using it unmodified in a subsequent shader.

The code being removed was added ~7 years ago in [this commit](https://github.com/xqemu/xqemu/commit/97be3f598683721bc43e354d7804221152e21f02)

From what I can see from the [HW test results](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-Fog_coord_vec4) the behavior on nv2a is more subtle than the cited `NV_vertex_program` documentation.

In practice the register more or less retains its value until it is explicitly modified.[The test](https://github.com/abaire/nxdk_pgraph_tests/blob/4cff2b2ebe49c29eb24caea9c058b3d6ddaa3a16/src/tests/fog_tests.cpp#L432) renders something with an explicitly set oFog.x, then renders again with a shader that does not modify oFog.x but references it in the pixel shader. The value carries over rather than being forced to 1.0.

~~Interestingly, this test is apparently not hermetic; running the other vec4_coord test (that explictly set various components of oFog) lead to unusual behavior where the fogging effect is not uniform across all vertices, despite never being set. Even more interestingly, re-running the test once it's in this state will often produce variations as to which vertices are apparently using a stale/incorrect fog value.~~ (See comment thread, this may have been due to HW parallelization and was resolved by doing repeated draws before using the shader that does not set the fog coord.)

### DIFF
--- a/hw/xbox/nv2a/shaders.c
+++ b/hw/xbox/nv2a/shaders.c
@@ -727,21 +727,7 @@ GLSL_DEFINE(texMat3, GLSL_C_MAT4(NV_IGRAPH_XF_XFCTX_T3MAT))
 "vec4 oB0 = vec4(0.0,0.0,0.0,1.0);\n"
 "vec4 oB1 = vec4(0.0,0.0,0.0,1.0);\n"
 "vec4 oPts = vec4(0.0,0.0,0.0,1.0);\n"
-/* FIXME: NV_vertex_program says: "FOGC is the transformed vertex's fog
- * coordinate. The register's first floating-point component is interpolated
- * across the assembled primitive during rasterization and used as the fog
- * distance to compute per-fragment the fog factor when fog is enabled.
- * However, if both fog and vertex program mode are enabled, but the FOGC
- * vertex result register is not written, the fog factor is overridden to
- * 1.0. The register's other three components are ignored."
- *
- * That probably means it will read back as vec4(0.0, 0.0, 0.0, 1.0) but
- * will be set to 1.0 AFTER the VP if it was never written?
- * We should test on real hardware..
- *
- * We'll force 1.0 for oFog.x for now.
- */
-"vec4 oFog = vec4(1.0,0.0,0.0,1.0);\n"
+"vec4 oFog = vec4(0.0,0.0,0.0,1.0);\n"
 "vec4 oT0 = vec4(0.0,0.0,0.0,1.0);\n"
 "vec4 oT1 = vec4(0.0,0.0,0.0,1.0);\n"
 "vec4 oT2 = vec4(0.0,0.0,0.0,1.0);\n"


PatrickvL:
> Interestingly, this test is apparently not hermetic; running the other vec4_coord test (that explictly set various components of oFog) lead to unusual behavior where the fogging effect is not uniform across all vertices, despite never being set. Even more interestingly, re-running the test once it's in this state will often produce variations as to which vertices are apparently using a stale/incorrect fog value.

This sounds a lot like the effect one would expect when multiple pixel shader pipelines process vertices, each retaining their own set of registers

abaire:
> This sounds a lot like the effect one would expect when multiple pixel shader pipelines process vertices, each retaining their own set of registers

Interesting point, I suppose it's possible that the HW is parallelizing both sets of draws and that the second set happens to run in an environment that was not stabilized by the first draw. Let me see if looping over the first draw a few times changes the behavior; I doubt the nv2a had much of a pipeline given how comparatively early it was in GPU development.


EDIT: Updated the test and it appears to be fixed. The parallelization seems to be just slightly more than the 4 triangle draw case I was throwing at it, looping that draw twice leads to a repeatable system. Thanks for the suggestion @PatrickvL !

Triticum0:
Don't worry =(, my mistake.

Triticum0:
@abaire This pr breaks fifa 2004 and 2005 I will send a renderdoc if you need it #1602 Edit: I created a capture I can seen on discord when you want to debug it.

Triticum0:
here  link to the renderdoc  https://drive.google.com/file/d/1TpeN_SXZlQjvy4fPAubMtqmgTuLIS-i8/view?usp=sharing

abaire:
@Triticum0 lets move this discussion to #1602. I think this PR just removed a default that happened to mask the real issue; I'll add notes there.

Issue #926, opened 2022-05-18
[Assertion] (index_count*vertex_size) == (pg->inline_array_length*4)


Triticum0:
### Title

[Conker: Live & Reloaded](https://xemu.app/titles/4d530051/#Conker-Live-Reloaded)
[Murakumo: Renegade Mech Pursuit](https://xemu.app/titles/55530016/#Murakumo-Renegade-Mech-Pursuit)
[Murakumo](https://xemu.app/titles/46530001/#Murakumo)
[Shenmue 2](https://xemu.app/titles/4d530032/#Shenmue-II)

### Bug Description

These Games Suffer from this assertion
https://github.com/mborgerson/xemu/blob/38a0e46f8c59c6da50029c1b53712070f211ac74/hw/xbox/nv2a/pgraph.c#L6785
 
This Assertion is hit when...

### Conker: Live & Reloaded
    When you face the giant poo boss - during the cutscene when the dung beetle tells you to run and multiple times after 
  

### Murakumo: Renegade Mech Pursuit/Murakumo
    When Going in-game after the intro cutscene

### Shenmue 2
    Boot game
    Start single player campaign (load save)
    make the riddle of the keys in the statue
    pass the animation
    fight against the henchmen
    pass the animation
    leave the room walk to the stairs when going up
    the error happens. 

### Expected Behavior

Should not crash at all

### xemu Version

0.7.13

### System Information

Field | Value
-- | --
OS | Windows 10 64-bit (21H1)
CPU | AMD Ryzen 5 2600 Six-Core Processor @ 3.40GHz
Graphics Device | AMD Radeon RX 5700 XT
Graphics Driver | 4.0.148202 Core Profile Context 21.3.2 27.20.15003.5017

### Additional Context

Realised there was 3 issue with the same assertion so I'm reducing them into one.Close #423  and #424

Fighter19:
In Conker that also happened the first time after an earthworm cutscene played.
Afterwards the next time I tried to boot the game it said my disc might be damaged.
Then after I restarted the emulator, the game booted as normal, but my savegame was gone,
telling me it was corrupted.

So I assume that maybe some autosave icon I've never seen caused that issue.

Triticum0:
#424 has all the times where it crashes. this is mostly for consideration of issues.

BatConker:
https://user-images.githubusercontent.com/68554598/170745589-7c15c71d-6b50-4d26-8fd7-a8a3733eb04e.mp4

[...]

PR #985, opened 2022-05-28
nv2a: Ignore excess inline array data

### LINKED ISSUE #424: Conker Live and Reloaded - Assertion Failed
### Bug Description

https://xemu.app/titles/4d530051/#Conker-Live-Reloaded (Duplicate of #423)

Conker has been crashing many times at random during my playthrough, however I have now reached a point at which is crashes consistently. It's when you are about to face the giant poo boss - during the cutscene when the dung beetle tells you to run. 

I have tried two different games in order to make sure it isn't the game itself and the crash happens every time regardless.

Message:
microsoft visual c++ runtime library
assertion failed
program ...
file ...
expression (index_count*vertex_size) ==
(pg->inline_array_length*4)
for information on how your program can cause an
assertion failure, see the visual c++ documentation on asserts

### Expected Behavior

Should not crash at all

### xemu Version

Version: 0.5.4-39-gedf3735718
Branch: master
Commit: edf3735...
Date: Tue Jul 20 17:39:54 UTC 2021

### System Information

OS: Windows 10 (64bit)
CPU: i7 7700k @ 4.20 GHz
GPU: Zotac GTX 1080 Amp Extreme!
Driver: 471.11

### Additional Context

None

abaire:
Fixes #424 
May also address #926 but I don't have those games to test

Testing on hardware shows that excess inline array data is gracefully ignored and discarded. The test draws triangle primitives, passing a triplet of valid vertices as well as a fourth and part of a fifth, then completes the draw. To verify that the excess data is dropped, it performs a second "triangles" draw with the remaining data, which does not result in a second triangle being drawn.

[Test](https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/tests/inline_array_size_mismatch.cpp)
[HW results](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-Inline_array_size_mismatch)

### DIFF
--- a/hw/xbox/nv2a/pgraph.c
+++ b/hw/xbox/nv2a/pgraph.c
@@ -6784,7 +6784,6 @@ static unsigned int pgraph_bind_inline_array(NV2AState *d)
 
     unsigned int vertex_size = offset;
     unsigned int index_count = pg->inline_array_length*4 / vertex_size;
-    assert((index_count*vertex_size) == (pg->inline_array_length*4));
 
     NV2A_DPRINTF("draw inline array %d, %d\n", vertex_size, index_count);
 
@@ -6792,7 +6791,7 @@ static unsigned int pgraph_bind_inline_array(NV2AState *d)
     glBindBuffer(GL_ARRAY_BUFFER, pg->gl_inline_array_buffer);
     glBufferData(GL_ARRAY_BUFFER, NV2A_MAX_BATCH_LENGTH * sizeof(uint32_t),
                  NULL, GL_STREAM_DRAW);
-    glBufferSubData(GL_ARRAY_BUFFER, 0, pg->inline_array_length*4, pg->inline_array);
+    glBufferSubData(GL_ARRAY_BUFFER, 0, index_count * vertex_size, pg->inline_array);
     pgraph_bind_vertex_attributes(d, 0, index_count-1, true, vertex_size,
                                   index_count-1);
 


Steff833:
It also fixes the stuttering Intro Movie on Conker Live and Reloaded

abaire:
> It also fixes the stuttering Intro Movie on Conker Live and Reloaded

@Steff833 Can you double check that with a comparison against the current release build? This change literally cannot have any effect except in cases where the assert was previously thrown (the assert would have failed on any case where this change takes effect). 
Unless you were previously using a build that had the assert disabled, I don't think it's possible for this to change performance, only to avoid an apparent crash.

Triticum0:
It Fixed Murakumo: Renegade Mech Pursuit have tested the jp version but most likely the same will test Shenmue 2 so will get back to you

Fighter19:
I didn't even have stuttering with the last release on flathub.

Steff833:
> > It also fixes the stuttering Intro Movie on Conker Live and Reloaded
> 
> @Steff833 Can you double check that with a comparison against the current release build? This change literally cannot have any effect except in cases where the assert was previously thrown (the assert would have failed on any case where this change takes effect). Unless you were previously using a build that had the assert disabled, I don't think it's possible for this to change performance, only to avoid an apparent crash.

You`re right sorry my bad. Didn`t check conker for longer time. The Intro video runs also flawlessly on recent builds. So it must have been fixed in some of the builds since Xemu changed the Interface and introduced the new config file system.

Issue #724, opened 2022-02-07
nv2a: stencil test with replace stencil op is not correctly applied


antangelo:
### Bug Description

Test case: https://github.com/antangelo/nxdk_pgraph_tests/blob/93f8613fc8b33358452e77f0ce96a617da59cf1e/src/tests/stencil_tests.cpp#L15

Draws a red square, then disables color writing, enables the stencil test and draws a smaller square. After this, draws a green square with same bounds as the red square. Expected output is a large red square and small embedded green square.

This particular test result uses the replace stencil op. My test case for other operations is still a work in progress; I'll be updating this issue as I test them.

### Expected Behavior

Xemu Behavior:
![StencilOnly_REPLACE_TestOn](https://user-images.githubusercontent.com/25046221/152711868-5f119c1f-a17b-4c2c-8167-e7b769cf263d.png)

Xbox Behavior:
![StencilOnly_REPLACE_TestOn](https://user-images.githubusercontent.com/25046221/152711888-1abb2c34-129f-4bd8-a546-3a46cdb461a9.png)

### xemu Version

0.6.2-64-g758e36d398

Not known if working previously

### System Information

OS: Arch Linux (5.16.3-arch1-1)
CPU: Intel Core i5-10400
GPU: NVIDIA GeForce RTX 3060 Ti/PCIe/SSE2
GL_VERSION: 4.0.0 NVIDIA 495.46

antangelo:
This issue appears to be influenced by compiler optimization in some capacity. Building with `--debug` on any commit before 5f14ee4a68ec4217952da655e0d073431867d349 (which changed the debug optimization level) appears to produce the correct result. Building with `--debug` after that commit exhibits the same incorrect behavior as above.

I have not yet found a commit that exhibits correct behavior with compiler optimizations applied.

antangelo:
This issue seems to be resolved

PR #989, opened 2022-05-28
nv2a: Gracefully ignore draws with no target bindings

### LINKED ISSUE #635: nv2a: color_binding || zeta_binding assertion failure
### Title

https://xemu.app/titles/5451001a/#Nickelodeon-SpongeBob-SquarePants-in-Battle-for-Bikini-Bottom
https://xemu.app/titles/4d570005/#Area-51
https://xemu.app/titles/5553001c/#Batman-Rise-of-Sin-Tzu
https://xemu.app/titles/5553004b/#Cold-Fear
https://xemu.app/titles/55530018/#Beyond-Good-Evil
### Bug Description

nv2a: color_binding || zeta_binding assertion failure from bubble moves.

### Expected Behavior
It shouldn't crash from bubble moves.

### xemu Version

v0.6.2-34-g62d04a3636
Was working in v.0.6.2-32-gd0fd0cb3e4

### System Information

OS: Windows 10
CPU: Intel(R) Xeon(R) CPU(TM) E5-2678 v3 @ 2.50GHz
GPU: NVIDIA Quadro P5000
GPU Driver: 462.96 NVIDIA

### Additional Context

_No response_

abaire:
Fixes #635 

[Test that triggers the assert without the fix](https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/tests/color_zeta_disable_tests.cpp)
[Tests of stencil behavior more generally](https://github.com/antangelo/nxdk_pgraph_tests/blob/stencil-tests/src/tests/stencil_tests.cpp)

### DIFF
--- a/hw/xbox/nv2a/pgraph.c
+++ b/hw/xbox/nv2a/pgraph.c
@@ -2685,6 +2685,7 @@ DEF_METHOD(NV097, SET_BEGIN_END)
     bool mask_red = control_0 & NV_PGRAPH_CONTROL_0_RED_WRITE_ENABLE;
     bool mask_green = control_0 & NV_PGRAPH_CONTROL_0_GREEN_WRITE_ENABLE;
     bool mask_blue = control_0 & NV_PGRAPH_CONTROL_0_BLUE_WRITE_ENABLE;
+    bool color_write = mask_alpha || mask_red || mask_green || mask_blue;
 
     bool depth_test = control_0 & NV_PGRAPH_CONTROL_0_ZENABLE;
     bool stencil_test =
@@ -2693,6 +2694,12 @@ DEF_METHOD(NV097, SET_BEGIN_END)
     if (parameter == NV097_SET_BEGIN_END_OP_END) {
 
         nv2a_profile_inc_counter(NV2A_PROF_BEGIN_ENDS);
+
+        if (!(pg->color_binding || pg->zeta_binding)) {
+            // TODO: Factor out the end handling (done in PR #805).
+            NV2A_GL_DGROUP_END();
+            return;
+        }
 
         assert(pg->shader_binding);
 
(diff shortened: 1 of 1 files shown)


Triticum0:
Do you need me to test the games?

mborgerson:
The assertion is there because I suspect we have missed something

abaire:
We could leave the assert in and put it behind the check to see if any of color/depth/stencil are write enabled; that would catch erroneous states and still work in the case that the bindings aren't necessary.

I have a HW test written, waiting for an nvtrace to complete so I can make sure it really behaves the way I expect, @antangelo mentioned running into errors on HW when testing this case.

abaire:
HW test fails in the same way that @antangelo encountered so there is something more interesting to this. I'll dig into it some more.

abaire:
I spent some more time investigating Spongebob and am fairly confident that I figured out why they're not getting an exception where my test (and @antangelo's) are. Details in my latest comment in #635, I have a [test case pending](https://github.com/abaire/nxdk_pgraph_tests/commit/f0bb7440a2c3f00f51df4945b35d045e427d27dd#diff-ecc0768980ac93f98ca7a964016bb20e0bb1b96d0933608607bbbdf04bcf8096R49) but @antangelo wrote a much more thorough test that I'd rather see merged into the test repo.

I'll leave this in draft mode until my results are reproduced by @antangelo.

antangelo:
[Test cases](https://github.com/antangelo/nxdk_pgraph_tests/blob/stencil-tests/src/tests/stencil_tests.cpp)

Results (on Xbox):
Stencil_REPLACE

with stencil test enabled:
Stencil_REPLACE_ST

Note that these tests are still not entirely accurate to hardware on xemu (see #724 for example, the behavior seems to have changed a bit since), but these issues are not related to this PR, so I can address them in a followup.

hw/xbox/nv2a/pgraph.c
```diff
@@ -2833,6 +2840,18 @@ DEF_METHOD(NV097, SET_BEGIN_END)
         assert(parameter <= NV097_SET_BEGIN_END_OP_POLYGON);
 
         pgraph_update_surface(d, true, true, depth_test || stencil_test);
+
+        if (!(color_write || depth_test || stencil_test)) {
```
mborgerson, line 2844: ~~Should this test against zeta write mask (like color), not just whether the depth/stencil tests are enabled?~~ Nevermind, I think this should be okay. If test is disabled I believe writes will also be disabled like OpenGL spec (but it would be good to confirm).

antangelo: Did a quick search through the pgraph logs I have:
- BFBB has depth mask off and stencil mask on
- Cold Fear has both depth and stencil masks on
- Batman has both depth and stencil masks on

The stencil tests I wrote exercise this case with the depth mask off and stencil mask on.

For the stencil test I believe OpenGL spec is to assume the stencil test has passed if the mask is disabled (see https://www.khronos.org/opengl/wiki/Stencil_Test ), and the pass op is applied. The stencil tests seem to indicate this is not the case on Xbox-- instead disabling the stencil test will not apply the pass op to the zeta buffer.

abaire: For the depth test case I have previously confirmed that disabling the test disables writes on nv2a (it broke a test I was writing that had erroneously left the test disabled when explicitly enabling the mask).

Issue #315, opened 2021-06-02
Meta: Shader compilation failures


Triticum0:
### Title Information
NHL-2002 - https://xemu.app/titles/45410005
<details close>
<summary>Failing Fragment Shader from NHL 2002 OLD </summary>
<br>
#version 330

struct VertexData {
  float inv_w;
  vec4 D0;
  vec4 D1;
  vec4 B0;
  vec4 B1;
  float Fog;
  vec4 T0;
  vec4 T1;
  vec4 T2;
  vec4 T3;
};
noperspective in VertexData g_vtx;
#define vtx g_vtx

out vec4 fragColor;

uniform vec4 fogColor;
float sign1(float x) {
    x *= 255.0;
    return (x-128.0)/127.0;
}
float sign2(float x) {
    x *= 255.0;
    if (x >= 128.0) return (x-255.5)/127.5;
               else return (x+0.5)/127.5;
}
float sign3(float x) {
    x *= 255.0;
    if (x >= 128.0) return (x-256.0)/127.0;
               else return (x)/127.0;
}
float sign3_to_0_to_1(float x) {
    if (x >= 0) return x/2;
           else return 1+x/2;
}
vec3 dotmap_zero_to_one(vec3 col) {
    return col;
}
vec3 dotmap_minus1_to_1_d3d(vec3 col) {
    return vec3(sign1(col.r),sign1(col.g),sign1(col.b));
}
vec3 dotmap_minus1_to_1_gl(vec3 col) {
    return vec3(sign2(col.r),sign2(col.g),sign2(col.b));
}
vec3 dotmap_minus1_to_1(vec3 col) {
    return vec3(sign3(col.r),sign3(col.g),sign3(col.b));
}
vec3 dotmap_hilo_1(vec3 col) {
    return col;
}
vec3 dotmap_hilo_hemisphere_d3d(vec3 col) {
    return col;
}
vec3 dotmap_hilo_hemisphere_gl(vec3 col) {
    return col;
}
vec3 dotmap_hilo_hemisphere(vec3 col) {
    return col;
}
const float[9] gaussian3x3 = float[9](
    1.0/16.0, 2.0/16.0, 1.0/16.0,
    2.0/16.0, 4.0/16.0, 2.0/16.0,
    1.0/16.0, 2.0/16.0, 1.0/16.0);
const vec2[9] convolution3x3 = vec2[9](
[...]

PR #992, opened 2022-05-28
nv2a: Assert on unimplemented writable const register path

abaire:
Produces a more meaningful error for the NHL 2002 shader compilation failure in #315

NHL2002 in particular seems to modify `c107` and then not use it, which makes me assume its value would persist across draws, so I think implementing this will need something more than trivial introduction of intermediary variables.


abaire:
Note: I don't have NHL2002 handy to test this, but from inspection it seems like the current code could never produce a compilable shader.

Triticum0:
Works
image

PR #1006, opened 2022-06-01
ui: show render scale and screen resolution to about window

### LINKED ISSUE #1000: Add render scale to about window (and compatibility reports)
### Feature Request

Frequently users report issues with performance or graphical glitches that only appear at certain render scales. It'd be very useful to have this as context in GitHub issues (copy pasted from the About screen info) and compatibility reports (via the automated reporter).

### Alternatives

_No response_

### Additional Context

_No response_

Fabxx:
Fixes: https://github.com/mborgerson/xemu/issues/1000

video: 

https://user-images.githubusercontent.com/30447649/174775650-37e6abde-83c0-4f11-af60-0c61ac32187a.mp4


ui/xui/main-menu.cc
```diff
@@ -794,24 +794,27 @@ void MainMenuSystemView::Draw()
 
 void MainMenuAboutView::Draw()
 {
-    static const char *build_info_text = NULL;
+     const char *build_info_text = NULL;
```
abaire: There is definitely no reason to make this non-static and recreate the build string on every draw, it's all determined at compile time and cannot change.

ui/xui/main-menu.cc
```diff
@@ -794,24 +794,27 @@ void MainMenuSystemView::Draw()
...
     if (build_info_text == NULL) {
         build_info_text = g_strdup_printf(
             "Version:      %s\nBranch:       %s\nCommit:       %s\nDate:         %s",
             xemu_version, xemu_branch, xemu_commit, xemu_date);
     }
 
-    static const char *sys_info_text = NULL;
+     const char *sys_info_text = NULL;
```
abaire: This change means you're fetching gl strings everytime the menu is drawn, which is probably not what you want to do. I think you'd be better off permanently caching the gl strings (which will never change) and then appending the bits that might change (scale and window dimensions) separately.

ui/xui/reporting.hh
```diff
@@ -31,6 +31,10 @@ public:
 	std::string xemu_branch;
 	std::string xemu_commit;
 	std::string xemu_date;
+	//not sure if report accepts unsigned int
+	unsigned int xemu_rendering_scale; 
+	//is it ok to save the data from SDL_GetCurrentDisplayMode in main_menu.cc directly into the class? same as the rendering scale so the data will be filled
```
abaire: Don't put the comment in the code, put the comment in the PR on the line in `reporting.cc` where you add the rendering scale.

ui/xui/reporting.cc
```diff
@@ -45,6 +45,7 @@ const std::string &CompatibilityReport::GetSerializedReport()
 		{"xemu_branch", xemu_branch},
 		{"xemu_commit", xemu_commit},
 		{"xemu_date", xemu_date},
+		{"xemu_rendering_scale", xemu_rendering_scale},
```
abaire: Do you plan on reporting width & height here as well?

ui/xui/main-menu.cc
Fabxx: If i leave this as static i can't update the resolution scaling factor value, it will keep always the initial scaling value that was stored, so even if a user tries to report with a different scale resolution but started xemu with 1x, the report will always contain that scale factor. So i either put back this to static and make a separate tiny buffer non static where i only get information on every frame for the resolution scale, or else dunno how you guys want this handled.

ui/xui/reporting.cc
Fabxx: If it's necessary for the report in the JSON then i'll put them

ui/xui/main-menu.cc
Fabxx: This was a misread for the buffers, will put it back as it was

ui/xui/main-menu.cc
antangelo: The scaling factor isn't included in `build_info_text`, only `sys_info_text`. So `static` should have no effect w.r.t that on `build_info_text`.

ui/xui/main-menu.cc
```diff
@@ -794,24 +794,27 @@ void MainMenuSystemView::Draw()
...
     if (sys_info_text == NULL) {
```
antangelo, line 839: If you are removing static from `sys_info_text` then you do not need this null check (as it will always be true).

ui/xui/main-menu.cc
```diff
@@ -794,7 +794,7 @@ void MainMenuSystemView::Draw()
 
 void MainMenuAboutView::Draw()
 {
-    static const char *build_info_text = NULL;
+    static  const char *build_info_text = NULL;
```
abaire: Remove the newly introduced space.

ui/xui/main-menu.cc
```diff
@@ -807,12 +807,15 @@ void MainMenuAboutView::Draw()
         const char *gl_version = (const char*)glGetString(GL_VERSION);
         const char *gl_renderer = (const char*)glGetString(GL_RENDERER);
         const char *gl_vendor = (const char*)glGetString(GL_VENDOR);
+        SDL_DisplayMode DM;
```
abaire: Resolution can change during the course of the program, you don't want this in your statically cached information, move it to the dynamic info calculated below.

ui/xui/main-menu.cc
```diff
@@ -850,11 +853,12 @@ void MainMenuAboutView::Draw()
 
     SectionTitle("System Information");
     ImGui::PushFont(g_font_mgr.m_fixed_width_font);
-    ImGui::InputTextMultiline("###systeminformation", (char *)sys_info_text,
-                              strlen(sys_info_text),
-                              ImVec2(-FLT_MIN, ImGui::GetTextLineHeight() * 8),
+    ImGui::InputTextMultiline("###systeminformation", (char *)content,
```
abaire: content is already a `char *`, this cast is unnecessary.

ui/xui/reporting.cc
```diff
@@ -45,6 +45,8 @@ const std::string &CompatibilityReport::GetSerializedReport()
...
+		{"Screen Resolution:" width, height},
```
abaire: Does this code compile?

ui/xui/reporting.hh
```diff
@@ -31,6 +31,8 @@ public:
 	std::string xemu_branch;
 	std::string xemu_commit;
 	std::string xemu_date;
+	unsigned int xemu_rendering_scale; 
+	int width, height;
```
abaire: Presumably negative sizes aren't allowed, any reason to have this be signed but rendering scale be unsigned?

ui/xui/main-menu.cc
```diff
@@ -850,11 +853,12 @@ void MainMenuAboutView::Draw()
...
+                              strlen(content),
+                              ImVec2(-FLT_MIN, ImGui::GetTextLineHeight() * 10),
                               ImGuiInputTextFlags_ReadOnly);
     ImGui::PopFont();
+    free(content);
```
antangelo: Memory allocated with `g_strdup_printf` (and other glib functions) should be freed with `g_free`

ui/xui/main-menu.cc
Fabxx: i did swap it to the dynamic buffer, but for some reason, despite changing resolution during xemu runtime, and then recalling the draw with the SDL display mode acquisition, the value it's not being updated in width and height

ui/xui/main-menu.cc
```diff
@@ -809,10 +809,15 @@ void MainMenuAboutView::Draw()
...
             "CPU:          %s\nOS Platform:  %s\nOS Version:   %s\nManufacturer: %s\n"
-            "GPU Model:    %s\nDriver:       %s\nShader:       %s",
+            "GPU Model:    %s\nDriver:       %s\nShader:       %s\n",
              xemu_get_cpu_info(), xemu_get_os_platform(), xemu_get_os_info(), gl_vendor,
              gl_renderer, gl_version, gl_shader_version);
     }
+    SDL_DisplayMode DM;
+    SDL_GetCurrentDisplayMode(0, &DM);
+
+    char *content = g_strdup_printf("%sScaling:      %dx\n" "Resolution:   %dx%d",
```
abaire: Nit: I'd remove the implicit string join between scaling and resolution, just make it a single string.

ui/xui/main-menu.cc
```diff
@@ -792,6 +792,24 @@ void MainMenuSystemView::Draw()
     }
 }
 
+static const char *get_display_mode(void)
```
abaire: This duplicates the behavior in `popup-menu.cc` and means we'll have to translate these strings twice, leading to a very good chance that they'll get out of sync. 

I'd suggest either creating a single helper function in a common place (maybe a new "ui-strings" file) or making the array in [popup-menu](https://github.com/mborgerson/xemu/blob/30829497d21ea8730b3cc1875cdedf67ce328b46/ui/xui/popup-menu.cc#L257) a global so that it can be accessed here via a trivial index instead of an if block or switch.

ui/xui/main-menu.cc
```diff
@@ -792,6 +792,24 @@ void MainMenuSystemView::Draw()
...
+{
+    const char *display_mode;
+
+    if (g_config.display.ui.fit == CONFIG_DISPLAY_UI_FIT_SCALE_16_9) {
```
abaire: FYI: with this many cases to handle, a `switch` block would be more in line with the prevailing xemu style (but don't do it, just index into an array as mentioned above).

ui/xemu.c
```diff
@@ -198,6 +198,11 @@ void sdl2_window_resize(struct sdl2_console *scon)
                       surface_height(scon->surface));
 }
 
+void xemu_get_window_size(window_data *data)
```
abaire: Don't define a structure just to avoid passing two pointers into this method. Just have it take one pointer to receive width, another to receive height.

ui/xui/main-menu.cc
```diff
@@ -27,6 +27,7 @@
 #include "misc.hh"
 #include "gl-helpers.hh"
 #include "reporting.hh"
+#include "popup-menu.hh"
```
abaire: Is this necessary if the home for the display mode function is this file?

ui/xui/main-menu.cc
```diff
@@ -792,6 +793,14 @@ void MainMenuSystemView::Draw()
     }
 }
 
+static const char *GetDisplayMode()
```
abaire: Update popup-menu.cc so it is also using this method so the strings don't need to be translated twice.

abaire (approved):

Fabxx:
Converting this to a draft since i'm reworking the code base to also put these details in the reporting system

PR #1024, opened 2022-06-04
ui: Make monitor text selectable

abaire:
ImGui has some unusual missing functionality around multiline text input, but that seems to be the only widget that would allow text selection. The alternative would be to add a button (or maybe a focus handler, I'm not that familiar w/ imgui) on the entire window that attempts to automatically copy to clipboard, but that feels like it'd be surprising to the user. Even as a frequent monitor user I still often reflexively try to select things and end up moving the window around.

This PR also bumps up the buffer size a bit; I ran into some issues where "help" commands were overflowing scrollback.


mborgerson:
Nice, thanks!

PR #1053, opened 2022-06-10
nv2a: Add RenderDoc support for Windows builds

Ernegien:
This should add RenderDoc support to Windows builds; it requires installing RenderDoc, uncommenting `#define DEBUG_NV2A_GL` within `hw\xbox\nv2a\debug.h`, appending `--debug --enable-renderdoc` to the end of the `build.sh` command arguments, and launching xemu.exe via RenderDoc to test.


meson.build
```diff
@@ -471,7 +471,7 @@ rt = cc.find_library('rt', required: false)
 libdl = not_found
 if 'CONFIG_PLUGIN' in config_host or get_option('renderdoc').enabled()
   libdl = cc.find_library('dl', required: false)
-  if not cc.has_function('dlopen', dependencies: libdl)
+  if targetos != 'windows' and not cc.has_function('dlopen', dependencies: libdl)
```
Ernegien, line 474: I'd assume we don't really need to check for `GetProcAddress` availability here.

meson.build
```diff
@@ -1234,7 +1234,7 @@ have_host_block_device = (targetos != 'darwin' or
 
 have_renderdoc = false
 if have_system and get_option('renderdoc').enabled()
-  if not cc.has_header('renderdoc_app.h')
+  if targetos != 'windows' and not cc.has_header('renderdoc_app.h')
```
Ernegien: Most of this should probably be removed if we end up including `renderdoc_app.h` within xemu. Not yet sure of the best way to check for `renderdoc.dll` availability via meson.

hw/xbox/nv2a/debug.c
```diff
@@ -72,15 +75,26 @@ void gl_debug_initialize(void)
...
+    if (renderdoc) {
+        pRENDERDOC_GetAPI RENDERDOC_GetAPI = (pRENDERDOC_GetAPI)GetProcAddress(
+            renderdoc, "RENDERDOC_GetAPI");
+#else
+    renderdoc = dlopen(renderdoc_lib, RTLD_NOW | RTLD_NOLOAD);
     if (renderdoc) {
         pRENDERDOC_GetAPI RENDERDOC_GetAPI = (pRENDERDOC_GetAPI)dlsym(
             renderdoc, "RENDERDOC_GetAPI");
+#endif
         int ret = RENDERDOC_GetAPI(eRENDERDOC_API_Version_1_1_2,
```
Ernegien, line 98: With this currently hard-coded, I suspect bundling `renderdoc_app.h` within xemu isn't that big of a deal just yet.

meson.build
abaire: You don't have to, renderdoc should take care of making sure it's available when launching the app. I think you should just rip this out in this PR since it's confusing to search for a system header that won't actually be used.

meson.build
abaire: Agreed, I believe it's part of Windows so I don't think a check would ever fail.

hw/xbox/nv2a/renderdoc_app.h
```diff
@@ -0,0 +1,707 @@
+/******************************************************************************
```
abaire: I wonder if this is the best home for the header; maybe we should have a `third_party` dir in the root?

Ernegien: Agreed, as it’s the more standard approach, just wasn’t sure if “polluting” the qemu root would be best versus this. I yield to Matt here.

mborgerson: /thirdparty or in /include would be fine, I'm not too particular about it

Ernegien: I opted for a thirdparty directory at the root which we can shove xxHash and any others in later commit(s).

mborgerson:
Nice, thanks!

PR #1055, opened 2022-06-10
nv2a: Fix clip region at renderscales > 1

### LINKED ISSUE #1054: `surface_scale_factor` application to clipping region yields incorrect clipping rect
### Bug Description

The calculation being done [in pgraph](https://github.com/mborgerson/xemu/blob/master/hw/xbox/nv2a/pgraph.c#L4131) to translate nv2a clipping regions into OpenGL does not properly handle the `surface_scale_factor` due to the way height is calculated.

Specifically, a clipping region 128x128 texture at 2x scale is (128,127) (clipping is inclusive) multiplied by 2, yielding (256,254), causing one extra row to be clipped. Going to 4x yields (512,508), etc...

This contributes to #735 and explains why artifacts reappear at higher render scales.

### Expected Behavior

Increasing render scale should not lead to extra pixels being clipped. The [Window Clip test](https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/tests/window_clip_tests.cpp) should match [HW](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-Window_clip) at every renderscale.

### xemu Version

Affects all recent xemu versions.

### System Information

System independent.

### Additional Context

_No response_

abaire:
Fixes #1054

With this change the output of the [Window Clip test](https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/tests/window_clip_tests.cpp) matches HW at renderscales 1, 4, and 6 (presumably at all scales).

[HW results](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-Window_clip)

### DIFF
--- a/hw/xbox/nv2a/pgraph.c
+++ b/hw/xbox/nv2a/pgraph.c
@@ -4108,23 +4108,26 @@ static void pgraph_shader_update_constants(PGRAPHState *pg,
     }
 
     /* Clipping regions */
-    int max_gl_height = pg->surface_binding_dim.height - 1;
+    unsigned int max_gl_width = pg->surface_binding_dim.width;
+    unsigned int max_gl_height = pg->surface_binding_dim.height;
+    pgraph_apply_scaling_factor(pg, &max_gl_width, &max_gl_height);
+
     for (i = 0; i < 8; i++) {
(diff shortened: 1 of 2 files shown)


abaire:
Note: I believe this PR will block the final version of #735 and will need to be merged first (otherwise 735 will break things in new and exciting ways as mentioned in the comment stream in that PR).

PR #1086, opened 2022-06-22
nv2a: Implement SET_SHADE_MODEL

### LINKED ISSUE #692: nv2a 0x97 -> 0x037c (glShadeModel?) unhandled
### Title

https://xemu.app/titles/53430007

### Bug Description

According to the debug output, 0x97 -> 0x037c is unhandled.
In the [NV20 nouveau sources](https://gitlab.freedesktop.org/mesa/mesa/-/blob/184a690fca700aa9ef2ae2e58b0f4bd86562aa6d/src/mesa/drivers/dri/nouveau/nv20_3d.xml.h#L323), this instruction is `SET_SHADE_MODEL`, and the values it can take matches what is received in the function calls.

### Expected Behavior

The NV2A instruction 0x037C should be handled.

### xemu Version

0.6.2-55-gcaf84884c4

### System Information

N/A

### Additional Context

_No response_

### LINKED ISSUE #912: Use of geometry shader for quads leads to behavior that diverges from HW in boundary conditions
### Bug Description

Discovered this while testing a fix for #801 (dealing with 0 w-param in vertex coords using the fixed function pipeline).

On hardware, using a 0 w-param in the upper left corner of a quad primitive with a checkerboard pattern texture leads to an unusual rendering

xemu uses a geometry shader to transform the quad into a pair of triangles, leading to very different output (this is from my WIP branch which still has a bug)
Screenshot_20220510_124459

Note that this is very similar to the way hardware renders a pair of triangles split in the same way as xemu expands the quad:
hw_bitri

### Expected Behavior

xemu output should match HW output.

### xemu Version

This is off of my work branch, master at the moment will display the same issue wrt. quad being rendered as a pair of triangles, but the actual visuals will differ due to the way the divide by zero in the test is handled.

Version:      0.7.3-36-g8e691278df
Branch:       work
Commit:       8e691278df10628c3fbd1f85c3b5c2de4f73bf00
Date:         Tue May 10 07:19:25 PM UTC 2022

### System Information

OS Platform:  Linux
OS Version:   Ubuntu 21.10
GPU Model:    NVIDIA GeForce GTX 1070/PCIe/SSE2
Driver:       4.0.0 NVIDIA 470.103.01

### Additional Context

[Test](https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/tests/w_param_tests.cpp)
[HW results](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-W_param)

abaire:
Fixes #692
Fixes #912 

Unfortunately this involves some non-trivial & ugly changes due to the fact that GL doesn't allow type qualifiers inside of structs, coupled with the fact that on the nv2a the shade model is only applied to a subset of the attributes.

[Test](https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/tests/shade_model_tests.cpp)
[HW results](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-Shade_model)


abaire:
Draft due to dependence on fix for #912 (the xemu results depart from HW for all of the complex primitives due to the geometry shader expansion). Might be worth merging and doing the geometry shader fix as a followup since it'll add some complexity (I guess it'll need to duplicate the flat shaded attributes across all generated provoking vertices to simulate a single primitive)

UPDATE: I'm wrong, even the basic triangles are using the wrong values, so there's something interesting about what the hardware considers the provoking vertex for flat shading (looks like it uses the first vertex in the primitive instead of the last).

abaire:
Adjusted the geometry shaders such that the provoking vertex matches hardware. Apart from the usual (practically invisible) differences in blending between my GTX1070 and Xbox 1.0, the output with this PR matches the hardware results, so this is ready for review. #912 is also fixed in this PR due to the change I made to the fixed orientation of the triangles when generating quads (they now match hardware).

mborgerson:
Thanks! Rebased and fixed merge conflicts

PR #1149, opened 2022-07-01
UI: Remember debug video size, position and state across restarts

7oxicshadow:
This pull request will remember the debug video window state across restarts. This includes if the advanced tab is active or not.

I was going to add validation of position against the application window size but it does not appear to be needed as ImGui seems to limit the position for us so that it is always recoverable.

We could consider having window size limitations again I don't think this is needed as ImGui should always put us in a position where it can be recovered.


ui/xui/debug.cc
```diff
@@ -230,16 +230,27 @@ DebugVideoWindow::DebugVideoWindow()
...
     m_transparent = false;
+    m_position_restored = false;
 }
 
 void DebugVideoWindow::Draw()
 {
     if (!m_is_open)
         return;
 
+    if( !m_position_restored )
```
abaire: Nit, this should be updated to match the prevailing style within the file (which is also the xemu style generally).

i.e.,

`if (!m_position_restored) {`

ui/xui/debug.cc
```diff
@@ -324,6 +338,15 @@ void DebugVideoWindow::Draw()
         }
 
         ImPlot::PopStyleVar(2);
+
+        ImVec2 dbgwinpos = ImGui::GetWindowPos();
+        g_config.display.debug.video_x_pos = dbgwinpos.x;
+        g_config.display.debug.video_y_pos = dbgwinpos.y;
+
+        ImVec2 dbgwinsize = ImGui::GetWindowSize();
+        g_config.display.debug.video_x_winsize = dbgwinsize.x;
```
abaire: Should this be dividing by the viewport scale before saving (or should the initial sizing not respect the scale)? If the scale is > 1, does this cause doubling on every restart?

ui/xui/debug.cc
7oxicshadow: Done. Thx

ui/xui/debug.cc
7oxicshadow: You are indeed correct, Good catch. I forgot about the scaling factor. 
Updated.

7oxicshadow:
I have added one additional minor change. When testing the scaling I noticed that the debug video window size does not scale when the elements inside it do. Now when changing the scale, it will force a window resize event so that everything scales accordingly.

ui/xui/debug.cc
```diff
@@ -230,16 +230,35 @@ DebugVideoWindow::DebugVideoWindow()
...
+    m_resize_init_complete = false;
+    prev_m_scale = g_viewport_mgr.m_scale;
 }
 
 void DebugVideoWindow::Draw()
 {
     if (!m_is_open)
         return;
 
+    if(!m_position_restored)
```
abaire: Sorry, still a bit off; the brace should be on the same line as the `if` (here and below). Braces only go on a new line for function bodies.

ui/xui/debug.hh
```diff
@@ -31,6 +31,9 @@ class DebugVideoWindow
 public:
     bool m_is_open;
     bool m_transparent;
+    bool m_position_restored;
+    bool m_resize_init_complete;
+    float prev_m_scale;
```
antangelo: nit: Should be `m_prev_scale`, the `m_` is a prefix for member fields.

ui/xui/debug.cc
```diff
@@ -230,16 +230,33 @@ DebugVideoWindow::DebugVideoWindow()
...
+    m_resize_init_complete = false;
+    prev_m_scale = g_viewport_mgr.m_scale;
 }
 
 void DebugVideoWindow::Draw()
 {
     if (!m_is_open)
         return;
 
+    if(!m_position_restored) {
```
antangelo: nit: Should have a space between `if` and the open bracket

antangelo (commented):
Commit title should follow usual standard, that is `<component>: <What was changed>`, in this case the component should be `ui`.
Other than that and a few nits, looks fine to me.

ui/xui/debug.cc
```diff
@@ -230,16 +230,33 @@ DebugVideoWindow::DebugVideoWindow()
...
 }
 
 void DebugVideoWindow::Draw()
 {
     if (!m_is_open)
         return;
 
+    if (!m_position_restored) {
+        ImGui::SetNextWindowPos(ImVec2(g_config.display.debug.video_x_pos, g_config.display.debug.video_y_pos), 
+                                ImGuiCond_Once, ImVec2(0, 0) );
```
antangelo: nit: Missed an extra space
```suggestion
                                ImGuiCond_Once, ImVec2(0, 0));
```

ui/xui/debug.cc
```diff
@@ -324,6 +344,15 @@ void DebugVideoWindow::Draw()
...
 
         ImPlot::PopStyleVar(2);
+
+        ImVec2 dbgwinpos = ImGui::GetWindowPos();
+        g_config.display.debug.video_x_pos = dbgwinpos.x;
+        g_config.display.debug.video_y_pos = dbgwinpos.y;
+
+        ImVec2 dbgwinsize = ImGui::GetWindowSize();
+        g_config.display.debug.video_x_winsize = dbgwinsize.x / g_viewport_mgr.m_scale;
+        g_config.display.debug.video_y_winsize = dbgwinsize.y / g_viewport_mgr.m_scale;
```
antangelo: nit: Variables don't follow snake_case naming convention
```suggestion
        ImVec2 debug_window_pos = ImGui::GetWindowPos();
        g_config.display.debug.video_x_pos = debug_window_pos.x;
        g_config.display.debug.video_y_pos = debug_window_pos.y;

        ImVec2 debug_window_size = ImGui::GetWindowSize();
        g_config.display.debug.video_x_winsize = debug_window_size.x / g_viewport_mgr.m_scale;
        g_config.display.debug.video_y_winsize = debug_window_size.y / g_viewport_mgr.m_scale;
```

antangelo (commented):
Couple more nits, otherwise the changes look OK to me

PR #1146, opened 2022-07-01
nv2a: Handle framebuffer CPU blit and PVIDEO only rendering

abaire:
Handles two edge cases:
1) CPU blits to the framebuffer without using 3D rendering or after 3D rendering was previously done with a surface shape that does not match the framebuffer (e.g., a different pitch).
2) Fullscreen PVIDEO rendering without any 3D rendering.

In both cases this change prevents the pgraph code from returning early,
bypassing the special case VGA handling in `sdl2_gl_refresh` and instead
using a special framebuffer texture to render the contents of VRAM.

Fixes #652
Fixes #1165
Fixes #2220
Fixes #2443

[Tests](https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/tests/antialiasing_tests.cpp)
[HW results](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-Antialiasing_tests)

PR #1178, opened 2022-07-08
nv2a: Handle SIZE_IN > SIZE_OUT case

### LINKED ISSUE #330: [Assertion] in_pitch >= in_width * 2
### Upon ejecting the disk
https://xemu.app/titles/4d53005f/#Voodoo-Vince
https://xemu.app/titles/54540004/#Serious-Sam

### Crashes After Finishing Intro
https://xemu.app/titles/4d57001c/#Midway-Arcade-Treasures
https://xemu.app/titles/41510003/#Sentou-Yousei-Yukikaze-Yousei-no-Mau-Sora 
(This should be in the intros, someone didn't file the report properly)
https://xemu.app/titles/57450001/#Ultimate-Beach-Soccer

### Bug Description
Known games with this issue:
- Midway Arcade Treasures* 
- Sentou Yousei Yukikaze Yousei no Mau Sora*
- Ultimate-Beach-Soccer*
- Voodoo Vince
- Serious Sam
- All games that use pvideo to display intro videos

The above games cause an assertion failure if you attempt to load an ISO while the game intro videos are playing
Additionally games with an Asterisk crash regardless of the game is ejected or not.
https://github.com/mborgerson/xemu/blob/72b511c091895b6ea267ad1305afd94ffe580135/hw/xbox/nv2a/pgraph.c#L4303

Reproduction:
1: Load the game and close Xemu so it runs it by default on the next boot
2: Open Xemu, and wait until the game's intro videos are playing
3: Load Disc (It can be any ISO, even the same game)
4: Assertion failed.
xemu: ../hw/xbox/nv2a/pgraph.c:4897: pgraph_render_display_pvideo_overlay: Assertion in_pitch >= in_width * 2' failed.

Reproduction Alternative:
1: Load game and close Xemu so it runs by default on next boot
2: Open Xemu and Eject disk when the game's intro videos are playing
3: Assertion failed.

Reproduction video on Voodoo Vince:

https://user-images.githubusercontent.com/105917902/169626030-474f3638-8323-4b2b-ab58-f0518077f34d.mp4

### Expected Behavior
I should be able to make it to the menus without a crash

<h3>Version</h3>
<ul>
</ul>
<h3>System Information</h3>

Field | Value
-- | --
OS | Windows 10 64-bit (21H1)

Additional Context
The games on the list crash automatically. I Can confirm from my testing. 
Also, I adapted so of this issue from [doinkers](https://github.com/doinkers) So I big thank for his help

abaire:
`NV_PVIDEO_SIZE_IN` is set to 0xFFFFFFFF during initialization and teardown of the PVIDEO overlay. In some cases this can happen before the overlay is explicitly stopped, leading to an assert.

On hardware any SIZE_IN > SIZE_OUT is silently capped to SIZE_OUT.

Fixes #330

[Test](https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/tests/pvideo_tests.cpp)

### DIFF
--- a/hw/xbox/nv2a/nv2a.c
+++ b/hw/xbox/nv2a/nv2a.c
@@ -278,6 +278,7 @@ static void nv2a_reset(NV2AState *d)
 
     memset(d->pfifo.regs, 0, sizeof(d->pfifo.regs));
     memset(d->pgraph.regs, 0, sizeof(d->pgraph.regs));
+    memset(d->pvideo.regs, 0, sizeof(d->pvideo.regs));
 
     d->pcrtc.start = 0;
     d->pramdac.core_clock_coeff = 0x00011C01; /* 189MHz...? */
--- a/hw/xbox/nv2a/nv2a_int.h
+++ b/hw/xbox/nv2a/nv2a_int.h
@@ -266,7 +266,9 @@ typedef struct PGRAPHState {
         GLuint pvideo_tex;
         GLint pvideo_enable_loc;
         GLint pvideo_tex_loc;
+        GLint pvideo_in_pos_loc;
         GLint pvideo_pos_loc;
+        GLint pvideo_scale_loc;
         GLint pvideo_color_key_enable_loc;
         GLint pvideo_color_key_loc;
         GLint palette_loc[256];
--- a/hw/xbox/nv2a/nv2a_regs.h
+++ b/hw/xbox/nv2a/nv2a_regs.h
@@ -655,6 +655,7 @@
 #   define NV_PVIDEO_POINT_IN_T                               0xFFFE0000
 #define NV_PVIDEO_DS_DX                                  0x00000938
 #define NV_PVIDEO_DT_DY                                  0x00000940
+#   define NV_PVIDEO_DIN_DOUT_UNITY                           0x00100000
 #define NV_PVIDEO_POINT_OUT                              0x00000948
 #   define NV_PVIDEO_POINT_OUT_X                              0x00000FFF
 #   define NV_PVIDEO_POINT_OUT_Y                              0x0FFF0000
--- a/hw/xbox/nv2a/pgraph.c
+++ b/hw/xbox/nv2a/pgraph.c
@@ -21,6 +21,8 @@
 
 #include "nv2a_int.h"
 
+#include <math.h>
+
 #include "nv2a_vsh_emulator.h"
 #include "s3tc.h"
 #include "ui/xemu-settings.h"
@@ -4998,7 +5000,9 @@ static void pgraph_init_display_renderer(NV2AState *d)
         "uniform sampler2D tex;\n"
         "uniform bool pvideo_enable;\n"
         "uniform sampler2D pvideo_tex;\n"
+        "uniform vec2 pvideo_in_pos;\n"
         "uniform vec4 pvideo_pos;\n"
+        "uniform vec3 pvideo_scale;\n"
         "uniform bool pvideo_color_key_enable;\n"
         "uniform vec4 pvideo_color_key;\n"
         "uniform vec2 display_size;\n"
@@ -5012,15 +5016,14 @@ static void pgraph_init_display_renderer(NV2AState *d)
         "    out_Color.rgba = texture(tex, texCoord);\n"
         "    if (pvideo_enable) {\n"
         "        vec2 screenCoord = gl_FragCoord.xy - 0.5;\n"
-        "        vec4 extent = vec4(pvideo_pos.xy, pvideo_pos.xy + pvideo_pos.zw);\n"
-        "        bvec4 clip = bvec4(lessThan(screenCoord, extent.xy),\n"
-        "                           greaterThan(screenCoord, extent.zw));\n"
-        "        if (!any(clip)) {\n"
-        "            vec2 videoCoord = (screenCoord - pvideo_pos.xy) / pvideo_pos.zw;\n"
-        "            videoCoord.y *= -1.0;\n"
-        "            if (!pvideo_color_key_enable || out_Color.rgba == pvideo_color_key) {\n"
-        "               out_Color.rgba = texture(pvideo_tex, videoCoord);\n"
-        "            }\n"
+        "        vec4 output_region = vec4(pvideo_pos.xy, pvideo_pos.xy + pvideo_pos.zw);\n"
+        "        bvec4 clip = bvec4(lessThan(screenCoord, output_region.xy),\n"
+        "                           greaterThan(screenCoord, output_region.zw));\n"
+        "        if (!any(clip) && (!pvideo_color_key_enable || out_Color.rgba == pvideo_color_key)) {\n"
+        "            vec2 out_xy = (screenCoord - pvideo_pos.xy) * pvideo_scale.z;\n"
+        "            vec2 in_st = (pvideo_in_pos + out_xy * pvideo_scale.xy) / textureSize(pvideo_tex, 0);\n"
+        "            in_st.y *= -1.0;\n"
+        "            out_Color.rgba = texture(pvideo_tex, in_st);\n"
         "        }\n"
         "    }\n"
         "}\n";
@@ -5029,7 +5032,9 @@ static void pgraph_init_display_renderer(NV2AState *d)
     pg->disp_rndr.tex_loc = glGetUniformLocation(pg->disp_rndr.prog, "tex");
     pg->disp_rndr.pvideo_enable_loc = glGetUniformLocation(pg->disp_rndr.prog, "pvideo_enable");
     pg->disp_rndr.pvideo_tex_loc = glGetUniformLocation(pg->disp_rndr.prog, "pvideo_tex");
+    pg->disp_rndr.pvideo_in_pos_loc = glGetUniformLocation(pg->disp_rndr.prog, "pvideo_in_pos");
     pg->disp_rndr.pvideo_pos_loc = glGetUniformLocation(pg->disp_rndr.prog, "pvideo_pos");
+    pg->disp_rndr.pvideo_scale_loc = glGetUniformLocation(pg->disp_rndr.prog, "pvideo_scale");
     pg->disp_rndr.pvideo_color_key_enable_loc = glGetUniformLocation(pg->disp_rndr.prog, "pvideo_color_key_enable");
     pg->disp_rndr.pvideo_color_key_loc = glGetUniformLocation(pg->disp_rndr.prog, "pvideo_color_key");
     pg->disp_rndr.display_size_loc = glGetUniformLocation(pg->disp_rndr.prog, "display_size");
@@ -5064,6 +5069,14 @@ static uint8_t *convert_texture_data__CR8YB8CB8YA8(const uint8_t *data,
     return converted_data;
 }
 
+static inline float pvideo_calculate_scale(unsigned int din_dout,
+                                           unsigned int output_size)
+{
+    float calculated_in = din_dout * (output_size - 1);
+    calculated_in = floorf(calculated_in / (1 << 20) + 0.5f);
+    return (calculated_in + 1.0f) / output_size;
+}
+
 static void pgraph_render_display_pvideo_overlay(NV2AState *d)
 {
     PGRAPHState *pg = &d->pgraph;
@@ -5076,7 +5089,7 @@ static void pgraph_render_display_pvideo_overlay(NV2AState *d)
     // stopping it.
     // Since the value seems to be set to 0xFFFFFFFF only in cases where the
     // content is not valid, it is probably good enough to treat it as an
-    // implicit stop. 
+    // implicit stop.
     bool enabled = (d->pvideo.regs[NV_PVIDEO_BUFFER] & NV_PVIDEO_BUFFER_0_USE)
         && d->pvideo.regs[NV_PVIDEO_SIZE_IN] != 0xFFFFFFFF;
     glUniform1ui(d->pgraph.disp_rndr.pvideo_enable_loc, enabled);
@@ -5093,12 +5106,11 @@ static void pgraph_render_display_pvideo_overlay(NV2AState *d)
     int in_height =
         GET_MASK(d->pvideo.regs[NV_PVIDEO_SIZE_IN], NV_PVIDEO_SIZE_IN_HEIGHT);
 
-    /* FIXME
     int in_s = GET_MASK(d->pvideo.regs[NV_PVIDEO_POINT_IN],
                         NV_PVIDEO_POINT_IN_S);
     int in_t = GET_MASK(d->pvideo.regs[NV_PVIDEO_POINT_IN],
                         NV_PVIDEO_POINT_IN_T);
-    */
+
     int in_pitch =
         GET_MASK(d->pvideo.regs[NV_PVIDEO_FORMAT], NV_PVIDEO_FORMAT_PITCH);
     int in_color =
@@ -5109,15 +5121,26 @@ static void pgraph_render_display_pvideo_overlay(NV2AState *d)
     unsigned int out_height =
         GET_MASK(d->pvideo.regs[NV_PVIDEO_SIZE_OUT], NV_PVIDEO_SIZE_OUT_HEIGHT);
 
+    float scale_x = 1.0f;
+    float scale_y = 1.0f;
+    unsigned int ds_dx = d->pvideo.regs[NV_PVIDEO_DS_DX];
+    unsigned int dt_dy = d->pvideo.regs[NV_PVIDEO_DT_DY];
+    if (ds_dx != NV_PVIDEO_DIN_DOUT_UNITY) {
+        scale_x = pvideo_calculate_scale(ds_dx, out_width);
+    }
+    if (dt_dy != NV_PVIDEO_DIN_DOUT_UNITY) {
+        scale_y = pvideo_calculate_scale(dt_dy, out_height);
+    }
+
     // On HW, setting NV_PVIDEO_SIZE_IN larger than NV_PVIDEO_SIZE_OUT results
     // in them being capped to the output size, content is not scaled. This is
     // particularly important as NV_PVIDEO_SIZE_IN may be set to 0xFFFFFFFF
     // during initialization or teardown.
     if (in_width > out_width) {
-        in_width = out_width;
+        in_width = floorf((float)out_width * scale_x + 0.5f);
     }
     if (in_height > out_height) {
-        in_height = out_height;
+        in_height = floorf((float)out_height * scale_y + 0.5f);
     }
 
     /* TODO: support other color formats */
@@ -5164,8 +5187,11 @@ static void pgraph_render_display_pvideo_overlay(NV2AState *d)
                  GL_UNSIGNED_BYTE, tex_rgba);
     g_free(tex_rgba);
     glUniform1i(d->pgraph.disp_rndr.pvideo_tex_loc, 1);
+    glUniform2f(d->pgraph.disp_rndr.pvideo_in_pos_loc, in_s, in_t);
     glUniform4f(d->pgraph.disp_rndr.pvideo_pos_loc,
                 out_x, out_y, out_width, out_height);
+    glUniform3f(d->pgraph.disp_rndr.pvideo_scale_loc,
+                scale_x, scale_y, 1.0f / pg->surface_scale_factor);
 }
 
 static void pgraph_render_display(NV2AState *d, SurfaceBinding *surface)
--- a/hw/xbox/nv2a/pgraph.c
+++ b/hw/xbox/nv2a/pgraph.c
@@ -5054,8 +5054,9 @@ static uint8_t *convert_texture_data__CR8YB8CB8YA8(const uint8_t *data,
     int x, y;
     for (y = 0; y < height; y++) {
         const uint8_t *line = &data[y * pitch];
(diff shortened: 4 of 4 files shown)


Triticum0:
It worked for https://xemu.app/titles/57450001/#Ultimate-Beach-Soccer

abaire:
> It worked for https://xemu.app/titles/57450001/#Ultimate-Beach-Soccer

Excellent. I was able to figure out why the tests were breaking, but fixing that also slightly changed the behavior; rather than immediately tearing down the overlay it looks like it just freezes the frame. Will push a new change in a couple hours when I can get back to this.

UPDATE: It doesn't freeze the frame, it looks like it actually just treats the input size as the same as the output size and that this is generally the case (setting SIZE_IN to 2x SIZE_OUT acts as though they're equal, it does not scale the content to fit SIZE_OUT)

Triticum0:
Didn't fix the issue with your latest commit. Tested all games with the previous version and it fixed them but with your latest commit, the fvm is never cleared from the screen, so is over the menus. Also, midway arcade still crashes with the assertion on the latest commit.

Triticum0:
This is what I get from midway arcade
```
  0x00007ff7054b751f in pgraph_render_display_pvideo_overlay (
    d=d@entry=0x21405219040) at ../hw/xbox/nv2a/pgraph.c:5114
        pg = 0x2140523bc98
        enabled = <optimized out>
        base = 0
        limit = 134217727
        offset = 60874752
        in_width = 720
        in_height = 480
        in_pitch = 1280
        in_color = <optimized out>
        out_width = 720
        out_height = 480
        out_x = 0
        out_y = 0
        color_key_enabled = <optimized out>
        color_key = <optimized out>
        end = <optimized out>

```

Triticum0:
This might also be useful 
 ```
    pg = 0x2140523bc98
        width = 720
        height = 480
        pline_offset = 3072
        pstart_addr = 15872000
        pline_compare = 1023
        line_offset = 1
        recreate = <optimized out>
        DrawBuffers = {36064}
#4  0x00007ff7054c21a4 in pgraph_gl_sync (d=d@entry=0x21405219040) at ../hw/xbox/nv2a/pgraph.c:5266
        pline_offset = 3072
        pstart_addr = 15872000
        pline_compare = 1023
        surface = 0x2140437b5b0
```

**Comment by Triticum0** [comment]
Also Trying to go in-game on Ultimate Beach Soccer by memory also shows the whole screen corrupted previously you could make it in-game and actually see what was going on.

**Comment by abaire** [comment]
> This is what I get from midway arcade
> 
> ```
>   0x00007ff7054b751f in pgraph_render_display_pvideo_overlay (
>     d=d@entry=0x21405219040) at ../hw/xbox/nv2a/pgraph.c:5114
>         pg = 0x2140523bc98
>         enabled = <optimized out>
>         base = 0
>         limit = 134217727
>         offset = 60874752
>         in_width = 720
>         in_height = 480
>         in_pitch = 1280
>         in_color = <optimized out>
>         out_width = 720
>         out_height = 480
>         out_x = 0
>         out_y = 0
>         color_key_enabled = <optimized out>
>         color_key = <optimized out>
>         end = <optimized out>
> ```

That one is unusual since the in_size is already equal to the out_size and all of the values other than pitch look reasonable. I think this is a different bug and should be filed separately. I'll write a separate test that lies about the pitch and see how the HW behaves.

I also just ordered the game and can look at it in a couple weeks when it arrives.


UPDATE: from my test it looks like the hardware can handle the pitch being < the expected pitch just fine. It still respects the pitch, so you can get a staircase effect.

**Comment by abaire** [comment]
> Also Trying to go in-game on Ultimate Beach Soccer by memory also shows the whole screen corrupted previously you could make it in-game and actually see what was going on.

Can you enable the `nv2a_reg_write` trace and capture the end of the video for one of those games? I need to see what is happening to the `PVIDEO` register when the overlay should be torn down.

As far as I can see from my test case, it is valid to give an incorrect (larger than output) size and the hardware does not tear down the display, so something else must be happening for these games that exhibit issues without the need to eject the disc or reset.

**Comment by Triticum0** [comment]
How do you specify with trace-event to only log pvideo register writes. Also, how do I step through the could without hitting an assertion? Do I have to place a breakpoint and then trace from there because when I run Xemu on symbols using gdb, it seems I can't type any commands while it running through the terminal?

**Comment by abaire** [comment]
> How do you specify with trace-event to only log pvideo register writes. Also, how do I step through the could without hitting an assertion? Do I have to place a breakpoint and then trace from there because when I run Xemu on symbols using gdb, it seems I can't type any commands while it running through the terminal?

In the monitor `trace-event nv2a_reg_write on` will turn on the reg write events.

If you're running with this PR it sounds like you shouldn't hit the assertion (for `Ultimate Beach Soccer`, let's ignore the Midway case for the moment, I'll have a prospective fix later as I can already see that the HW handles pitch < expected gracefully).

**Comment by Triticum0** [comment]
Here you go. I figure out how to do it eventually. Is stepping through the code per frame possible without your debug pr. Thank for the help
[Ultimate-Beach-Soccer.log](https://github.com/mborgerson/xemu/files/9075395/Ultimate-Beach-Soccer.log)

**Comment by abaire** [comment]
> Here you go. I figure out how to do it eventually. Is stepping through the code per frame possible without your debug pr. Thank for the help [Ultimate-Beach-Soccer.log](https://github.com/mborgerson/xemu/files/9075395/Ultimate-Beach-Soccer.log)

You can in GDB; set a breakpoint in `nv2a_profile_flip_stall` in pgraph.c and it'll be close to the same behavior as with my debugger window.

**Comment by Triticum0** [comment]
how do you do  that

**Comment by abaire** [comment]
> how do you do that

Ping #general or #help, it's not something related to this code review :)

Can you check Midway again with the latest version of this PR? I confirmed that a pitch less than a full row is supported by the hardware and removed the assert. The output doesn't exactly match HW but is close.


Also looking at the log you posted, I don't see an explicit teardown of the PVIDEO, but I do see some writes to PVIDEO registers that I've not seen before so I'm guessing that may be what's missing. Will write another test case to see what happens when I duplicate what I see logged.

UPDATE: Writing to those registers has no effect, so I'm not sure what causes the PVIDEO to be hidden in that game. As far as I can see the overlay should remain up forever...

**Comment by abaire** [comment]
@Triticum0 can you test all games again with the commit I just now pushed (sorry to keep asking for retests!)

I don't see anything in the Ultimate Beach Soccer log that looks like it should tear down the overlay, but I also have checked a number of games and only see 0xFFFFFFFF being used when video is about to be shown (immediately replaced with a valid size) or about to be torn down (promptly followed by a write to the STOP register with the low bit set). Thus I think it's safe for now to treat it as a command that should just hide the video. If we find a real world case where this causes a problem (e.g., a video overlay blinking or disappearing) we can dig into how UBS is causing the overlay to be hidden and remove the FIXME that I added.

**Comment by Triticum0** [comment]
I didn't mention this earlier for Ultimate Beach Soccer but the fvm is not even rendered on the screen correctly

https://user-images.githubusercontent.com/71127352/178083212-f3e22ce6-71a8-41b3-a157-1506c8d59e8b.mp4

https://www.youtube.com/watch?v=tbx3iUc26kk

**Comment by Triticum0** [comment]
I retest tomorrow.

**Comment by Triticum0** [comment]
Won't be able to test until later tonight, but I should also mention the misalignment of the FMV on masters only happens once the FMV ends or if you try to the bailout of the FMV manually to go to the menus, which causes the assert to hit. In the recent pr where you try to handle the SIZE_IN > SIZE_OUT case is what causes the misalignment to happen during the whole FMV.

**Comment by abaire** [comment]
> Won't be able to test until later tonight, but I should also mention the misalignment of the FMV on masters only happens once the FMV ends or if you try to the bailout of the FMV manually to go to the menus, which causes the assert to hit. In the recent pr where you try to handle the SIZE_IN > SIZE_OUT case is what causes the misalignment to happen during the whole FMV.

If you can let me know if the behavior is the same after the latest commit I can take a look. From the log you sent I'm not sure how this could happen, as the size in is clearly less than the size out except for the end of the video when it sets it to 0xFFFFFFFF

```
nv2a_reg_write PVIDEO [0x704] = 0x00000000: Stop: Overlay active: False, Stop immediately: True
nv2a_reg_write PVIDEO [0xb00] = 0x00000000: color key: 0x00000000
nv2a_reg_write PVIDEO [0x920] = 0x0337d000: offset[0]: Offset: 0x000cdf40
nv2a_reg_write PVIDEO [0x930] = 0x00000000: point_in[0]: 0, 0
nv2a_reg_write PVIDEO [0x928] = 0x024002d0: size_in[0]: 360 x 576
nv2a_reg_write PVIDEO [0x938] = 0x001200cd: ds/dx[0]: 0x00024019 (147481)
nv2a_reg_write PVIDEO [0x940] = 0x001334e9: dt/dy[0]: 0x0001334e (78670)
nv2a_reg_write PVIDEO [0x948] = 0x00000000: point_out[0]: 0, 0
nv2a_reg_write PVIDEO [0x950] = 0x01e00280: size_out[0]: 640 x 480
nv2a_reg_write PVIDEO [0x958] = 0x000105c0: format[0]: Pitch: 1472, LE_CR8YB8CB8YA8, Display always
nv2a_reg_write PVIDEO [0x908] = 0x07ffffff: limit[0]: 2097151 (0x001fffff)
nv2a_reg_write PVIDEO [0x140] = 0x00000001: Interrupt enabled: Buffer0: True, Buffer1: False
nv2a_reg_write PVIDEO [0x700] = 0x00000001: Use: Buffer0: True, Buffer1: False
```

So for the frames in the log, the change should have no effect at all until the part that would've asserted. It may be that there is some initial frame that isn't in the log that has things mismatched, but the output size and position are controlled by the out register and not influenced by this PR.

If it's still doing this on with the latest commit, can you get a `nv2a_reg_write` trace of the entire video playback? (Start the trace before you start the video)

**Comment by Triticum0** [comment]
Here is the log from the Xbox logo to the menu. Note this is not using you lastest commit
[Second round.txt](https://github.com/mborgerson/xemu/files/9077879/Second.round.txt)

**Comment by Triticum0** [comment]
Hear it is with the you lastest commit:
[Ultimate-Beach-Soccer-Latest.txt](https://github.com/mborgerson/xemu/files/9077946/Ultimate-Beach-Soccer-Latest.txt)
[Sentou-Yousei-Yukikaze-Yousei-no-Mau-Sora-latest.txt](https://github.com/mborgerson/xemu/files/9077948/Sentou-Yousei-Yukikaze-Yousei-no-Mau-Sora-latest.txt)
[midway-arcade-treasures-latest.txt](https://github.com/mborgerson/xemu/files/9077961/midway-arcade-treasures-latest.txt)

Note: lastest commit caused flicking in all Fmv's I tested

**Comment by abaire** [comment]
Thanks!

> Note: lastest commit caused flicking in all Fmv's I tested

This is almost certainly because this PR doesn't include the fix for #1049. If you patch this PR on top of #1168 (which fixes the incorrect handling of the stop register) I suspect it won't flicker anymore.


I also suspect the incorrect rendering may be because xemu doesn't do anything with these registers:
```
#define NV_PVIDEO_DS_DX                                  0x00000938
#define NV_PVIDEO_DT_DY                                  0x00000940
```

Given the names they seem likely to be a ratio between the input and output sizes. I'll leave this in draft mode until I have a chance to figure out what those registers are actually doing when the values aren't set to `0x00100000` (which appears to be the case in every game I've tested that uses PVIDEO, but is not the case in some of the logs you've posted)



UPDATE: Confirmed that these control scaling of the video. They are a simple ratio of `(input - 1) << 20 / (output - 1)`.

**Comment by abaire** [comment]
@Triticum0 for Ultimate Beach Soccer, can you enable `nv2a_pgraph_method_unhandled` tracing and capture a log for the end of the video?

`trace-event nv2a_pgraph_method_unhandled on` and also `trace-event nv2a_reg_write on` at the same time.

I'm wondering if there's some pgraph command that has a side effect of disabling the pvideo overlay since I can't see anything in that log that makes me think it'd hide automatically.

**Comment by Triticum0** [comment]
Here you go:
[Ultimate-Beach-Soccer.txt](https://github.com/mborgerson/xemu/files/9078014/Ultimate-Beach-Soccer.txt)

**Comment by abaire** [comment]
> Here you go: [Ultimate-Beach-Soccer.txt](https://github.com/mborgerson/xemu/files/9078014/Ultimate-Beach-Soccer.txt)

Thanks, I see some unhandled things after the last PVIDEO frame but there is a bunch of register writing going on in between so I'm not convinced they're related.

Mind doing one more?
```
trace-event nv2a_reg_write on
trace-event nv2a_pgraph_method_*  on
```

This will create a lot of output so it might be _very_ slow when you turn everything on.

**Comment by Triticum0** [comment]
Here
[Ultimate-Beach-Soccer1.txt](https://github.com/mborgerson/xemu/files/9078029/Ultimate-Beach-Soccer1.txt)

**Comment by theboy181** [comment]
Tested a bunch of games that started with videos, and was able to eject, or start a new game without issue! Great Job, and thanks!

**Comment by abaire** [comment]
> Here [Ultimate-Beach-Soccer1.txt](https://github.com/mborgerson/xemu/files/9078029/Ultimate-Beach-Soccer1.txt)

@Triticum0 I messed up and had you put an extra undescore so it didn't actually capture the interesting stuff, sorry!

Can you do one more?

```
trace-event nv2a_reg_write on
trace-event nv2a_pgraph_*  on
```

The second trace is more generic and will capture more data.

abaire:
@Triticum0 I just pushed another commit that handles the DS_DX and DT_DY registers, can you see if it fixes the misalignment you saw in Ultimate.../Sentou/Midway? All of those have non-unity scale factors.

Triticum0:
Yeah sure. I couldn't test with pgraph enabled it keeps on freezing after a couple of frames, not sure why. So I'm probably going to try Linux to see if the issue is resolved on that platform. Give me an hour and I give you what results occur from the commit

Triticum0:
@abaire Yeah it not completely fixed.

https://user-images.githubusercontent.com/71127352/178354362-eb956b29-1ace-44f9-874d-f82d68f6b486.mp4

[123.txt](https://github.com/mborgerson/xemu/files/9087431/123.txt)

abaire:
> @abaire Yeah it not completely fixed.

~~Thanks, I'll see if I can build a better test case. Before this last push was it the same behavior where the video wrapped around or was it something different?~~ I have a test case that reproduces the problem, will push a new fix in a bit.

Also not sure if the 123.txt was intended to have the pgraph methods or not, but it just seems to be the register writes.

Triticum0:
As I said I can you pgraph methods as it is so unstable that it freezes after a few steps so I can't take a full log. I could only do it for about for couple of frames max but I don't know if it will help

Triticum0:
I will send to txt file of one on the first frame of the intro and one on the last using pgraph.

abaire:
> I will send to txt file of one on the first frame of the intro and one on the last using pgraph.

All I need is the end of the video, I'm looking for whatever is causing the overlay to be hidden, since the 0xFFFFFFFF hack that I have in there right now does not match hardware behavior. If you can capture the last frame of the video and the first frame where the video disappears, that should give me all the info we can get from the pgraph log.

Triticum0:
Here you go the first one is not important.
[first.txt](https://github.com/mborgerson/xemu/files/9087772/first.txt)
[important.txt](https://github.com/mborgerson/xemu/files/9087773/important.txt)

abaire:
Surprisingly when I run my test on HW w/ a 720x576 input into a 640x480 output I do see wraparound in the X, but no wraparound on the Y. When I run it in xemu both wrap around (like the video you posted).

@Triticum0 is this a PAL game running on an NTSC-configured image by any chance? I'm curious as to how this game behaves on actual HW (running on an NTSC machine, assuming your image is NTSC based on the 480p output instead of 576p)

abaire:
> Here you go the first one is not important.

This is perfect, thanks! I see it's doing a bunch of interesting 0x62 and 0x9F commands which are related to 2D surfaces. It is possible this triggers the hiding of the overlay, I'll see if I can reproduce that next.

abaire:
@Triticum0 there was a silly mistake in the way I was doing the scaling, fixed now. I don't think it matches hardware but hopefully looks better now.

Triticum0:
Still has the orginal issue as before im using a usa verison i redumped it off my xbox again just to make sure the dump wasn't causing any issue hopfully this log will be more useful
 
Here another log
[first.txt](https://github.com/mborgerson/xemu/files/9088114/first.txt)

https://user-images.githubusercontent.com/71127352/178374732-6fed84e7-f815-462d-810e-6d23685da943.mp4

Triticum0:
If you still can't figure out the issue, I can paypal you the money to buy the game if this is not working for you. It only 25 dollar on ebay which i don't mind paying for it.

abaire:
I'm confused as to how it's behaving so differently from my tests using the exact same parameters.. You don't have to capture any more logs, none of the changes I'm making to attempt to fix the scaling will have any effect on what is logged so there's nothing useful in there.

What are your View menu settings (resolution scale and display mode)?

Triticum0:
Also this fixes breaks other games with seeing over y axis top boarders as well as misalignment on fmv other than the games affected.

abaire:
> Also this fixes breaks other games with seeing over y axis top boarders as well as misalignment on fmv other than the games affected.

Can you give me a few examples? That'll probably help me figure out where things are different. I've been using Serious Sam as a spot test, the intro FMV looks correct to me, as does the video that plays on the loading screen.

abaire:
Update: I found a video that ends up being wrong, the THQ intro for Spongebob. Will debug and push another fix.

Triticum0:
scaled and 1x internal resoultion. The seeing over y axis top boarders only occurs on pal region game i tested call England International Football on the codemaster intro

abaire:
There was a conflict between the way I was blindly capping width/height in the older commit and the correct handling of the dIn/dOut registers. The output of the pgraph test now matches HW and the Spongebob THQ video (720x480=>640x480) is no longer wrapping incorrectly.

@Triticum0 can you check again with this new commit?

Triticum0:
yes, give me 15 mins

Triticum0:
I retested it looks correct there are just now 4 diffrent screens simultaneously 

https://user-images.githubusercontent.com/71127352/178389268-85446566-9f96-4d72-90ad-8bb2bb87a36f.mp4

abaire:
> I retested it looks correct there are just now 4 diffrent screens simultaneously
> 
>  2022-07-12.02-34-27.-.Copy.mp4

I went ahead and ordered the game, it's not clear to me why the behavior differs so much between what you're seeing and what I see in other games and my test program. Will come back to this in a couple weeks when it arrives.

I am somewhat curious what you see if you run the "PVIDEO" > "PAL into NTSC" test from https://github.com/abaire/nxdk_pgraph_tests/releases, but I expect I'll be able to figure out whatever is broken once I can test directly.

Triticum0:
It is the same thing that happens with the video in the in the commit
xemu-2022-07-12-02-54-28

abaire:
Thanks for testing, there is something very different between the behavior on your machine and my my two, I don't see doubling on either.

I assume you're still testing in Windows on your nvidia card?

Just to rule it out, can you also try rebooting?

Triticum0:
yes, I'm using Nvidia 30 series and windows updated to the latest drivers + reboot did nothing.

abaire:
> yes, I'm using Nvidia 30 series and windows updated to the latest drivers + reboot did nothing.

Thanks for checking. I have one last thing to try, I don't expect it to make any difference but want to rule out the possibility that `textureSize` does something different in your config than mine. Mind doing one last test?

Thanks again for all your help remote-debugging this :)

abaire:
@Triticum0 I was able to reproduce the quadrupling of output by setting my render scale to 2x, can you double check that your render scale was still set to 1x in your last round of testing?

I also just pushed a fix that handles higher scales, so hopefully that'll also make the problem go away.

Triticum0:
@abaire Yeah can say with full confidence that you fixed all the issues that I was having on the pr. It seems to match hardware now. Thank you for your patients and hard work.
PR:

https://user-images.githubusercontent.com/71127352/178830230-d2629740-926a-44c0-bf0c-0393a81ff15e.mp4


Hardware:
https://www.youtube.com/watch?v=tbx3iUc26kk

Note: retested all games on the list not work properly and other issues i have are fixed

Triticum0:
Note: special case doesn't fix [Broken-Sword-The-Sleeping-Dragon](https://xemu.app/titles/5451001d/#Broken-Sword-The-Sleeping-Dragon) and [Disney-s-The-Haunted-Mansion](https://xemu.app/titles/544d000d/#Disney-s-The-Haunted-Mansion). haven't test with both pr combined and squash commits when made the test build don't know if it has an effect?

abaire:
> Note: special case doesn't fix [Broken-Sword-The-Sleeping-Dragon](https://xemu.app/titles/5451001d/#Broken-Sword-The-Sleeping-Dragon) and [Disney-s-The-Haunted-Mansion](https://xemu.app/titles/544d000d/#Disney-s-The-Haunted-Mansion). haven't test with both pr combined and squash commits when made the test build don't know if it has an effect?

I suspect they're just different underlying issues, can you file separate title bugs for them? 
The fact that Broken Sword interacts poorly with #1146 is in indication that it's doing some interesting things with CPU blitting which may contribute to or cause whatever the current behavior is. I'm not sure what the behavior is for the haunted mansion so I'd need to see a video of that to guess at what might be happening (ideally in a new title-specific bug).

Triticum0:
Edit #1183 has I video so you can look at what it looks like. Ping me if you need any help

PR #1180, opened 2022-07-11
nv2a: Fix missing flat qualifier for inv_w param

### LINKED ISSUE #1179: Blinx: Corrupt shadows due to incorrect qualifier on `vtx_inv_w` in flat shading mode
### Title

https://xemu.app/titles/4d530013/#Blinx-The-Time-Sweeper

### Bug Description

The shadows in Blinx generate incorrect geometry due to an issue with the qualifier on the inverse w param in the vertex shader.

Pic from Degorth

### Expected Behavior

The shadows should render normally (as they did prior to 0.7.52

### xemu Version

Anything post 0.7.52, which introduced the regression

### System Information

n/a. This is not system specific

### Additional Context

_No response_

abaire:
`vtx_inv_w` should also be flat, since interpolating it causes incorrect values
for the flat shaded colors.

Fixes #1179

[Test](https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/tests/shade_model_tests.cpp)
[HW results](https://github.com/abaire/nxdk_pgraph_tests_golden_results/wiki/Results-Shade_model)

### DIFF
--- a/hw/xbox/nv2a/psh.c
+++ b/hw/xbox/nv2a/psh.c
@@ -801,10 +801,17 @@ static MString* psh_convert(struct PixelShader *ps)
 
     /* calculate perspective-correct inputs */
     MString *vars = mstring_new();
-    mstring_append(vars, "vec4 pD0 = vtxD0 / vtx_inv_w;\n");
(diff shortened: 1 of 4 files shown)


abaire:
Still some differences in non-triangle primitives that need to be worked out.

PR #1235, opened 2022-08-15
vl: Initialize usb daughterboard at machine start

antangelo:
Creating the USB daughterboard when the UI initializes input prevents `-device` command line arguments from working correctly (as the daughterboard does not exist when they are parsed, they will not succeed).

Moving this into `vl.c` creates the device on machine start, so the ports will be available for other `-device` commands to use immediately.

### DIFF
--- a/softmmu/vl.c
+++ b/softmmu/vl.c
@@ -2966,6 +2966,10 @@ void qemu_init(int argc, char **argv, char **envp)
     fake_argv[fake_argc++] = strdup("-display");
     fake_argv[fake_argc++] = strdup("xemu");
 
+    // Create USB Daughterboard for 1.0 Xbox. This is connected to Port 1 of the Root hub.
+    fake_argv[fake_argc++] = strdup("-device");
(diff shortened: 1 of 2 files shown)


HadetTheUndying:
I have built this PR locally and tested the following devices

MadCatz Racing Wheel
Megajockey-9000 (Steel Battalion Controller)
Generic DDR Dance Mats
Duke Controller
Controller S
Lightgun ( it passes through, don't think it'll work on my screen )

EDIT: All of these devices passed through and worked, The Megajockey had some interesting input lag/delay  problems, and I was not able to get in-game in Steel Battalion because of some rendering asserts that I'll open issues for if this PR is merged. Line of Contact was having some interesting networking issues on Kai, so i was unable to get in-game as well, I will try that again later but the controller works as expected. 

The emulator will occasionally spit out this assert:
```
xemu: ../hw/usb/core.c:520: usb_cancel_packet: Assertion `usb_packet_is_inflight(p)' failed.
Aborted
```

PR #1272, opened 2022-10-03
ui: Display MCPX, BIOS md5 hashes in About view

antangelo:
Displays the hashes of the MCPX and BIOS ROM files (if available) in the about view. Should remove the need for users to determine this hash themselves in the event that they are debugging mismatched or corrupted ROM files.

### DIFF
--- a/ui/xui/main-menu.cc
+++ b/ui/xui/main-menu.cc
@@ -782,10 +782,12 @@ void MainMenuSystemView::Draw()
     if (FilePicker("MCPX Boot ROM", &g_config.sys.files.bootrom_path,
                    rom_file_filters)) {
         m_dirty = true;
+        g_main_menu.UpdateAboutViewConfigInfo();
     }
     if (FilePicker("Flash ROM (BIOS)", &g_config.sys.files.flashrom_path,
                    rom_file_filters)) {
         m_dirty = true;
+        g_main_menu.UpdateAboutViewConfigInfo();
     }
     if (FilePicker("Hard Disk", &g_config.sys.files.hdd_path,
                    qcow_file_filters)) {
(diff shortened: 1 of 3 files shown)


ui/xui/main-menu.cc
```diff
@@ -797,6 +797,25 @@ void MainMenuSystemView::Draw()
...
+static gchar *get_file_checksum(const char *path)
+{
+    auto *checksum = g_checksum_new(G_CHECKSUM_MD5);
+
+    auto *file = qemu_fopen(path, "rb");
+    if (!file) return nullptr;
+
+    guchar buf[512];
+    size_t nread;
+    while ((nread = fread(buf, sizeof(buf), 1, file))) {
```
mborgerson: This can fail for files with size not a multiple of 512 bytes (unusual, but possible), but this can be made to support arbitrary file sizes by swapping the second and third parameters and dropping the `* sizeof(buf)` factor below

ui/xui/main-menu.cc
```diff
@@ -819,6 +838,26 @@ void MainMenuAboutView::Draw()
...
+            bootrom_checksum = g_strdup("None");
+        }
+
+        gchar *flash_rom_checksum = get_file_checksum(g_config.sys.files.flashrom_path);
+        if (!flash_rom_checksum) {
+            flash_rom_checksum = g_strdup("None");
+        }
+
+        config_info_text = g_strdup_printf(
+                "MCPX Bootrom md5 Hash:      %s\n"
```
mborgerson: * Let's call it 'MCPX Boot ROM', to be consistent with the System configuration tab
* Let's capitalize 'MD5' here and elsewhere

ui/xui/main-menu.cc
```diff
@@ -797,6 +797,25 @@ void MainMenuSystemView::Draw()
     }
 }
 
+static gchar *get_file_checksum(const char *path)
```
mborgerson: * This function probably belongs somewhere else. Can you find a better spot for it?
* nit: Preferably the function name would mention MD5 specifically so the caller doesn't have to read the code

ui/xui/main-menu.cc
```diff
@@ -819,6 +838,26 @@ void MainMenuAboutView::Draw()
              gl_renderer, gl_version, gl_shader_version);
     }
 
+    static const char *config_info_text = NULL;
+    if (config_info_text == NULL) {
+        gchar *bootrom_checksum = get_file_checksum(g_config.sys.files.bootrom_path);
```
mborgerson: If someone changes the file after starting xemu, the hash will be different. What do you think about calculating the hashes on startup, when we load the file?

ui/xui/main-menu.cc
antangelo: Fixed

ui/xui/main-menu.cc

ui/xui/main-menu.cc
antangelo: Moved into `misc.hh`. I couldn't think of a better spot otherwise, since most utility source files are OS/platform specific.

ui/xui/main-menu.cc
antangelo: I added an update function that is called when either file is changed (in the system settings), and recomputes the hashes.

PR #1277, opened 2022-10-13
build: optimize for x86-64-v3 rather than ivybridge

LukeUsher:
This optimizes for a generic x64 target with `AVX, AVX2, BMI1, BMI2, F16C, FMA, LZCNT, MOVBE, XSAVE x86-64-v4: AVX512F, AVX512BW, AVX512CD, AVX512DQ, AVX512VL` rather than Ivy Bridge specifically

This should prevent Intel specific optimisations that harm AMD processors, while still taking advantage of newer instruction sets.

Note that this does require at least GCC11 or CLANG12, so it may not build with older toolchains.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -231,7 +231,7 @@ case "$platform" in # Adjust compilation options based on platform
         export LDFLAGS="-arch ${target_arch} \
                         -isysroot ${sdk}"
         if [ "$target_arch" == "x86_64" ]; then
-            sys_cflags='-march=ivybridge'
+            sys_cflags='-march=x86-64-v3'
         fi
         sys_ldflags='-headerpad_max_install_names'
         export PKG_CONFIG_PATH="${lib_prefix}/lib/pkgconfig"


GXTX:
~~I do not have a CPU which supports AVX2 (an Ivy Bridge Xeon), can we guarantee that AVX2 is not being used in places with this `march`? Or should we not consider 'older' users at this point?~~

Apple only build flag, doesn't effect me, ignore.

LukeUsher:
Crap I missed that, it shouldn't be an apple only flag; it's generic to GCC >= 11 and clang >= 12, I misread the build script.

If, however, x86-64-v3 is an issue due to using AVX2, perhaps we can use x86-64-v2 instead?

GXTX:
> If, however, x86-64-v3 is an issue due to using AVX2, perhaps we can use x86-64-v2 instead?

AVX is *probably* useful and is supported on Sandy Bridge Intel and above as well as Bulldozer AMD (about the same release date as Sandy Bridge) chips, so perhaps v2 + AVX if there's such a flag for specific extensions? I think it would complicate things too much to have a non-AVX and a AVX, or generic 'newer' procs build tag for *little* benefit.

LukeUsher:
Hmm, I'll close this PR for now; the current option may be the best trade-off for compatibility and performance, despite having intel-specific optimizations. 

I don't think it's worth cutting off older users for this, when it currently runs fine for them.

MasonT8198:
> Hmm, I'll close this PR for now; the current option may be the best trade-off for compatibility and performance, despite having intel-specific optimizations. 
> 
> 
> 
> I don't think it's worth cutting off older users for this, when it currently runs fine for them.

You could make a mention of this on the performance considerations page of the site, may prove useful to someone who wants to try to augment this themselves

Spidy123222:
> > If, however, x86-64-v3 is an issue due to using AVX2, perhaps we can use x86-64-v2 instead?
> 
> AVX is _probably_ useful and is supported on Sandy Bridge Intel and above as well as Bulldozer AMD (about the same release date as Sandy Bridge) chips, so perhaps v2 + AVX if there's such a flag for specific extensions? I think it would complicate things too much to have a non-AVX and a AVX, or generic 'newer' procs build tag for _little_ benefit.

I currently have a bulldozer it runs slowly but runs fine on some things. I researched flags and there does seem to be one by adding -mavx adds the avx specific extention. But why not do the opposite having x86-64-v3 or v4 and adding SSE3/4. PCSX2 currently has 2 separated builds for SSE4 and AVX2 in their develop page on their site so maybe it could be done similar here. I don’t think would know performance difference until tried on both intel and amd by comparing the differences with or without.

https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html

PR #1301, opened 2022-12-01
Implement keyboard rebinding support

### LINKED ISSUE #136: Implement gamepad button/axis remapping
hi

maybe can add option for change keyboard buttons

thanks

Fabxx:
Special thanks to:

toxicshadow, dracc and antangelo for suggestions.
SSUPI for correcting some errors in the SDL_Event handling.

Partially fixes #136  for keyboard rebinding. Controller rebinding will come in another PR

-Allows you to rebind keyboard controls

-Controls are stored in the .toml file configuration for next boots (no need to remap)

-Checks if a key has already been bound to any previous key to avoid conflicts.

-Uses same window for the controller, no extra UI windows.

-A white text under the toggles  tells to the user what key for xemu he is binding for each key.

-Button press,  rumble, analog movement and trigger animations during remapping are disabled to avoid confusing visual conflicts with already mapped keys before changing them.

-When remapping, a "Abort rebinding" toggle will appear, so the user can manually stop the binding process. This will not alter the already done bindings just in case the user presses the toggle by mistake.

-The auto-bind option has been removed, instead there's a "Reset controls to default" option, so the user can reset his controls while running xemu, and doesn't need to reboot to get the controls updated at runtime. The .toml file is updated on reboot with defaults if a remapping is not done after this option.

-While remapping, the "Reset controls to default" toggle will disappear.

-As soon the user by mistake changes window or closes the menu, the remapping process is aborted and we restore the default keys, since SDL still continues to catch events if outside of that window, meaning that the user will involountary map the keys even if it doesn't want to.

-The ESC button will NOT make the user exit the window while remapping. The button can be mapped as well.

Showcase (updated at 6/12/2022 14:56pm):

https://user-images.githubusercontent.com/30447649/205931385-f2626989-a80e-4cf4-bfbf-b7bb9babb23e.mp4

### DIFF
--- a/ui/xemu-input.c
+++ b/ui/xemu-input.c
@@ -422,9 +422,6 @@ void xemu_input_keyboard_rebind(const SDL_Event *ev)
 {   
     //Check if the user aborts the remapping process, if so, reset the default mapping. and save the defaults.
     if (abort_rebinding) {
-        char *buf = g_strdup_printf("WARNING: Rebinding process aborted. Restoring default bindings...");
-        xemu_queue_notification(buf);
-        free(buf);
         xemu_input_restore_defaults();
         is_remapping_active = false;
         abort_rebinding = false;
--- a/ui/xemu-input.c
+++ b/ui/xemu-input.c
@@ -429,9 +429,8 @@ void xemu_input_keyboard_rebind(const SDL_Event *ev)
         abort_rebinding = false;
     }
     
-    if(ev->type == SDL_KEYDOWN){
+    if (ev->type == SDL_KEYDOWN) {
         sdl_kbd_scancode_map[currently_remapping] = ev->key.keysym.scancode;
-
         //check for duplicated keybindings, if found, rebind that button.
         for (size_t i = 0; i < currently_remapping; i++) {
             if (sdl_kbd_scancode_map[currently_remapping] == sdl_kbd_scancode_map[i]) {
@@ -444,18 +443,18 @@ void xemu_input_keyboard_rebind(const SDL_Event *ev)
             }
         }
             
-        if( (sdl_kbd_scancode_map[currently_remapping] < SDL_SCANCODE_UNKNOWN) || 
-            (sdl_kbd_scancode_map[currently_remapping] >= SDL_NUM_SCANCODES) ) {
+        if ((sdl_kbd_scancode_map[currently_remapping] < SDL_SCANCODE_UNKNOWN) || 
+            (sdl_kbd_scancode_map[currently_remapping] >= SDL_NUM_SCANCODES)) {
             char *buf = g_strdup_printf("WARNING: Keyboard scancode out of range, try another key.");
             xemu_queue_notification(buf);
             free(buf);
             currently_remapping--;
-    }            
+        }            
+        
         currently_remapping++;
 
         //If the user has mapped all the buttons, store the new bindings.
-
-        if(currently_remapping == 25) {
+        if (currently_remapping == 25) {
             g_config.input.keyboard_controller_scancode_map.a            = sdl_kbd_scancode_map[0];
             g_config.input.keyboard_controller_scancode_map.b            = sdl_kbd_scancode_map[1];
             g_config.input.keyboard_controller_scancode_map.x            = sdl_kbd_scancode_map[2]; 
--- a/ui/xemu-input.c
+++ b/ui/xemu-input.c
@@ -455,7 +455,7 @@ void xemu_input_keyboard_rebind(const SDL_Event *ev)
 
         //If the user has mapped all the buttons, store the new bindings.
 
-        if(currently_remapping == 25){
+        if(currently_remapping == 25) {
             g_config.input.keyboard_controller_scancode_map.a            = sdl_kbd_scancode_map[0];
             g_config.input.keyboard_controller_scancode_map.b            = sdl_kbd_scancode_map[1];
             g_config.input.keyboard_controller_scancode_map.x            = sdl_kbd_scancode_map[2]; 
--- a/ui/xemu-input.c
+++ b/ui/xemu-input.c
@@ -369,9 +369,8 @@ void xemu_input_keyboard_rebind(const SDL_Event *ev, ControllerState *state)
             {
                 if (sdl_kbd_scancode_map[currently_remapping] == sdl_kbd_scancode_map[i])
                 {
-                    char buf[40];
-                    snprintf(buf, sizeof(buf), "WARNING: Keybind already in use, try another key.");
-                    xemu_queue_notification(buf);
+                    duplicate_found = true;
+                    already_mapped = i;
                     currently_remapping--;
                     break;
                 }
--- a/ui/xemu-input.c
+++ b/ui/xemu-input.c
@@ -298,9 +298,11 @@ void xemu_input_process_sdl_events(const SDL_Event *event)
         }
     } else if (event->type == SDL_CONTROLLERDEVICEREMAPPED) {
         DPRINTF("Controller Remapped: %d\n", event->cdevice.which);
-    } else if (is_remapping_active && event->type == SDL_KEYDOWN) {
+    } else if (is_remapping_active) {
         xemu_input_keyboard_rebind(event);
-    }  
+    } else if (restore_controls) {
+        xemu_input_restore_defaults();
+    }
 }
 
 void xemu_input_update_controller(ControllerState *state)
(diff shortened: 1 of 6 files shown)


ui/xemu-input.c
```diff
@@ -353,6 +356,65 @@ void xemu_input_update_sdl_kbd_controller_state(ControllerState *state)
...
+void xemu_input_keyboard_rebind(const SDL_Event *ev, ControllerState *state)
+{
+    sdl_kbd_scancode_map[currently_remapping] = SDL_SCANCODE_UNKNOWN;
+    
+        if(ev->type == SDL_KEYDOWN){
+            sdl_kbd_scancode_map[currently_remapping] = ev->key.keysym.scancode;
+
+            //check for duplicated keybindings, if found, rebind that button.
+            for (size_t i = 0; i < currently_remapping; i++)
+            {
```
antangelo: nit: Formatting should be consistent, with braces on the same line except for function definitions. This appears a few times throughout the diff.

ui/xui/gl-helpers.cc
```diff
@@ -519,7 +520,9 @@ void RenderController(float frame_x, float frame_y, uint32_t primary_color,
         frame_x += ((float)(rand() % 5)-2.5) * (1-t);
         frame_y += ((float)(rand() % 5)-2.5) * (1-t);
         rumble_l = rumble_r = sin_wav;
+        }
```
antangelo, line 523: nit: Code above this end brace should be indented to match the rest of the added `if` block.

ui/xui/gl-helpers.cc
```diff
@@ -534,14 +537,19 @@ void RenderController(float frame_x, float frame_y, uint32_t primary_color,
...
+        for (int i = 0; i < 12; i++) {
+            if (state->buttons & (1 << i)) {
+                RenderDecal(g_decal_shader, frame_x + buttons[i].x,
+                            frame_y + buttons[i].y, buttons[i].w, buttons[i].h, 0,
+                            0, 1, 1, 0, 0, primary_color + 0xff);
+            }
         }
     }
 
+
```
antangelo: nit: Extra newline (in a few places).

ui/xui/main-menu.cc
```diff
@@ -263,6 +268,47 @@ void MainMenuInputView::Draw()
...
+        }
+
+        if (g_config.input.auto_bind)
+        {
+            is_auto_binding_active = true;
+            is_remapping_active = false;
+        }
+    }
+
+    if (is_auto_binding_active)
```
antangelo: This variable shouldn't need to exist, since it's always equal to `g_config.input.auto_bind`

ui/xui/main-menu.cc
```diff
@@ -263,6 +268,47 @@ void MainMenuInputView::Draw()
...
+
+    if (is_remapping_active && !g_config.input.auto_bind)
+    {
+        ImGui::Text("Press the key you want to bind for: %s", bindings[currently_remapping]);
+    }
+
+    if (duplicate_found)
+    {
+        char buf[40];
+        snprintf(buf, sizeof(buf), "WARNING: Keybind already in use for: %s\n, try another key.", bindings[already_mapped]);
```
antangelo: This buffer is not big enough to hold that string. Recommend switching to `g_strdup_printf`.

ui/xui/main-menu.cc
```diff
@@ -263,6 +268,47 @@ void MainMenuInputView::Draw()
     Toggle("Background controller input capture",
            &g_config.input.background_input_capture,
            "Capture even if window is unfocused (requires restart)");
+
+    //Interface and checks for keyboard remapping
+    
+    if (ImGui::IsItemClicked(ImGui::Button("Rebind Controls")))
```
antangelo: You should only need `if (ImGui::Button())` here, no need for `IsItemClicked`. Also, I think this should be moved up into the controller section, and only be visible if the keyboard is bound.

Same nit as above with brace placement.

ui/xemu-input.c
```diff
@@ -353,6 +356,65 @@ void xemu_input_update_sdl_kbd_controller_state(ControllerState *state)
     if (kbd[sdl_kbd_scancode_map[24]]) state->axis[CONTROLLER_AXIS_RTRIG] = 32767;
 }
 
+void xemu_input_keyboard_rebind(const SDL_Event *ev, ControllerState *state)
```
antangelo: The `state` arg seems to be unused and should be removed.

ui/xemu-input.c
```diff
@@ -298,7 +298,10 @@ void xemu_input_process_sdl_events(const SDL_Event *event)
         }
     } else if (event->type == SDL_CONTROLLERDEVICEREMAPPED) {
         DPRINTF("Controller Remapped: %d\n", event->cdevice.which);
-    }
+    } else if (is_remapping_active){
```
antangelo: Should check if the key event is a `SDL_KEYDOWN` event here, instead of within the rebind function, to prevent it from eating any unexpected events.

ui/xemu-input.c
```diff
@@ -353,6 +356,65 @@ void xemu_input_update_sdl_kbd_controller_state(ControllerState *state)
...
+                    duplicate_found = true;
+                    already_mapped = i;
+                    currently_remapping--;
+                    break;
+                }
+            }
+            
+            if( (sdl_kbd_scancode_map[currently_remapping] < SDL_SCANCODE_UNKNOWN) || 
+                (sdl_kbd_scancode_map[currently_remapping] >= SDL_NUM_SCANCODES) ) {
+                fprintf(stderr, "WARNING: Keyboard controller map scancode out of range (%d) : Disabled\n", sdl_kbd_scancode_map[currently_remapping]);
```
antangelo: Most people are not likely to see this error message if printed to stderr (as an example, Windows users would only notice it if they checked the log file, and you might not see it at all if you aren't running xemu from the terminal on other operating systems). I would just replace it with an assert.

antangelo (commented):
Left a few comments. Another thing I noticed while using it:

* While remapping keys, it's possible for the keyboard inputs to be used to navigate the UI. I think this should be disabled while remapping.
* The need to disable auto-bind is rather unintuitive. If there is some bug where the config is overwritten by auto-bind being on, then I think it should be fixed instead of requiring it to be disabled.

ui/xemu-input.c
Fabxx: Will correct the indentation, as for the auto-bind it's not a bug but it's actually how it works by xemu itself, if you use auto-binding after mapping your keyboard and you reboot xemu without toggling the setting, you will overwrite your previous configuration

ui/xemu-input.c
Fabxx: yeah it was a debug thing that i forgot to make visible to the user in case of invalid scancode. If it's fine i qill apply the modifications tomorrow by showing this error via xemu notifications

ui/xemu-input.c
Fabxx: Wouldn't this conflict with all types of events by using SDL_KEYDOWN instead of expecting if the rebinding has been enabled, and only in that case capture the events from the keyboard?

ui/xemu-input.c
Fabxx: It was planned to add controller remapping as well, but yeah in this PR it will be removed

Fabxx:
The modifications are tested and complete, feel free to test it and report if you agree with the choices applied to the input management and code design

ElTioRata:
Does this includes rebinding buttons to mouse?

Fabxx:
> Does this includes rebinding buttons to mouse?

No, of course it can be implemented by just catching the SDL event for the mouse press and bind it to that scancode, but the right mouse button is used to access the UI. Besides i think it's more proper to implement the mouse binding only on analog axis movement since that's the most useful thing when playing with both keyboard and mouse. If the devs agree with this thinkering i might implement it in this PR already. UPDATE: If we want to make sure that the user can use the ESC Button and the mouse buttons/movement we should:

-Disable the esc button for the UI permanently, i already have an idea for it.
-Disable the right click mouse button for the UI permanently

This is needed if we have to guarantee the usage of all keys for the user while playing and not having interferences with the UI, not to mention if i add mouse remapping as well that could be very useful. 

I can already do the code but i need a response to understand how i have to move. So guarantee freedom with keyboard + mouse but disable button UI interaction, and just let the user move the mouse/use the controller to access the UI, or it stays like so.

UPDATE: I guess i'll leave it as it is to avoid a mess. The code works as intended for the keyboard rebinding, mouse/controller might come in another PR later.

ElTioRata:
imagen
Is this insanely high disk usage supposed to be normal?

ElTioRata:
Tried the build for a few weeks, works fine. However, I think it would be a lot more convenient if you could remap specific buttons separately instead of going through the whole remap process. It would also be nice if you could add key mapping profiles as well, since not all games are comfortable to play with the same keys.

Fabxx:
> Tried the build for a few weeks, works fine. However, I think it would be a lot more convenient if you could remap specific buttons separately instead of going through the whole remap process. It would also be nice if you could add key mapping profiles as well, since not all games are comfortable to play with the same keys.

For the profiles might be doable, for the specific buttons i have to see

Fabxx:
The system works as intended, but in the future it will be made more flexible, but for now it's possible to simply bind all the keys to the combination wanted.

ui/xemu-input.c
```diff
@@ -298,6 +298,10 @@ void xemu_input_process_sdl_events(const SDL_Event *event)
         }
     } else if (event->type == SDL_CONTROLLERDEVICEREMAPPED) {
         DPRINTF("Controller Remapped: %d\n", event->cdevice.which);
+    } else if (is_remapping_active) {
+        xemu_input_keyboard_rebind(event);
+    } else if (restore_controls) {
+        xemu_input_restore_defaults();
```
antangelo, line 304: Why is this done here instead of being called directly when the restore button is pressed?

ui/xemu-input.c
```diff
@@ -353,6 +357,138 @@ void xemu_input_update_sdl_kbd_controller_state(ControllerState *state)
...
+    char *buf = g_strdup_printf("INFO: Keys restored to default.");
+    xemu_queue_notification(buf);
+    free(buf);
+
+    restore_controls = false;
+}
+
+void xemu_input_keyboard_rebind(const SDL_Event *ev)
+{   
+    //Check if the user aborts the remapping process.
```
antangelo, line 423: nit: This comment adds no value

ui/xui/gl-helpers.cc
```diff
@@ -501,25 +501,28 @@ void RenderController(float frame_x, float frame_y, uint32_t primary_color,
...
-        frame_x += ((float)(rand() % 5)-2.5) * (1-t);
-        frame_y += ((float)(rand() % 5)-2.5) * (1-t);
-        rumble_l = rumble_r = sin_wav;
+            // Add a little extra flare: wiggle the frame around while we rumble
+            frame_x += ((float)(rand() % 5)-2.5) * (1-t);
+            frame_y += ((float)(rand() % 5)-2.5) * (1-t);
+            rumble_l = rumble_r = sin_wav; 
+        }
     }
+
```
antangelo, line 525: nit: Extra newline

ui/xui/gl-helpers.cc
```diff
@@ -534,51 +537,70 @@ void RenderController(float frame_x, float frame_y, uint32_t primary_color,
...
+    }
+
     RenderDecal(g_decal_shader, (int)(c_x - w / 2.0f + 10.0f * rstick_x),
                 (int)(c_y - h / 2.0f + 10.0f * rstick_y), w, h,
                 tex_items[obj_rstick].x, tex_items[obj_rstick].y, w, h,
-                (state->buttons & CONTROLLER_BUTTON_RSTICK) ? secondary_color :
+                (state->buttons & rstick_press_anim) ? secondary_color :
                                                               primary_color,
-                (state->buttons & CONTROLLER_BUTTON_RSTICK) ? primary_color :
+                (state->buttons & rstick_press_anim) ? primary_color :
```
antangelo, line 597: Same as above

ui/xui/gl-helpers.cc
```diff
@@ -534,51 +537,70 @@ void RenderController(float frame_x, float frame_y, uint32_t primary_color,
...
 
-    // Render left thumbstick
+    // Render left thumbstick, do not make move/highlight while remapping.
     float w = tex_items[obj_lstick].w;
     float h = tex_items[obj_lstick].h;
     float c_x = frame_x+lstick_ctr.x;
     float c_y = frame_y+lstick_ctr.y;
     float lstick_x = (float)state->axis[CONTROLLER_AXIS_LSTICK_X]/32768.0;
     float lstick_y = (float)state->axis[CONTROLLER_AXIS_LSTICK_Y]/32768.0;
+    int lstick_press_anim = CONTROLLER_BUTTON_LSTICK;
```
antangelo, line 560: Should this take the remapping into account?

ui/xui/gl-helpers.cc
```diff
@@ -534,51 +537,70 @@ void RenderController(float frame_x, float frame_y, uint32_t primary_color,
...
+    }
+
     RenderDecal(g_decal_shader, (int)(c_x - w / 2.0f + 10.0f * lstick_x),
                 (int)(c_y - h / 2.0f + 10.0f * lstick_y), w, h,
                 tex_items[obj_lstick].x, tex_items[obj_lstick].y, w, h,
-                (state->buttons & CONTROLLER_BUTTON_LSTICK) ? secondary_color :
+                (state->buttons & lstick_press_anim) ? secondary_color :
                                                               primary_color,
-                (state->buttons & CONTROLLER_BUTTON_LSTICK) ? primary_color :
+                (state->buttons & lstick_press_anim) ? primary_color :
```
antangelo, line 573: Instead of using `SDL_SCANCODE_UNKNOWN` it'd be clearer to condition these on `!is_remapping_active && (state->buttons & ...)`

ui/xui/main-menu.cc
```diff
@@ -955,6 +997,7 @@ MainMenuScene::MainMenuScene()
 
     m_current_view_index = 0;
     m_next_view_index = m_current_view_index;
+
```
antangelo, line 1000: nit: Extra newline

ui/xui/main-menu.cc
```diff
@@ -258,11 +268,43 @@ void MainMenuInputView::Draw()
...
+            restore_controls = true;
+        }
+    }
+
+    if (is_remapping_active) {
+        if (Toggle("Abort rebinding", &abort_rebinding, "Abort the rebinding process.")) {
+            abort_rebinding = true;
+        }
+    ImGui::SetKeyboardFocusHere(1);
+    ImGui::Text("\nPress the key you want to bind for: %s", bindings[currently_remapping]);
```
antangelo, line 306: nit: Incorrect indentation

ui/xui/main-menu.cc
```diff
@@ -258,11 +268,43 @@ void MainMenuInputView::Draw()
...
+    
+    if (!is_remapping_active) {
+        if (Toggle("Reset controls to default", &restore_controls,
+                   "Resets the keyboard mapping to default. (No reboot required)")) {
+            restore_controls = true;
+        }
+    }
+
+    if (is_remapping_active) {
+        if (Toggle("Abort rebinding", &abort_rebinding, "Abort the rebinding process.")) {
```
antangelo, line 302: Why is this a toggle? Seems more suited for a button.

ui/xemu-input.c
```diff
@@ -353,6 +357,138 @@ void xemu_input_update_sdl_kbd_controller_state(ControllerState *state)
...
+        is_remapping_active = false;
+        abort_rebinding = false;
+    }
+    
+    if (ev->type == SDL_KEYDOWN) {
+        sdl_kbd_scancode_map[currently_remapping] = ev->key.keysym.scancode;
+        //check for duplicated keybindings, if found, rebind that button.
+        for (size_t i = 0; i < currently_remapping; i++) {
+            if (sdl_kbd_scancode_map[currently_remapping] == sdl_kbd_scancode_map[i]) {
+                already_mapped = i;
```
antangelo, line 437: This variable isn't used anywhere else, seems like it can be removed.

ui/xui/main-menu.cc
```diff
@@ -38,6 +38,16 @@
 #include "../xemu-xbe.h"
 
 MainMenuScene g_main_menu;
+bool is_remapping_active = false;
+bool duplicate_found = false;
```
antangelo, line 42: This variable is unused

ui/xui/main-menu.cc
```diff
@@ -258,11 +268,43 @@ void MainMenuInputView::Draw()
...
+        is_remapping_active = true;
+    } 
+
+    if (g_config.general.last_viewed_menu_index != 1 && is_remapping_active) {
+        abort_rebinding = true;
+        restore_controls = true;
+    }
+    
+    if (!is_remapping_active) {
+        if (Toggle("Reset controls to default", &restore_controls,
```
antangelo, line 295: Same comment about being a toggle as below

ui/xui/main-menu.cc
```diff
@@ -38,6 +38,16 @@
 #include "../xemu-xbe.h"
 
 MainMenuScene g_main_menu;
+bool is_remapping_active = false;
```
antangelo, line 41: Globals should be prefixed with `g_`

ui/xemu-input.c
```diff
@@ -353,6 +357,138 @@ void xemu_input_update_sdl_kbd_controller_state(ControllerState *state)
...
+    sdl_kbd_scancode_map[17]           = SDL_SCANCODE_F; 
+    sdl_kbd_scancode_map[18]           = SDL_SCANCODE_D; 
+    sdl_kbd_scancode_map[19]           = SDL_SCANCODE_W;
+    sdl_kbd_scancode_map[20]           = SDL_SCANCODE_I;
+    sdl_kbd_scancode_map[21]           = SDL_SCANCODE_J;
+    sdl_kbd_scancode_map[22]           = SDL_SCANCODE_L;
+    sdl_kbd_scancode_map[23]           = SDL_SCANCODE_K;
+    sdl_kbd_scancode_map[24]           = SDL_SCANCODE_O;
+    
+    g_config.input.keyboard_controller_scancode_map.a            = SDL_SCANCODE_A;
```
antangelo, line 388: Create a function that takes `sdl_kbd_scancode_map` as a parameter and sets the `g_config` values like you do when binding is complete, then reuse that here to clean this up.

ui/xui/main-menu.cc
```diff
@@ -38,6 +38,16 @@
...
+bool duplicate_found = false;
+bool restore_controls = false;
+bool abort_rebinding = false;
```
antangelo, line 44: You can detect the abort case if you keep a static variable `in_progress` in `xemu_input_keyboard_rebind` that tracks this:

- Initialize to false
- If `is_remapping_active` is true, and `in_progress` is false, set it to true
- If `is_remapping_active` is false, and `in_progress` is true, this is the abort case. Only change you have to add is to set `in_progress` to false after handling
- If both are false do nothing, same as now.
- When finished remapping, set `in_progress` to false.

You can rename it to something better if you want, just a suggestion. This removes the need for an extra global.

Fabxx:
closing this, please follow #1516 since it's in a better state.

tigerbugyou:
How do I remap the keyboard?

abaire:
> How do I remap the keyboard?

Please check the docs before pinging old closed issues: https://xemu.app/docs/controller/#remapping-keyboard-buttons-advanced

PR #1304, opened 2022-12-15
UX: Temporarily disable DSP toggle

Fabxx:
A lot of times the user reports crashes regarding xemu, mostly because it enables the DSP option which is not implemented yet. To avoid this i've temporarily disabled the DSP toggle by commenting the code, and resetting the `use_dsp` flag to false just in case in a previous boot it was activated.

DSP


mborgerson:
Thanks for the PR, but I do not plan to merge this. I want the option to remain available.

allkern:
Uh, DSP processing *is* implemented

PR #1307, opened 2022-12-19
UI: Hide mouse cursor option

### LINKED ISSUE #825: Hide Mouse Cursor
### Feature Request

Is it possible to hide the mouse cursor?

Thanks

Fabxx:
Fixes #825 

Hides the cursor when option is enabled and outside of UI:

https://user-images.githubusercontent.com/30447649/212203521-c04e0454-9d48-469f-bcea-58c00f5d53f0.mp4

### DIFF
--- a/config_spec.yml
+++ b/config_spec.yml
@@ -136,6 +136,9 @@ display:
     show_notifications:
       type: bool
       default: true
+    hide_cursor:
+      type: bool
+      default: true
     use_animations:
       type: bool
       default: true
--- a/ui/xui/main-menu.cc
+++ b/ui/xui/main-menu.cc
@@ -316,6 +316,8 @@ void MainMenuDisplayView::Draw()
            "Show main menu bar when mouse is activated");
     Toggle("Show notifications", &g_config.display.ui.show_notifications,
            "Display notifications in upper-right corner");
+    Toggle("Hide mouse cursor", &g_config.display.ui.hide_cursor,
+           "Hide the mouse cursor when it is not moving");
 
     int ui_scale_idx;
     if (g_config.display.ui.auto_scale) {
--- a/ui/xui/main.cc
+++ b/ui/xui/main.cc
@@ -256,6 +256,16 @@ void xemu_hud_render(void)
         }
     }
 
+    static uint32_t last_mouse_move = 0;
+    if (g_input_mgr.MouseMoved()) {
+        last_mouse_move = now;
+    }
+
+    // FIXME: Handle time wrap around
+    if (g_config.display.ui.hide_cursor && (now - last_mouse_move) > 3000) {
+        ImGui::SetMouseCursor(ImGuiMouseCursor_None);
+    }
+
     if (!ImGui::IsWindowFocused(ImGuiFocusedFlags_AnyWindow) &&
         !g_scene_mgr.IsDisplayingScene()) {
 


antangelo (approved):

antoniodesousa:
I hope this gets merged soon...

config_spec.yml
```diff
@@ -133,6 +133,9 @@ display:
     show_menubar:
       type: bool
       default: true
+    hide_cursor:
+      type: bool
+      default: true
```
mborgerson, line 141: Should this be enabled by default?

ui/xui/main-menu.cc
```diff
@@ -314,6 +314,9 @@ void MainMenuDisplayView::Draw()
     SectionTitle("Interface");
     Toggle("Show main menu bar", &g_config.display.ui.show_menubar,
            "Show main menu bar when mouse is activated");
+    Toggle("Hide mouse cursor", &g_config.display.ui.hide_cursor,
+            "Hide the cursor when menubar disappears and outside UI.");
```
mborgerson: Other settings captions don't normally end with a period

ui/xui/main.cc
```diff
@@ -252,6 +252,9 @@ void xemu_hud_render(void)
             ShowMainMenu();
             ImGui::PopStyleColor();
         } else {
+            if (g_config.display.ui.hide_cursor && !g_scene_mgr.IsDisplayingScene()) {
```
mborgerson: Because of where this is placed, this hides the cursor only if the menu bar is shown. What if the menu bar is configured to be hidden also? This code will never trigger and the mouse cursor will not be hidden in this case.

mborgerson (commented):
I like the feature, but it needs some minor changes before being merged

config_spec.yml
antangelo: I think having it enabled by default is fair behavior, it mimics the behavior of youtube (which hides the cursor over the video player) and games that don't involve the cursor and hide it during gameplay. I'm not sure what other emulators do, however.

antoniodesousa: Ryujinx has multiple settings for the cursor:
- Always visible
- Hide on idle (after 5 sec or so)
- Always hidden

Personally, I really like the hide on idle option.

Fabxx: Let me know when an agreement on this has been reached before i touch the code.

ui/xui/main.cc
Fabxx: if you agree i can add a simple check on the alpha state of the menu bar with an OR, so i can either check this, or if the bar has become invisible to hide the cursor, so if you check the hide bar option, it can hide the cursor based directly on that state, and it won't conflict with the menu open since then the menu state is active

config_spec.yml
mborgerson: > I think having it enabled by default is fair behavior, it mimics the behavior of youtube (which hides the cursor over the video player) and games that don't involve the cursor and hide it during gameplay.

Sounds reasonable. Let's have it hide on idle by default then.

ui/xui/main.cc
mborgerson: We don't need to consider the alpha state of the menu bar. Just hide the cursor if it is not being moved after some period of time.

Fabxx: I've reworked the code, but i still have to figure out how to add a small delay after the cursor stopped moving before hiding the cursor.

PR #1308, opened 2022-12-20
UI: Improve user experience when AV Pack is set to "None"

### LINKED ISSUE #1186: Improve user experience when AV Pack is set to "None"
### Feature Request

It is possible to set the AV Pack to "None", leading to a "The guest has not initialized the display" message with no indication as to why.

We should
1) Update the default display message to indicate that the user should change their AV Pack setting (it's only useful for testing behavior when there is no AV Pack i.e., if you're not a developer it's probably an error to select it)

2) Consider adding a warning in the settings page explaining that selecting "None" means that there will be no output.

3) Consider renaming "None" in the UI to "No video" so it's extremely clear that it's probably not what a typical user wants.

### Alternatives

_No response_

### Additional Context

This came up in a #help chat in Discord: https://discord.com/channels/680221390359887933/722155756622708744/997648642838118524

Fabxx:
Fixes #1186 

https://github.com/xemu-project/xemu/assets/30447649/df0475e2-a22b-4805-8ec4-069faabbd4fb


GXTX:
There's nothing inherently wrong with warning users about this, however I think a better approach would be to remove the option from the UI. Having no avpack connected is not beneficial to any end user and is only useful to a developer testing how the kernel and software react to that situation. People who need this functionality can already do this via command line launch options and via config.

halotroop2288:
> not beneficial to any end user

I counter that it would make blindfolded playthroughs more authentic! 😸
And editing a config by hand is just bad UX, I think this is better.

antangelo:
I would prefer that we keep the option in the UI, I agree that having it there is better UX than requiring a command line argument even for developer use. As an alternative, we can add a warning above or below the setting in the UI if `None` is selected that indicates there will not be any video output, which should be clear enough on that front (in addition to the changing of the uninitialized display text).

dracc:
> > not beneficial to any end user
> 
> I counter that it would make blindfolded playthroughs more authentic! 😸 And editing a config by hand is just bad UX, I think this is better.

Current implementation doesn't just disable video output, but emulates what happens if you haven't connected the AV Pack. What happens is the front LEDs flash orange and green and execution stops - not at all helpful for speedrunning.

ui/console.c
```diff
@@ -1560,13 +1560,14 @@ DisplaySurface *qemu_create_placeholder_surface(int w, int h,
     pixman_image_t *glyph;
     int len, x, y, i;
 
+    //Horizontal center x on 640x480 is 5.
```
antangelo: What's the relevance of this comment?

ui/console.c
```diff
@@ -1560,13 +1560,14 @@ DisplaySurface *qemu_create_placeholder_surface(int w, int h,
...
     len = strlen(msg);
-    x = (w / FONT_WIDTH  - len) / 2;
-    y = (h / FONT_HEIGHT - 1)   / 2;
+    x = (w / FONT_WIDTH  - 70) / 2;
```
antangelo: Instead of hard-coding the line length, we should use the computed length of the longest line (i.e. allow msg to contain newlines and handle them appropriately).

ui/console.c
```diff
@@ -1560,13 +1560,14 @@ DisplaySurface *qemu_create_placeholder_surface(int w, int h,
...
+    y = (h / FONT_HEIGHT - 1) / 2;
     for (i = 0; i < len; i++) {
-        glyph = qemu_pixman_glyph_from_vgafont(FONT_HEIGHT, vgafont16, msg[i]);
+        glyph  = qemu_pixman_glyph_from_vgafont(FONT_HEIGHT, vgafont16, msg[i]);
         qemu_pixman_glyph_render(glyph, surface->image, &fg, &bg,
-                                 x+i, y, FONT_WIDTH, FONT_HEIGHT);
+                                 x+i % 70, y + i / 70, FONT_WIDTH, FONT_HEIGHT);
```
antangelo: Similar to above, should update the logic to handle arbitrary newlines so that is more versatile.

mborgerson:
I agree a better default uninitialized screen would be nice, but I think we can do better than more console text here. How about just warning on the settings page when None is selected

halotroop2288:
Did you just revert the whole PR?

Fabxx:
Main code has been updated

ui/xui/main-menu.cc
```diff
@@ -989,6 +989,10 @@ void MainMenuSystemView::Draw()
         ImGui::TextColored(ImVec4(1,0,0,1), "Application restart required to apply settings");
     }
 
+    if ((int)g_config.sys.avpack == 6) {
```
antangelo: Can we use the enum value here instead of comparing as an integer?

Fabxx: > Can we use the enum value here instead of comparing as an integer?

done

Fabxx:
all the modifications were applied, should be ready to merge

mborgerson:
Thanks!

PR #1309, opened 2022-12-25
UX: Get windows product and build number properly

Fabxx:
instead of returning "windows", based on the kernel version, we obtain the product name and the build number displayed in `winver` command. 

For modern systems the check is made on "10.0" value according to the windows NT kernel version list.

Screenshot (via wine, in this case wine uses "Software" instead of "SOFTWARE"):
immagine

Wine windows XP
immagine

Wine windows Vista
immagine

Wine windows 7
immagine

Wine windows 8
immagine

Wine windows 8.1
immagine


ui/xemu-os-utils-windows.c
```diff
@@ -19,10 +19,42 @@
...
+
+                //If it's lower (win 8 and below until XP) get the build descriptor from CSDBuildNumber.
+                } else { 
+                    RegQueryValueExW(keyHandle, L"CSDVersion", NULL, NULL, (LPBYTE)windows_build, &string2);
+                }
+
+            RegQueryValueExW(keyHandle, L"ProductName", NULL, NULL, (LPBYTE)windows_name, &string3); 
+            RegCloseKey(keyHandle);
+        } else {
+            const char *error = "Couldn't access system information!";
```
GXTX: Fallback to just replying Windows.

ui/xemu-os-utils-windows.c
```diff
@@ -19,10 +19,42 @@
...
+                }
+
+            RegQueryValueExW(keyHandle, L"ProductName", NULL, NULL, (LPBYTE)windows_name, &string3); 
+            RegCloseKey(keyHandle);
+        } else {
+            const char *msg = "Windows";
+            return msg;
+        }
+    }
+    buffer = g_strdup_printf("%ls %ls", windows_name, windows_build);
```
mborgerson: * What if any of the `RegQueryValueExW` calls fail? The buffers will be initialized with garbage
* This line should be moved up. (`buffer` does not need to be re-initialized, and `windows_name` and `windows_build` may be uninitialized at that time leading to potential crash)

ui/xemu-os-utils-windows.c
```diff
@@ -19,10 +19,42 @@
...
 #include <windows.h>
+#include <glib.h>
+#include <glib/gprintf.h>
+#include <wchar.h>
 
 const char *xemu_get_os_info(void)
-{
-	return "Windows";
+{ 
+	HKEY keyHandle;
```
mborgerson: nit: Indentation

ui/xemu-os-utils-windows.c
```diff
@@ -19,10 +19,42 @@
...
+
+                //If it's lower (win 8 and below until XP) get the build descriptor from CSDVersion.
+                } else { 
+                    RegQueryValueExW(keyHandle, L"CSDVersion", NULL, NULL, (LPBYTE)windows_build, &string2);
+                }
+
+            RegQueryValueExW(keyHandle, L"ProductName", NULL, NULL, (LPBYTE)windows_name, &string3); 
+            RegCloseKey(keyHandle);
+        } else {
+            const char *msg = "Windows";
```
mborgerson: nit: For brevity, you may simply `return "Windows";`

ui/xemu-os-utils-windows.c
```diff
@@ -19,10 +19,42 @@
...
+    if (buffer == NULL) {
+        //Get Current Kernel version number.
+        if (RegOpenKeyExW(HKEY_LOCAL_MACHINE, L"SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion", 0,
+                          KEY_QUERY_VALUE, &keyHandle) == ERROR_SUCCESS) {
+            RegQueryValueExW(keyHandle, L"CurrentVersion", NULL, NULL, (LPBYTE)version_number, &string);
+            
+            /* if version number is 10.0 (win 8.1/10/11), get the build number from the DisplayVersion Registry.
+               Reference: https://en.wikipedia.org/wiki/Windows_NT */
+            if (wcscmp(version_number, L"10.0") == 0) {
+                RegQueryValueExW(keyHandle, L"DisplayVersion", NULL, NULL, (LPBYTE)windows_build, &string1);
```
mborgerson: The [MSDN page](https://learn.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regqueryvalueexw) for `RegQueryValueExW` recommends:

> To ensure that any string values (REG_SZ, REG_MULTI_SZ, and REG_EXPAND_SZ) returned are null-terminated, use the [RegGetValue](https://learn.microsoft.com/en-us/windows/desktop/api/winreg/nf-winreg-reggetvaluea) function.

Any reason not to use `RegGetValue` here?

ui/xemu-os-utils-windows.c
```diff
@@ -19,10 +19,42 @@
...
+#include <wchar.h>
 
 const char *xemu_get_os_info(void)
-{
-	return "Windows";
+{ 
+	HKEY keyHandle;
+    WCHAR windows_build[1024], windows_name[1024], version_number[1024];
+    static const char *buffer = NULL; 
+    DWORD string = 1024, string1 = 1024, string2 = 1024, string3 = 1024;
```
mborgerson: nit: `string{, 1, 2, 3}` could be better named; these are buffer sizes (in bytes) before call, and size of copied contents after call to `RegQueryValueExW`

ui/xemu-os-utils-windows.c
```diff
@@ -19,10 +19,42 @@
...
+
+    if (buffer == NULL) {
+        //Get Current Kernel version number.
+        if (RegOpenKeyExW(HKEY_LOCAL_MACHINE, L"SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion", 0,
+                          KEY_QUERY_VALUE, &keyHandle) == ERROR_SUCCESS) {
+            RegQueryValueExW(keyHandle, L"CurrentVersion", NULL, NULL, (LPBYTE)version_number, &string);
+            
+            /* if version number is 10.0 (win 8.1/10/11), get the build number from the DisplayVersion Registry.
```
mborgerson: According to the linked page Windows 8.1 has a version number of 6.3?

ui/xemu-os-utils-windows.c
```diff
@@ -19,10 +19,42 @@
...
+               Reference: https://en.wikipedia.org/wiki/Windows_NT */
+            if (wcscmp(version_number, L"10.0") == 0) {
+                RegQueryValueExW(keyHandle, L"DisplayVersion", NULL, NULL, (LPBYTE)windows_build, &string1);
+
+                //If it's lower (win 8 and below until XP) get the build descriptor from CSDVersion.
```
mborgerson: * What happens when the next version of Windows comes out and the number is >10.0?
* nit: Comment should be in associated block of code
* nit: Indentation

ui/xemu-os-utils-windows.c
Fabxx: No reason, but looking through the window API was a mess, seeing how often they changed functions and behaviors, that one seemed fine for this purpose. Since you have found a better function i'll use that instead.

ui/xemu-os-utils-windows.c
Fabxx: thought about that but i thinked it wouldn't look good, if you like that way i can do that.

ui/xemu-os-utils-windows.c
Fabxx: my bad, will move above in the NULL verification

ui/xemu-os-utils-windows.c
Fabxx: aside from better renaming, do i have to do anything else? I'm not understanding

ui/xemu-os-utils-windows.c
Fabxx: Depends on how MS will handle the registry key for the newer version, but so far they seemed to do it the same way from 8.1 and ongoing. I might try to check directly the retreived string of the OS, and based on that for each case i check the related registry in case MS changes that again, but still then you will need to add each new string to check. I can't really do that much about it here because that's how MS thinked the system. If i have to check something i need the exact string comparison/build number comparison

ui/xemu-os-utils-windows.c
```diff
@@ -19,10 +19,45 @@
...
+    HKEY keyhandle; 
+    WCHAR current_version[1024], current_build[1024], product_name[1024]; 
+    WCHAR version_size = 1024, build_size = 1024, product_size = 1024;
+
+    if (buffer == NULL) {
+        //Get Current Kernel version number.
+       if (RegOpenKeyExW(HKEY_LOCAL_MACHINE, L"SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion", 0,
+                          KEY_QUERY_VALUE, &keyhandle) == ERROR_SUCCESS) {
+
+            RegGetValueW(HKEY_LOCAL_MACHINE, L"SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion",
```
mborgerson: Please handle the case that RegGetValueW fails; if so, the g_strdup_printf below may read from uninitialized buffers, possibly crashing the app

ui/xemu-os-utils-windows.c
```diff
@@ -19,11 +19,50 @@
...
+    static const char *buffer = NULL;
+    HKEY keyhandle; 
+    WCHAR current_version[1024], current_build[1024], product_name[1024]; 
+    WCHAR version_size = 1024, build_size = 1024, product_size = 1024;
+
+    if (buffer == NULL) {
+        //Get Current Kernel version number.
+       if (RegOpenKeyExW(HKEY_LOCAL_MACHINE, L"SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion", 0,
+                          KEY_QUERY_VALUE, &keyhandle) == ERROR_SUCCESS) {
+            if (RegGetValueW(HKEY_LOCAL_MACHINE, L"SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion",
```
antangelo: Instead of having multiple nested `if` blocks, I suggest splitting all of these checks into a static helper function. Then, you can instead have something like:
```C
if (...() != ERROR_SUCCESS) return NULL;
if (...() != ERROR_SUCCESS) return NULL;
...
return g_strdup_printf(...);
```

You can then call this helper function once in the `if (buffer == NULL)` block, and set `buffer` depending on the result.

ui/xemu-os-utils-windows.c
```diff
@@ -19,11 +19,50 @@
...
+            if (RegGetValueW(HKEY_LOCAL_MACHINE, L"SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion", 
+                             L"ProductName", RRF_RT_REG_SZ, (LPVOID)NULL, &product_name, (LPDWORD)&product_size == ERROR_SUCCESS)) {
+                if (RegGetValueW(HKEY_LOCAL_MACHINE, L"SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion", 
+                                 L"CurrentVersion", RRF_RT_REG_SZ, (LPVOID)NULL, &current_version, (LPDWORD)&version_size == ERROR_SUCCESS)) {
+
+                        /* if version number is 6.3/10.0 (8.1/10/11), get the build number from the DisplayVersion Registry.
+                            Reference: https://en.wikipedia.org/wiki/Windows_NT */
+                    if ((wcscmp(current_version, L"10.0") == 0) || (wcscmp(current_version, L"6.3") == 0)) {
+                        if (RegGetValueW(HKEY_LOCAL_MACHINE, L"SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion", 
+                                    L"DisplayVersion", RRF_RT_REG_SZ, (LPVOID)NULL, &current_build, (LPDWORD)&build_size == ERROR_SUCCESS)) {
```
antangelo: The comparison here (and in a couple spots above) looks like it's not being applied to the result of the function call, but to the last argument (i.e. you're passing `(LPDWORD) &build_size == ERROR_SUCCESS` as an argument instead of checking the return value).

ui/xemu-os-utils-windows.c
```diff
@@ -19,11 +19,50 @@
...
+                            buffer = g_strdup_printf("%ls %ls", product_name, current_build);
+                            //If it's lower (win 8 and below until XP) get the build descriptor from CSDVersion.
+                        } else if (RegGetValueW(HKEY_LOCAL_MACHINE, L"SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion",
+                                        L"CSDVersion", RRF_RT_REG_SZ, (LPVOID)NULL, &current_build, (LPDWORD)&build_size)) {
```
antangelo: Is this missing an `ERROR_SUCCESS` check?

ui/xemu-os-utils-windows.c
Fabxx: i have to make the `ifs` nested because they depend on the first `regkeyopen` operation, if i don't open the key to look for the subkeys it will return garbage in the `g_strdup_printf` for not being able to read the values, and each function needs to redirect to the default fallback as requested because we don't know if a key works and the others don't. Besides splitting the checks in another function just to make another function look cleaner looks redundant to me, and also i can't return NULL as the default fallback requested was `"windows"`, and i don't like the fact to type multiple times `return "Windows"` when can be done only once.  Also note that this check returns already implicitly the state of the operation, hence why the last if is made directly on that state returned.

antangelo: The `RegOpenKeyExW` functions are filling the string and size buffers when they return success, and those variables are not limited to the scope of the `if` statement that makes the function call, so it's perfectly fine to not nest here. Those variables only have invalid values if the `RegOpenKeyExW` call does not return success, but if you return from the function after checking this, you will never run into a scenario where you are using invalid values.

The point of splitting checks into separate functions (or splitting functions altogether) is to take one function that is difficult to understand, and split it into two functions that are easy to understand. Generally you don't want to have too much nesting either, as it can get unwieldy, so it's a good sign that things should be split up.

You can still return the default value of `"Windows"` from the `xemu_get_os_info` function if the helper returns null, and you don't have to do it more than once:

```C
static const char *get_detailed_windows_version(); // returns NULL on failure
const char *xemu_get_os_info()
{
  static const char *buffer = NULL;
  if (!buffer) {
    buffer = get_detailed_windows_version();
    if (!buffer) buffer = "Windows";
  }

  return buffer;
}
```

ui/xemu-os-utils-windows.c
```diff
@@ -19,11 +19,59 @@
...
+    WCHAR current_version[1024], current_build[1024], product_name[1024]; 
+    WCHAR version_size = 1024, build_size = 1024, product_size = 1024;
+
+    //If one of the registry open fails, return Windows as a fallback.
+    if (RegOpenKeyExW(HKEY_LOCAL_MACHINE, L"SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion", 0,
+                    KEY_QUERY_VALUE, &keyhandle) != ERROR_SUCCESS) 
+                    return NULL;
+
+    if (RegGetValueW(HKEY_LOCAL_MACHINE, L"SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion", 
+                        L"ProductName", RRF_RT_REG_SZ, (LPVOID)NULL, &product_name, (LPDWORD)&product_size) != ERROR_SUCCESS)
```
antangelo: nit: These should be aligned consistently

ui/xemu-os-utils-windows.c
```diff
@@ -19,11 +19,59 @@
...
+
+    if (RegGetValueW(HKEY_LOCAL_MACHINE, L"SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion", 
+                        L"ProductName", RRF_RT_REG_SZ, (LPVOID)NULL, &product_name, (LPDWORD)&product_size) != ERROR_SUCCESS) 
+                        return NULL; 
+        
+    if (RegGetValueW(HKEY_LOCAL_MACHINE, L"SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion", 
+                        L"CurrentVersion", RRF_RT_REG_SZ, (LPVOID)NULL, &current_version, (LPDWORD)&version_size) != ERROR_SUCCESS) 
+                        return NULL;
+
+                /* if version number is 6.3/10.0 (8.1/10/11), get the build number from the DisplayVersion Registry.
```
antangelo: nit: Comments not indented correctly

ui/xemu-os-utils-windows.c
```diff
@@ -19,11 +19,59 @@
...
+    if (buffer == NULL) {
+        buffer = get_windows_build_info();
+        
+        if (buffer == NULL) {
+            buffer = "Windows";
+        }
+    }
+    return buffer;
 }
+
```
antangelo: nit: Extra newline

antangelo (commented):
A couple more nits, but it otherwise looks fine to me

mborgerson:
@Fabxx I made some simplifications over here: https://github.com/xemu-project/xemu/tree/fix/win_os_info

 Would you mind taking a look and helping to test it?

Fabxx:
> @Fabxx I made some simplifications over here: https://github.com/xemu-project/xemu/tree/fix/win_os_info
> 
>  Would you mind taking a look and helping to test it?

Sure, will do when i get home. Will let you know within the day

mborgerson:
Thanks!

Issue #1128, opened 2022-06-28
Add user interface for attaching XMU, XBLC


mborgerson:
### Feature Request

Related to #90, #389 and soon #1126. Breaking out because this deserves its own thread. I have some initial ideas about how the UI should work for attaching these controller peripherals, and for supporting emulation of multiple devices, but they need more refinement. If you choose to begin working on this, please work with me to make sure we align on a direction to expedite merge.

I also have some artwork for XMUs that I made a while back in the style of the existing outline art to be integrated into the UI: 
<img src=https://user-images.githubusercontent.com/8210/176090896-26e77b89-ad57-4b5b-9ba4-3e65bc77e613.png width=200 />

Possible interface mockup:

![image](https://user-images.githubusercontent.com/8210/176982909-df37eaba-797f-455d-b07d-80c919e0948c.png)
![image](https://user-images.githubusercontent.com/8210/176983237-59ef77fd-59a7-4c29-974a-d724654522ee.png)
![image](https://user-images.githubusercontent.com/8210/176983579-46ff8dfa-99f1-4209-8803-655fcbe68443.png)
![image](https://user-images.githubusercontent.com/8210/176984011-9d05ff92-cf7f-4b4d-b6b3-f30d4351a501.png)

Or in an orientation with one peripheral over the other.

binarymaster:
It would be nice also to allow attaching HID keyboard and tablet mouse, really handy for testing desktop operating systems ported to Xbox (e.g. Xbox-Linux, ReactOS).

Issue #389, opened 2021-06-30
Adding missing peripherals 


Triticum0:
### Feature Request
There are a couple of missing peripherals which need to be implemented into Xemu to make some games become playable and would improve the emulator as a whole. Here is a full list of all the ones I know that need to be added.

- [ ] Gametrak
- [x] Xbox Communicator Set (#1126)
- [x] Xbox-Compatible Microphone
- [ ] MadCatz XBOX Steering Wheel
- [ ] Steel Battalion Controller
- [ ] Mad Catz Xbox Blaster
- [ ] Silent Scope Light Rifle
- [ ] Dance Dance Revolution Game Pad
- [x] Xbox Memory Card https://github.com/xemu-project/xemu/pull/1315

UI for above peripherals: #1128

### Games which become playable
Real World Golf
N.U.D.E.@: Natural Ultimate Digital Experiment
Silent Scope Collection
Steel Battalion
Karaoke Revolution
Karaoke Revolution Party
Dance Dance Revolution Ultramix Series  (Not required To play)
Dance UK
https://xemu.app/titles/53540001/#Zillernet (needs mic)

### Additional Context

I'm not sure if there other games that have been omitted by mistake, which are not on the list, and If any peripherals are already implemented, feel free to comment below to correct me.
Source:https://gamicus.fandom.com/wiki/List_of_Xbox_peripherals

GXTX:
There's nothing stopping you from using these already. You plug them into your computer and tell xemu to use them via command line.

There's some extra steps you must go through to get it working on Windows (iirc) but Linux is straight forward.

Triticum0:
Is there a guide on how to do this on windows and does this include Gametrak and Steel Battalion Controller because pcsx2 hasn't even implemented Gametrak
Also XMU hasn't been implemented https://github.com/mborgerson/xemu/issues/90

GXTX:
Read the QEMU man pages on USB pass through, and XMU support is already there (it's just a generic storage device), just again not via the UI.

mborgerson:
Pass-thru will work for some peripherals. For light guns it will probably not work, as they use they are expected to use the yellow wire, however these can be emulated.

xemu will eventually expose an interface for configuring XMUs+pass-thru+emulated peripherals. In the mean time, we consider the path "advanced" and don't really provide user support for it. That said, I'd accept a PR for the website which plainly describes how to do it for users who want to get their hands dirty.

Triticum0:
[...]

PR #1315, opened 2023-01-02
UI: XMU Settings

faha223:
Added support for creating and mounting/unmounting XMU images in the Input settings menu.


ui/xemu-input.c
```diff
@@ -516,3 +701,56 @@ int xemu_input_get_test_mode(void)
...
+    uint32_t signature;
+    uint32_t volume_id;
+    uint32_t sectors_per_cluster;
+    uint32_t root_cluster;
+    uint16_t unknown1;
+    uint8_t  padding[4078];
+};
+#pragma pack()
+
+bool xemu_new_xmu(const char* filename, unsigned int size)
```
GXTX: This should be unneeded. The dashboard properly formats entirely blank images.

faha223: The user isn't guaranteed to have a dashboard installed, and this saves the user an extra step

faha223: If you want me to remove it I will. I just think it would be incomplete without it, and it might make things difficult for users who don't have a dashboard in their HDD image.

faha223:
This is related to these issues:
https://github.com/xemu-project/xemu/issues/1128
https://github.com/xemu-project/xemu/issues/389

ui/xui/main-menu.cc
```diff
@@ -257,6 +260,183 @@ void MainMenuInputView::Draw()
...
+                const char *id = g_strdup_printf("MU_%d%c File Path", active + 1, 'A' + i);
+                const char *xmu_port_path = NULL;
+                if(xmu->filename == NULL)
+                    xmu_port_path = strdup("");
+                else 
+                    xmu_port_path = strdup(xmu->filename);
+                if (FilePicker(id, &xmu_port_path, img_file_filters)) {
+                    xemu_input_bind_xmu(active, i, xmu_port_path);
+                }
+                free((void*)id);
```
antangelo: This should be freed with `g_free` (since it's allocated with `g_strdup`)

ui/xemu-input.c
```diff
@@ -178,6 +198,25 @@ int xemu_input_get_controller_default_bind_port(ControllerState *state, int star
...
 
+void xemu_save_peripheral_settings(int player_index, int peripheral_index, int peripheral_type, const char *peripheral_parameter)
+{
+    fprintf(stderr, "saving peripheral settings for port %d%c: type = %d, filename = %s\r\n", player_index+1, 'A' + peripheral_index, peripheral_type, peripheral_parameter == NULL ? "NULL" : peripheral_parameter);
+
+    int *peripheral_type_ptr = peripheral_types_settings_map[player_index][peripheral_index];
+    const char **peripheral_param_ptr = peripheral_params_settings_map[player_index][peripheral_index];
+
+    if(peripheral_type_ptr != NULL && peripheral_param_ptr != NULL)
+    {
```
antangelo: nit (here and throughout): xemu conventions (in most files) are to have braces on the same line, except for function definitions (which are on a newline). Generally, you should follow whatever style the surrounding code does.

ui/xemu-input.c
```diff
@@ -178,6 +198,25 @@ int xemu_input_get_controller_default_bind_port(ControllerState *state, int star
...
 }
 
+void xemu_save_peripheral_settings(int player_index, int peripheral_index, int peripheral_type, const char *peripheral_parameter)
+{
+    fprintf(stderr, "saving peripheral settings for port %d%c: type = %d, filename = %s\r\n", player_index+1, 'A' + peripheral_index, peripheral_type, peripheral_parameter == NULL ? "NULL" : peripheral_parameter);
+
+    int *peripheral_type_ptr = peripheral_types_settings_map[player_index][peripheral_index];
+    const char **peripheral_param_ptr = peripheral_params_settings_map[player_index][peripheral_index];
+
+    if(peripheral_type_ptr != NULL && peripheral_param_ptr != NULL)
```
antangelo: You can probably assert on these, since they point to the config they should never be null in practice

ui/xemu-input.c
```diff
@@ -178,6 +198,25 @@ int xemu_input_get_controller_default_bind_port(ControllerState *state, int star
...
+    {
+        *peripheral_type_ptr = peripheral_type;
+        if(*peripheral_param_ptr != NULL)
+            free(*peripheral_param_ptr);
+        *peripheral_param_ptr = 
+            peripheral_parameter == NULL ? strdup("") : strdup(peripheral_parameter);
```
antangelo: This pattern is repeated a few times, might be worth modifying `xemu_settings_set_string` to accept a null string parameter and use that instead

ui/xemu-input.c
```diff
@@ -255,6 +298,30 @@ void xemu_input_process_sdl_events(const SDL_Event *event)
...
             snprintf(buf, sizeof(buf), "Connected '%s' to port %d", new_con->name, port+1);
             xemu_queue_notification(buf);
+
+            // Try to bind peripherals back to controller
+            enum peripheral_type peripheralType = PERIPHERAL_NONE;
+            const char *param = NULL;
+
+            for(int i = 0; i < 2; i++)
+            {
+                peripheralType = (enum peripheral_type)(*peripheral_types_settings_map[port][i]);
```
antangelo: Not a huge issue, but we may want to validate that `*(peripheral_types_settings_map[port][i])` is in the expected range for the enum, in case an invalid value is saved to the config at any point.

ui/xemu-input.c
```diff
@@ -255,6 +298,30 @@ void xemu_input_process_sdl_events(const SDL_Event *event)
             char buf[128];
             snprintf(buf, sizeof(buf), "Connected '%s' to port %d", new_con->name, port+1);
             xemu_queue_notification(buf);
+
+            // Try to bind peripherals back to controller
+            enum peripheral_type peripheralType = PERIPHERAL_NONE;
+            const char *param = NULL;
```
antangelo: nit: These definitions can be moved into the for loop to limit scope

ui/xemu-input.c
```diff
@@ -255,6 +298,30 @@ void xemu_input_process_sdl_events(const SDL_Event *event)
...
+                if(peripheralType == PERIPHERAL_XMU) {
+                    if(param != NULL && strlen(param) > 0) {
+                        // This is an XMU and needs to be bound to this controller
+                        bound_controllers[port]->PeripheralTypes[i] = (enum peripheral_type)peripheralType;
+                        bound_controllers[port]->Peripherals[i] = malloc(sizeof(XmuState));
+                        memset(bound_controllers[port]->Peripherals[i], 0, sizeof(XmuState));
+                        xemu_input_bind_xmu(port, i, param);
+
+                        char buf[128];
+                        snprintf(buf, 128, "Connected XMU %s to port %d%c", param, port + 1, 'A' + i);
```
antangelo: This can be truncated if the path is longer than ~100 characters, not sure if this will be frequently encountered in practice but it may be worth considering.

ui/xemu-input.c
```diff
@@ -516,3 +703,56 @@ int xemu_input_get_test_mode(void)
 {
     return test_mode;
 }
+
+#define FATX_SIGNATURE 0x58544146
```
antangelo: Might be better to move the FATX code into its own file

ui/xemu-input.c
```diff
@@ -516,3 +703,56 @@ int xemu_input_get_test_mode(void)
...
+
+bool xemu_new_xmu(const char* filename, unsigned int size)
+{
+    unsigned int PartitionId = (unsigned int)rand();
+    unsigned int SectorsPerCluster = 0x04;
+    unsigned int RootDirectoryCluster = 0x01;
+    unsigned char zero = 0x00;
+    unsigned int empty_fat = 0xfffffff8;
+
+    FILE *fp = fopen(filename, "wb");
```
antangelo: `fopen` should be replaced with `qemu_fopen`

ui/xui/main-menu.cc
```diff
@@ -257,6 +260,183 @@ void MainMenuInputView::Draw()
...
+                    ImGui::PushID(86 + j);
+                    const char *selectable_label = peripheralTypeNames[j];
+                    char buf[128];
+
+                    if (ImGui::Selectable(selectable_label, is_selected)) {
+                        if(bound_state->Peripherals[i] != NULL)
+                        {
+                            if(bound_state->PeripheralTypes[i] == PERIPHERAL_XMU)
+                            {
+                                fprintf(stderr, "Another peripheral was already bound. Unplugging\r\n");
```
antangelo: Are these prints intended to provide context for the user? If so, then it may be better to show a UI notification instead, since not all users will be able to see stderr (Windows, for instance). If they are for debugging, then I think they can be removed or placed behind some sort of tracing flag.

ui/xui/main-menu.cc
```diff
@@ -257,6 +260,183 @@ void MainMenuInputView::Draw()
...
+                                xemu_input_unbind_xmu(active, i);
+                            }
+                        }
+                        bound_state->PeripheralTypes[i] = (enum peripheral_type)j;
+                        if(j == PERIPHERAL_XMU) {
+                            bound_state->Peripherals[i] = malloc(sizeof(XmuState));
+                            memset(bound_state->Peripherals[i], 0, sizeof(XmuState));
+                        } else if(bound_state->Peripherals[i] != NULL) {
+                            fprintf(stderr, "'None' was selected. Deleting the XmuState\r\n");
+                            //free((void*)bound_state->Peripherals[i]);
```
antangelo: Does the absence of this free leak memory?

ui/xui/main-menu.cc
faha223: These were for debugging and weren't supposed to end up in the PR

ui/xui/main-menu.cc
faha223: This was commented out when I was tracking down a bug that turned out not to be related to this. Thanks for catching this.

ui/xemu-input.c
faha223: I moved this to ui\fatx.c with a fatx.h header that's included in main-menu.cc. I put it here to keep it close to xemu-input.c. Is this a good place for it or can you think of a better one?

antangelo: `/ui/thirdparty` or `/thirdparty` may be slightly better (since it's from an external lib)

faha223: I moved it to ui\thirdparty\fatx

ui/xemu-input.c
faha223: For validating the enum, should I add an enum value with a name like PERIPHERAL_TYPE_COUNT as the highest value of the enum and just make sure the value is between 0 and PERIPHERAL_TYPE_COUNT?

And should I make it an assert, or should I make it correct the value to PERIPHERAL_NONE if it's out of bounds?

antangelo: That approach to validation sounds good to me. Either an assert or correcting the value is fine, the latter may be more user friendly but this isn't a case that most people will run into anyway, so I don't think it's a huge deal.

ui/xemu-input.c
```diff
@@ -507,6 +579,122 @@ void xemu_input_bind(int index, ControllerState *state, int save)
...
+    xmu->filename = strdup(filename);
+
+    const int port_map[4] = {3, 4, 1, 2};
+    const int xmu_map[2] = {2, 3};
+    char *tmp;
+
+    static int id_counter = 0;
+    tmp = g_strdup_printf("xmu_%d", id_counter++);
+    
+    // Add the file as a drive (fixme: Specify the format as raw so that qemu doesn't have to guess)
```
antangelo: Is this fixme still relevant? Looks like you're specifying the format already.

ui/xemu-input.c
```diff
@@ -507,6 +579,122 @@ void xemu_input_bind(int index, ControllerState *state, int save)
...
+    if(xmu != NULL)
+    {
+        if(xmu->dev != NULL) {
+            qdev_unplug((DeviceState*)xmu->dev, &error_abort);
+            object_unref(OBJECT(xmu->dev));
+            xmu->dev = NULL;
+        }
+
+        if(xmu->filename != NULL) {
+            free((void*)xmu->filename);
```
antangelo: If you use `g_strdup` to set the filename then you don't need the null check here

ui/xemu-input.h
```diff
@@ -88,6 +99,9 @@ typedef struct ControllerState {
     SDL_JoystickID      sdl_joystick_id;
     SDL_JoystickGUID    sdl_joystick_guid;
 
+    enum peripheral_type PeripheralTypes[2];
+    void               *Peripherals[2];
```
antangelo: nit: Inconsistent alignment

ui/xui/main-menu.cc
```diff
@@ -257,6 +262,177 @@ void MainMenuInputView::Draw()
...
+                            g_viewport_mgr.Scale(ImVec2(0, 12)));
+        ImGui::Columns(2, "mixed", false);
+
+        xmu_fbo->Target();
+        id = (ImTextureID)(intptr_t)xmu_fbo->Texture();
+
+        const char *img_file_filters = ".img Files\0*.img\0All Files\0*.*\0";
+        const char *comboLabels[2] = { "###PeripheralTypesA", "###PeripheralTypesB" };
+        for (int i = 0; i < 2; i++) {
+
```
antangelo: nit: Extra newline

ui/xui/main-menu.cc
```diff
@@ -257,6 +262,177 @@ void MainMenuInputView::Draw()
...
+            enum peripheral_type selectedType = bound_state->PeripheralTypes[i];
+            const char *peripheralTypeNames[2] = { "None", "Memory Unit" };
+            const char *selectedPeripheralType = peripheralTypeNames[selectedType];
+            ImGui::SetNextItemWidth(-FLT_MIN);
+            if (ImGui::BeginCombo(comboLabels[i], selectedPeripheralType, ImGuiComboFlags_NoArrowButton))
+            {
+                // Handle all available input devices
+                for(int j = 0; j < 2; j++) {
+                    bool is_selected = selectedType == j;
+                    ImGui::PushID(86 + j);
```
antangelo: Where does `86` come from?

ui/thirdparty/fatx/fatx.c
```diff
@@ -0,0 +1,51 @@
...
+            fwrite(&zero, 1, 1, fp);
+
+        fflush(fp);
+        fclose(fp);
+
+        return true;
+    }
+
+    return false;
+}
```
antangelo: nit: Files should end with newline

ui/xui/main-menu.cc
```diff
@@ -257,6 +262,177 @@ void MainMenuInputView::Draw()
...
+            ImGui::SetCursorPosX(
+                ImGui::GetCursorPosX() +
+                (int)((ImGui::GetColumnWidth() - xmu_w * g_viewport_mgr.m_scale -
+                    2 * port_padding * g_viewport_mgr.m_scale) /
+                    2));
+
+            selectedType = bound_state->PeripheralTypes[i];
+            if(selectedType == PERIPHERAL_XMU) {
+                // We are using the same texture for all buttons, but ImageButton
+                // uses the texture as a unique ID. Push a new ID now to resolve
```
antangelo: Is this referring to the ID that is pushed later in this block (as opposed to now)?

ui/xui/main-menu.cc
```diff
@@ -257,6 +262,177 @@ void MainMenuInputView::Draw()
...
+                    }
+                }
+                ImGui::PopID();
+
+                const char *id = g_strdup_printf("MU_%d%c File Path", active + 1, 'A' + i);
+                const char *xmu_port_path = NULL;
+                if(xmu->filename == NULL)
+                    xmu_port_path = strdup("");
+                else 
+                    xmu_port_path = strdup(xmu->filename);
```
antangelo: Should use `g_strdup` to match `g_free` below

ui/xui/main-menu.cc
```diff
@@ -257,6 +262,177 @@ void MainMenuInputView::Draw()
...
+                                xemu_input_unbind_xmu(active, i);
+                            }
+                        }
+                        bound_state->PeripheralTypes[i] = (enum peripheral_type)j;
+                        if(j == PERIPHERAL_XMU) {
+                            bound_state->Peripherals[i] = malloc(sizeof(XmuState));
+                            memset(bound_state->Peripherals[i], 0, sizeof(XmuState));
+                        } else if(bound_state->Peripherals[i] != NULL) {
+                            // 'None' was selected. Deleting the XmuState
+                            free((void*)bound_state->Peripherals[i]);
```
antangelo: Might be worth using `g_malloc/g_free` here as well to be consistent throughout

ui/xui/main-menu.cc
```diff
@@ -257,6 +262,177 @@ void MainMenuInputView::Draw()
...
+                    int flags = NOC_FILE_DIALOG_SAVE;
+                    const char *new_path = PausedFileOpen(flags, img_file_filters, NULL, "xmu.img");
+
+                    if (new_path) {
+                        if(qemu_access(new_path, F_OK) == 0) {
+                            // The file already exists
+                            char *msg = g_strdup_printf("%s already exists", new_path);
+                            xemu_queue_notification(msg);
+                            g_free(msg);
+                        } else  {
```
antangelo: nit: There's an extra space after the `else`

ui/xui/main-menu.cc
```diff
@@ -257,6 +262,177 @@ void MainMenuInputView::Draw()
...
+                            // Create an 8MB formatted XMU image. We can create smaller or larger ones, but 8 MB is the default
+                            if(create_fatx_image(new_path, 8*1024*1024)) {
```
antangelo: nit (here and in a few other places): Should be a space between `if/for/while` and the open bracket

antangelo (commented):
A few more comments, mostly nits

ui/xemu-input.c
faha223: Yeah, that's not relevant anymore

ui/xui/main-menu.cc
faha223: It was a random offset value I chose to make sure the ImGui::PushID on that line wouldn't conflict with the one for the New Image button or the Clear Selection button. It seems I was wrong and I didn't need to worry about that. Thanks for calling my attention to this

ui/thirdparty/fatx/fatx.c
```diff
@@ -0,0 +1,51 @@
...
+        memset(&superblock, 0xff, sizeof(struct fatx_superblock));
+
+        superblock.signature = FATX_SIGNATURE;
+        superblock.sectors_per_cluster = 4;
+        superblock.volume_id = (uint32_t)rand();
+        superblock.root_cluster = 1;
+        superblock.unknown1 = 0;
+
+        // Write the fatx superblock.
+        fwrite(&superblock, sizeof(superblock), 1, fp);
```
mborgerson, line 36: This implementation depends on host byte order being little endian. Can you rework it so it will also work on BE hosts?

ui/thirdparty/fatx/fatx.c
```diff
@@ -0,0 +1,51 @@
...
+
+        // Write the FAT
+        fwrite(&empty_fat, sizeof(empty_fat), 1, fp);
+
+        // Fill the rest of the space with zeros
+        for (unsigned int i = ftell(fp); i < size; i++)
+            fwrite(&zero, 1, 1, fp);
```
mborgerson: We can avoid this loop and writing a bunch of zeros with seek/truncate, which will create a sparse file (logically, but not physically allocating space for all the zeros)

ui/xemu-input.c
```diff
@@ -507,6 +579,120 @@ void xemu_input_bind(int index, ControllerState *state, int save)
...
+    // Create the device
+    QemuOpts *opts = qemu_opts_from_qdict(qemu_find_opts("device"), qdict2, &error_abort);
+
+    DeviceState *dev = qdev_device_add(opts, &error_abort);
+    assert(dev);
+
+    xmu->dev = (void*)dev;
+
+    // Unref for eventual cleanup
+    //g_free(dinfo);
```
mborgerson: ?

ui/xemu-input.c
```diff
@@ -178,6 +198,22 @@ int xemu_input_get_controller_default_bind_port(ControllerState *state, int star
...
+    int *peripheral_type_ptr = peripheral_types_settings_map[player_index][peripheral_index];
+    const char **peripheral_param_ptr = peripheral_params_settings_map[player_index][peripheral_index];
+
+    assert(peripheral_type_ptr);
+    assert(peripheral_param_ptr);
+    
+    *peripheral_type_ptr = peripheral_type;
+    xemu_settings_set_string(peripheral_param_ptr, peripheral_parameter == NULL ? "" : peripheral_parameter);
+
+    fprintf(stderr, "saved peripheral settings for port %d%c: type = %d, filename = %s\r\n", player_index+1, 'A' + peripheral_index, peripheral_type, peripheral_parameter == NULL ? "NULL" : peripheral_parameter);
```
mborgerson: Don't need this level of debug info normally

config_spec.yml
```diff
@@ -25,6 +25,51 @@ input:
     port2: string
     port3: string
     port4: string
+  peripherals:
+    port1:
+      peripheral_type_0:
```
mborgerson: nit: The schema could be a little smaller

Some thoughts: [genconfig](https://github.com/mborgerson/genconfig) is the config system used in xemu
* genconfig does support arrays, which could simplify this schema a bit, but might complicate the code a little too
* genconfig system has an enum type which would be preferable for config checking, or a string type could be used to identify the device type if necessary, enums in gen-config aren't easily re-used

or if you want to keep it fixed as-is, it could be simplified like:

```
    port2:
      peripheral_type_0: integer
      peripheral_param_0: string
      peripheral_type_1: integer
      peripheral_param_1: string
```

nit: The manual refers to these peripheral slots as "Expansion Slot A" (top) and "Expansion Slot B" (bottom, nearest the cable)

ui/xemu-input.c
```diff
@@ -507,6 +579,120 @@ void xemu_input_bind(int index, ControllerState *state, int save)
...
+                        return;
+                    }
+                }
+            }
+        }
+    }
+
+    xmu->filename = strdup(filename);
+
+    const int port_map[4] = {3, 4, 1, 2};
```
mborgerson: This is duplicated with `xemu_input_bind`, could you refactor this mapping out so it can be shared by both functions in this file

ui/xui/main-menu.cc
```diff
@@ -260,6 +265,176 @@ void MainMenuInputView::Draw()
...
+
+                    if (new_path) {
+                        if (qemu_access(new_path, F_OK) == 0) {
+                            // The file already exists
+                            char *msg = g_strdup_printf("%s already exists", new_path);
+                            xemu_queue_notification(msg);
+                            g_free(msg);
+                        } else {
+                            // Create an 8MB formatted XMU image. We can create smaller or larger ones, but 8 MB is the default
+                            if (create_fatx_image(new_path, 8*1024*1024)) {
```
mborgerson: Let's move this number out to a constant definition and drop the above comment

ui/xui/main-menu.cc
```diff
@@ -260,6 +265,176 @@ void MainMenuInputView::Draw()
...
+                    xmu_display_size =
+                        ImVec2(xmu_w * g_viewport_mgr.m_scale,
+                            xmu_h * g_viewport_mgr.m_scale);
+                }
+
+                ImGui::SetCursorPosX(
+                    ImGui::GetCursorPosX() +
+                    (int)((ImGui::GetColumnWidth() - xmu_display_size.x) / 2.0));
+
+                ImGui::Text("Peripheral %c", 'A' + i);
```
mborgerson: "Expansion Slot" is the manual terminology

ui/xemu-input.h
```diff
@@ -88,6 +99,9 @@ typedef struct ControllerState {
     SDL_JoystickID      sdl_joystick_id;
     SDL_JoystickGUID    sdl_joystick_guid;
 
+    enum peripheral_type PeripheralTypes[2];
```
mborgerson: nit: identifier style

ui/xui/main-menu.cc
```diff
@@ -260,6 +265,176 @@ void MainMenuInputView::Draw()
...
+                if (ImGui::Button("New Image", ImVec2(250, 0)))
+                {
+                    int flags = NOC_FILE_DIALOG_SAVE;
+                    const char *new_path = PausedFileOpen(flags, img_file_filters, NULL, "xmu.img");
+
+                    if (new_path) {
+                        if (qemu_access(new_path, F_OK) == 0) {
+                            // The file already exists
+                            char *msg = g_strdup_printf("%s already exists", new_path);
+                            xemu_queue_notification(msg);
```
mborgerson: There's `xemu_queue_error_message` for errors

ui/xui/main-menu.cc
```diff
@@ -260,6 +265,176 @@ void MainMenuInputView::Draw()
...
+                    xmu_port_path = g_strdup("");
+                else 
+                    xmu_port_path = g_strdup(xmu->filename);
+                if (FilePicker(id, &xmu_port_path, img_file_filters)) {
+                    xemu_input_bind_xmu(active, i, xmu_port_path);
+                }
+                g_free((void*)id);
+                g_free((void*)xmu_port_path);
+
+                ImGui::PushID(i+2); // Add 2 because ID value of i was already used and the for loop iterating on i stops at 2
```
mborgerson: Should be able to have just the one PushID at the start and PopID at the end of the loop, dropping the middle Pop/Push

faha223: I removed those lines and it doesn't appear to have broken anything

ui/xui/main-menu.cc
faha223: Thanks, I changed over the xemu_queue_notification calls to xemu_queue_error_message where I was using it for error messages.

ui/xemu-input.h
faha223: I changed it to peripheral_types and peripherals. Is that ok?

ui/thirdparty/fatx/fatx.c
faha223: I changed this to fseek to size - sizeof(unsigned byte), and I just write one byte of 0x00 at the end. I initially tried ftruncate(size) but that didn't do the job.

config_spec.yml
faha223: I simplified it as you suggested. I used the word peripheral as it refers to the device plugged into the expansion slot rather than the slot itself, but I can change this if you'd like

ui/xemu-input.c
faha223: I pulled this out into a static const in[] and referenced that in both places

ui/xui/main-menu.cc
faha223: I moved this to a preprocessor definition called DEFAULT_XMU_SIZE

ui/xui/main-menu.cc
faha223: I changed the labels in the UI to follow the Expansion Slot wording, but I left the variables saying peripheral as they refer to the device rather than the slot in which the device is installed

ui/thirdparty/fatx/fatx.c
faha223: I have changed the code to use cpu_to_le32(u32) on the unsigned integers that were being written to the superblock. I think that will work but I don't have a BE machine to test on. It doesn't break this functionality on my machine.

ui/xemu-input.c
faha223: dinfo is allocated in drive_new by a call to g_malloc0. I don't use it again, but I'm not sure what to do with it. Calling g_free here on dinfo causes a segfault when unmounting the XMU image. I didn't mean to leave this commented.

antangelo: It's tied to the blockdev and freed when the blockdev is deleted. You shouldn't have to do anything

ui/xui/main-menu.cc
```diff
@@ -260,6 +267,173 @@ void MainMenuInputView::Draw()
...
+                if (ImGui::Button("New Image", ImVec2(250, 0)))
+                {
+                    int flags = NOC_FILE_DIALOG_SAVE;
+                    const char *new_path = PausedFileOpen(flags, img_file_filters, NULL, "xmu.img");
+
+                    if (new_path) {
+                        if (qemu_access(new_path, F_OK) == 0) {
+                            // The file already exists
+                            char *msg = g_strdup_printf("%s already exists", new_path);
+                            xemu_queue_error_message(msg);
```
mborgerson, line 410: As an alternative to showing an error message in this way, we can request that the native file dialog prompt users for overwrite by adding the flag `NOC_FILE_DIALOG_OVERWRITE_CONFIRMATION` with a bitwise OR above on line 383

ui/xui/main-menu.cc
```diff
@@ -260,6 +267,173 @@ void MainMenuInputView::Draw()
...
+                    }
+                }
+
+                const char *id = g_strdup_printf("MU_%d%c File Path", active + 1, 'A' + i);
+                const char *xmu_port_path = NULL;
+                if (xmu->filename == NULL)
+                    xmu_port_path = g_strdup("");
+                else 
+                    xmu_port_path = g_strdup(xmu->filename);
+                if (FilePicker(id, &xmu_port_path, img_file_filters)) {
```
mborgerson: If the user presses the "X" button on the file picker widget, this will cause xemu to crash because file will be unspecified when binding the xmu. In this case, we can unbind the xmu and eliminate the need for the "Clear Selection" button below

ui/xemu-input.c
```diff
@@ -255,6 +293,39 @@ void xemu_input_process_sdl_events(const SDL_Event *event)
             char buf[128];
             snprintf(buf, sizeof(buf), "Connected '%s' to port %d", new_con->name, port+1);
             xemu_queue_notification(buf);
+
+            // Try to bind peripherals back to controller
+            for(int i = 0; i < 2; i++) {
```
mborgerson: Unfortunately binding here wont detect the keyboard binds. We can probably just refactor this code out and add a call in `xemu_input_bind`

ui/xemu-input.c
```diff
@@ -255,6 +293,39 @@ void xemu_input_process_sdl_events(const SDL_Event *event)
...
+                enum peripheral_type peripheralType = (enum peripheral_type)(*peripheral_types_settings_map[port][i]);
```
mborgerson: nit: style

ui/xui/main-menu.cc
```diff
@@ -260,6 +267,173 @@ void MainMenuInputView::Draw()
...
+                // If mounted
+                XmuState *xmu = (XmuState*)bound_state->peripherals[i];
+                if (xmu->filename != NULL && strlen(xmu->filename) > 0) {
+                    RenderXmu(x, y, 0x81dc8a00, 0x0f0f0f00);
+
+                } else {
+                    RenderXmu(x, y, 0x1f1f1f00, 0x0f0f0f00);
+                }
+
+                ImVec2 cur = ImGui::GetCursorPos();
```
mborgerson: Set but unused

ui/xui/main-menu.cc
```diff
@@ -260,6 +267,173 @@ void MainMenuInputView::Draw()
...
+                    xmu_display_size =
+                        ImVec2(xmu_w * g_viewport_mgr.m_scale,
+                            xmu_h * g_viewport_mgr.m_scale);
+                }
+
+                ImGui::SetCursorPosX(
+                    ImGui::GetCursorPosX() +
+                    (int)((ImGui::GetColumnWidth() - xmu_display_size.x) / 2.0));
+
+                ImGui::Text("Expansion Slot %c", 'A' + i);
```
mborgerson: Let's drop this extra text

ui/xui/main-menu.cc
```diff
@@ -260,6 +267,173 @@ void MainMenuInputView::Draw()
...
+                                // Show alert message
+                                char *msg = g_strdup_printf("Unable to create XMU image at %s", new_path);
+                                xemu_queue_error_message(msg);
+                                g_free(msg);
+                            }
+                        }
+                    }
+                }
+
+                const char *id = g_strdup_printf("MU_%d%c File Path", active + 1, 'A' + i);
```
mborgerson: Let's change this to something simpler, like "XMU" or "Image" or something. The FilePicker icon already tells the user that it's a file path

ui/xui/main-menu.cc
faha223: This doesn't appear to work on Windows. I changed it to work this way, but when testing it just overwrites the file without a confirmation and, after looking through the implementation of that function, it doesn't appear that NOC_FILE_DIALOG_OVERWRITE_CONFIRMATION is used in the Windows implementation.

ui/xemu-input.c
faha223: I refactored the code into a new function called xemu_input_rebind_xmu(int port) but calling it from xemu_input_bind causes xemu to crash when switching between Keyboard and Xbox Controller. This is because unbinding the controller unplugs the XMUs, and trying to bind the XMUs again right after unplugging them causes an error because the file is still in use.

Instead I call this function where the code was originally and also on line 180 just after it auto-binds the keyboard.

ui/xemu-input.c
faha223: I renamed this variable to peripheral_type, though that might be confusing because it's the same as its type.

ui/xui/main-menu.cc
faha223: I changed it from "MU_1A File Path" to "Image"

ui/xui/main-menu.cc
faha223: I made the X button unbind the XMU and clear the filename, and I removed the "Clear Selection" button

ui/xui/main-menu.cc
faha223: Yes. I've moved this comment to the relevant code.

ui/xui/main-menu.cc
mborgerson: Oh wow, you are right! It looks like it's also missing on macOS, or maybe its the default.

On Windows we will need to support `NOC_FILE_DIALOG_OVERWRITE_CONFIRMATION` and if set pass the `OFN_OVERWRITEPROMPT` flag in when creating the save prompt. Can you try it?

faha223: That worked, I committed the change

ui/xemu-input.c
```diff
@@ -254,6 +293,7 @@ void xemu_input_process_sdl_events(const SDL_Event *event)
             char buf[128];
             snprintf(buf, sizeof(buf), "Connected '%s' to port %d", new_con->name, port+1);
             xemu_queue_notification(buf);
+            xemu_input_rebind_xmu(port);
         }
     } else if (event->type == SDL_CONTROLLERDEVICEREMOVED) {
```
antangelo, line 312: Do we need to unbind the XMU and free the peripherals in this path?

ui/xemu-input.c
```diff
@@ -429,6 +469,11 @@ void xemu_input_bind(int index, ControllerState *state, int save)
     if (bound_controllers[index]) {
         assert(bound_controllers[index]->device != NULL);
         Error *err = NULL;
+
+        // Unbind any XMUs
+        for(int i = 0; i < 2; i++)
+            xemu_input_unbind_xmu(index, i);
```
antangelo: Do we need to free the XMU/peripherals in this path?

ui/xemu-input.c
```diff
@@ -506,6 +550,154 @@ void xemu_input_bind(int index, ControllerState *state, int save)
...
+        }
+
+        const char *param = *peripheral_params_settings_map[port][i];
+
+        if(peripheral_type == PERIPHERAL_XMU) {
+            if(param != NULL && strlen(param) > 0) {
+                // This is an XMU and needs to be bound to this controller
+                if(qemu_access(param, F_OK) == 0) {
+                    bound_controllers[port]->peripheral_types[i] = peripheral_type;
+                    bound_controllers[port]->peripherals[i] = malloc(sizeof(XmuState));
```
antangelo: `malloc` is used to allocate peripherals here, but `g_malloc` is used in the UI. They must be consistent.

ui/xemu-input.c
```diff
@@ -460,7 +505,6 @@ void xemu_input_bind(int index, ControllerState *state, int save)
         bound_controllers[index] = state;
         bound_controllers[index]->bound = index;
 
-        const int port_map[4] = {3, 4, 1, 2};
         char *tmp;
 
         // Create controller's internal USB hub.
```
antangelo, line 537: In my testing if you disconnect and reconnect a controller in the input UI the XMU is not rebound. Do we need to rebind peripherals here?

ui/xui/main-menu.cc
```diff
@@ -260,6 +267,156 @@ void MainMenuInputView::Draw()
...
+                    xmu_display_size =
+                        ImVec2(xmu_w * g_viewport_mgr.m_scale,
+                            xmu_h * g_viewport_mgr.m_scale);
+                }
+
+                ImGui::SetCursorPosX(
+                    ImGui::GetCursorPosX() +
+                    (int)((ImGui::GetColumnWidth() - xmu_display_size.x) / 2.0));
+
+                // We are using the same texture for all buttons, but ImageButton
```
antangelo: What is this comment referring to?

ui/xemu-input.c
```diff
@@ -506,6 +550,154 @@ void xemu_input_bind(int index, ControllerState *state, int save)
...
+                        xemu_queue_notification(buf);
+                        g_free(buf);
+                        return;
+                    }
+                }
+            }
+        }
+    }
+
+    xmu->filename = strdup(filename);
```
antangelo: This should be `g_strdup` to follow the `g_free` used on it.

antangelo (commented):
Please run `clang-format` on the code (see https://xemu.app/docs/dev/#contributing for details on how to do that).

There are also a few spots where `camelCase` variable naming conventions are used, can you update these to follow xemu's `lower_snake_case` convention?

ui/xui/main-menu.cc
faha223: This comment was referring to some unnecessary code that was removed. I have removed the comment

ui/xemu-input.c
faha223: I'm not sure if I understand what you're asking. 

When qdev_unplug is called in xemu_input_bind to free up the underlying QDEV device, it implicitly frees any peripherals connected to the hub. So, we could just set the peripherals[0]->dev and peripherals[1]->dev to NULL and peripheral_types[0] and peripheral_types[1] to PERIPHERAL_NONE. We can't bind them back again when we plug the new controller in though because qdev_unplug doesn't release the XMU image files immediately.

This gave me an idea though.

It might be useful if, instead of freeing the entire QEMU device when we swap controllers, we just reuse the same unerlying QEMU device and move it to the new ControllerState.

This is what it would look like:

If bound_controllers[index] is not null and state is not null, then we're going to be replacing an existing device with a new device. So, instead of freeing the existing device and creating a new one, we call a swap function that does the following:

1. If state is already bound to another port then call xemu_input_bind(state->bound, NULL, 1) to unbind it from the other port.
2. Set state->device to bound_controllers[index]->device
3. Set state->bound to bound_controllers[index]->bound
4. Set bound_controllers[index]->device to NULL
5. Iterate over bound_controllers[index]->periperals and bound_controllers[index]->peripheral_types
    4a. for each expansion slot, copy the peripheral and peripheral_type from bound_controllers[index] to state, then set bound_controllers[index]->peripherals[expansion_slot] to NULL and set bound_controllers[index]->peripheral_types[expansion_slot] to PERIPHERAL_NONE
6. Set bound_controllers[index]->bound to -1
7. Set bound_controllers[index] to state
8. Save the changes. In the config, only the controller should have changed.

Concerns:
In this instance, if we call xemu_input_bind(index, bound_controllers[index], 1) in the current implementation, then it would free the exiting XID device from QEMU and create a new one. If we do that with this proposed implementation without any changes then we'll corrupt the ControllerState for this port. So, a check for bound_controllers[index] == state will need to be done and we'll need to decide to either leave the device alone (which is not what we currently do but which would preserve the state of perpherals), or do what we currently do which is free up the device (and the connected peripherals), and then reconnect the device, losing the peripherals in the process.

Any thoughts on this?

ui/xemu-input.c
faha223: This is how it behaves when we unplug a device and plug it back in with auto-bind enabled so we should probably do that here as well. 

If I remember correctly, this will cause Xemu to crash if the user selects a different controller with one already connected though because QEMU will not have finished releasing the XMU image after freeing the existing controller when it goes to rebind the XMU to the hub created for the new controller. If we want to do this then we might want to consider reusing the QEMU device that's already bound when changing from one controller to another instead freeing it and creating a new one.

ui/xemu-input.c
faha223: I'm not sure I understand. Are you asking in relation to the SDL_CONTROLLERDEVICEREMOVED event or the xemu_input_rebind_xmu call?

antangelo: This is in relation to the device removed event. When the device is removed, the controller state is freed, so any allocated peripherals will be leaked.

ui/xemu-input.c
antangelo: Initially I thought this might be another leak like the above case, but on second inspection it might not be since the underlying state isn't freed.

From a technical standpoint I don't see an issue with your idea. From a UX standpoint, the current behavior when switching controllers is similar to switching on real hardware (i.e. unplugging one controller and plugging in another one). I don't suspect most users would care about preserving this, but it does add an extra step if this is wanted behavior.

If you did want to attempt it, I would suggest sending it as a separate patch and getting Matt's opinion on the UX (leaving out the peripheral part of it and adjusting one or the other patch appropriately depending on which lands first).

ui/xemu-input.c
faha223: It will unbind the XMU/peripherals when we call xemu_input_bind(iter->bind, NULL, 0)

ui/xui/main-menu.cc
```diff
@@ -260,6 +267,164 @@ void MainMenuInputView::Draw()
...
+                    const char *selectable_label = peripheral_type_names[j];
+
+                    if (ImGui::Selectable(selectable_label, is_selected)) {
+                        if (bound_state->peripherals[i] != NULL) {
+                            if (bound_state->peripheral_types[i] ==
+                                PERIPHERAL_XMU) {
+                                // Another peripheral was already bound.
+                                // Unplugging
+                                xemu_input_unbind_xmu(active, i);
+                            }
```
antangelo, line 318: Do we need to free `bound_state->peripherals[i]` here?

ui/xemu-input.c
faha223: Oh, I think I see what you mean. It's not the QEMU device that's being leaked, it's the memory for the peripheral state that's being leaked.

antangelo: Unless I'm missing it though, the `XmuState` structs themselves are not freed by `xemu_input_unbind_xmu` (they are only freed in the UI if the user selects None under the peripheral).

ui/xui/main-menu.cc
faha223: Yes, otherwise the existing XmuState will be lost when a new one is allocated

ui/xemu-input.c
faha223: I'm beginning to think that maybe I should change that. I'm going through and fixing these memory leaks and running out of places where I'm not freeing the XmuState immediately after unbinding it

faha223: I decided to leave the freeing of the `XmuState` structs out of the `xemu_input_unbind_xmu` function because there are 2 places where I want to unbind the XMU but I don't want to free it: When deselecting the image, and when changing the image. Another option would be to add a boolean argument to the function that lets the caller specify whether or not to free the `XmuState` and set the peripheral type to `PERIPHERAL_NONE`

antangelo:
I've created https://github.com/xemu-project/xemu/issues/1572 to track the detach/reattach issue as a followup

PR #1330, opened 2023-01-13
ci: Generate AppImage

### LINKED ISSUE #1103: Build AppImage releases
### Feature Request

It would be nice if xemu released AppImage versions for Linux. It's something we can easily do in the CI pipeline to provide consistent/reproducible/portable releases to people running different distributions.

### Alternatives

_No response_

### Additional Context

_No response_

mborgerson:
Closes #1103

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Clone tree
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
       with:
         fetch-depth: 0
     # On push to master, increment patch version and create a new tag on release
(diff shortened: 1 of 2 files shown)


j8r:
We can then use this same file for the Flatpak.
Also note that `appdata.xml` extenaion has been deprecates in favor of `metainfo.xml`, see https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html.

j8r:
We can then use this same file for the Flatpak.
Also note that `appdata.xml` extension has been deprecated in favor of `metainfo.xml`, see https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html.

mborgerson:
> We can then use this same file for the Flatpak. Also note that `appdata.xml` extension has been deprecated in favor of `metainfo.xml`, see https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html.

Yeah definitely, although I don't know if flathub requires up to date releases section, and I don't want to create extra churn to track it on xemu master branch if so

And appimage packing apparently expects appdata.xml. Since both are accepted generally, we'll just go with appdata.xml until appimage tools advance

j8r:
Where to host the screenshots? I think they can have their place on the website too, then we will be able to use their URL on the appdata file.

mborgerson:
xemu-website seems like a reasonable spot for the screenshots to live

PR #1357, opened 2023-01-18
Add screenshots to appdata file

j8r:
This file will then be used in https://github.com/flathub/app.xemu.xemu.
The empty spaces have been removed automatically by my editor.

### DIFF
--- a/xemu.appdata.xml
+++ b/xemu.appdata.xml
@@ -1,24 +1,33 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <component type="desktop-application">
   <id>app.xemu.xemu</id>
-  
+
   <name>xemu</name>
   <summary>Original Xbox Emulator</summary>
-  
+
   <metadata_license>CC0-1.0</metadata_license>
   <project_license>GPL-2.0-only</project_license>
-  
+
   <supports>
     <control>pointing</control>
     <control>keyboard</control>
     <control>gamepad</control>
   </supports>
-  
+
   <description>
     <p>
        A free and open-source application that emulates the original Microsoft Xbox game console. Supports connecting up to 4 controllers for local play, networking for multiplayer, resolution scaling, and more.
     </p>
   </description>
+
+  <screenshots>
+    <screenshot type="default"><image>https://xemu.app/screenshots/0.png</image></screenshot>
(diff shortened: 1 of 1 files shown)


xemu.appdata.xml
```diff
@@ -1,32 +1,40 @@
...
-  
+
   <description>
     <p>
        A free and open-source application that emulates the original Microsoft Xbox game console. Supports connecting up to 4 controllers for local play, networking for multiplayer, resolution scaling, and more.
     </p>
   </description>
 
+  <screenshots>
+    <screenshot type="default"><image>https://xemu.app/screenshots/5.png</image></screenshot>
```
mborgerson: Why screenshots 51234 and not 012345?

j8r: I missed the 0.
And for the 5th, it shows the emator's UI, I think it maybe better as a first screenshot.

j8r: I missed the 0.
And for the 5th, it shows the emulator's UI, I think it maybe better as a first screenshot.

mborgerson: A fair argument, but the core focus and most recognizable imagery will be on the Xbox software, especially the Xbox Dashboard. I think people care less about the emulator settings UI, so let's include it, but not first.

j8r: Right, the best would be to have a windowed screenshot, then we can see it with a game running, not just a capture.
It should show some part of the emulator UI in my opinion

j8r: The Xbox dashboard is good too, let's do that.

mborgerson: [AppStream policy](https://freedesktop.org/software/appstream/docs/chap-Quickstart.html) for full screen apps like games, and by extension emulators which are often run in full screen, is to not include window chrome so let's follow that guidance.

However I did include some window chrome on the xemu website to make it clear it's a screenshot from xemu and add some visual style.

mborgerson:
Thanks!

xemu.appdata.xml
j8r: The xemu top menu could be added? Then, we know it is not a game but inside an emulator, and it will still be fullscreen.

PR #1394, opened 2023-02-14
smc: Implement read/write of SMC error code storage

LoveMHz:
Implements the read/write logic for the SMC error code storage. As detailed on the xboxdevwiki, https://xboxdevwiki.net/PIC#Command_overview, these SMC registers are used to store error codes.

This seems mostly unused on retail kernels, but the SMC logic is still valid (read/write). Written values survive SMC reset and the initial uninitialized read value is always zero. Verified on hardware revisions 1.0 - 1.6


mborgerson:
Thanks!

PR #1403, opened 2023-03-02
RPM Package

### LINKED ISSUE #1012: RPM package
### Feature Request

Hello,

would it be possible to pack it in RPM format? We already have a DEB package and flatpak, so I don’t think RPM would be a problem.

Thank you

### Alternatives

_No response_

### Additional Context

_No response_

Fabxx:
Fixes #1012 
thanks to @SSUPII for help

-Creates a RPM package with github actions.
-Checks if a new version is available by parsing commit number
-Updates the field "Version" in the spec file automatically

### DIFF
--- a/RPM_package.yml
+++ b/RPM_package.yml
@@ -1,12 +1,16 @@
 name: Xemu RPM
-on: push
+on: 
+  push:
+    tags:
+      - v*
 
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-
+    - run:  bash RPM_check_version.sh
+    
     - name: xemu RPM package
       id: rpm
       uses: xemu-project/xemu
--- a/RPM_package.yml
+++ b/RPM_package.yml
@@ -1,0 +1,2 @@
+- name: RPM Build
+  uses: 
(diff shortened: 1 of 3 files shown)


cangarujack:
i know that this is wrong place for ask a request, but it's possible to add mouse support for GHOST SQUAD arcade?

RPM_package.yml
```diff
@@ -0,0 +1,24 @@
+name: Xemu RPM
```
antangelo, line 1: Why is this done separately instead of being part of the packaging steps in the main CI pipeline? This is also in the wrong directory in any case, so it won't run.

RPM_package.yml
```diff
@@ -0,0 +1,24 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - run:  bash RPM_check_version.sh
+    
+    - name: xemu RPM package
+      id: rpm
+      uses: xemu-project/xemu
+      with:
+        spec_file: "RPM_package.spec"
```
antangelo, line 18: How does this work? I don't think the `xemu-project/xemu` action exists

PR #1416, opened 2023-04-02
Set correct version for macOS bundles

### LINKED ISSUE #1344: Wrong version information in macOS download 0.7.77 (and before)
### Bug Description

download https://github.com/xemu-project/xemu/releases/download/v0.7.77/xemu-macos-universal-release.zip contains version 1 instead of 0.7.77 - the` info.plist` should contain the correct version in: 

```
  <key>CFBundleShortVersionString</key>
  <string>1</string>
…
  <key>CFBundleVersion</key>
  <string>1</string>
```

### Expected Behavior

```
  <key>CFBundleShortVersionString</key>
  <string>0.7.77</string>
…
  <key>CFBundleVersion</key>
  <string>0.7.77</string>
```

### xemu Version

0.7.77 

### System Information

MacOS 12.6

### Additional Context

_No response_

Florin9doi:
Closes #1344

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -77,6 +77,9 @@ package_macos() {
 
     cp Info.plist dist/xemu.app/Contents/
 
+    plutil -replace CFBundleShortVersionString -string $(cat ${project_source_dir}/XEMU_VERSION | cut -f1 -d-) dist/xemu.app/Contents/Info.plist
+    plutil -replace CFBundleVersion            -string $(cat ${project_source_dir}/XEMU_VERSION | cut -f1 -d-) dist/xemu.app/Contents/Info.plist
+
     codesign --force --deep --preserve-metadata=entitlements,requirements,flags,runtime --sign - "${exe_path}"
     python3 ./scripts/gen-license.py --version-file=macos-libs/$target_arch/INSTALLED > dist/LICENSE.txt
 }


.github/workflows/build.yml
```diff
@@ -270,6 +270,9 @@ jobs:
         echo -e "\nCompiler Cache Stats:"
         ccache -s
         pushd dist
+        export VERSION=$(cat ../XEMU_VERSION)
+        plutil -replace CFBundleShortVersionString -string $VERSION xemu.app/Contents/Info.plist
```
mborgerson: * Let's move this to the `build.sh` script, so those building from source will have proper version numbers too on their locally built bundle.
* According to the docs for [CFBundleShortVersionString](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleshortversionstring) and [CFBundleVersion](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleversion):
   > The required format is three period-separated integers, such as 10.14.1. The string can only contain numeric characters (0-9) and periods.

   Which works for xemu release versions which have this format, but will be incorrect for build artifacts and modified source builds which have a format that looks like `0.7.86-2-g2ca90d65f2`. You could just `cut -f1 -d-` the version string to make it compliant before updating the plist.

Florin9doi:
fixed

build.sh
```diff
@@ -77,6 +77,9 @@ package_macos() {
 
     cp Info.plist dist/xemu.app/Contents/
 
+    plutil -replace CFBundleShortVersionString -string $(cat ${project_source_dir}/XEMU_VERSION | cut -f1 -d-) dist/xemu.app/Contents/Info.plist
+    plutil -replace CFBundleVersion            -string $(cat ${project_source_dir}/XEMU_VERSION)               dist/xemu.app/Contents/Info.plist
```
mborgerson: We may have the wrong version format here:

> This key is a machine-readable string composed of one to three period-separated integers, such as 10.14.1. The string can only contain numeric characters (0-9) and periods. 
https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleversion

mborgerson:
Thanks for the contribution

PR #1447, opened 2023-06-04
ui: Add snapshot management UI cont'd

mborgerson:
antangelo's awesome PR #782 with a few patches


ui/xui/main-menu.cc
```diff
@@ -727,173 +709,177 @@ void MainMenuSnapshotsView::SnapshotBigButton(QEMUSnapshotInfo *snapshot, const
...
+    bool load = ImGui::Button("###button", ImVec2(-FLT_MIN, fmax(thumbnail_size.y + style.FramePadding.y * 2, ts_title.y + ts_sub.y + style.FramePadding.y * 3)));
     if (load) {
         xemu_snapshots_load(snapshot->name, &err);
         if (err) {
             xemu_queue_error_message(error_get_pretty(err));
             error_free(err);
         }
     }
+    bool options_key_pressed = ImGui::IsKeyPressed(ImGuiKey_GamepadFaceLeft);
+    bool show_snapshot_context_menu = ImGui::IsItemHovered() &&
```
antangelo: Should we have some sort if indicator that this context menu exists? Maybe an arrow pointing left could work, otherwise I don't know if someone using a controller will know that `GamepadFaceLeft` will open the context menu (or otherwise know that these buttons are right-clickable when I don't believe anything else in the UI is).

ui/xui/main-menu.cc
```diff
@@ -727,173 +709,177 @@ void MainMenuSnapshotsView::SnapshotBigButton(QEMUSnapshotInfo *snapshot, const
...
+        (ImGui::IsMouseReleased(ImGuiMouseButton_Right) ||
+         ImGui::IsKeyPressed(ImGuiKey_GamepadFaceLeft));
```
antangelo: Is there a reason this is distinct from `options_key_pressed`?

ui/xui/gl-helpers.hh
```diff
@@ -47,6 +47,8 @@ void RenderController(float frame_x, float frame_y, uint32_t primary_color,
                       uint32_t secondary_color, ControllerState *state);
 void RenderControllerPort(float frame_x, float frame_y, int i,
                           uint32_t port_color);
-void RenderFramebuffer(GLint tex, int width, int height, bool flip, bool apply_scaling_factor = true);
-bool ExtractFramebufferPixels(GLuint tex, bool flip, std::vector<uint8_t> &png, int width = 0, int height = 0);
+void RenderFramebuffer(GLint tex, int width, int height, bool flip);
+void RenderFramebuffer(GLint tex, int width, int height, bool flip, float scale[2]);
+bool RenderFramebufferToPng(GLuint tex, bool flip, std::vector<uint8_t> &png, int max_width = 0, int max_height = 0);
 void SaveScreenshot(GLuint tex, bool flip);
+void ScaleDimensions(int src_width, int src_height, int max_width, int max_height, int *out_width, int *out_height);
```
antangelo: nit: Missing newline

ui/xui/main-menu.cc
mborgerson: Totally agree. Removed the buttons to save screen space.

In the popup menu you'll see control annotations, which I want to surface on the main menu view as well. Still needs to be done and made context sensitive, which I'll save for follow up work

ui/xui/main-menu.cc
mborgerson: Nope

PR #1452, opened 2023-06-12
nv2a: Fix SET_ANTI_ALIASING_CONTROL

mborgerson:
- Rename from SET_SMOOTHING_CONTROL
- Use correct register


hw/xbox/nv2a/nv2a.c
```diff
@@ -517,7 +517,6 @@ static const VMStateDescription vmstate_nv2a = {
         VMSTATE_INT32_ARRAY(pgraph.gl_draw_arrays_start, NV2AState, 1250),
         VMSTATE_INT32_ARRAY(pgraph.gl_draw_arrays_count, NV2AState, 1250),
         VMSTATE_UINT32_ARRAY(pgraph.regs, NV2AState, 0x2000),
-        VMSTATE_BOOL(pgraph.smoothing_enabled, NV2AState),
```
antangelo, line 520: Should this warrant a snapshot version bump?

mborgerson: I bypassed the release for the commit that introduced this and now that it's removed it shouldn't have an impact

antangelo: Ah I didn't catch that it was just introduced, thanks for clarifying

PR #1467, opened 2023-06-29
UX: Add Overclock setting and Override CPU clock-speed setting

### LINKED ISSUE #1429: [Feature Request] virtual cpu clock control
### Feature Request

I was tweaking with xemu’s cpu clock speed and it acts like how other emulators do for games. Was able to gain performance lowering clock speed of virtual cpu on some games and worse on others and also caused more instability fps which is expected. It’s not ideal but would be interesting feature to add to ui or in the toml to change the clock speed of cpu (by percentage like 100% is default and above and lower is different) for variable refresh rate games by raising the clock speed and or games that might benefit from lowering cpu clock. 

It’s more of a hack but there is definitely people that would use it. If it was added I would not accept issue reports if it’s not default clock as it could have bugs on some games but also could activate frame skip for some games to run full speed for some.

this is how dolphin explains it. (Goes from 5% to 400% as a example which 400% would be 4x default essentially)
IMG_6937

### Alternatives

_No response_

### Additional Context

Using my steamdeck I am able to get 4x performance on menus (4fps before and getting 18 fps after) and 2x certain areas by lowering the cpu clock by 50%. Certain races it makes it playable. This is where I changed it. https://github.com/xemu-project/xemu/blob/9e1f5f2f19a8b0e880e034d455b359d426f6d386/hw/i386/x86.c#L526

And this is what I did to get 50% (I was testing beforehand to 200% so that why it’s not the same as default but if you do 366666666.5 x 2. It will equal default.) https://github.com/Spidy123222/xemu/pull/3/commits/a1a7c813c98cc443dba64327ee7be9817e55a065

Spidy123222:
Fixes https://github.com/xemu-project/xemu/issues/1429

This adds the ability to override the current clock speed and adjust it to help optimize a game. 

Changing this value like other emulators when increasing the value increases the clock-speed cpu timing which can benefit variable-refresh rate games or mods that require higher clocks and or lower can trigger a games frame skip feature if it has one to increase performance.

Adjusting clock-speed can sometimes cause issues in some games so is to be warned about doing it and maybe shouldn’t be used in compatibility reports.

Increasing clock-speed seems to get some games to allow to go over 60 fps which could be useful if the game properly handles it.

A game I find improvements on personally is in Burnout 3 turning it down makes it more usable in menus on steamdeck from 4fps to 23fps adjusting to 47% and a bit ingame. In Jet Set Radio future with a 10 fps increase in more demanding areas.

IMG_7386

### DIFF
--- a/config_spec.yml
+++ b/config_spec.yml
@@ -253,3 +253,9 @@ perf:
   cache_shaders:
     type: bool
     default: true
+  override_clockspeed: 
+    type: bool
+    default: false
+  cpu_clockspeed:
+    type: number
+    default: 0.5
--- a/config_spec.yml
+++ b/config_spec.yml
@@ -258,4 +258,4 @@ perf:
     default: false
   cpu_clockspeed:
     type: number
-    default: 0.5
+    default: 1
--- a/config_spec.yml
+++ b/config_spec.yml
@@ -256,6 +256,6 @@ perf:
   override_clockspeed: 
     type: bool
     default: false
-  cpu_clockspeed:
+  cpu_clockspeed_scale:
     type: number
     default: 1
--- a/hw/i386/x86.c
+++ b/hw/i386/x86.c
@@ -541,6 +541,7 @@ uint64_t cpu_get_tsc(CPUX86State *env)
     
     if (g_config.perf.override_clockspeed) {
         cpu_clock_hz *= clock_multiplier;
+        printf(clock_multiplier)
     }
     printf(cpu_clock_hz)
     return muldiv64(qemu_clock_get_ns(QEMU_CLOCK_VIRTUAL), cpu_clock_hz, NANOSECONDS_PER_SECOND);
--- a/hw/i386/x86.c
+++ b/hw/i386/x86.c
@@ -59,6 +59,9 @@
 #include CONFIG_DEVICES
 #include "kvm/kvm_i386.h"
 
+#include "ui/xemu-settings.h"
+
+
 /* Physical Address of PVH entry point read from kernel ELF NOTE */
 static size_t pvh_start_addr;
 
@@ -532,8 +535,18 @@ static long get_file_size(FILE *f)
 uint64_t cpu_get_tsc(CPUX86State *env)
 {
 #ifdef XBOX
-    return muldiv64(qemu_clock_get_ns(QEMU_CLOCK_VIRTUAL), 733333333,
-                    NANOSECONDS_PER_SECOND);
+    int DEFAULT_CPU_CLOCK = 733333333; /* 733333333 hz */
+    float clockOutput;
+    float OVERCLOCK_VALUE = g_config.perf.cpu_clockspeed; /* 0.5 is 100% */
+    float PERCENTAGE_OUTPUT = OVERCLOCK_VALUE * 2;
+    
+    if (g_config.perf.override_clockspeed) {
+        float clockOutput = DEFAULT_CPU_CLOCK * PERCENTAGE_OUTPUT;
+        return muldiv64(qemu_clock_get_ns(QEMU_CLOCK_VIRTUAL), clockOutput, NANOSECONDS_PER_SECOND);
+    } else {
+        float clockOutput = DEFAULT_CPU_CLOCK;
+        return muldiv64(qemu_clock_get_ns(QEMU_CLOCK_VIRTUAL), clockOutput, NANOSECONDS_PER_SECOND);
+    }
 #else
     return cpus_get_elapsed_ticks();
 #endif
--- a/hw/i386/x86.c
+++ b/hw/i386/x86.c
@@ -540,13 +540,11 @@ uint64_t cpu_get_tsc(CPUX86State *env)
     float OVERCLOCK_VALUE = g_config.perf.cpu_clockspeed; /* 0.5 is 100% */
     float PERCENTAGE_OUTPUT = OVERCLOCK_VALUE * 2;
     
+    float cpu_clock_hz = 733333333;
     if (g_config.perf.override_clockspeed) {
-        float clockOutput = DEFAULT_CPU_CLOCK * PERCENTAGE_OUTPUT;
-        return muldiv64(qemu_clock_get_ns(QEMU_CLOCK_VIRTUAL), clockOutput, NANOSECONDS_PER_SECOND);
-    } else {
-        float clockOutput = DEFAULT_CPU_CLOCK;
-        return muldiv64(qemu_clock_get_ns(QEMU_CLOCK_VIRTUAL), clockOutput, NANOSECONDS_PER_SECOND);
+      cpu_clock_hz *= clock_multiplier;
     }
+    return muldiv64(qemu_clock_get_ns(QEMU_CLOCK_VIRTUAL), cpu_clock_hz, NANOSECONDS_PER_SECOND);
 #else
     return cpus_get_elapsed_ticks();
 #endif
--- a/hw/i386/x86.c
+++ b/hw/i386/x86.c
@@ -541,9 +541,9 @@ uint64_t cpu_get_tsc(CPUX86State *env)
     
     if (g_config.perf.override_clockspeed) {
         cpu_clock_hz *= clock_multiplier;
-        printf(clock_multiplier)
+        printf(clock_multiplier);
     }
-    printf(cpu_clock_hz)
+    printf(cpu_clock_hz);
     return muldiv64(qemu_clock_get_ns(QEMU_CLOCK_VIRTUAL), cpu_clock_hz, NANOSECONDS_PER_SECOND);
 #else
     return cpus_get_elapsed_ticks();
--- a/hw/i386/x86.c
+++ b/hw/i386/x86.c
@@ -536,8 +536,8 @@ static long get_file_size(FILE *f)
 uint64_t cpu_get_tsc(CPUX86State *env)
 {
 #ifdef XBOX
-    float clock_multiplier = g_config.perf.cpu_clockspeed * 2.0;
-    float cpu_clock_hz = 733333333.0;
+    float clock_multiplier = g_config.perf.cpu_clockspeed;
+    int cpu_clock_hz = 733333333;
     
     if (g_config.perf.override_clockspeed) {
         cpu_clock_hz *= clock_multiplier;
--- a/hw/i386/x86.c
+++ b/hw/i386/x86.c
@@ -59,8 +59,9 @@
 #include CONFIG_DEVICES
 #include "kvm/kvm_i386.h"
 
+#ifdef XBOX
 #include "ui/xemu-settings.h"
-
+#endif
 
 /* Physical Address of PVH entry point read from kernel ELF NOTE */
 static size_t pvh_start_addr;
@@ -535,14 +536,12 @@ static long get_file_size(FILE *f)
 uint64_t cpu_get_tsc(CPUX86State *env)
 {
 #ifdef XBOX
-    int DEFAULT_CPU_CLOCK = 733333333; /* 733333333 hz */
-    float clockOutput;
-    float OVERCLOCK_VALUE = g_config.perf.cpu_clockspeed; /* 0.5 is 100% */
-    float PERCENTAGE_OUTPUT = OVERCLOCK_VALUE * 2;
-    
+    float clockOutput; 
+    float clock_multiplier = g_config.perf.cpu_clockspeed * 2;
     float cpu_clock_hz = 733333333;
+    
     if (g_config.perf.override_clockspeed) {
-      cpu_clock_hz *= clock_multiplier;
+        cpu_clock_hz *= clock_multiplier;
     }
     return muldiv64(qemu_clock_get_ns(QEMU_CLOCK_VIRTUAL), cpu_clock_hz, NANOSECONDS_PER_SECOND);
 #else
--- a/hw/i386/x86.c
+++ b/hw/i386/x86.c
@@ -541,9 +541,7 @@ uint64_t cpu_get_tsc(CPUX86State *env)
     
     if (g_config.perf.override_clockspeed) {
         cpu_clock_hz *= clock_multiplier;
-        printf(clock_multiplier);
     }
-    printf(cpu_clock_hz);
     return muldiv64(qemu_clock_get_ns(QEMU_CLOCK_VIRTUAL), cpu_clock_hz, NANOSECONDS_PER_SECOND);
 #else
     return cpus_get_elapsed_ticks();
--- a/hw/i386/x86.c
+++ b/hw/i386/x86.c
@@ -536,7 +536,6 @@ static long get_file_size(FILE *f)
 uint64_t cpu_get_tsc(CPUX86State *env)
 {
 #ifdef XBOX
-    float clockOutput; 
     float clock_multiplier = g_config.perf.cpu_clockspeed * 2;
     float cpu_clock_hz = 733333333;
     
--- a/hw/i386/x86.c
+++ b/hw/i386/x86.c
@@ -536,11 +536,10 @@ static long get_file_size(FILE *f)
 uint64_t cpu_get_tsc(CPUX86State *env)
 {
 #ifdef XBOX
-    float clock_multiplier = g_config.perf.cpu_clockspeed;
     int cpu_clock_hz = 733333333;
     
     if (g_config.perf.override_clockspeed) {
-        cpu_clock_hz *= clock_multiplier;
+        cpu_clock_hz *= g_config.perf.cpu_clockspeed_scale;
     }
     return muldiv64(qemu_clock_get_ns(QEMU_CLOCK_VIRTUAL), cpu_clock_hz, NANOSECONDS_PER_SECOND);
 #else
--- a/hw/i386/x86.c
+++ b/hw/i386/x86.c
@@ -536,12 +536,13 @@ static long get_file_size(FILE *f)
 uint64_t cpu_get_tsc(CPUX86State *env)
 {
 #ifdef XBOX
-    float clock_multiplier = g_config.perf.cpu_clockspeed * 2;
-    float cpu_clock_hz = 733333333;
+    float clock_multiplier = g_config.perf.cpu_clockspeed * 2.0;
+    float cpu_clock_hz = 733333333.0;
     
     if (g_config.perf.override_clockspeed) {
         cpu_clock_hz *= clock_multiplier;
     }
+    printf(cpu_clock_hz)
     return muldiv64(qemu_clock_get_ns(QEMU_CLOCK_VIRTUAL), cpu_clock_hz, NANOSECONDS_PER_SECOND);
 #else
     return cpus_get_elapsed_ticks();
--- a/ui/xui/compat.cc
+++ b/ui/xui/compat.cc
@@ -218,9 +218,8 @@ void CompatibilityReporter::Draw()
             is_open = false;
         }
     }
-    if (g_config.perf.override_clockspeed)
-    {
-        ImGui::Text("Reports using Emulated CPU clock override is not allowed.");
+    if (g_config.perf.override_clockspeed) {
+        ImGui::Text("Reports cannot be made while using altered CPU clock speeds");
         ImGui::PopItemFlag();
         ImGui::PopStyleVar();
     }
--- a/ui/xui/compat.cc
+++ b/ui/xui/compat.cc
@@ -206,8 +206,7 @@ void CompatibilityReporter::Draw()
     ImGui::SetCursorPosX(ImGui::GetWindowWidth()-(120+10)*g_viewport_mgr.m_scale);
 
     ImGui::SetItemDefaultFocus();
-    if (g_config.perf.override_clockspeed)
-    {
+    if (g_config.perf.override_clockspeed) {
         ImGui::PushItemFlag(ImGuiItemFlags_Disabled, true);
         ImGui::PushStyleVar(ImGuiStyleVar_Alpha, ImGui::GetStyle().Alpha * 0.5f);
     }
--- a/ui/xui/compat.cc
+++ b/ui/xui/compat.cc
@@ -203,13 +203,17 @@ void CompatibilityReporter::Draw()
         ImGui::SameLine();
     }
 
-    ImGui::SetCursorPosX(ImGui::GetWindowWidth()-(120+10)*g_viewport_mgr.m_scale);
-
-    ImGui::SetItemDefaultFocus();
     if (g_config.perf.override_clockspeed) {
         ImGui::PushItemFlag(ImGuiItemFlags_Disabled, true);
         ImGui::PushStyleVar(ImGuiStyleVar_Alpha, ImGui::GetStyle().Alpha * 0.5f);
+
+        ImGui::Text("Reports cannot be made while using altered CPU clock speeds");
+        ImGui::SameLine();
     }
+
+    ImGui::SetCursorPosX(ImGui::GetWindowWidth()-(120+10)*g_viewport_mgr.m_scale);
+
+    ImGui::SetItemDefaultFocus();
     if (ImGui::Button("Send", ImVec2(120*g_viewport_mgr.m_scale, 0))) {
         did_send = true;
         send_result = report.Send();
--- a/ui/xui/compat.cc
+++ b/ui/xui/compat.cc
@@ -206,6 +206,11 @@ void CompatibilityReporter::Draw()
     ImGui::SetCursorPosX(ImGui::GetWindowWidth()-(120+10)*g_viewport_mgr.m_scale);
 
     ImGui::SetItemDefaultFocus();
+    if (g_config.perf.override_clockspeed)
+    {
+        ImGui::PushItemFlag(ImGuiItemFlags_Disabled, true);
+        ImGui::PushStyleVar(ImGuiStyleVar_Alpha, ImGui::GetStyle().Alpha * 0.5f);
+    }
     if (ImGui::Button("Send", ImVec2(120*g_viewport_mgr.m_scale, 0))) {
         did_send = true;
         send_result = report.Send();
@@ -213,7 +218,12 @@ void CompatibilityReporter::Draw()
             is_open = false;
         }
     }
-
+    if (g_config.perf.override_clockspeed)
+    {
+        ImGui::Text("Reports using Emulated CPU clock override is not allowed.");
+        ImGui::PopItemFlag();
+        ImGui::PopStyleVar();
+    }
     ImGui::End();
 }
 
--- a/ui/xui/compat.cc
+++ b/ui/xui/compat.cc
@@ -222,7 +222,6 @@ void CompatibilityReporter::Draw()
         }
     }
     if (g_config.perf.override_clockspeed) {
-        ImGui::Text("Reports cannot be made while using altered CPU clock speeds");
         ImGui::PopItemFlag();
         ImGui::PopStyleVar();
     }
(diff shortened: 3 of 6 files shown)


hw/i386/x86.c
```diff
@@ -56,6 +56,9 @@
 #include CONFIG_DEVICES
 #include "kvm/kvm_i386.h"
 
+#include "ui/xemu-settings.h"
+
```
antangelo: nit: Extra newline. Also should enclose this in a `#ifdef XBOX` guard

hw/i386/x86.c
```diff
@@ -523,8 +526,18 @@ static long get_file_size(FILE *f)
 uint64_t cpu_get_tsc(CPUX86State *env)
 {
 #ifdef XBOX
-    return muldiv64(qemu_clock_get_ns(QEMU_CLOCK_VIRTUAL), 733333333,
-                    NANOSECONDS_PER_SECOND);
+    int DEFAULT_CPU_CLOCK = 733333333; /* 733333333 hz */
+    float clockOutput;
```
antangelo: nit: Unused variable

hw/i386/x86.c
```diff
@@ -523,8 +526,18 @@ static long get_file_size(FILE *f)
...
+    float OVERCLOCK_VALUE = g_config.perf.cpu_clockspeed; /* 0.5 is 100% */
+    float PERCENTAGE_OUTPUT = OVERCLOCK_VALUE * 2;
```
antangelo: You can just combine these into one variable, there's no point in separating them. Also the comment here adds little value, you can just name it accordingly (e.g. `clock_multiplier`).
Also note that the case is incorrect, xemu uses snake_case for variables.

hw/i386/x86.c
```diff
@@ -523,8 +526,18 @@ static long get_file_size(FILE *f)
 uint64_t cpu_get_tsc(CPUX86State *env)
 {
 #ifdef XBOX
-    return muldiv64(qemu_clock_get_ns(QEMU_CLOCK_VIRTUAL), 733333333,
-                    NANOSECONDS_PER_SECOND);
+    int DEFAULT_CPU_CLOCK = 733333333; /* 733333333 hz */
```
antangelo: This variable (and comment) is unnecessary, you can just use the constant directly (see below for why this is the case).

hw/i386/x86.c
```diff
@@ -523,8 +526,18 @@ static long get_file_size(FILE *f)
...
+    float OVERCLOCK_VALUE = g_config.perf.cpu_clockspeed; /* 0.5 is 100% */
+    float PERCENTAGE_OUTPUT = OVERCLOCK_VALUE * 2;
+    
+    if (g_config.perf.override_clockspeed) {
+        float clockOutput = DEFAULT_CPU_CLOCK * PERCENTAGE_OUTPUT;
+        return muldiv64(qemu_clock_get_ns(QEMU_CLOCK_VIRTUAL), clockOutput, NANOSECONDS_PER_SECOND);
+    } else {
+        float clockOutput = DEFAULT_CPU_CLOCK;
+        return muldiv64(qemu_clock_get_ns(QEMU_CLOCK_VIRTUAL), clockOutput, NANOSECONDS_PER_SECOND);
+    }
```
antangelo, line 543: Let's refactor this to remove duplicate code

```suggestion
    float cpu_clock_hz = 733333333;
    if (g_config.perf.override_clockspeed) {
      cpu_clock_hz *= clock_multiplier;
    }
    return muldiv64(qemu_clock_get_ns(QEMU_CLOCK_VIRTUAL), cpu_clock_hz, NANOSECONDS_PER_SECOND);
```

ui/xui/main-menu.cc
```diff
@@ -62,6 +62,15 @@ void MainMenuGeneralView::Draw()
     Toggle("Cache shaders to disk", &g_config.perf.cache_shaders,
            "Reduce stutter in games by caching previously generated shaders");
 
+    Toggle("Emulated CPU clock override", &g_config.perf.override_clockspeed,
+           "Enables to override default CPU clock speed");
+    
+    char buf[32];
+    snprintf(buf, sizeof(buf), "Clock Speed (%d%%)",
+             (int)(g_config.perf.cpu_clockspeed * 200));
+    Slider("Virtual CPU clock", &g_config.perf.cpu_clockspeed, buf);
```
antangelo: I don't like how the clockspeed needs to be scaled by 2. I think we should add an overload to Slider to be able to take a min/max threshold such that the clockspeed config entry is 1.0 at the default clock speed.

ui/xui/main-menu.cc
```diff
@@ -62,6 +62,15 @@ void MainMenuGeneralView::Draw()
     Toggle("Cache shaders to disk", &g_config.perf.cache_shaders,
            "Reduce stutter in games by caching previously generated shaders");
 
+    Toggle("Emulated CPU clock override", &g_config.perf.override_clockspeed,
+           "Enables to override default CPU clock speed");
+    
+    char buf[32];
+    snprintf(buf, sizeof(buf), "Clock Speed (%d%%)",
```
antangelo: Would be nice to also display the actual clock speed in Hz alongside the percentage

ui/xui/main-menu.cc
```diff
@@ -62,6 +62,15 @@ void MainMenuGeneralView::Draw()
     Toggle("Cache shaders to disk", &g_config.perf.cache_shaders,
            "Reduce stutter in games by caching previously generated shaders");
 
+    Toggle("Emulated CPU clock override", &g_config.perf.override_clockspeed,
```
antangelo, line 69: I don't think having the toggle and slider out at the same time is the best choice. Some ideas:

- Remove the toggle, with 0.5 being the same as disabling. Probably need a reset button for accuracy.
- Toggle can hide the slider

antangelo (commented):
We should prevent compatibility reports from being submitted if the clock speed is not stock.

Any fixes to games as a result of adjusting clock speeds are (most probably) hacks. I imagine the only non-hacky utility this change adds is in testing CPU overclocking and replacements.

Spidy123222:
> We should prevent compatibility reports from being submitted if the clock speed is not stock.
> 

I am not quite sure how to do that as i am new to this codebase and also to C. But to be able to know it's enabled is as simple as using the override toggle's state as it's required for it to be enabled for it to work.

> Any fixes to games as a result of adjusting clock speeds are (most probably) hacks. I imagine the only non-hacky utility this change adds is in testing CPU overclocking and replacements. 

Yeah that's reasonable and makes sense as the report system is meant for mimicking the actual hardware as close as possible but having this can help optimize game and test overclocks also. On dolphin's one there is more detail on it about not reporting compatibility using their enabled. 

I couldn't fit more text than a few words so I would of added more detail on how it works but that's not entirely important to have and it can be on the site.

antangelo:
See `ui/xui/reporting.cc` and `ui/xui/compat.cc` for the reporting system.

My main point there is that "optimizing" games by altering the emulated clock speed is not correct. Issues should be resolved by moving the emulation closer to that of original hardware, not farther.

ui/xui/main-menu.cc
Spidy123222: I’m not fond of it either but I couldn’t figure how to solve that really.

ui/xui/main-menu.cc
Spidy123222: Agreed but how could that be displayed as my original intention was to also have that.

ui/xui/main-menu.cc
Spidy123222: I feel having them both open at same time can make it so you can disable the toggle and able to change clock without doing it live. Hiding if it’s 100% would cause glitching if it’s hiding it when it’s there. Not quite sure that’s the best approach.

A reset button would be nice true. Likely all sliders should have one tbh.

hw/i386/x86.c
Spidy123222: Compiler was complaining that it wasn't defined being first time as it called it and putting it outside of the if statement is what fixed it

hw/i386/x86.c
Spidy123222: Latest commits should solve it

hw/i386/x86.c
Spidy123222: Latest commits should solve it

hw/i386/x86.c
Spidy123222: Latest commit should solve this nit

hw/i386/x86.c
Spidy123222: Removed in latest commits

ui/xui/main-menu.cc
BiatuAutMiahn: Just a thought, why not have a drop down with the format 0.5x (366MHz), 1.0x (733MHz), ...?

Spidy123222: Because some games sometimes work well in a specific percentage making premade values making it difficult if you do one for every value. If you reference dolphin it's this way for that reason also.

ui/xui/main-menu.cc
Spidy123222: Managed to get it displayed in latest commit!

ui/xui/main-menu.cc
Spidy123222: With the latest commit I made, the slider will snap if it gets very close to 100% to make it snap to the correct frequency which makes it back to default without a problem. I think with this approach it makes it good enough to not need a reset button and it actually works better on gamepads than before.

ui/xui/main-menu.cc
Spidy123222: > I don't like how the clockspeed needs to be scaled by 2. I think we should add an overload to Slider to be able to take a min/max threshold such that the clockspeed config entry is 1.0 at the default clock speed.

I figured out what I did wrong and was able to fix it and slider system is identical to audio and no longer multiplied by 2 in latest commit. I like the threshold thing but I think it's a bit out of my skill range for now and besides I solved that as it was me just not understanding how the sliders worked before.

binarymaster:
Would this setting be affected by a request coming from the guest code / software side, such as https://github.com/GXTX/XboxOverclock ?

Spidy123222:
> Would this setting be affected by a request coming from the guest code / software side, such as https://github.com/GXTX/XboxOverclock ?

I haven’t tried it but you can try it out. It might just default back to what xemu says it should be.

config_spec.yml
```diff
@@ -206,3 +206,9 @@ perf:
   cache_shaders:
     type: bool
     default: true
+  override_clockspeed: 
+    type: bool
+    default: false
+  cpu_clockspeed:
+    type: number
+    default: 0.5
```
antangelo: Default should now be 1

ui/xui/compat.cc
```diff
@@ -206,14 +206,24 @@ void CompatibilityReporter::Draw()
     ImGui::SetCursorPosX(ImGui::GetWindowWidth()-(120+10)*g_viewport_mgr.m_scale);
 
     ImGui::SetItemDefaultFocus();
+    if (g_config.perf.override_clockspeed)
+    {
```
antangelo: nit: Brace style should be on the same line (here and below)

ui/xui/main-menu.cc
Spidy123222: Should be fixed by latest commit by doing overloading like mentioned

ui/xui/compat.cc
```diff
@@ -206,14 +206,24 @@ void CompatibilityReporter::Draw()
...
         did_send = true;
         send_result = report.Send();
         if (send_result) {
             is_open = false;
         }
     }
-
+    if (g_config.perf.override_clockspeed)
+    {
+        ImGui::Text("Reports using Emulated CPU clock override is not allowed.");
```
antangelo: Change wording to "Reports cannot be made while using altered CPU clock speeds" with no period at the end.

ui/xui/main-menu.cc
```diff
@@ -62,6 +62,20 @@ void MainMenuGeneralView::Draw()
...
+           "Enables to override default CPU clock speed");
+    
+    char buf[32];
+    float min = 0.01;
+    float max = 2;
+    float gpspeed = 0.01;
```
antangelo: Pass these as arguments directly instead of defining constants

ui/xui/main-menu.cc
```diff
@@ -354,9 +368,12 @@ void MainMenuAudioView::Draw()
 {
     SectionTitle("Volume");
     char buf[32];
+    float min = 0;
+    float max = 1;
+    float gpspeed = 0.05;
     snprintf(buf, sizeof(buf), "Limit output volume (%d%%)",
              (int)(g_config.audio.volume_limit * 100));
-    Slider("Output volume limit", &g_config.audio.volume_limit, buf);
+    Slider("Output volume limit", &g_config.audio.volume_limit, min, max, gpspeed, buf);
```
antangelo: Instead of passing these values here, let's overload the Slider function so that the defaults are `min = 0.f`, `max = 1.f`, and `step = 0.05f`

ui/xui/widgets.cc
```diff
@@ -222,8 +222,9 @@ bool Toggle(const char *str_id, bool *v, const char *description)
     return status;
 }
 
-void Slider(const char *str_id, float *v, const char *description)
+void Slider(const char *str_id, float *v, float min, float max, float gpspeed, const char *description)
```
antangelo: I don't think `gpspeed` is a great name for this parameter, maybe something like `increment` or `step` is more clear

ui/xui/widgets.cc
```diff
@@ -222,8 +222,9 @@ bool Toggle(const char *str_id, bool *v, const char *description)
...
 {
+    float x = *v / max;
```
antangelo: This is correct only if `min = 0`. The general formula is `x = (*v - min) / (max - min)` for `*v` in the interval [min, max].
We should clamp `*v` to this interval beforehand in case it isn't.

ui/xui/widgets.cc
```diff
@@ -283,13 +284,14 @@ void Slider(const char *str_id, float *v, const char *description)
...
-
+    
     if (ImGui::IsItemActive()) {
         ImVec2 mouse = ImGui::GetMousePos();
-        *v = GetSliderValueForMousePos(mouse, slider_pos, slider_size);
+        x = GetSliderValueForMousePos(mouse, slider_pos, slider_size);
     }
-    *v = fmax(0, fmin(*v, 1));
-    DrawSlider(*v, ImGui::IsItemHovered() || ImGui::IsItemActive(), slider_pos,
+    x = fmax(0, fmin(x, 1));
```
antangelo: We can clean this up a bit using C++17, include `<algorithm>` and then use `x = std::clamp(x, 0.f, 1.f)`

ui/xui/widgets.cc
```diff
@@ -283,13 +284,14 @@ void Slider(const char *str_id, float *v, const char *description)
...
+    *v = fmax(min,fmin(x * max,max));
```
antangelo: This line should be `*v = x * (max - min) + min`, no need to clamp as we've already clamped x to the interval [0, 1] above.

ui/xui/widgets.hh
```diff
@@ -34,7 +34,7 @@ float GetSliderValueForMousePos(ImVec2 mouse, ImVec2 pos, ImVec2 size);
 void DrawSlider(float v, bool hovered, ImVec2 pos, ImVec2 size);
 void DrawToggle(bool enabled, bool hovered, ImVec2 pos, ImVec2 size);
 bool Toggle(const char *str_id, bool *v, const char *description = nullptr);
-void Slider(const char *str_id, float *v, const char *description = nullptr);
+void Slider(const char *str_id, float *v, float min, float max, float gpspeed, const char *description = nullptr);
```
antangelo: You can set defaults for `min, max, gpspeed` here in a similar way to how `description` defaults to `nullptr`. You may want to change the order of these parameters, because if you want to provide `description` then you will have to provide `min, max, gpspeed` regardless of their defaults due to limitations of the language.

ui/xui/compat.cc
Spidy123222: Has been changed in latest commit.

ui/xui/main-menu.cc
Spidy123222: Thanks for the help on this. I have tried doing that before but managed to finally get it to do it that way.

ui/xui/widgets.cc
Spidy123222: I changed the name to increment.

ui/xui/main-menu.cc
Spidy123222: Simple fix moved to argument.

ui/xui/widgets.cc
Spidy123222: I was kind of worried that it would happen but didn’t think of a different equation and thought it would be fine. This helps and definitely works better as there’s no dead zone at beginning if minimum value was high enough.

ui/xui/widgets.cc
Spidy123222: Latest commit should solve this

ui/xui/widgets.cc
Spidy123222: Yeah that’s defiantly more clean and I’ll use that! It’s in latest commit now.

ui/xui/widgets.hh
Spidy123222: They have been set now. Had trouble in past but was more pointer issues than anything so since that’s gone this started working so added it back which I didn’t think it would but it did. Thanks for telling this as I preferred having a default value.

ui/xui/compat.cc
Spidy123222: Fixed nit in latest commit.

ui/xui/main-menu.cc
```diff
@@ -62,6 +62,15 @@ void MainMenuGeneralView::Draw()
...
            "Reduce stutter in games by caching previously generated shaders");
 
+    Toggle("Emulated CPU clock override", &g_config.perf.override_clockspeed,
+           "Enables to override default CPU clock speed");
+    
+    char buf[32];
+    snprintf(buf, sizeof(buf), "Clock Speed %d%% (%.2f MHz)", (int)(g_config.perf.cpu_clockspeed * 100), (733333333 * g_config.perf.cpu_clockspeed)/1000000);
+    Slider("Virtual CPU clock", &g_config.perf.cpu_clockspeed, buf,0.01f , 2.f, 0.01f);
+
+    if ((g_config.perf.cpu_clockspeed-0.999)*(g_config.perf.cpu_clockspeed-1.009) <= 0) {g_config.perf.cpu_clockspeed = 1;}
```
antangelo: What's the point of this?

ui/xui/main-menu.cc
```diff
@@ -62,6 +62,15 @@ void MainMenuGeneralView::Draw()
     Toggle("Cache shaders to disk", &g_config.perf.cache_shaders,
            "Reduce stutter in games by caching previously generated shaders");
 
+    Toggle("Emulated CPU clock override", &g_config.perf.override_clockspeed,
+           "Enables to override default CPU clock speed");
+    
+    char buf[32];
+    snprintf(buf, sizeof(buf), "Clock Speed %d%% (%.2f MHz)", (int)(g_config.perf.cpu_clockspeed * 100), (733333333 * g_config.perf.cpu_clockspeed)/1000000);
+    Slider("Virtual CPU clock", &g_config.perf.cpu_clockspeed, buf,0.01f , 2.f, 0.01f);
```
antangelo: nit: Spacing on arguments doesn't follow style of surrounding code

ui/xui/widgets.cc
```diff
@@ -283,13 +284,14 @@ void Slider(const char *str_id, float *v, const char *description)
             ImGui::NavMoveRequestCancel();
         }
     }
-
+
```
antangelo: nit: Spacing

ui/xui/main-menu.cc
Spidy123222: To keep accuracy when your at 100% that it will be 733.33….MHz. I feel it was a better alternative than doing a reset button. It will snap to that value making it easy to go back to default.

ui/xui/main-menu.cc
Spidy123222: Idk it looks like how audio portion looks like.

antangelo: This part: `buf,0.01f ,`

ui/xui/main-menu.cc
antangelo: Can you make this a bit more clear? Maybe change the check to `std::abs(g_config.perf...- 1.f) <= 0.005f` and fix the brace formatting.
Also, should make sure this can be escaped with a controller, as if the increment is too low you will be stuck at 1.

ui/xui/main-menu.cc
binarymaster: Just use a code suggestion:
```suggestion
    Slider("Virtual CPU clock", &g_config.perf.cpu_clockspeed, buf, 0.01f, 2.f, 0.01f);
```

Spidy123222: > This part: `buf,0.01f ,`

Oh that makes it more clear. Sure thing 👍

ui/xui/widgets.cc
Spidy123222: Fixed in latest commit and I also rebased.

ui/xui/main-menu.cc
Spidy123222: > Can you make this a bit more clear? Maybe change the check to `std::abs(g_config.perf...- 1.f) <= 0.005f` and fix the brace formatting.
> Also, should make sure this can be escaped with a controller, as if the increment is too low you will be stuck at 1.

So I managed to get abs working in some form at some point. Using fabs seemed to work better for this and now is working. It’s in latest commit and I rebased.

antangelo: Just curious, in what way did `fabs` work better? They generate the same code on my machine.

There's a minor edge case here in that if the config is edited manually to set the clock override to a value within this interval, that value will be overwritten to snap back to 1 despite the user not touching the slider (and it will not snap to one until this screen is opened). I personally don't think this is worth the complexity to solve though.

ui/xui/compat.cc
```diff
@@ -206,14 +206,22 @@ void CompatibilityReporter::Draw()
     ImGui::SetCursorPosX(ImGui::GetWindowWidth()-(120+10)*g_viewport_mgr.m_scale);
 
     ImGui::SetItemDefaultFocus();
+    if (g_config.perf.override_clockspeed) {
+        ImGui::PushItemFlag(ImGuiItemFlags_Disabled, true);
+        ImGui::PushStyleVar(ImGuiStyleVar_Alpha, ImGui::GetStyle().Alpha * 0.5f);
+    }
```
antangelo: Small change so that the text and button are aligned:

```suggestion
    if (g_config.perf.override_clockspeed) {
        ImGui::PushItemFlag(ImGuiItemFlags_Disabled, true);
        ImGui::PushStyleVar(ImGuiStyleVar_Alpha, ImGui::GetStyle().Alpha * 0.5f);

        ImGui::Text("Reports cannot be made while using altered CPU clock speeds");
        ImGui::SameLine();
    }

    ImGui::SetCursorPosX(ImGui::GetWindowWidth()-(120+10)*g_viewport_mgr.m_scale);

    ImGui::SetItemDefaultFocus();
```

and then remove the `ImGui::Text(...)` on line 221.

ui/xui/main-menu.cc
Spidy123222: > Just curious, in what way did `fabs` work better? They generate the same code on my machine.
> 
> There's a minor edge case here in that if the config is edited manually to set the clock override to a value within this interval, that value will be overwritten to snap back to 1 despite the user not touching the slider (and it will not snap to one until this screen is opened). I personally don't think this is worth the complexity to solve though.

So when I was testing using normal abs it would some reason not allow to remove slider but changing to fabs worked. Maybe could of been a bug during compile, Not sure. Yeah doing it this way does have effect that it will do it directly to the value.

hw/i386/x86.c
```diff
@@ -532,8 +536,13 @@ static long get_file_size(FILE *f)
 uint64_t cpu_get_tsc(CPUX86State *env)
 {
 #ifdef XBOX
-    return muldiv64(qemu_clock_get_ns(QEMU_CLOCK_VIRTUAL), 733333333,
-                    NANOSECONDS_PER_SECOND);
+    float clock_multiplier = g_config.perf.cpu_clockspeed;
+    int cpu_clock_hz = 733333333;
+    
+    if (g_config.perf.override_clockspeed) {
+        cpu_clock_hz *= clock_multiplier;
```
antangelo: nit: `clock_multiplier` doesn't need to be a variable, can just replace it with the config value cast to float

ui/xui/widgets.cc
```diff
@@ -222,8 +223,8 @@ bool Toggle(const char *str_id, bool *v, const char *description)
     return status;
 }
 
-void Slider(const char *str_id, float *v, const char *description)
-{
+void Slider(const char *str_id, float *v, const char *description, float min, float max, float increment) {
```
antangelo: Functions should have opening brace on the next line

config_spec.yml
```diff
@@ -206,3 +206,9 @@ perf:
   cache_shaders:
     type: bool
     default: true
+  override_clockspeed: 
+    type: bool
+    default: false
+  cpu_clockspeed:
```
antangelo: nit: Would recommend renaming to `cpu_clockspeed_multiplier` or `cpu_clockspeed_scale` to reflect that this is a multiplier as opposed to the raw value

ui/xui/main-menu.cc
```diff
@@ -62,6 +62,17 @@ void MainMenuGeneralView::Draw()
     Toggle("Cache shaders to disk", &g_config.perf.cache_shaders,
            "Reduce stutter in games by caching previously generated shaders");
 
+    Toggle("Emulated CPU clock override", &g_config.perf.override_clockspeed,
+           "Enables to override default CPU clock speed");
+    
+    char buf[32];
+    snprintf(buf, sizeof(buf), "Clock Speed %d%% (%.2f MHz)", (int)(g_config.perf.cpu_clockspeed * 100), (733333333 * g_config.perf.cpu_clockspeed)/1000000);
```
antangelo: nit: Inconsistent spacing between the `/` operands in the last argument and the `*` operands. Either have a space around the `/` or no space around the `*`s.

ui/xui/widgets.cc
Spidy123222: Ah sorry I will fix that. I’ve been doing swift lately which has functions formatted this way.

hw/i386/x86.c
Spidy123222: Fixed in latest commit

ui/xui/widgets.cc
Spidy123222: Fixed in latest commit

ui/xui/main-menu.cc
Spidy123222: Fixed in latest commit. I decided to space it out since nearby stuff were similar and to keep formatting the same.

config_spec.yml
Spidy123222: Fixed in latest commit

antangelo (dismissed):

pmsobrado:
Was this ever implemented, or is this still active? Any current way to increase CPU?

Spidy123222:
> Was this ever implemented, or is this still active? Any current way to increase CPU?

This was never merged which is why it's open.

pmsobrado:
Sure, I'll wait patiently then :)

PR #1510, opened 2023-07-29
ci: Add clang-format lint check

antangelo:
Checks formatting and lints for PR changes in CI, using this github action: https://github.com/marketplace/actions/c-c-linter

I left the default clang-tidy checks enabled, if this is undesirable they can be set to read from a .clang-tidy file, although there isn't one in the repo currently.

I've also set it to comment on the pull request thread with the results, so that they are more visible. However, this does introduce noise to the PR thread so I'm open to disabling it. In either case, it will comment on the most recent commit with the results.

Unfortunately there aren't many actions available to run clang-format on only the files changed in a PR from what I can find, and this one does not indicate the exact formatting errors that are found (only that the files are not formatted correctly).

A dummy PR with failing lints is available here: https://github.com/antangelo/xemu/pull/3

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,36 @@
+name: Lint
+
+on:
+  pull_request:
+    paths-ignore:
+      - '.github/**'
+      - '!.github/workflows/lint.yml'
+      - 'README.md'
+      - 'ubuntu-win64-cross/**'
+
+jobs:
+  Check:
+    name: Check formatting and lints
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+    steps:
+      - name: Clone tree
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Check formatting/lints
+        uses: cpp-linter/cpp-linter-action@v2.13.4
+        id: linter
+        env:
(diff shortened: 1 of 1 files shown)


mborgerson:
On first look I didn't like the reporting of the tool much (e.g. reporting failure without saying what specifically was wrong, soliciting feedback, etc). Apparently there is a feature now that can leave detailed code review comments at https://cpp-linter.github.io/cpp-linter-action/#only-clang-format. What do you think?

antangelo:
That looks cool, I've updated this patch and the test PR to use that flag instead of the thread comment. I did notice advice against using both the format review and tidy review at the same time in their docs (https://cpp-linter.github.io/cpp-linter-action/pr-review-caveats/#posts-a-new-review-on-each-run), which may hinder extension to clang-tidy if we choose to use it at some point in the future, but those issues can be examined later (if any).

I initially didn't think having the suggestions in the PR were particularly necessary, since it's relatively simple to run `git clang-format ...` on a patch to have it fix issues automatically, but it seems in practice we have PRs where clang-format is run on entire files, and formatting changes are intermixed with functional changes (which is undesirable).

mborgerson:
I missed your update! Looks cool. Rebased and pinned the actions. Let's give it a go.

PR #1478, opened 2023-07-15
UI: Device emulation

faha223:
Added functionality to emulate an Xbox Controller S or a Steel Battalion controller in addition to the existing emulated Xbox controller

PR #1516, opened 2023-08-07
ui: Support changing controller/keyboard mappings

### LINKED ISSUE #136: Implement gamepad button/axis remapping

antangelo:
Adds a configuration scheme and UI for remapping keyboard/gamepad inputs to the emulated controller. I tried to keep it as modular as possible so that #1478 can extend this system with other devices and the other devices can be mapped independently.

Resolves #136 and supersedes #1301 

Leaving in draft initially as this requires some genconfig PRs to be merged-- the genconfig submodule is currently pointed to my fork. I've formatted the PR using clang-format, which is why it differs at points from neighboring code. I am leaving it as-is because formatting code I haven't touched will make this already large change harder to review, but can do so if needed.

image

### DIFF
--- a/config_spec.yml
+++ b/config_spec.yml
@@ -55,6 +55,7 @@ input:
   auto_bind:
     type: bool
     default: true
+  # Deprecated: use 'enable_rumble' on the controller mapping settings instead
   allow_vibration:
     type: bool
     default: true
--- a/config_spec.yml
+++ b/config_spec.yml
@@ -136,6 +136,91 @@ input:
     rtrigger:
       type: integer
       default: 18 # w
+  gamepad_mappings:
+    type: array
+    items:
+      gamepad_id: string
+      enable_rumble:
+        type: bool
+        default: true
+      # steel_battalion_mapping:
+      # ...
+      controller_mapping:
+        a:
+          type: integer
+          default: 0
+        b:
+          type: integer
+          default: 1
+        x:
+          type: integer
+          default: 2
+        y:
+          type: integer
+          default: 3
+        back:
+          type: integer
+          default: 4
+        guide:
+          type: integer
+          default: 5
+        start:
+          type: integer
+          default: 6
+        lstick_btn:
+          type: integer
+          default: 7
+        rstick_btn:
+          type: integer
+          default: 8
+        lshoulder:
+          type: integer
+          default: 9
+        rshoulder:
+          type: integer
+          default: 10
+        dpad_up:
+          type: integer
+          default: 11
+        dpad_down:
+          type: integer
+          default: 12
+        dpad_left:
+          type: integer
+          default: 13
+        dpad_right:
+          type: integer
+          default: 14
+        axis_left_x:
+          type: integer
+          default: 0
+        axis_left_y:
+          type: integer
+          default: 1
+        axis_right_x:
+          type: integer
+          default: 2
+        axis_right_y:
+          type: integer
+          default: 3
+        axis_trigger_left:
+          type: integer
+          default: 4
+        axis_trigger_right:
+          type: integer
+          default: 5
+        invert_axis_left_x:
+          type: bool
+          default: false
+        invert_axis_left_y:
+          type: bool
+          default: false
+        invert_axis_right_x:
+          type: bool
+          default: false
+        invert_axis_right_y:
+          type: bool
+          default: false
 
 display:
   renderer:
--- a/ui/xemu-controllers.cc
+++ b/ui/xemu-controllers.cc
@@ -1,0 +1,102 @@
+/*
+ * xemu Controller Binding Management
+ *
+ * Copyright (C) 2020-2023 Matt Borgerson
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "xemu-controllers.h"
+#include "xemu-settings.h"
+#include <assert.h>
+#include <limits>
+#include <string>
+
+std::pair<bool, bool>
+ControllerKeyboardRebindingMap::ConsumeRebindEvent(SDL_Event *event)
+{
+    if (event->type == SDL_KEYDOWN) {
+        *(g_keyboard_scancode_map[table_row]) = event->key.keysym.scancode;
+
+        return { true, true };
+    }
+
+    return { false, false };
+}
+
+std::pair<bool, bool>
+ControllerGamepadRebindingMap::ConsumeRebindEvent(SDL_Event *event)
+{
+    if (event->type == SDL_CONTROLLERDEVICEREMOVED) {
+        if (state->sdl_joystick_id == event->cdevice.which) {
+            return { false, true };
+        }
+    } else if (event->type == SDL_CONTROLLERBUTTONUP && table_row < 15 &&
+               seen_key_down) {
+        // Bind on controller up ensures the UI does not immediately respond
+        // once the new binding is applied
+
+        if (state->sdl_joystick_id != event->cbutton.which) {
+            return { false, false };
+        }
+
+        int *button_map[15] = {
+            &state->controller_map->controller_mapping.a,
+            &state->controller_map->controller_mapping.b,
+            &state->controller_map->controller_mapping.x,
+            &state->controller_map->controller_mapping.y,
+            &state->controller_map->controller_mapping.back,
+            &state->controller_map->controller_mapping.guide,
+            &state->controller_map->controller_mapping.start,
+            &state->controller_map->controller_mapping.lstick_btn,
+            &state->controller_map->controller_mapping.rstick_btn,
+            &state->controller_map->controller_mapping.lshoulder,
+            &state->controller_map->controller_mapping.rshoulder,
+            &state->controller_map->controller_mapping.dpad_up,
+            &state->controller_map->controller_mapping.dpad_down,
+            &state->controller_map->controller_mapping.dpad_left,
+            &state->controller_map->controller_mapping.dpad_right,
+        };
+
+        *(button_map[table_row]) = event->cbutton.button;
+
+        return { true, true };
+    } else if (event->type == SDL_CONTROLLERBUTTONDOWN && table_row < 15) {
+        // If we are rebinding with a controller, we should not consume the key
+        // up event from activating the button
+        seen_key_down = true;
+    } else if (event->type == SDL_CONTROLLERAXISMOTION && table_row >= 15 &&
+               std::abs(event->caxis.value >> 1) >
+                   std::numeric_limits<Sint16>::max() >> 2) {
+        // FIXME: Allow face buttons to map to axes
+        if (state->sdl_joystick_id != event->caxis.which) {
+            return { false, false };
+        }
+
+        int *axis_map[6] = {
+            &state->controller_map->controller_mapping.axis_left_x,
+            &state->controller_map->controller_mapping.axis_left_y,
+            &state->controller_map->controller_mapping.axis_right_x,
+            &state->controller_map->controller_mapping.axis_right_y,
+            &state->controller_map->controller_mapping.axis_trigger_left,
+            &state->controller_map->controller_mapping.axis_trigger_right,
+        };
+
+        *(axis_map[table_row - 15]) = event->caxis.axis;
+
+        return { true, true };
+    }
+
+    return { false, false };
+}
--- a/ui/xemu-controllers.cc
+++ b/ui/xemu-controllers.cc
@@ -39,35 +39,35 @@ ControllerKeyboardRebindingMap::ConsumeRebindEvent(SDL_Event *event)
     return RebindEventResult::Ignore;
 }
 
-RebindEventResult
-ControllerGamepadRebindingMap::HandleButtonEvent(SDL_ControllerButtonEvent *event)
+RebindEventResult ControllerGamepadRebindingMap::HandleButtonEvent(
+    SDL_ControllerButtonEvent *event)
 {
-  if (m_state->sdl_joystick_id != event->which) {
-    return RebindEventResult::Ignore;
-  }
+    if (m_state->sdl_joystick_id != event->which) {
+        return RebindEventResult::Ignore;
+    }
 
-  int *button_map[controller_button_count] = {
-    &m_state->controller_map->controller_mapping.a,
-    &m_state->controller_map->controller_mapping.b,
-    &m_state->controller_map->controller_mapping.x,
-    &m_state->controller_map->controller_mapping.y,
-    &m_state->controller_map->controller_mapping.back,
-    &m_state->controller_map->controller_mapping.guide,
-    &m_state->controller_map->controller_mapping.start,
-    &m_state->controller_map->controller_mapping.lstick_btn,
-    &m_state->controller_map->controller_mapping.rstick_btn,
-    &m_state->controller_map->controller_mapping.lshoulder,
-    &m_state->controller_map->controller_mapping.rshoulder,
-    &m_state->controller_map->controller_mapping.dpad_up,
-    &m_state->controller_map->controller_mapping.dpad_down,
-    &m_state->controller_map->controller_mapping.dpad_left,
-    &m_state->controller_map->controller_mapping.dpad_right,
-  };
+    int *button_map[controller_button_count] = {
+        &m_state->controller_map->controller_mapping.a,
+        &m_state->controller_map->controller_mapping.b,
+        &m_state->controller_map->controller_mapping.x,
+        &m_state->controller_map->controller_mapping.y,
+        &m_state->controller_map->controller_mapping.back,
+        &m_state->controller_map->controller_mapping.guide,
+        &m_state->controller_map->controller_mapping.start,
+        &m_state->controller_map->controller_mapping.lstick_btn,
+        &m_state->controller_map->controller_mapping.rstick_btn,
+        &m_state->controller_map->controller_mapping.lshoulder,
+        &m_state->controller_map->controller_mapping.rshoulder,
+        &m_state->controller_map->controller_mapping.dpad_up,
+        &m_state->controller_map->controller_mapping.dpad_down,
+        &m_state->controller_map->controller_mapping.dpad_left,
+        &m_state->controller_map->controller_mapping.dpad_right,
+    };
 
-  // FIXME: Allow face buttons to map to axes
-  if (m_table_row >= controller_button_count) {
-    return RebindEventResult::Ignore;
-  }
+    // FIXME: Allow face buttons to map to axes
+    if (m_table_row >= controller_button_count) {
+        return RebindEventResult::Ignore;
+    }
 
     // If we only track up events, then we might rebind to a button
     // that was already held down when the rebinding event began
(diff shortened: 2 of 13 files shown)


UserGit-2021:
Hello! Good job! Will there be mouse support?

antangelo:
> Hello! Good job! Will there be mouse support?

Mouse support is out of scope for this PR

ui/xui/main-menu.cc
```diff
@@ -268,6 +323,154 @@ void MainMenuInputView::Draw()
...
+        "Start",
+        "Left Stick Button",
+        "Right Stick Button",
+        "White",
+        "Black",
+        "DPad Up",
+        "DPad Down",
+        "DPad Left",
+        "DPad Right",
+        "Left Stick Axis+",
```
mborgerson: Left Stick "Axis+" and "Axis-" are a bit confusing. Why not X and Y here?

ui/xui/main-menu.cc
```diff
@@ -260,6 +282,39 @@ void MainMenuInputView::Draw()
...
 
+    SectionTitle("Mapping");
+
+    float p = ImGui::GetFrameHeight() * 0.3;
+    ImGui::PushStyleVar(ImGuiStyleVar_CellPadding, ImVec2(p, p));
+    if (ImGui::BeginTable("input_remap_tbl", 3,
+                          ImGuiTableFlags_RowBg | ImGuiTableFlags_Borders)) {
+        ImGui::TableSetupColumn("Button");
+        ImGui::TableSetupColumn("Mapped to");
+        ImGui::TableSetupColumn("Remap");
```
mborgerson: Here it is called 'Remap', but then the button text calls it 'Rebind'. Let's pick either 'bind' or 'map'. 'Bind' is also used when talking about pairing host device to emulated input device below, so it is overloaded now. We need to distinguish between controller pairing and controller input mapping.

ui/xui/main-menu.cc
```diff
@@ -260,6 +282,39 @@ void MainMenuInputView::Draw()
...
     ImGui::SetCursorPos(pos);
 
+    SectionTitle("Mapping");
+
+    float p = ImGui::GetFrameHeight() * 0.3;
+    ImGui::PushStyleVar(ImGuiStyleVar_CellPadding, ImVec2(p, p));
+    if (ImGui::BeginTable("input_remap_tbl", 3,
+                          ImGuiTableFlags_RowBg | ImGuiTableFlags_Borders)) {
+        ImGui::TableSetupColumn("Button");
+        ImGui::TableSetupColumn("Mapped to");
```
mborgerson: I think the 'Button' and 'Mapped to' language here is confusing. Because input flows from real hardware peripheral through to emulated input device then to game input stack, I think 'Mapped to' should be the emulated input button/axis.

We could be more clear changing the columns to something like 'Host Input' and 'Emulated Input', or <Name of Real Peripheral> and 'Emulated Xbox Controller'.

ui/xemu-controllers.cc
```diff
@@ -0,0 +1,189 @@
...
+        };
+
+        *(button_map[table_row]) = event->cbutton.button;
+
+        return { true, true };
+    } else if (event->type == SDL_CONTROLLERBUTTONDOWN && table_row < 15) {
+        // If we are rebinding with a controller, we should not consume the key
+        // up event from activating the button
+        seen_key_down = true;
+    } else if (event->type == SDL_CONTROLLERAXISMOTION && table_row >= 15) {
```
mborgerson: We need some kind of threshold for activation here, like 50%. Joysticks can be pretty sensitive and the slightest touch will register as an event here.

ui/xemu-controllers.cc
```diff
@@ -0,0 +1,189 @@
...
+    }
+}
+
+void xemu_settings_save_controller_binding(const char *guid,
+                                           struct controller_binding *cb)
+{
+    const char *base_path = get_binding_path();
+
+    assert(strlen(guid) > 0);
+    char *binding_path = g_strdup_printf("%s/%s.toml", base_path, guid);
```
mborgerson: I assume storing non-keyboard mappings in separate files was done to work around genconfig limitations. From a user perspective, it seems weird to have them in separate spots. Should we try to move these input mappings into xemu.toml or do you think it's better to have them separated like this

ui/xui/main-menu.cc
antangelo: I've replaced them with X and Y

ui/xui/main-menu.cc
antangelo: I've removed this column entirely to simplify the table. Instead, the button text is now just the host controller input that maps to the corresponding guest input.

ui/xui/main-menu.cc

ui/xemu-controllers.cc
antangelo: Done, added a 50% activation threshold

ui/xemu-controllers.cc
antangelo: I didn't want to overly clutter `xemu.toml`, as there are many entries for the various mappings and if a user has multiple controllers then the file can fill up quickly (especially with the addition of more devices later on, or if we support setting mappings per title).

There were genconfig limitations either way (which I resolved separately to allow for multiple configs in the same program), but those required to support maps with string keys would be more involved as well.

mborgerson: I expect most people have at most a handful of devices they intend to use, so I don't think storing button mappings in xemu.toml would add much clutter in the typical case. Would you consider looking into migrating device input mapping into the main config?

AkihitoZero5454:
Can we also have an option to switch the controller image to Controller S?

ui/xemu-controllers.cc
antangelo: I have a solution that stores the bindings in an array in the main config. It depends on https://github.com/mborgerson/genconfig/pull/10

I also looked into adding support for tables with arbitrary string keys to genconfig but there's some added complexity there with serialization and efficiently adding nodes to the map without invalidating pointers to the previous nodes within the serialized struct. I don't personally think this complexity is worthwhile if there won't usually be many controllers in the array, as you say, but can finish this approach if you think it's better.

ElTioRata:
It is planned to add mapping profiles? That way they can be assigned per-game.

ui/xemu-controllers.cc
antangelo: Done, using the aforementioned array solution.

faha223:
This section takes up a lot of space vertically in the UI. Would it be better to have the grid in a collapsible container?

MasonT8198:
This could use a rebase, and also, I second that the UI rebind should be a smaller element with a slider that way you can go through it, that's a long boi

config_spec.yml
```diff
@@ -128,6 +128,91 @@ input:
...
       type: integer
       default: 18 # w
+  gamepad_mappings:
+    type: array
+    items:
+      gamepad_id: string
+      enable_rumble:
+        type: bool
+        default: true
+      # steel_battalion_mapping:
```
abaire, line 147: super nit: remove until the controller is emulated?

ui/xemu-controllers.cc
```diff
@@ -0,0 +1,102 @@
+/*
+ * xemu Controller Binding Management
+ *
+ * Copyright (C) 2020-2023 Matt Borgerson
```
abaire: nit: Update?

ui/xemu-controllers.cc
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    return { false, false };
+}
+
+std::pair<bool, bool>
+ControllerGamepadRebindingMap::ConsumeRebindEvent(SDL_Event *event)
+{
+    if (event->type == SDL_CONTROLLERDEVICEREMOVED) {
+        if (state->sdl_joystick_id == event->cdevice.which) {
```
abaire: Should a ControllerGamepadRebindingMap ever respond to events that are not for its joystick ID? 

Should we just have a return early check to cover everything (looks like button down events aren't filtered, which seems odd to me)?

ui/xemu-controllers.cc
```diff
@@ -0,0 +1,102 @@
...
+    } else if (event->type == SDL_CONTROLLERBUTTONUP && table_row < 15 &&
+               seen_key_down) {
+        // Bind on controller up ensures the UI does not immediately respond
+        // once the new binding is applied
+
+        if (state->sdl_joystick_id != event->cbutton.which) {
+            return { false, false };
+        }
+
+        int *button_map[15] = {
```
abaire: nit: could we use a union with an array in the struct to avoid having to do this indirection? If somebody reordered entries in the future it would be easy to miss this and cross-map entries.

ui/xemu-controllers.h
```diff
@@ -0,0 +1,82 @@
...
+public:
+    std::pair<bool, bool> ConsumeRebindEvent(SDL_Event *event) override;
+    ControllerGamepadRebindingMap(int table_row, ControllerState *state)
+        : RebindingMap(table_row), state{ state }, seen_key_down{ false }
+    {
+    }
+};
+
+extern "C" {
+#endif
```
abaire: nit: maybe worth tagging with `// #ifdef __cplusplus` since it's a long ways away from the opening?

ui/xemu-controllers.h
```diff
@@ -0,0 +1,82 @@
...
+    {
+        return table_row;
+    }
+
+    virtual ~RebindingMap()
+    {
+    }
+
+protected:
+    int table_row;
```
abaire: nit: elsewhere it seems like the xemu C++ convention is to have members before methods (including in this file below)

ui/xemu-controllers.h
```diff
@@ -0,0 +1,82 @@
...
+    // consume: Whether the SDL_Event should not propagate to the UI
+    // cancel: Whether this rebinding map should be cancelled
+    virtual std::pair<bool, bool> ConsumeRebindEvent(SDL_Event *event) = 0;
+
+    int GetTableRow() const
+    {
+        return table_row;
+    }
+
+    virtual ~RebindingMap()
```
abaire: @mborgerson do we have a baseline requirement on C++ language version? I.e., can this be `= default` or does xemu still want to support the really ancient specs?

ui/xui/main-menu.cc
```diff
@@ -75,6 +78,24 @@ void MainMenuGeneralView::Draw()
     //        "Limit DVD/HDD throughput to approximate Xbox load times");
 }
 
+bool MainMenuInputView::ConsumeRebindEvent(SDL_Event *event)
+{
+    if (!rebinding)
+        return false;
+
+    auto [consume, cancel] = rebinding->ConsumeRebindEvent(event);
```
abaire: To my comment before re: min C++ version, if we're using `auto` I think we can use `default` for uninteresting constructors/destructors?

ui/xui/main-menu.cc
```diff
@@ -48,6 +48,9 @@ MainMenuScene g_main_menu;
 
 MainMenuTabView::~MainMenuTabView() {}
 void MainMenuTabView::Draw() {}
```
abaire: @mborgerson is there a style reason not to inline trivial methods in the header files? Would cut down on some boilerplate

ui/xui/main-menu.cc
```diff
@@ -439,6 +461,38 @@ void MainMenuInputView::Draw()
         ImGui::Columns(1);
     }
 
+    SectionTitle("Mapping");
+
+    float p = ImGui::GetFrameHeight() * 0.3;
```
abaire: In response to the (entirely valid, in my opinion) comments about this being a long list, we could wrap it in a `CollapsibleHeader` as a tiny change: https://github.com/xemu-project/xemu/commit/a4a55ed7d2314bb6e9702a550384a03a2cf45da7

ui/xui/main-menu.cc
```diff
@@ -447,6 +501,155 @@ void MainMenuInputView::Draw()
...
+        "Left Stick Axis X",
+        "Left Stick Axis Y",
+        "Right Stick Axis X",
+        "Right Stick Axis Y",
+        "Left Trigger Axis",
+        "Right Trigger Axis",
+    };
+
+    bool is_keyboard = state->type == INPUT_DEVICE_SDL_KEYBOARD;
+    int table_rows = is_keyboard ? 25 : 21;
```
abaire: nit: `ARRAY_SIZE` rather than hardcode (not that they're likely to change)?

ui/xemu-controllers.cc
abaire: I missed that the mapping is defined in the autogen'd config and would require chnges to `gen_config.py` to generate a union.

ui/xui/main-menu.cc
mborgerson: Nope, go for it

ui/xemu-controllers.h
mborgerson: xemu builds today with C++17. I'm not too concerned with maintaining support for building with old tools

ui/xui/main-menu.cc
abaire: Cool, I'll take a pass once this one is merged to avoid complicating rebasing it unnecessarily. Since it'd just be cleanup I'd be inclined to do a `clang-format` of just the affected files (if you don't beat me to it with a larger pass) so we can get that out of the way and hide both from blame.

TheSloth1218:
Any news on this addition I really wish it was implemented asap?

config_spec.yml
antangelo: There are other spots in the UI code that comment out options for work-in-progress features (e.g. the disk IO throttling setting switch), so this is intended to follow that precedent (but I don't feel that strongly about it).

ui/xemu-controllers.cc
antangelo: Done

ui/xemu-controllers.cc
antangelo: The `which` member depends on the event type (i.e. device, button, or axis), and I didn't want to rely on the SDL event struct layouts being the same for each event. I have separated this function out by event to make it a bit cleaner.

ui/xemu-controllers.h
antangelo: Done

ui/xemu-controllers.h
antangelo: Done, reordered and updated the names to follow xemu convention.

ui/xemu-controllers.h
antangelo: Updated to `= default;`

ui/xui/main-menu.cc
antangelo: Thanks, I've cherry-picked your change in to this set. I opted to move the controller axis inversion and rumble settings out of the header, so only the mapping table is hidden by default.

ui/xui/main-menu.cc
antangelo: `ARRAY_SIZE` from qemu headers doesn't seem to compile here, so I've opted to use `std::size` instead for the same effect.

ui/xui/main-menu.cc
abaire: Tracked in #2441

ui/xemu-controllers.h
```diff
@@ -0,0 +1,85 @@
...
+
+extern "C" {
+#endif // __cplusplus
+
+extern int *g_keyboard_scancode_map[25];
+
+GamepadMappings *xemu_settings_load_gamepad_mapping(const char *guid);
+
+#ifdef __cplusplus
+}
```
abaire: super nit: Consider adding "  // extern "C"" in case more C-accessible items are added in the future pushing this further from the opening of the block avove?

ui/xemu-settings.cc
```diff
@@ -254,3 +255,24 @@ void remove_net_nat_forward_ports(unsigned int index)
...
+            return mapping;
+        }
+    }
+
+    auto cnode = config_tree.child("input")->child("gamepad_mappings");
+    cnode->update_from_struct(&g_config);
+    cnode->free_allocations(&g_config);
+    cnode->children.push_back(*cnode->array_item_type);
+    auto &e = cnode->children.back();
+    e.child("gamepad_id")->set_string(std::string(guid));
```
abaire: Super nit: Does the string need explicit construction?

ui/xui/main-menu.hh
```diff
@@ -36,6 +38,7 @@ class MainMenuTabView
 public:
     virtual ~MainMenuTabView();
     virtual void Draw();
+    virtual void Hide();
```
abaire: Nit: Just inline the nop implementation here?

abaire (commented):
A few tiny nitpicks, otherwise this looks good to me

ui/xemu-controllers.h
antangelo: Done

ui/xemu-settings.cc
antangelo: It does not, I've updated it to be implicit.

ui/xui/main-menu.hh
antangelo: Done

abaire (approved):
lgtm

ui/xui/main-menu.cc
```diff
@@ -506,6 +566,156 @@ void MainMenuInputView::Draw()
...
+{
+    m_rebinding = nullptr;
+}
+
+void MainMenuInputView::PopulateTableController(ControllerState *state)
+{
+    if (!state)
+        return;
+
+    static const char *kbd_map[25] = {
```
mborgerson: Can we order these to be more similar? e.g. here DPad follows Y, but below it follows Black.

ui/xui/main-menu.cc
```diff
@@ -496,6 +516,46 @@ void MainMenuInputView::Draw()
...
+
+            PopulateTableController(bound_state);
+
+            ImGui::EndTable();
+        }
+        ImGui::PopStyleVar();
+    }
+
+    if (bound_state && bound_state->type == INPUT_DEVICE_SDL_GAMECONTROLLER) {
+        Toggle("Enable Rumble", &bound_state->controller_map->enable_rumble);
```
mborgerson, line 542: We now confusingly have two rumble toggles in the configuration UI:
- Enable Rumble
- Controller vibration

Can we consolidate them?

ui/xui/main-menu.cc
```diff
@@ -496,6 +516,46 @@ void MainMenuInputView::Draw()
         ImGui::Columns(1);
     }
 
+    SectionTitle("Mapping");
+    ImVec4 tc = ImGui::GetStyle().Colors[ImGuiCol_Header];
+    tc.w = 0.0f;
+    ImGui::PushStyleColor(ImGuiCol_Header, tc);
+
+    if (ImGui::CollapsingHeader("Button Mapping")) {
```
mborgerson: nit: Since we map more than just buttons, prefer a more general phrase like "Input mapping"

ui/xui/main-menu.cc
antangelo: Which one do you want to keep? This PR adds a per-controller setting, whereas the existing one is global.

mborgerson: How about we just go with the per-controller setting

ui/xui/main-menu.cc
```diff
@@ -496,6 +516,46 @@ void MainMenuInputView::Draw()
...
+        Toggle("Invert Left Y Axis",
+               &bound_state->controller_map->controller_mapping
+                    .invert_axis_left_y);
+        Toggle("Invert Right X Axis",
+               &bound_state->controller_map->controller_mapping
+                    .invert_axis_right_x);
+        Toggle("Invert Right Y Axis",
+               &bound_state->controller_map->controller_mapping
+                    .invert_axis_right_y);
+    }
```
mborgerson, line 555: Let's add an option to reset input mappings to default settings

ui/xui/main-menu.cc
antangelo: Done. The old ordering followed the sequence in `ControllerState::buttons`, so this update required unrolling a loop in `xemu_input_update_sdl_kbd_controller_state` to handle the updated offsets (which now looks more similar to the gamepad code).

ui/xui/main-menu.cc
antangelo: Done. I have removed the `allow_vibration` setting from the UI and added logic that migrates the setting to any controllers connected to xemu on the first run after updating. The setting is then set to default, removing it from the user config.

ui/xui/main-menu.cc

ui/xui/main-menu.cc
antangelo: Done

mborgerson:
Great, thanks!

Issue #1239, opened 2022-08-18
Passed Through USB devices do not match hardware behavior


HadetTheUndying:
### Bug Description

While I was testing Steel Battalion: Line of Contact I noticed some peculiar issues.

1. Inputs appear to be getting pressed when know buttons, or keys are pressed.
2.  Inputs appear to reset(flicker) or trigger the wrong inputs in the controller calibration menu
3.  it also may be effecting my racing wheel as i noticed it felt like games were occasionally steering against me(Returning to center)

Video of the problem:

Uploading 2022-08-18_17-01-52.mp4…
EDIT: Apparently the video did not upload here is a link to the same video I uploaded in Discord 
https://cdn.discordapp.com/attachments/1009927154529742879/1009946148087279626/2022-08-18_17-01-52.mp4
.

### Expected Behavior

This is a video of the correct behavior on hardware. I do not believe this is a Steel Battalion specific issue because of what i mentioned with my racing wheel, however it is the only game I have that I can quickly test with.

https://user-images.githubusercontent.com/28496334/185505797-8c5bd4ca-7dfc-4c00-9534-cb5cdf6950af.mp4

### xemu Version

Version:      0.7.56-15-g940bee452c
Branch:       master
Commit:       940bee452cc34127ecf3f364c3c5f52e4e6a80a4
Date:         Thu Aug 18 08:13:32 PM UTC 2022

### System Information

CPU:          AMD Ryzen 7 2700X Eight-Core Processor         
OS Platform:  Linux
OS Version:   Void Linux
Manufacturer: AMD
GPU Model:    AMD Radeon RX Vega (vega10, LLVM 12.0.1, DRM 3.46, 5.18.16_1)
Driver:       4.6 (Core Profile) Mesa 22.1.7
Shader:       4.60

### Additional Context

I have tested this scenario in both the original and Line of Contact's calibration screens. It also causes inputs to be hit multiple times instead of just once. Could this be because I'm using a USB 3 port for my USB adapter? I only have Front panel USB, and I'm not sure if I have back panel USB 2 at all. The controller works fine and behaves correctly when using the Windows Drivers in a VM from the same port however. Could xemu not be syncing correctly?

GXTX:
Isn't pass through disabled?

HadetTheUndying:
> Isn't pass through disabled?

[...]

PR #1531, opened 2023-09-23
OHCI: Fixed a bug in the OHCI implementation from QEMU

faha223:
This fixes the issue with USB passthrough in which a race condition is created when using the async api in libusb to write to and read from multiple endpoints simultaneously

Addresses this bug:
https://github.com/xemu-project/xemu/issues/1239


GXTX:
Since this an issue with QEMU, should this perhaps be posted to the ML for comments / future merge?

faha223:
Yes, it should. I'll need to look into what their patch submission process is

hw/usb/hcd-ohci.h
```diff
@@ -84,13 +98,11 @@ typedef struct OHCIState {
 
     /* Active packets.  */
     uint32_t old_ctl;
-    USBPacket usb_packet;
-    uint8_t usb_buf[8192];
-    uint32_t async_td;
-    bool async_complete;
+    // union  { struct USBActivePacket *tqh_first; QTailQLink tqh_circ; }
```
mborgerson: ?

hw/usb/hcd-ohci.c
```diff
@@ -1999,9 +2096,6 @@ const VMStateDescription vmstate_ohci_state = {
         VMSTATE_UINT32(hreset, OHCIState),
         VMSTATE_UINT32(htest, OHCIState),
         VMSTATE_UINT32(old_ctl, OHCIState),
-        VMSTATE_UINT8_ARRAY(usb_buf, OHCIState, 8192),
```
mborgerson, line 2032: * How are in-flight packets handled now during vmsave/load events?
* The structure change breaks snapshot compatibility

hw/usb/hcd-ohci.c
```diff
@@ -1890,9 +1970,10 @@ void usb_ohci_init(OHCIState *ohci, DeviceState *dev, uint32_t num_ports,
     ohci->localmem_base = localmem_base;
 
     ohci->name = object_get_typename(OBJECT(dev));
-    usb_packet_init(&ohci->usb_packet);
-
-    ohci->async_td = 0;
+    
+    // Inialize the QTAILQ_HEAD. QTAILQ_HEAD_INITIALIZER doesn't work here
```
mborgerson: How about `QTAILQ_INIT`

hw/usb/hcd-ohci.c
```diff
@@ -1906,10 +1987,26 @@ void ohci_sysbus_die(struct OHCIState *ohci)
 {
     trace_usb_ohci_die();
 
+    ohci_clear_active_packets(ohci);
     ohci_set_interrupt(ohci, OHCI_INTR_UE);
     ohci_bus_stop(ohci);
 }
 
+void ohci_clear_active_packets(struct OHCIState *ohci)
```
mborgerson, line 2051: Why `clear` and not `cancel`?

hw/usb/hcd-ohci.h
```diff
@@ -33,7 +33,21 @@ typedef struct OHCIPort {
...
+typedef struct USBActivePacket USBActivePacket;
+typedef struct OHCIState OHCIState;
+
+struct USBActivePacket {
+    USBEndpoint *ep;
+    USBPacket usb_packet;
+    uint8_t usb_buf[8192];
+    uint32_t async_td;
+    bool async_complete;
+    OHCIState *ohci;
```
mborgerson, line 46: Do we need `OHCIState` pointer here?

hw/usb/hcd-ohci.c
```diff
@@ -1751,11 +1823,19 @@ static void ohci_child_detach(USBPort *port1, USBDevice *dev)
...
-        usb_cancel_packet(&ohci->usb_packet);
-        ohci->async_td = 0;
+    USBActivePacket *iter, *iter2;
+    QTAILQ_FOREACH_SAFE(iter, &ohci->active_packets, next, iter2) {
+        if(iter->usb_packet.ep->dev == dev) {
+            if (iter->async_td &&
+                usb_packet_is_inflight(&iter->usb_packet) &&
+                iter->usb_packet.ep->dev == dev) {
+                usb_cancel_packet(&iter->usb_packet);
+                iter->async_td = 0;
```
mborgerson: Unnecessary assignment, packet free'd below

hw/usb/hcd-ohci-pci.c
```diff
@@ -87,9 +87,12 @@ static void usb_ohci_exit(PCIDevice *dev)
...
     ohci_bus_stop(s);
 
-    if (s->async_td) {
-        usb_cancel_packet(&s->usb_packet);
-        s->async_td = 0;
+    USBActivePacket *iter;
+    QTAILQ_FOREACH(iter, &s->active_packets, next) {
+        if (iter->async_td) {
+            usb_cancel_packet(&iter->usb_packet);
+            iter->async_td = 0;
```
mborgerson: Should you clean up packets

hw/usb/hcd-ohci.h
```diff
@@ -33,7 +33,21 @@ typedef struct OHCIPort {
     uint32_t ctrl;
 } OHCIPort;
 
-typedef struct OHCIState {
+typedef struct USBActivePacket USBActivePacket;
+typedef struct OHCIState OHCIState;
+
+struct USBActivePacket {
```
mborgerson, line 40: Since this is basically 1:1 with endpoints, does it makes sense to just move the async tracking into `USBEndpoint` and avoid all the list iteration?

hw/usb/hcd-ohci.c
```diff
@@ -909,6 +906,35 @@ static int ohci_service_td(OHCIState *ohci, struct ohci_ed *ed)
...
+    USBActivePacket *packet = NULL;
+    USBActivePacket *iter;
+    QTAILQ_FOREACH(iter, &ohci->active_packets, next) {
+        if(iter->ep == ep) {
+            packet = iter;
+            break;
+        }
+    }
+
+    // A packet for this endpoint doesn't exist yet. Make one
```
mborgerson: nit: File uses `/* this style of comment */`

hw/usb/hcd-ohci.c
```diff
@@ -1751,11 +1823,19 @@ static void ohci_child_detach(USBPort *port1, USBDevice *dev)
...
     OHCIState *ohci = port1->opaque;
 
-    if (ohci->async_td &&
-        usb_packet_is_inflight(&ohci->usb_packet) &&
-        ohci->usb_packet.ep->dev == dev) {
-        usb_cancel_packet(&ohci->usb_packet);
-        ohci->async_td = 0;
+    USBActivePacket *iter, *iter2;
+    QTAILQ_FOREACH_SAFE(iter, &ohci->active_packets, next, iter2) {
+        if(iter->usb_packet.ep->dev == dev) {
```
mborgerson: nit: Convention is to include space between `if` and `(`

mborgerson:
> This fixes the issue with USB passthrough in which a race condition is created when using the async api in libusb to write to and read from multiple endpoints simultaneously

Can you share more specifics about the race condition to help me understand the root cause of the issue?

faha223:
Every USBEndpoint that shares a single OCHI Controller shares a single buffer for reading and writing data.
The libsub async API returns after sending a request and responses are handled later.

If a read request is sent and a write request is sent after the read request is sent but before the response comes in, then the read request has a chance to return data from the write request. I observed this while using the Steel Battalion controller but also while using the gamepad. 

Converting the libusb passthrough code to use the sync API fixes the issue by blocking on every read and write request. This introduces latency so fixing the code to work properly with the async API is preferred.

hw/usb/hcd-ohci.h
faha223: USBEndpoint is shared between OHCI, EHCI, UHCI, and XHCI. Moving the async tracking into USBEndpoint would make sense from the perspective of OHCI, but wouldn't be used by EHCI, UHCI, or XHCI

hw/usb/hcd-ohci.h
faha223: It's used in ohci_async_complete_packet when calling ohci_process_lists. This is the only place it is used. 

We get USBActivePacket by using container_of on the USBPacket argument. It might be possible to use container_of to get the OHCI pointer we need here if we can use one of the QTAILQ macros to get to the head of the list.

hw/usb/hcd-ohci.c
faha223: clear because this function makes the list empty. I am not opposed to changing it to cancel

hw/usb/hcd-ohci.c
faha223: I hadn't tried it. It appears to work, Thanks!

hw/usb/hcd-ohci.h
faha223: This comment shows what QTAILQ_HEAD(, USBActivePacket) expands to. I didn't mean to include this.

LoveMHz:
And here I was, spending my day thinking I borked something in my USB driver.

This PR appears to resolve the issue stated in #1239. I was able to verify this with Stellar's Xbox One driver where we were running into endpoint data appearing in endpoint interrupts under xemu. With the patch, I was able to verify that the correct behavior matches that of our hw tests.

I look forward to reviewing the patch once @mborgerson's comments are addressed.

@faha223 would you mind dropping me an email? (dustin@makemhz.com). Thank you for your contribution.

hw/usb/hcd-ohci-pci.c
faha223: No.

I removed this foreach loop because the call right after this (ohci_stop_endpoints) actually calls (ohci_clear_active_packets) and does the work of this foreach loop already.

hw/usb/hcd-ohci.c
faha223: When loading a snapshot, `vmstate_load_state` invokes `pre_load` function pointer in the `VMStateDescription` if one was defined. Then it copies the data stored for each field in the `VMState` into the appropriate memory as defined in the `VMStateDescription`. Then it invokes the `post_load` function pointer in the `VMStateDescription` if one was defined.

Saving a snapshot does the same thing, but using the `pre_save` and `post_save` function pointers.

These fields are not defined in `vmstate_ohci_state` so there's no logic that runs when loading or saving a snapshot that interacts with these fields. 

I suspect there's a real possibility that the `VMState` saving/restoring the `async_td` and `usb_packet` fields in snapshots can cause problems, but the structure change that would be needed to save and load the new list of `async_packets` would break snapshot compatibility.

I tested saving and loading snapshots with a libusb controller attached, both while interacting with the controller and while not interacting with the controller and it seemed to be mostly stable. I was able to crash the emulator during vmload a few times but the point of failure was clearing out the existing async packets before restoring the snapshot data. It didn't have anything to do with the actual snapshot data.

I've added back the fields I removed from `OHCIState` and `vmstate_ohci_state` since their removal breaks snapshot compatibility.

hw/usb/hcd-ohci.c
```diff
@@ -1135,11 +1161,62 @@ static int ohci_service_ed_list(OHCIState *ohci, uint32_t head)
...
+            default:
+                continue;
+            }
+
+            dev = ohci_find_device(ohci, OHCI_BM(ed.flags, ED_FA));
+            if (dev != NULL) {
+                ep = usb_ep_get(dev, pid, OHCI_BM(ed.flags, ED_EN));
+
+                if (ep != NULL) {
+                    USBActivePacket *iter;
```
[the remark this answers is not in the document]
faha223: The variable is being declared at the top of the scope in which it lives.

hw/usb/hcd-ohci.c
```diff
@@ -937,6 +940,35 @@ static int ohci_service_td(OHCIState *ohci, struct ohci_ed *ed)
...
+
+    /* A packet for this endpoint doesn't exist yet. Make one */
+    if (packet == NULL) {
+        packet = g_malloc(sizeof(USBActivePacket));
+        usb_packet_init(&packet->usb_packet);
+        packet->ep = ep;
+        packet->async_complete = false;
+        packet->async_td = 0;
+        packet->ohci = ohci;
+        QTAILQ_INSERT_TAIL(&ohci->active_packets, packet, next);
```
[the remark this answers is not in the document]
faha223: I moved the variable declarations to the end of the existing variable declarations at the start of the function body

faha223:
I'm working on submitting a patch email to the QEMU mailing list. I'm still not completely satisfied with the way this change solves the problem, but if it gets merged upstream then there will be no need to keep this PR.

In the mean time, I noticed something interesting while I was trying to reproduce the issue in QEMU. It only happens if I explicitly use pci-ohci to add USB to the VM. If I use piix3-usb-uhci or qemu-xhci then I don't have this issue. I don't fully understand why that's possible, and I suspect that changing Xemu to use a different USB implementation wouldn't work (the kernel doesn't support USB 2.0 and later), but it may point to a better solution.

Issue #1234, opened 2022-08-15
USB pass through


GXTX:
### Feature Request

USB pass through is currently disabled and the docker image is missing libusb to be able to build for it.

### Additional Context

https://github.com/libusb/libusb/issues/692#issuecomment-584509601

PR #1532, opened 2023-09-23
Build: Add libusb support to builds

faha223:
Changed the meson.build: in the following ways
Changed the libusb dependency to have 'required' set to true if enabled, otherwise false. This ensures that the build system looks for the library but doesn't fail if the library isn't found.

Updated the dockerfile in the following ways:
Changed the mxe commit to the latest one because of an issue with zlib caused by a dead link
Added libusb1 to the mxe packages installed

Relevant issue:
https://github.com/xemu-project/xemu/issues/1234


faha223:
I converted this to a draft because the mac arm64 build is broken and I'm not sure how best to fix it.

This is the problem:
ld: warning: ignoring file /usr/local/Cellar/libusb/1.0.26/lib/libusb-1.0.dylib, building for macOS-arm64 but attempting to link with file built for macOS-x86_64

mborgerson:
> I converted this to a draft because the mac arm64 build is broken and I'm not sure how best to fix it.
> 
> This is the problem: ld: warning: ignoring file /usr/local/Cellar/libusb/1.0.26/lib/libusb-1.0.dylib, building for macOS-arm64 but attempting to link with file built for macOS-x86_64

xemu macos builds pull library dependencies from macports, both x86-64 and arm64 for cross compilation. You can try adding libusb here https://github.com/xemu-project/xemu/blob/master/scripts/download-macos-libs.py

faha223:
The the mac version is now building on both x86_64 and arm64 and the compile log indicates that libusb is enabled on both builds.

I verified that the Ubuntu Release build is working with USB Passthrough via the monitor. using the following commands with auto-bind off:
device_add usb-hub,port=1.3
device_add usb-host,vendorid=0x045e,productid=0x0289,port=1.3.1

The Windows version builds, but is waiting on an updated docker image with the libusb1 package installed before it'll build with libusb passthrough enabled

mborgerson:
You'll also want to update https://github.com/xemu-project/xemu/blob/master/debian/control

mborgerson:
We'll need to update the cross build container image tag after its done building

PR #1544, opened 2023-10-25
.clang-tidy: Add basic .clang-tidy for identifier naming

antangelo:
Adds a `.clang-tidy` config that covers most identifier naming conventions. Ideally we can offload this component of PR reviews to CI along with formatting in #1510 .

Unfortunately it doesn't seem like clang-tidy supports having a separate option for global static variables vs global non-static variables (from what I can tell the `StaticVariable` option only applies to locals), so one limitation of this config is that it will recommend the `g_` prefix for global static variables. The alternative is not to enforce a naming convention on the clang-tidy level, but we'd have to continue doing so manually in PR reviews.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,24 @@
+---
+Checks: '-*,readability-identifier-naming'
+WarningsAsErrors: true
+CheckOptions:
+  - key:    readability-identifier-naming.VariableCase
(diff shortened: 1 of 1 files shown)


mborgerson:
Let's give it a shot! Thanks. Sorry for the delay.

Issue #1563, opened 2023-11-27
Request: Option to open the emulator without automatically running the last game


Shideravan:
### Feature Request

I would like there to be an option for the emulator to load clean, without running the last software. This would make it easier to choose different games each time it is started.

BiatuAutMiahn:
like an `-eject_boot` parameter or something to start as if pressing the eject button on the console as compared to the power button.

mborgerson:
Sounds good. Send a PR @BiatuAutMiahn ?

BiatuAutMiahn:
`-dvd_path` would be the best option. See in https://xemu.app/docs/cli/

PR #1564, opened 2023-11-28
Handle tray eject/load via smc

BiatuAutMiahn:
As per #1563, allow user to specify `-no_dvd` to clear the dvd_path on startup.

### DIFF
--- a/subprojects/volk.wrap
+++ b/subprojects/volk.wrap
@@ -1,4 +1,4 @@
 [wrap-git]
 url = https://github.com/zeux/volk
-revision = 0b17a763ba5643e32da1b2152f8140461b3b7345
+revision = 3ca312a4f38baa63d8006b6905abbeeb89c8087d
 depth = 1


dracc:
Is this different from `-dvd_path ""`?

BiatuAutMiahn:
`-dvd_path ""` did not clear the path in my testing.

GXTX:
`-no-dvd` I'd assume we aren't attaching the drive (which could be useful for developers), not that we're claiming there's no disc inserted.

BiatuAutMiahn:
> `-no-dvd` I'd assume we aren't attaching the drive (which could be useful for developers), not that we're claiming there's no disc inserted.

It's only clearing the path. Maybe I should implement `-no_dvd_drive` for testing as you suggested?

```c
    if (strlen(dvd_path) > 0) {
        // Allow clearing the dvd path on boot.
        for (int i = 1; i < argc; i++) {
            if (argv[i] && strcmp(argv[i], "-no_dvd") == 0) {
                argv[i] = NULL;
                dvd_path = "";
                xemu_settings_set_string(&g_config.sys.files.dvd_path, "");
                break;
            }
        }
    }
```

BiatuAutMiahn:
How's this?

```c
    bool no_dvd_drive = false;
    bool no_dvd = false;

    // debug only.
#ifdef DEBUG_XEMU_C
    for (int i = 1; i < argc; i++) {
        if (argv[i] && strcmp(argv[i], "-no_dvd_drive") == 0) {
            argv[i] = NULL;
            no_dvd_drive = true;
            break;
        }
    }
#endif

    const char *dvd_path = g_config.sys.files.dvd_path;
    if (!no_dvd_drive) {
        // Allow clearing the dvd path on boot.
        for (int i = 1; i < argc; i++) {
            if (argv[i] && strcmp(argv[i], "-no_dvd") == 0) {
                argv[i] = NULL;
                no_dvd = true;
                dvd_path = "";
                xemu_settings_set_string(&g_config.sys.files.dvd_path, "");
                break;
            }
        }
        if (!no_dvd) {
            // Allow overriding the dvd path from command line
            for (int i = 1; i < argc; i++) {
                if (argv[i] && strcmp(argv[i], "-dvd_path") == 0) {
                    argv[i] = NULL;
                    if (i < argc - 1 && argv[i+1]) {
                        dvd_path = argv[i+1];
                        argv[i+1] = NULL;
                    }
                    break;
                }
            }
            if (strlen(dvd_path) > 0) {
                if (xemu_check_file(dvd_path) || strcmp(dvd_path, hdd_path) == 0) {
                    char *msg = g_strdup_printf("Failed to open DVD image file '%s'. Please check machine settings.", dvd_path);
                    xemu_queue_error_message(msg);
                    g_free(msg);
                    dvd_path = "";
                }
            }
        }

        // Always populate DVD drive. If disc path is the empty string, drive is
        // connected but no media present.
        fake_argv[fake_argc++] = strdup("-drive");
        char *escaped_dvd_path = strdup_double_commas(dvd_path);
        fake_argv[fake_argc++] = g_strdup_printf("index=1,media=cdrom,file=%s",
            escaped_dvd_path);
        free(escaped_dvd_path);
    }
```

mborgerson:
I thought an eject boot option was going to be added. This is different than starting with no dvd in the tray. `-dvd_path ""` should just load with no dvd in the tray (if it doesn't then it's a bug and should also be fixed).

BiatuAutMiahn:
> I thought an eject boot option was going to be added. This is different than starting with no dvd in the tray. `-dvd_path ""` should just load with no dvd in the tray (if it doesn't then it's a bug and should also be fixed).

I think this is what you had in mind

Edit:
also the `-dvd_path` bit is bugged. If any args follow the `-dvd_path`, they get appended to the dvd_path string. However the "" works fine just doesn't update the config.

hw/xbox/smbus_xbox_smc.c
```diff
@@ -264,6 +294,10 @@ static void smbus_smc_realize(DeviceState *dev, Error **errp)
     smc->cmd = 0;
     smc->error_reg = 0;
 
+    if (object_property_get_bool(qdev_get_machine(), "eject-after-boot", NULL)) {
+        smc->scratch_reg = SMC_REG_SCRATCH_EJECT_AFTER_BOOT;
```
antangelo: This and the one below it should be bitwise or operations (i.e. `|=`). As it is now, if you set both eject after boot and short animation, the first will not apply.

hw/xbox/smbus_xbox_smc.c
```diff
@@ -37,11 +37,24 @@
...
+
+#include "../../xemu-config.h"
+
+extern struct config g_config;
+
+#ifdef __cplusplus
+}
+#endif
+
+#define DEBUG
```
antangelo: DEBUG should not be set in-tree

hw/xbox/smbus_xbox_smc.c
```diff
@@ -37,11 +37,24 @@
...
 #define XBOX_SMC(obj) OBJECT_CHECK(SMBusSMCDevice, (obj), TYPE_XBOX_SMC)
 
-// #define DEBUG
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include "../../xemu-config.h"
+
+extern struct config g_config;
```
antangelo: This should just be included from `xemu-settings.h`, you should not have to define it again.

hw/xbox/smbus_xbox_smc.c
```diff
@@ -37,11 +37,24 @@
 #include "smbus.h"
 #include "sysemu/runstate.h"
 #include "hw/qdev-properties.h"
+#include "ui/xemu-settings.h"
 
 #define TYPE_XBOX_SMC "smbus-xbox-smc"
 #define XBOX_SMC(obj) OBJECT_CHECK(SMBusSMCDevice, (obj), TYPE_XBOX_SMC)
 
-// #define DEBUG
+#ifdef __cplusplus
```
antangelo: C++ guards are only necessary in header files, this file will never be built in a C++ context.

hw/xbox/smbus_xbox_smc.c
```diff
@@ -140,6 +157,19 @@ static int smc_write_data(SMBusDevice *dev, uint8_t *buf, uint8_t len)
         }
         break;
 
+    case SMC_REG_TRAYEJECT:
+        if (buf[0]) {
```
antangelo: Cromwell uses `0x0c01` to eject the disc tray. Do we know that other values of the second byte also have the same behavior?

antangelo (commented):
Please run clang-format on the code as per https://xemu.app/docs/dev/#contributing

hw/xbox/smbus_xbox_smc.c
BiatuAutMiahn: Good point, didn't realize that it was just unnecessary when we were only setting short animation.

hw/xbox/smbus_xbox_smc.c
BiatuAutMiahn: done

hw/xbox/smbus_xbox_smc.c
BiatuAutMiahn: At one point I ran into a linker issue and this was an attempt to resolve, it was indeed unnecessary.

hw/xbox/smbus_xbox_smc.c
BiatuAutMiahn: Not currently, no.

antangelo: Can you assert that the values are as expected then (i.e. that the buffer is either 0 or 1)?

hw/xbox/smbus_xbox_smc.c
```diff
@@ -140,6 +145,19 @@ static int smc_write_data(SMBusDevice *dev, uint8_t *buf, uint8_t len)
...
+            const char *path = g_config.sys.files.dvd_path;
+            qmp_blockdev_change_medium(true, "ide0-cd1", false, NULL, path,
+                                       false, "", false, false, false, 0,
+                                       &error);
+        } else {
+            xemu_settings_set_string(&g_config.sys.files.dvd_path, "");
```
antangelo: This breaks disc loading from the UI, because the Xbox software issues a tray eject in response to the eject button interrupt, after which this code ejects the disc that was just loaded. We can probably avoid this if we ignore the first eject write after loading a disc with the UI, @mborgerson WDYT?

hw/xbox/smbus_xbox_smc.c
```diff
@@ -24,19 +24,20 @@
...
-#include "qemu/config-file.h"
+#include "hw/qdev-properties.h"
 #include "qapi/error.h"
+#include "qemu/config-file.h"
+#include "qemu/option.h"
 #include "sysemu/block-backend.h"
 #include "sysemu/blockdev.h"
+#include "sysemu/runstate.h"
 #include "sysemu/sysemu.h"
+#include "ui/xemu-settings.h"
```
antangelo: Please revert the header order changes, some qemu headers depend on ordering. I think this is an issue with our clang-format config and I'll fix it.

hw/xbox/smbus_xbox_smc.c
BiatuAutMiahn: Afaik when you physically press the eject button on the console, and when software executes the eject routine the SMC_REG_TRAYEJECT is called and the smc tells the drive to load/eject the tray. the following byte is true if the tray has closed/loaded, and is false when the tray is ejected.

hw/xbox/smbus_xbox_smc.c
BiatuAutMiahn: done

hw/xbox/smbus_xbox_smc.c
BiatuAutMiahn: imo we should have the UI logic send smc command `0x0c00` for eject and `0x0c01` for load. Then in the UI toggle between a single Load/Eject option based on the tray state. And let the smc do the ejecting/loading of the image.

hw/xbox/smbus_xbox_smc.c
BiatuAutMiahn: per [https://xboxdevwiki.net/PIC](https://xboxdevwiki.net/PIC) they state:

```
Command | Description
0x01 | PIC version string counter reset
0x02 | reset and power off control
0x05 | power fan mode (0 = automatic; 1 = custom speed from reg0x06)
0x06 | power fan speed (0..~50)
0x07 | LED mode (0 = automatic; 1 = custom sequence from reg 0x08)
0x08 | LED flashing sequence
0x0C | tray eject (0 = eject; 1 = load)
0x0E | another scratch register? seems like an error code.
0x19 | reset on eject (0 = enable; 1 = disable)
0x1A | interrupt enable (write 0x01 to enable; can't disable onceenabled)
0x1B | scratch register for the original kernel
0x20 | response to PIC challenge (written first)
0x21 | response to PIC challenge (written second)
```
`0x0C | tray eject (0 = eject; 1 = load)`

antangelo: That is true if 0 or 1 is written to that byte, but there's nothing that stops software from writing other values and the current implementation will treat all such values as if they are 1. My thought is just that the code should be more specific about what it accepts (and I think an assert is fine as I doubt anything will actually send values other than 0 or 1, and if it did, the assert would indicate that it's worth researching further).

hw/xbox/smbus_xbox_smc.c
antangelo: This is more or less what the UI does, but it doesn't wait for the software to respond to the eject press and does the load at the same time. If you implemented it this way, you would have to track the tray state between the UI simulating the eject button press and tray close along with the corresponding software responses, which is more complicated. I also don't know if there's a reason the UI does everything inline other than this SMC logic not being implemented; if there is software that doesn't respond to SMC events, then disc loading would no longer work under this implementation.

I don't think the UX would be great if we required that the user eject the image through the UI before being able to load a new image and simulate the tray closing. The load button should handle everything necessary on its own.

mborgerson:
Please revert non-essential (eg formatting, .gitignore) changes

BiatuAutMiahn:
> Please revert non-essential (eg formatting, .gitignore) changes

done

bomkz:
@BiatuAutMiahn I'm no C++ developer so I may be wrong, but after trying to compile your PR as a test, I believe xbox_smc_tray_eject needs to be added to smbus.h, aside from this, without disabling the warning as error, it will not compile due to using Object variable as SMBusDevice argument in the last line of the function xbox_smc_tray_eject.

PR #1589, opened 2024-02-17
monitor: Add support for write command

jamesbrq:
Adds a new command to monitor named ```write|w```. The syntax is ```write addr size data```. This adds the capability to write up to 4 bytes of data at a time to the xbox.ram memory region.


mborgerson:
Thank you for the PR

I was certain this must exist in QEMU monitor already, and was totally surprised not to find it! I usually just use GDB to do this, but I think it would be nice for users to be able to easily poke memory in the monitor too, for playing with games and whatnot, and the feature is not too complex. In the future maybe we should just bake in a GDB client for people to use.

In the mean time, can you please expand the PR to support writing to both physical and virtual arbitrary addresses, regardless of any particular internal region. You can model this after the other related monitor / GDB stub commands.

This feature might also be good for upstream QEMU, if you're interested in submitting to qemu-devel.

jamesbrq:
I added an extra command, so there is one for virtual and one for physical now.

I'm assuming the syntax and everything is okay but if you think anything else needs to be changed let me know.
I'll also have to look into upstream qemu as well, but I'll see if I can get around to it. At the very least having it on xemu will be a nice addition. I can finally write to ram without having to pause the whole emulator lol

monitor/hmp-cmds.c
```diff
@@ -14,20 +14,18 @@
...
-#include "net/eth.h"
 #include "chardev/char.h"
-#include "sysemu/block-backend.h"
-#include "sysemu/runstate.h"
-#include "qemu/config-file.h"
-#include "qemu/option.h"
-#include "qemu/timer.h"
-#include "qemu/sockets.h"
-#include "qemu/help_option.h"
+#include "exec/memory.h"
```
mborgerson, line 18: Please revert any non-essential changes to this header order

softmmu/memory.c
```diff
@@ -25,15 +25,16 @@
...
 #include "exec/memory-internal.h"
 #include "exec/ram_addr.h"
+#include "hw/boards.h"
+#include "hw/core/cpu.h"
+#include "migration/vmstate.h"
+#include "qemu/accel.h"
 #include "sysemu/kvm.h"
 #include "sysemu/runstate.h"
 #include "sysemu/tcg.h"
-#include "qemu/accel.h"
```
mborgerson: Same comment as above re header order

softmmu/memory.c
```diff
@@ -3595,6 +3596,28 @@ void mtree_info(bool flatview, bool dispatch_tree, bool owner, bool disabled)
...
+{
+    MemoryRegion *sm = get_system_memory();
+    MemoryRegion *mr;
+    uint8_t *buf = ptr;
+    CPUState *cs = qemu_get_cpu(0);
+    if (is_physical) {
+        QTAILQ_FOREACH (mr, &sm->subregions, subregions_link) {
+            if (strcmp(memory_region_name(mr), "xbox.ram") == 0) {
+                uint8_t *ram_ptr = qemu_map_ram_ptr(mr->ram_block, addr);
+                memcpy(ram_ptr, buf, len);
```
mborgerson, line 3610: Is there not an existing physical memory write function somewhere? 

Anyway, we should check here that we don't accidentally write out of bounds

PR #1592, opened 2024-02-26
Build: Fix the build for macos

faha223:
Changed Darwin target for Arm64 to 21 because gettext-runtime doesn't support Darwin 20 for arm64 anymore
Added logic to attempt to use packages for darwin_any and noarch before giving up


scripts/download-macos-libs.py
```diff
@@ -44,6 +44,17 @@ def get_latest_pkg_filename_url(self, pkg_name):
...
+		
+		if len(pkgs) < 1:
+			pkgs = re.findall(pkg_name + r'[\w\.\-\_\+]*?\.darwin_any\.' + self._arch + r'\.tbz2', pkg_list)
+		if len(pkgs) < 1:
+			pkgs = re.findall(pkg_name + r'[\w\.\-\_\+]*?\.' + self._darwin_target + r'\.noarch\.tbz2', pkg_list)
+		if len(pkgs) < 1:
+			pkgs = re.findall(pkg_name + r'[\w\.\-\_\+]*?\.darwin_any\.noarch\.tbz2', pkg_list)
+		
+		if len(pkgs) < 1:
+			print(f'    [*] [ERROR] Unable to find version of {pkg_name} compatible with {self._darwin_target}.{self._arch}')
```
mborgerson, line 56: In this case, we will crash on line 58 below

faha223: That's the line it was crashing on before. Should the build continue after failing to retrieve one or more dependencies? I thought it would be better to fail earlier, with an error message detailing why.

mborgerson: No, we should not allow the code to just crash after detecting this case. Exiting with an error code is fine, and it looks like you've done this since I last commented, so thanks!

Issue #1492, opened 2023-07-21
Enable additional architectural optimizations in release


mborgerson:
### Feature Request

* We should optimize releases to run well on the majority of target machines. Thankfully the [Steam Hardware Survey](https://store.steampowered.com/hwsurvey/Steam-Hardware-Software-Survey-Welcome-to-Steam) can enlighten us to which architectural features likely xemu users have.
   * Consider creating an xemu hardware survey to get a more accurate picture of what machines xemu is actually running on.
* xemu does already opportunistically generate guest emulation code using hw simd features, but is otherwise compiled without additional architecture flags, with exception of macOS x86-64 builds which currently targets IvyBridge.
* Default architecture features can be listed with `gcc -Q --help=target`. Looks like for x86-64, SSE2 is enabled by default. We probably want to enable SSE4.2/AVX2/etc.
* We should be careful not to force these compiler optimizations when building from source, e.g. #1485.
* Consider providing legacy hardware releases for older machines.
   * Consider macOS Universal style multiversion binary.
* Add startup feature check that provides a helpful error message--not crashing on unsupported instruction.

insaneninja117:
When it comes to supported instruction sets:

- The very first iterations of AMD Athlon 64 have SSE2 as their maximum
- Later AMD Athlon 64 and All Intel 64 Capable CPUS support up to SSE3
- First iteration of Core 2 Duo and Core 2 Quad introduced SSSE3
- Second Iteration of Core 2 Duo and Core 2 Quad introduced SSE4.1
- All Intel Core ix and AMD FX and newer CPUs support SSE4.2
- Intel Sandy Bridge (Core ix 2xxx series) and AMD FX (Bulldozer) introduced AVX
- AVX2 introduced with Intel Haswell (Core ix 4xxx series) and final iteration of AMD FX (Piledriver architecture)

Bumping up the minimum instruction set from SSE2 to 3 would lose virtually zero users (but the gains would likely be small), SSE4.2 has been around for almost 15 years now and AVX2 10 so it is a good idea to introduce these optimisations. But if it is of little trouble then keeping a lower end .exe would be best.

Spidy123222:
> When it comes to supported instruction sets:
> 
> * The very first iterations of AMD Athlon 64 have SSE2 as their maximum
> * Later AMD Athlon 64 and All Intel 64 Capable CPUS support up to SSE3
> * First iteration of Core 2 Duo and Core 2 Quad introduced SSSE3
[...]

PR #1608, opened 2024-03-19
Faster swizzle code

NZJenkins:
This speeds up swizzling/unswizzling textures.

`fill_pattern` showed up in profiling Fable: The Lost Chapters.

~~It seems to implement the equivalent of the `pdep` instruction.
It turns out there is a slightly faster method to do this in a fixed number of steps.
This increases fps a bit in certain scenes in Fable TLC.~~

~~Also use native `pdep` if available.~~

Instead of calculating these offsets from masks, we can increment through the offsets.
See https://fgiesen.wordpress.com/2011/01/17/texture-tiling-and-swizzling/

I wrote a small testbench for different methods to ensure it's working the same way (not integrated into xemu)
It should be around 30X faster and is matching both swizzle/unswizzle behaviour of the original code.
```
width: 256 (67MB)
height: 256
depth: 256
bpp: 4
iterations: 10
...
simple_avg_ms: 1284.55 (old code)
...
fabien_avg_ms: 36.2146  (new code)
```

Gives a few more FPS in certain situations in Fable TLC:
<details>
  <summary>Fable TLC FPS comparison</summary>
  
Before:
master

After:
quick_swizzle

</details>

### DIFF
--- a/hw/xbox/nv2a/pgraph/swizzle.c
+++ b/hw/xbox/nv2a/pgraph/swizzle.c
@@ -26,10 +26,23 @@
 
 #include "swizzle.h"
 
-/* This should be pretty straightforward.
- * It creates a bit pattern like ..zyxzyxzyx from ..xxx, ..yyy and ..zzz
- * If there are no bits left from any component it will pack the other masks
- * more tighly (Example: zzxzxzyx = Fewer x than z and even fewer y)
+/*
+ * Helpers for converting to and from swizzled (Z-ordered) texture formats.
+ * Swizzled textures store pixels in a more cache-friendly layout for rendering
+ * than linear textures.
+ * Width, height, and depth must be powers of two.
+ * See also:
+ * https://en.wikipedia.org/wiki/Z-order_curve
+ */
+
+/*
+ * Create masks representing the interleaving of each linear texture dimension (x, y, z).
+ * These can be used to map linear texture coordinates to a swizzled "Z" offset.
+ * For example, a 2D 8x32 texture needs 3 bits for x, and 5 bits for y:
+ * mask_x:  00010101
+ * mask_y:  11101010
+ * mask_z:  00000000
+ * for "Z": yyyxyxyx
  */
 static void generate_swizzle_masks(unsigned int width,
                                    unsigned int height,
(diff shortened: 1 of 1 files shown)


hw/xbox/nv2a/swizzle.c
```diff
@@ -55,32 +59,59 @@ static void generate_swizzle_masks(unsigned int width,
...
+        mp = mp ^ (mp << 2);
+        mp = mp ^ (mp << 4);
+        mp = mp ^ (mp << 8);
+        mp = mp ^ (mp << 16);
+        mv = mp & mask; // Bits to move.
+        expand_mask->moves[i] = mv;
+        mask = (mask ^ mv) | (mv >> (1 << i)); // Compress m.
+        mk = mk & ~mp;
+    }
+
```
mborgerson: nit: space

hw/xbox/nv2a/swizzle.c
```diff
@@ -93,8 +124,11 @@ void swizzle_box(
     unsigned int slice_pitch,
     unsigned int bytes_per_pixel)
 {
-    uint32_t mask_x, mask_y, mask_z;
-    generate_swizzle_masks(width, height, depth, &mask_x, &mask_y, &mask_z);
+    expand_mask mask_x, mask_y, mask_z;
+    generate_swizzle_masks(width, height, depth, &mask_x.mask, &mask_y.mask, &mask_z.mask);
+    generate_expand_mask_moves(&mask_x);
```
mborgerson: Calculation unnecessary in case of pdep instruction?

hw/xbox/nv2a/swizzle.c
```diff
@@ -55,32 +59,59 @@ static void generate_swizzle_masks(unsigned int width,
...
 }
 
-/* This fills a pattern with a value if your value has bits abcd and your
- * pattern is 11010100100 this will return: 0a0b0c00d00
+/* Move low bits to positions given by a mask, while keeping bits in order.
+ * e.g.
+ * expand(0000abcd, 10011010) = a00bc0d0.
+ *
+ * Implementation from Hacker's delight chapter 7 "Expand"
+ * https://stackoverflow.com/questions/77834169/what-is-a-fast-fallback-algorithm-which-emulates-pdep-and-pext-in-software
```
mborgerson: Citing the book should be enough, don't need the stackoverflow.com link

hw/xbox/nv2a/swizzle.c
```diff
@@ -55,32 +59,59 @@ static void generate_swizzle_masks(unsigned int width,
...
+        mv = mp & mask; // Bits to move.
+        expand_mask->moves[i] = mv;
+        mask = (mask ^ mv) | (mv >> (1 << i)); // Compress m.
+        mk = mk & ~mp;
+    }
+
+}
+
+static uint32_t expand(uint32_t x, expand_mask* expand_mask) {
+    #ifdef __BMI2__
```
mborgerson: nit: Readability would be better with these macros unindented

mborgerson (commented):
Looks like a big improvement in this swizzle logic performance, nice! In addition to the minor feedback on code, a couple things:
* Similar to the benchmark, please also test to confirm that the swizzle/unswizzle results are the same in all cases: current impl, new faster sw impl, and new instruction impl
* Please determine how well this instruction will be supported on user systems. Find out which generation of x86 processors includes this instruction, and if it is relatively commonplace now, confirm that our release builds are actually building for it (related: #1492)

hw/xbox/nv2a/swizzle.c
```diff
@@ -55,32 +59,59 @@ static void generate_swizzle_masks(unsigned int width,
...
-            result |= value & 1 ? bit : 0;
-            value >>= 1;
-        }
-        bit <<= 1;
+typedef struct expand_mask {
+    uint32_t mask;
+    uint32_t moves[5];
+} expand_mask;
+
+static void generate_expand_mask_moves(expand_mask* expand_mask) {
```
mborgerson: nit: Asterisk goes next to identifier

mborgerson:
Sorry it took me a while to get around to this. I've rebased and force-pushed to your feature branch. After some testing  and benchmarking, I measured a 15x speedup (0.15 GiB/s &rarr; 2.5 GiB/s). Great job! Do you want to add your name to the copyright header?

mborgerson (approved):

mborgerson:
Thanks again!

PR #1613, opened 2024-03-23
Update metainfo.xml information

j8r:
Some new information is needed by the appstream linter.

metainfo.xml extension is preferable over appdata.xml, see https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html#spec-appdata-introduction.

This fixes the Flatpak build.

### DIFF
--- a/xemu.metainfo.xml
+++ b/xemu.metainfo.xml
@@ -46,4 +46,7 @@
   <provides>
     <binary>xemu</binary>
   </provides>
+  <developer id="xemu.app">
+    <name>xemu project</name>
+  </developer>
 </component>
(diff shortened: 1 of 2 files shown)


xemu.metainfo.xml
```diff
@@ -46,4 +46,7 @@
   <provides>
     <binary>xemu</binary>
   </provides>
+  <developer id="xemu.app">
+    <name>Matt Borgerson</name>
```
mborgerson: Let's just say "xemu project" here instead

mborgerson:
Thanks!

Issue #339, opened 2021-06-06
Jet Set Radio Future: Audio Crackling


RedxYeti:
https://xemu.app/titles/4d53003d/
https://xemu.app/titles/49470018/

I tested this on both to try and rule out a bad rip of my games, and was able to reproduce the audio distortion in both versions.

### Bug Description
The audio cracks/distorts during specific times, most notably right after jumping.
It also happens if you skate through a large crowd of pedestrians

Steps to reproduce this issue:
1. Load up a character and jump, or skake through a large group of people

I can upload a short video if need be.

### Version

Version: 0.5.2-26-g0a8e9b8db3
Branch:  master
Commit:  0a8e9b8db31ee120428246265ebbf743088b429f
Date:    Fri Jun  4 19:15:47 UTC 2021

### System Information

| OS			| win10
| CPU			| i9 9900k@4.9ghz
| Graphics Device	| 2080ti fe
| Graphics Driver	| Nvidia 466.63

### Additional Context
Im using the NA versions of each game posted

MasonT8198:
This is a known issue in the compatibility report, however thank you for reporting this as it helps track the issues!

RedxYeti:
Oh sorry you're right. I started with the dual game and didn't read the issues on the standalone game. My bad.

HelgenX:
I'm just bumping this out of curiosity of there's any updates on this, because the game honestly runs better than the official Xbox lol, but the audio crackling/skipping is the only thing that's a problem for me. Amazing work thus far, nonetheless!

MasonT8198:
> I'm just bumping this out of curiosity of there's any updates on this, because the game honestly runs better than the official Xbox lol, but the audio crackling/skipping is the only thing that's a problem for me. Amazing work thus far, nonetheless!

No there haven't been any updates, however when matt or someone else gets to making audio improvements this should alleviate the crackling issue

HelgenX:
> > I'm just bumping this out of curiosity of there's any updates on this, because the game honestly runs better than the official Xbox lol, but the audio crackling/skipping is the only thing that's a problem for me. Amazing work thus far, nonetheless!
> 
> No there haven't been any updates, however when matt or someone else gets to making audio improvements this should alleviate the crackling issue

Awesome, thanks for the acknowledgement and great work!

mborgerson:
When it is fixed this issue will be closed and you'll get a notification. There are higher priority tasks to work on at the moment, but it'll definitely get fixed.

[...]

PR #1616, opened 2024-03-23
mcpx: Mitigate ADPCM audio crackling

kumpulak:
Adds a workaround for audio crackling that frequently occurs when non-streaming ADPCM voices are requested to be deactivated before finishing using `NV1BA0_PIO_VOICE_OFF.` Fixes the issue #339 most likely similar issues with other games too.

I noticed the `NV1BA0_PIO_VOICE_OFF` arrives to the `vp_write` callback after something has already written garbage data to the voice processor memory region. At that point it would not be possible to retroactively clear the output buffers, since the voices have already been mixed together. So I created a workaround which detects this garbage data before it is decoded and written to sample buffers.

This will have a very minimal performance impact (could not notice at all when profiling). It is however a hack around a root issue I do not have the expertise to debug. In ideal scenario the `NV1BA0_PIO_VOICE_OFF` would arrive in time before the memory contains this garbage data. 

I tested the fix with USA version of standalone Jet Set Radio Future, where both mono and stereo ADPCM voices exhibited this issue. Now both are fixed and there are no other downsides that I could find. I can't come up with a scenario where a valid ADPCM header would contain all identical non-zero values and still represent audio someone wants to hear 🙂

I tested the code on Ubuntu 22.04 (x86), Windows 11 (x86) and M1 Mac (arm64), identical behavior on those platforms.

### DIFF
--- a/hw/xbox/mcpx/apu.c
+++ b/hw/xbox/mcpx/apu.c
@@ -1981,6 +1981,21 @@ static int voice_get_samples(MCPXAPUState *d, uint32_t v, float samples[][2],
                             ldl_le_phys(&address_space_memory, addr);
                         linear_addr += 4;
                     }
+                    /* WAR: Deactivate voice if ACPM header values are non-zero
+                     * and identical. Something overwrites voice memory region
+                     * before NV1BA0_PIO_VOICE_OFF is set. Mitigates loud
+                     * crackling produced by decoding/playing such data.
+                     */
+                    if (adpcm_block[0] != 0) {
+                        uint32_t diff = 0;
+                        for (uint8_t i = 1; i < 8; i++) {
+                            diff |= adpcm_block[i] ^ adpcm_block[0];
+                        }
+                        if (diff == 0) {
+                            voice_off(d, v);
+                            return -1;
+                        }
+                    }
                 }
                 adpcm_decode_block(adpcm_decoded, (uint8_t *)adpcm_block,
                                    block_size, channels);


MasonT8198:
The logic/code seem sound to me, I can't think of any case where it would end up muting any audio that would be intended to be heard by the player...

Although as another test case I'd be interested to see if it also applies to https://youtu.be/t08VA1_sflM?t=27 which affects for PGR1/2 on multi car collision.

kumpulak:
> The logic/code seem sound to me, I can't think of any case where it would end up muting any audio that would be intended to be heard by the player...
> 
> Although as another test case I'd be interested to see if it also applies to https://youtu.be/t08VA1_sflM?t=27 which affects for PGR1/2 on multi car collision.

The RalliSport Challenge seems to use normal PCM instead of ADPCM for the engine sounds/other effects so this doesn't fix anything there. I took a quick look and it seems all of the different engine sounds are playing at the same time, no wonder it sounds horrible when they are mixed together 😅. Maybe the developers were misusing the APU hardware in a way that is not currently supported.

mborgerson:
Thank you for the PR, I appreciate the debugging effort and the provided workaround. However, I think we should really avoid merging such workarounds and try to understand and fix the problem properly instead of patching around it

user18081972:
This is currently the best way to play Jet Set Radio Future, so i would like to use [this PR](kumpulak:mitigate-adpcm-crackling)
Any chance someone could upload a compiled version of it here?

EntranceJew:
> This is currently the best way to play Jet Set Radio Future, so i would like to use this PR Any chance someone could upload a compiled version of it here?

I tried to create some builds of my own for Windows using this but they completely fail both in github's CI and when I follow the build steps under WSL (it builds for Ubuntu fine on my native linux box) -- although I don't know how this code was tested if github's actions can't complete the build process with any version of the toolchain I've forced it to utilize. If you just want an arm64 ubuntu build I can get you that.

The problem I'm getting both locally under WSL and on github's actions:
```
[891/1500] Compiling C object libchardev.fa.p/chardev_char.c.obj
[892/1500] Compiling Windows resource version.rc
FAILED: version.rc_version.o
/opt/mxe/usr/bin/x86_64-w64-mingw32.static-windres -I./ -I/xemu/. -I/xemu/build/. ../version.rc version.rc_version.o --preprocessor-arg=-MD --preprocessor-arg=-MQversion.rc_version.o --preprocessor-arg=-MFversion.rc_version.o.d
/opt/mxe/usr/bin/x86_64-w64-mingw32.static-windres: ../version.rc:6: syntax error
[893/1500] Generating i386-softmmu-gdbstub-xml.c with a custom command (wrapped by meson to capture output)
[894/1500] Compiling C object libchardev.fa.p/chardev_char-console.c.obj
[895/1500] Compiling C object libchardev.fa.p/chardev_char-win.c.obj
[896/1500] Compiling C object libchardev.fa.p/chardev_char-win-stdio.c.obj
[897/1500] Compiling C object libhwcore.fa.p/hw_core_bus.c.obj
[898/1500] Compiling C object libhwcore.fa.p/hw_core_vmstate-if.c.obj
[899/1500] Compiling C object libhwcore.fa.p/hw_core_reset.c.obj
[900/1500] Compiling C object libhwcore.fa.p/hw_core_qdev-properties.c.obj
[901/1500] Compiling C object libhwcore.fa.p/hw_core_irq.c.obj
[902/1500] Compiling C object libhwcore.fa.p/hw_core_resettable.c.obj
[903/1500] Compiling C object libhwcore.fa.p/hw_core_hotplug.c.obj
[904/1500] Compiling C object libhwcore.fa.p/hw_core_clock.c.obj
[905/1500] Compiling C object libhwcore.fa.p/hw_core_qdev-clock.c.obj
[906/1500] Compiling C object libhwcore.fa.p/hw_core_qdev.c.obj
[907/1500] Compiling C object libhwcore.fa.p/hw_core_qdev-hotplug.c.obj
ninja: build stopped: subcommand failed.
make[1]: *** [Makefile:165: run-ninja] Error 1
make[1]: Leaving directory '/xemu/build'
make: *** [GNUmakefile:11: qemu-system-i386w.exe] Error 2

real    1m0.438s
user    12m45.901s
sys     2m18.179s
```

This is even [after merging master into a copy of this branch](https://github.com/EntranceJew/xemu/tree/mitigate-adpcm-crackling), and trying with all these versions of the toolchain image:
- sha-8152913
- sha-8ef45cc
- sha-158cc0d
- sha-d0d3e7b

As an aside, I understand the motivation for not wanting to compromise on the accuracy of the project, but I don't think it would hurt to implement this behavior behind a CLI switch argument until someone _does_ actually figure out what is going on with the ADPCM problems either in code or by studying an xbox under a microscope and finding out that player 1's controller completes a physical circuit which suppresses this behavior.

user18081972:
> If you just want an arm64 ubuntu build I can get you that.

I dont have any arm64 hardware to play it on.

@kumpulak mentioned in the original post that he tested it on x86 Windows, so it must be possible to compile it. Maybe he is willing to share his builds with this PR applied.

socialenigma-og:
This commit fixes stereo separation/mixing in BLiNX: The Time Sweeper, when real-time DSP is enabled.
Test case: Approach one of the enemies - Chrono Blob - and rotate the camera while (it) bounces.

(wearing head or earphones makes it easier to perceive)

bigleanator:
For anyone else who is having trouble or (can't be bothered) compiling. I got github actions to work correctly on [the fork that's on my account](https://github.com/bigleanator/xemu-jsrf-fix). It incorporates this change while being up to date with upstream. Ill try keep it updated (no promises).

PR #1617, opened 2024-03-24
Net: Search for npcap instead of winpcap based files

MasonT8198:
If you install NPcap without checking the install in winpcap api-compliant mode, it will not use wpcap.dll and instead use its default packet.dll for the lib, and not populate wpcap.dll.

Edit: My apologies, the xemu cross-compiler docker container hogs about 12.5 gbs of ram on my laptop causing it to crawl, so I'm using the current runners to generate the working build!

### DIFF
--- a/winpcap-loader/loader.c
+++ b/winpcap-loader/loader.c
@@ -21,7 +21,7 @@
 #include <pcap/pcap.h>
 #include <string.h>
 
-const char *lib_not_loaded_err = "winpcap library is not loaded";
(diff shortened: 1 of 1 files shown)


MasonT8198:
Turns out my initial prognosis of why this doesn't work was incorrect... I'll continue to diagnose and come up with a solution... this current impl works however is hit or miss due to the differences in non-winpcap api mode

MasonT8198:
Remarking as draft until I can think of a better way to tackle this issue.

MasonT8198:
Closing until I can think of a better way to handle this issue.

Issue #524, opened 2021-11-02
Halo 2: Flickering/depth issues on Apple M1 Macs


mborgerson:
### Title

https://xemu.app/titles/4d530064/#Halo-2

### Bug Description

Distinct from already known depth issues with Halo 2.

Possibly related to #523 

Main menu blue overlay is broken, scene elements flicker:

![Screen Shot 2021-11-02 at 12 52 39 AM](https://user-images.githubusercontent.com/8210/139806894-5805e747-98cf-4ad0-a663-42bb54dce656.png)

Chief arm/hand model flickers:

![Screen Shot 2021-11-02 at 12 58 16 AM](https://user-images.githubusercontent.com/8210/139807718-7b407d9d-7f25-4422-ab15-779392afa99d.png)

### Expected Behavior

<FIXME>

### xemu Version

v0.6.1-33-gbbcd5763e6

### System Information

2020 M1 Mac mini
macOS 11.6

camden-smallwood:
I decided to take a look at this over the weekend and was able to implement something that fixed a large chunk of the missing UI overlays on my M1 Mac mini. This does not address the flickering meshes or transparent meshes issues (which I believe to be related to a precision issue). There may be other issues that were addressed by the fix, these were just the things I noticed offhand.

The fix can be applied by replacing these lines in [./hw/xbox/nv2a/vsh.c#L852-L854](https://github.com/xemu-project/xemu/blob/94d826a4f125d755d6d37069ad7084bfde33d650/hw/xbox/nv2a/vsh.c#L852-L854) with the following:

```C
"  if (clipRange.y > clipRange.x)\n"
"    oPos.z = (oPos.z - clipRange.x)/(0.5 * (clipRange.y - clipRange.x)) - 1;\n"
"  else\n"
"    oPos.z = (oPos.z - clipRange.y)/(0.5 * (clipRange.x - clipRange.y)) - 1;\n"/
```

### xemu Version
v0.7.120

### System Information
2020 M1 Mac mini
macOS 14.1.1

### Current Behavior

Missing blue background on mainmenu:
![image](https://github.com/xemu-project/xemu/assets/52939755/28734483-f110-4870-b11e-8a138bfefc43)

Missing profile icons:
![image](https://github.com/xemu-project/xemu/assets/52939755/dbbb6cb5-3acc-4541-9d00-4b5050744b34)

Missing UI elements:
[...]

Issue #1223, opened 2022-08-05
Almost Black Screen on Mac M1/Steam Deck (rdna2) and Intel iGPU -  Burnout3


lehrbua:
### Title

https://xemu.app/titles/4541005b/#Burnout-3-Takedown

### Bug Description

Only Race is visible and car select screen without hud
Whole Menu + Intro is black.

### Expected Behavior

Game is visible.

### xemu Version

0.7.67

### System Information

OS Platform:  macOS
OS Version:   Version 12.4 (Build 21F79)
Manufacturer: Apple
GPU Model:    Apple M1 Max
Driver:       4.1 Metal - 76.3
Shader:       4.10

Triticum0:
Please have a title

Triticum0:
Note: this only affects M1 mac, played this game on Nvidia GPU works fine

Spidy123222:
This also affects Steamdecks and intel HD graphics to my knowledge. Also, you need a proper title for the issue as no one will know what this is or is a mistake message. 

Referenced issue. https://github.com/mborgerson/xemu/issues/571

Spidy123222:
> Note: this only affects M1 mac, played this game on Nvidia GPU works fine

This also affects Steamdecks (I asked some steamdeck users to test in emudeck discord) and intel HD graphics to my knowledge. Also, they need a proper title for the issue as no one will know what this is or is a mistake message. 

Referenced issue of similar issue. https://github.com/mborgerson/xemu/issues/571

zaylivinglive:
Confirming that this issue is also on steam deck. Starting the game will have the loading screen and bar, and after that intro and menu is completely black with only the background music playing.

deefha:
Confirming @zaylivinglive report on my Steam Deck too (xemu 0.7.70). Can I help somehow? I am here if more info is needed.

OutRunner:
Could this be the same issue that Ninja Gaiden Black has? [https://github.com/mborgerson/xemu/issues/991](https://github.com/mborgerson/xemu/issues/991)

TaraSophieDev:
Having the same Issue on my Steam deck, the Loading Bar is showing but after that, there is only black screen, everything else works tho.

mbriar:
Can someone on Steam Deck try setting `vs_position_always_invariant=true` as environment variable or someone on mac this patch:  
```
diff --git a/hw/xbox/nv2a/shaders.c b/hw/xbox/nv2a/shaders.c
[...]

PR #1620, opened 2024-03-26
nv2a: Fix vertex Z calculation using clip range

camden-smallwood:
As mentioned in [this comment](https://github.com/xemu-project/xemu/issues/524#issuecomment-2019334483), this fix restores a number of missing UI elements in Halo 2 on Mac.

### System Information
2020 M1 Mac mini
macOS 14.1.1

### Fixed Behavior

Restored blue background on mainmenu:
image

Restored profile icons:
image

Restored UI elements:
image

### DIFF
--- a/hw/xbox/nv2a/vsh.c
+++ b/hw/xbox/nv2a/vsh.c
@@ -849,8 +849,10 @@ void vsh_translate(uint16_t version,
         mstring_append(body, "  oPos.z = oPos.w;\n");
     }
     mstring_append(body,
-        "  if (clipRange.y != clipRange.x) {\n"
-        "    oPos.z = (oPos.z - clipRange.x)/(0.5*(clipRange.y - clipRange.x)) - 1;\n"
+        "  if (clipRange.y > clipRange.x) {\n"
+        "    oPos.z = (oPos.z - clipRange.x)/(0.5 * (clipRange.y - clipRange.x)) - 1;\n"
+        "  } else {\n"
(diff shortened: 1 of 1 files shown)


Spidy123222:
This Pr fixes other games also it seems like burnout 3's menus (pitch black for most part in xemu before). I will have to test this on my steam deck as it also has these issues as from what others saw that it mostly was on integrated graphics devices or certain AMD gpu's https://github.com/xemu-project/xemu/issues/1223. 
image
image

Spidy123222:
so I tested the pull request on steam deck. doesn't fix it there unfortunately.

HadetTheUndying:
> so I tested the pull request on steam deck. doesn't fix it there unfortunately.

Does the SteamDeck issue persist if you switch to zink as your GL driver? I have a feeling this is still that same mesa issue that we've been dealing with for over a year now.

Spidy123222:
> > so I tested the pull request on steam deck. doesn't fix it there unfortunately.
> 
> 
> 
> Does the SteamDeck issue persist if you switch to zink as your GL driver? I have a feeling this is still that same mesa issue that we've been dealing with for over a year now. 

Zink just breaks the graphics nowadays. Not sure why and this doesn't change that either. Both macOS and steam deck can be fixed and broken at same time with an environment variable that's not correct solution. As Matt said above the pr change could be wrong.

ws141:
> so I tested the pull request on steam deck. doesn't fix it there unfortunately.

Thanks for reporting I'm trying to get Burnout 3 going on my Steam Deck as well. ☹️

Fabxx:
i've tested the first splinter cell with this PR, unfortunately you can still see through walls and objects

Issue #354, opened 2021-06-18
Crashes without an available audio device


mborgerson:
### Bug Description
If the system does not have an available audio device, xemu will bail after failed initialization.

Steps to reproduce this issue:
1. Have a system with no available output device
2. Start xemu

### Expected Behavior

xemu should boot, without audio.

### Version

* Version: a7f3f8f21221174efb65fdc0b8f7e2dd234c7412
* Last-known Working: N/A

### System Information

N/A

### Additional Context

N/A

GXTX:
Something like this might work putting after the initial SDL init.

```c
    if (SDL_GetNumAudioDevices(0) == 0) {
        SDL_QuitSubSystem(SDL_INIT_AUDIO);

        if (SDL_AudioInit("dummy") < 0) {
            fprintf(stderr, "Failed to initalize 'dummy' audio device\n");
            exit(1);
        }
    }
```

abaire:
Marking as related to the general request for a null audio backend in #2331

Presumably if we had a null audio option, we'd just default to that if no device is detected.

PR #1623, opened 2024-03-28
apu.c: Kill audio subsystem if there are no playback devices

MasonT8198:
fixes issue #354

@GXTX 's initial hypothesis was correct, simply killing the audio subsystem if there are no available audio devices does work...

in the event the dummy audio device does fail to load, assert and kill the program, however in my testing this *shouldn't* occur

### DIFF
--- a/hw/xbox/mcpx/apu.c
+++ b/hw/xbox/mcpx/apu.c
@@ -2631,6 +2631,16 @@ void mcpx_apu_init(PCIBus *bus, int devfn, MemoryRegion *ram)
         exit(1);
     }
 
+    if (SDL_GetNumAudioDevices(0) == 0) {
+        SDL_QuitSubSystem(SDL_INIT_AUDIO);
+        
+        if (SDL_AudioInit("dummy") < 0) {
+            fprintf(stderr, "Failed to initalize 'dummy' audio device\n");
+            assert(!"SDL_AudioInit failed!");
+            exit(1);
+        }
+    }
+    
     SDL_AudioDeviceID sdl_audio_dev;
     sdl_audio_dev = SDL_OpenAudioDevice(NULL, 0, &sdl_audio_spec, NULL, 0);
     if (sdl_audio_dev == 0) {


antangelo:
I tried testing this on linux by stopping pipewire, but the assertion still trips. `SDL_GetNumAudioDevices(0)` reports value `2`.

As an aside, what happens if there is no audio device on loading xemu, but then one is connected after the fact? Is xemu able to output to the newly connected device?

MasonT8198:
> I tried testing this on linux by stopping pipewire, but the assertion still trips. `SDL_GetNumAudioDevices(0)` reports value `2`.
> 
> 
> 
> As an aside, what happens if there is no audio device on loading xemu, but then one is connected after the fact? Is xemu able to output to the newly connected device?

I've been unable to repro on my machine as to that specific assertion, even so, I'm not sure why'd it report 2... unless it was grabbing input devices that it couldn't access.

I'm unsure as to how SDL handles that, in most games that use SDL for audio, basically always has to restart to detect new devices.. could be wrong.

As an aside, it's been a minute since I looked at the xemu UI, Using SDL for an audio device selector would be nice.

mborgerson:
> I tried testing this on linux by stopping pipewire, but the assertion still trips. SDL_GetNumAudioDevices(0) reports value 2.

In this case, how about we detect the failed open, then fall back to re-open on the dummy driver

> As an aside, what happens if there is no audio device on loading xemu, but then one is connected after the fact? Is xemu able to output to the newly connected device?

This would be nice to have, but not necessarily in this PR

MasonT8198:
> > I tried testing this on linux by stopping pipewire, but the assertion still trips. SDL_GetNumAudioDevices(0) reports value 2.
> 
> 
> 
> In this case, how about we detect the failed open, then fall back to re-open on the dummy driver
> 
> 
> 
> > As an aside, what happens if there is no audio device on loading xemu, but then one is connected after the fact? Is xemu able to output to the newly connected device?
> 
> 
> 
> This would be nice to have, but not necessarily in this PR

I'm not entirely sure how to do either of these with the current impl of the subsystem.

Also, reopening the audio subsystem on reconnection on all platforms seems to have to have some instance of the subsystem being open...

I don't want to periodically check for audio... possibly maybe an audio menu in the UI with the default option of NULL audio devices on launch may be pertinent... and upon selection it reenables the audio subsys, otherwise you'll add extra overhead periodically trying to scan for audio devices

MasonT8198:
Closing for now.

PR #1627, opened 2024-03-30
UI: Add games directory and allow switching of games via controller in popup-menu

TrentSeed:
At the moment, switching between games in Xemu typically requires using a mouse/keyboard. When using a game controller, it's possible to open the popup menu, and select "Load Disc...", however that launches a native file browser which does not have controller input support, requiring use of a mouse/keyboard to select the target file. The goal of this PR is to make switching between games a bit more convenient and possible using a controller.

This PR introduces a new setting called "Games directory" within Settings > General > Miscellaneous. By default there is no value specified, and when clicked the user is presented with a file browser to select the desired directory containing .iso game files.

When the popup menu is activated (via right click, or start/back from an Xbox Controller S), there is now a "Games" submenu button that when selected, will present the user with a sorted list of games based on the configured directory. Selecting a game will result in Xemu loading the corresponding disc file. In the event no .iso files are detected in the directory, or if the directory is still undefined, there is an option "No games found" that when selected will open the main settings UI so that the user may adjust the games directory as needed.

All Settings

Root Popup Menu

Games Popup Menu - With .iso Files Found

Games Popup Menu - No Files Found / Directory Not Set

### DIFF
--- a/ui/xui/popup-menu.cc
+++ b/ui/xui/popup-menu.cc
@@ -17,21 +17,21 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 //
 #include "ui/xemu-notifications.h"
-#include "popup-menu.hh"
-#include "../xemu-snapshots.h"
-#include "IconsFontAwesome6.h"
-#include "actions.hh"
-#include "font-manager.hh"
-#include "input-manager.hh"
-#include "main-menu.hh"
-#include "misc.hh"
-#include "scene-manager.hh"
-#include "viewport-manager.hh"
-#include "xemu-hud.h"
-#include <filesystem>
-#include <map>
 #include <string>
 #include <vector>
+#include <filesystem>
+#include <map>
+#include "misc.hh"
+#include "actions.hh"
+#include "font-manager.hh"
+#include "viewport-manager.hh"
+#include "scene-manager.hh"
+#include "popup-menu.hh"
+#include "input-manager.hh"
+#include "xemu-hud.h"
+#include "IconsFontAwesome6.h"
+#include "../xemu-snapshots.h"
+#include "main-menu.hh"
 
 PopupMenuItemDelegate::~PopupMenuItemDelegate() {}
 void PopupMenuItemDelegate::PushMenu(PopupMenu &menu) {}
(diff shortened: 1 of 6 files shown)


AMS-ABSTAND (approved):

ui/xui/popup-menu.cc
```diff
@@ -338,9 +340,58 @@ class SettingsPopupMenu : public virtual PopupMenu {
...
+                if (std::filesystem::is_regular_file(entry_path) &&
+                    entry_path.extension() == ".iso") {
+                    sorted_file_names.insert(
+                        { entry_path.stem().string(), entry_path });
+                }
+            }
+        }
+
+        for (const auto &[label, file_path] : sorted_file_names) {
+            if (PopupMenuButton(label, ICON_FA_GAMEPAD)) {
```
mborgerson: Should probably use the disc icon instead of a gamepad

ui/xui/popup-menu.cc
```diff
@@ -338,9 +340,58 @@ class SettingsPopupMenu : public virtual PopupMenu {
...
+
+        if (m_focus && !m_pop_focus) {
+            ImGui::SetKeyboardFocusHere();
+        }
+
+        const char *games_dir = g_config.general.games_dir;
+        std::filesystem::path directory(games_dir);
+        std::multimap<std::string, std::string> sorted_file_names;
+
+        if (std::filesystem::is_directory(directory)) {
```
mborgerson: We should avoid querying the filesystem in the draw loop

ui/xui/popup-menu.cc
```diff
@@ -17,19 +17,21 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 //
 #include "ui/xemu-notifications.h"
-#include <string>
-#include <vector>
-#include "misc.hh"
+#include "popup-menu.hh"
```
mborgerson: Please revert any unnecessary header order changes

mborgerson (changes requested):

ui/xui/popup-menu.cc
TrentSeed: Sounds good, icon has been updated to ICON_FA_COMPACT_DISC

ui/xui/popup-menu.cc
TrentSeed: File system is no longer queried in the draw loop

ui/xui/popup-menu.cc
TrentSeed: Done

mborgerson:
Thanks for addressing the review feedback. I've rebased onto the latest version of the `master` branch. Can you please fix the build issues, then I'll happily merge.

mborgerson:
Sorry it's taken a while to come back to this. I'll close in favor of #2449 for rebase and minor improvements, but your authorship will be retained! Thanks for the improvement.

PR #1628, opened 2024-04-01
Update User Interface & Add Additional Options To IR Scaling, UI Scaling, & Aspect Ratio settings

Joshua-1248:
Adds additional options for Increase Internal Resolution, UI Scale, & Aspect Ratio settings in the main-menu.cc file.

Some minor changes have been made to the descriptions.

### DIFF
--- a/ui/xui/main-menu.cc
+++ b/ui/xui/main-menu.cc
@@ -1175,7 +1175,7 @@ void MainMenuSnapshotsView::Draw()
         const char *msg;
         if (g_snapshot_mgr.m_snapshots_len) {
             if (!m_search_buf.empty()) {
-                msg = "Select Create to create new snapshot";
+                msg = "Press Create to create new snapshot";
             } else {
                 msg = "No snapshots match filter criteria";
             }
--- a/ui/xui/main-menu.cc
+++ b/ui/xui/main-menu.cc
@@ -53,23 +53,23 @@ void MainMenuGeneralView::Draw()
 {
 #if defined(_WIN32)
     SectionTitle("Updates");
-    Toggle("Check for updates", &g_config.general.updates.check,
-           "Check for updates whenever xemu is opened");
+    Toggle("Check For Updates", &g_config.general.updates.check,
+           "Checks for updates whenever xemu is opened");
 #endif
 
 #if defined(__x86_64__)
     SectionTitle("Performance");
-    Toggle("Hard FPU emulation", &g_config.perf.hard_fpu,
+    Toggle("Hard FPU Emulation", &g_config.perf.hard_fpu,
            "Use hardware-accelerated floating point emulation (requires restart)");
 #endif
 
-    Toggle("Cache shaders to disk", &g_config.perf.cache_shaders,
-           "Reduce stutter in games by caching previously generated shaders");
+    Toggle("Cache Shaders To Disk", &g_config.perf.cache_shaders,
+           "Reduces stutter in games by caching previously generated shaders");
 
     SectionTitle("Miscellaneous");
-    Toggle("Skip startup animation", &g_config.general.skip_boot_anim,
-           "Skip the full Xbox boot animation sequence");
-    FilePicker("Screenshot output directory", &g_config.general.screenshot_dir,
+    Toggle("Fast Boot", &g_config.general.skip_boot_anim,
+           "Skips the Xbox startup animation");
+    FilePicker("Screenshot Output Directory", &g_config.general.screenshot_dir,
                NULL, true);
     // toggle("Throttle DVD/HDD speeds", &g_config.general.throttle_io,
     //        "Limit DVD/HDD throughput to approximate Xbox load times");
@@ -440,9 +440,9 @@ void MainMenuInputView::Draw()
     }
 
     SectionTitle("Options");
-    Toggle("Auto-bind controllers", &g_config.input.auto_bind,
+    Toggle("Auto-Bind Controllers", &g_config.input.auto_bind,
            "Bind newly connected controllers to any open port");
-    Toggle("Background controller input capture",
+    Toggle("Background Controller Input Capture",
            &g_config.input.background_input_capture,
            "Capture even if window is unfocused (requires restart)");
 }
@@ -452,7 +452,7 @@ void MainMenuDisplayView::Draw()
     SectionTitle("Quality");
     int rendering_scale = nv2a_get_surface_scale_factor() - 1;
     if (ChevronCombo("Internal resolution scale", &rendering_scale,
-                     "1x\0"
+                     "1x (Default)\0"
                      "2x\0"
                      "3x\0"
                      "4x\0"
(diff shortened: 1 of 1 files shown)


ui/xui/main-menu.cc
```diff
@@ -53,23 +53,23 @@ void MainMenuGeneralView::Draw()
 {
 #if defined(_WIN32)
     SectionTitle("Updates");
-    Toggle("Check for updates", &g_config.general.updates.check,
-           "Check for updates whenever xemu is opened");
+    Toggle("Check For Updates", &g_config.general.updates.check,
```
MasonT8198: Nit: “Check for Updates” would be the grammatically correct notation

ui/xui/main-menu.cc
```diff
@@ -53,23 +53,23 @@ void MainMenuGeneralView::Draw()
...
+           "Checks for updates whenever xemu is opened");
```
MasonT8198: Nit: “Check for updates whenever xemu is opened” is also the correct notation, why was this changed?

ui/xui/main-menu.cc
```diff
@@ -53,23 +53,23 @@ void MainMenuGeneralView::Draw()
...
 #if defined(__x86_64__)
     SectionTitle("Performance");
-    Toggle("Hard FPU emulation", &g_config.perf.hard_fpu,
+    Toggle("Hard FPU Emulation", &g_config.perf.hard_fpu,
            "Use hardware-accelerated floating point emulation (requires restart)");
 #endif
 
-    Toggle("Cache shaders to disk", &g_config.perf.cache_shaders,
-           "Reduce stutter in games by caching previously generated shaders");
+    Toggle("Cache Shaders To Disk", &g_config.perf.cache_shaders,
```
MasonT8198: Nit: “Cache Shaders to Disk” otherwise this is fine

ui/xui/main-menu.cc
```diff
@@ -53,23 +53,23 @@ void MainMenuGeneralView::Draw()
...
+           "Reduces stutter in games by caching previously generated shaders");
 
     SectionTitle("Miscellaneous");
-    Toggle("Skip startup animation", &g_config.general.skip_boot_anim,
-           "Skip the full Xbox boot animation sequence");
-    FilePicker("Screenshot output directory", &g_config.general.screenshot_dir,
+    Toggle("Fast Boot", &g_config.general.skip_boot_anim,
```
MasonT8198: Nit: “Skip Startup Animation” makes more sense than “Fast Boot”, in the context of what the config option, although I see what you’re getting across, however if these were going to be changed, do note that the `&g_config` lines should be changed to mimic the change in name scheme

ui/xui/main-menu.cc
```diff
@@ -462,40 +462,77 @@ void MainMenuDisplayView::Draw()
...
                      "1920x1080\0"
+                     "1920x1440\0"
+                     "2048x1152\0"
+                     "2048x1536\0"
                      "2560x1440\0"
                      "2560x1600\0"
                      "2560x1920\0"
                      "3840x2160\0",
+                     "3840x2880\0",
                      "Select preferred startup window size")) {
```
MasonT8198, line 525: Nit: This is **a lot** of config options, making the scrollbar look somewhat cluttered, maybe it’d be worth considering to use a selection of general options, and adding a custom toggle where you can type in what resolution that you would like to use?

Furthermore, some of these config options make no real sense? As they would make most games look unplayable

ui/xui/main-menu.cc
```diff
@@ -462,40 +462,77 @@ void MainMenuDisplayView::Draw()
...
     }
-    Toggle("Vertical refresh sync", &g_config.display.window.vsync,
-           "Sync to screen vertical refresh to reduce tearing artifacts");
+    Toggle("VSync", &g_config.display.window.vsync,
```
MasonT8198: Nit: I believe that the general notation of vsync is `V-Sync` however I’m amenible to the difference

ui/xui/main-menu.cc
```diff
@@ -1049,7 +1097,7 @@ void MainMenuSnapshotsView::Draw()
 
     if (snapshot_with_create_name_exists && ImGui::IsItemHovered()) {
         ImGui::SetTooltip("A snapshot with the name \"%s\" already exists. "
-                          "This button will overwrite the existing snapshot.",
```
MasonT8198: Nit: The plain description for this does work pretty well, I wouldn’t change it

ui/xui/main-menu.cc
```diff
@@ -462,40 +462,77 @@ void MainMenuDisplayView::Draw()
...
     }
 
     SectionTitle("Window");
     bool fs = xemu_is_fullscreen();
-    if (Toggle("Fullscreen", &fs, "Enable fullscreen now")) {
+    if (Toggle("Fullscreen", &fs, "Enable Fullscreen")) {
         xemu_toggle_fullscreen();
     }
-    Toggle("Fullscreen on startup",
+    Toggle("Fullscreen On Startup",
```
MasonT8198: Nit: “Fullscreen on Startup”

ui/xui/main-menu.cc
```diff
@@ -462,40 +462,77 @@ void MainMenuDisplayView::Draw()
...
+                     "19x\0",
+                     "20x\0",
+                     "Increase surface scaling factor for higher visual quality")) {
         nv2a_set_surface_scale_factor(rendering_scale+1);
     }
 
     SectionTitle("Window");
     bool fs = xemu_is_fullscreen();
-    if (Toggle("Fullscreen", &fs, "Enable fullscreen now")) {
+    if (Toggle("Fullscreen", &fs, "Enable Fullscreen")) {
```
MasonT8198, line 481: I like this change, makes more sense

ui/xui/main-menu.cc
```diff
@@ -1127,7 +1175,7 @@ void MainMenuSnapshotsView::Draw()
         const char *msg;
         if (g_snapshot_mgr.m_snapshots_len) {
             if (!m_search_buf.empty()) {
-                msg = "Press Create to create new snapshot";
+                msg = "Select Create to create new snapshot";
```
MasonT8198: Nit: select and press do the same thing, also the msg doubling of `create to create` adds some redundant language to the message

Perhaps something along the lines of `Select Create to Generate New Snapshot`

Also this change doesn’t follow your prior commit theme of capitalizing statements.

MasonT8198 (changes requested):
Reviewed…

Also, when submitting PRs we follow a general commit name structure

So in this case you edited main-menu.cc so the commit would be

`main-menu.cc: Clarify menu descriptions` or something along those lines.

I do agree that some of these descriptions could use a little touching up though.

ui/xui/main-menu.cc
Joshua-1248: Gotcha. Changed it to lower-case.

ui/xui/main-menu.cc
Joshua-1248: Was not sure which to choose. Will leave it as "Check" instead of "Checks" now that you mention it.

ui/xui/main-menu.cc
Joshua-1248: Done

ui/xui/main-menu.cc
Joshua-1248: Gotcha.

Restored default descriptor to "Skip Startup Animation"

ui/xui/main-menu.cc
Joshua-1248: A custom resolution config option would definitely work better in that regard, both for reducing clutter & for allowing the user to set the resolution to anything that fits their needs.

ui/xui/main-menu.cc
Joshua-1248: Changed to V-Sync

ui/xui/main-menu.cc
Joshua-1248: Reverted description to original one

ui/xui/main-menu.cc
Joshua-1248: Done

ui/xui/main-menu.cc
Joshua-1248: Done

mborgerson:
I appreciate the PR, but I disagree with these changes and will be closing the PR. The capitalization of menu option text is intentional and I don't wish for it to change, the plethora of added options don't seem necessary to me and will not work correctly as written, and the other misc. changes really seem like typos.

Issue #55, opened 2020-10-30
Halo 2: Able to see other objects through objects close to you


Alastair0100:
Also can depend on viewing angle aswell as how close objects are to you from my testing

Title: 4d530064

![image](https://user-images.githubusercontent.com/40572266/97756179-5d9e5f00-1af2-11eb-8a3a-0af083ec9825.png)
![image](https://user-images.githubusercontent.com/40572266/97756201-6858f400-1af2-11eb-91a9-5894d618aa83.png)
![image](https://user-images.githubusercontent.com/40572266/97756212-6bec7b00-1af2-11eb-8968-8966bf63237d.png)
![image](https://user-images.githubusercontent.com/40572266/97756221-6f800200-1af2-11eb-9fc6-327b9d5b6d46.png)

Shoegzer:
Confirmed in testing here too.

Relevant test specs:
* OS: Linux Mint 20.3 / kernel 5.131
* GPU: NVidia GTX 3070 / driver 470.86
* Commit tested: 3fd15218c76cedd89496db637c7ef5369ef35a63

Triticum0:
This also affects Wallace & Gromit in Project Zoo, Miami Vice, Tom Clancy's Splinter Cell and Serious Sam

BiatuAutMiahn:
In #266, I mention when this behaviour started in Halo 2.

RageXbox:
Here’s a log for this.
[Halo 2 seethrough.log](https://github.com/mborgerson/xemu/files/7901151/Halo.2.seethrough.log)

abaire:
This looks like it may be a depth buffer precision issue, stepping through the draws in renderdoc I see depth buffer values for the more distant draws that are suspiciously close to but less than the nearby correct nearer geometry with odd discontinuities (e.g., if I follow the plane of a wall that is being cut out, I see values that follow a linear trend then jump to higher values where the earlier draw is peeking through, then resume within the expected interpolated range after the slice).

Halo2 uses a non-standard clip range:
```
nv2a_pgraph_method 0: NV20_KELVIN_PRIMITIVE<0x97> -> NV097_SET_CLIP_MIN<0x394> (0x437FFFFF => 255.999985)
nv2a_pgraph_method 0: NV20_KELVIN_PRIMITIVE<0x97> -> NV097_SET_CLIP_MAX<0x398> (0x4B7FFFFF => 16777215.000000)
```
though it at least uses the same range throughout all of the draws in the frame I traced.

[...]

PR #1644, opened 2024-05-09
Halo 2 skybox fix (& realtime lighting shadow casting in other games)

polymetal0:
Clamping depth value to 1 apparently fixes this. Skybox is also affected by the see through issue:

EDIT: In case you are caching shaders in disk, you will need to delete your shader cache file.

Captura3
Captura3mal
Captura4
Captura4mal
Captura
Capturamal
Captura2
Captura2mal

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,9 @@
 /GNUmakefile
 /build/
 /.cache/
+/ccache/
 /.vscode/
+/.vs/
 *.pyc
 .sdk
 .stgit-*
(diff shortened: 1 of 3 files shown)


Shoegzer:
Fantastic!  And yeah, the see-through textures issue still exists as per #55 but this is quite a welcome fix.

polymetal0:
I noticed that this changes also restore dynamic shadows (tested on Conker and Halo CE).

Some screenshots:

Latest release:
conker
halo

Clamping z to 1:
conker2
halo 2

GXTX:
Ideally you'd test with https://github.com/abaire/nxdk_pgraph_tests

polymetal0:
Thanks! I don't expect these changes to make it to the main branch anyway, but I hope someone with more understanding of the project can get a better clue about where the issue might be.

PR #1659, opened 2024-06-10
Implement XDVD security challenge responses

### LINKED ISSUE #1036: DVD Authentication does not work
Currently, we don't properly emulate the DVD authentication which means you need a modded kernel ( usually referred to as bios) which does allow you to run copied games.

Here is some more info: http://xboxdevwiki.net/Xbox_Game_Disc

https://github.com/JayFoxRox/xqemu-espes/pull/23 but I'm currently not actively working on it.

### Additional Context

This was ripped directly fromhttps://github.com/xqemu/xqemu/issues/26 So big thank to JayFoxRox for the information.

Ryzee119:
This PR implements the DVD security challenge/response mechanism to the best of my knowledge used on a Xbox console between the DVD drive and the Xbox. 

This allows loading retail Xbox titles both in 'redump' format and xiso format with an unmodified retail BIOS.

For this to work, you need the security info from your original game disk. There is two methods:
1. `SS.bin` from a compatible 'Kreon' drive using the method on redump.org <http://wiki.redump.org/index.php?title=Microsoft_Xbox_and_Xbox_360_Dumping_Guide>, 
2. `dvd_layout.bin` from your Xbox and original disk using these homebrew apps <https://github.com/Ryzee119/OGX_CHALLENGE_TABLE/releases/tag/build-202310141001> or <https://gist.github.com/Ernegien/bef054a43213c52c2bef8d94bf9f51cb>  

These two dumps have the same info but in a slightly different format. As I understand `SS.bin` represents what is physically on the DVD, `dvd_layout.bin` is the packet that is formatted and sent by the DVD drive firmware. This PR detects either and converts accordingly.

To provide this file to xemu, there is currently no GUI element and is done by command line

`xemu.exe -dvd_security_path "SS.bin"`. The same arg is used for `"dvd_layout.bin"`.

**Notes from my testing:**
1. I had pretty good success using one common SS.bin/.dvd_layout.bin dump with many games. However later bioses seemed to only accept SS.bin dumps from that specific title. This could be improved security or an issue in my implementation?
2. The Xbox didn't seem to care if the DVD is 'redump' or XISO style aslong as the challenges are responded to correctly. Your mileage may vary.
3. Using a retail BIOS comes with the restrictions of the retail system such as region restrictions, and you wont be able to run homebrew (FTP etc)

**References**
1. https://gist.github.com/Ernegien/bef054a43213c52c2bef8d94bf9f51cb (Major shout out the xbox7887 for this!)
2. https://wayback-api.archive.org/web/20240316195746/https://multimedia.cx/eggs/xbox-sphinx-protocol/
4. https://xboxdevwiki.net/DVD_Drive
5. https://xboxdevwiki.net/Xbox_Game_Disc

This process just follows the text book challenge response process from the Xbox. There may be some unusual aspects that are missed and I only have a smallish selection of disks to test with. I pulled in some public domain RC4 and SHA1 libraries, not sure if they should live in the xdvd folder or not.

**Future work suggestions**
1. Generate SS.bin automatically?
2. Add to GUI and somehow link to specific titles

image

Fixes #1036

### DIFF
--- a/hw/ide/atapi.c
+++ b/hw/ide/atapi.c
@@ -1101,6 +1101,12 @@ static void cmd_read(IDEState *s, uint8_t* buf)
     }
 
     lba = ldl_be_p(buf + 2);
+
+#ifdef XBOX
+    lba = xdvd_get_lba_offset(&s->xdvd_security, total_sectors, lba);
+    total_sectors = xdvd_get_sector_cnt(&s->xdvd_security, total_sectors);
+#endif
+
     if (lba >= total_sectors || lba + nb_sectors - 1 >= total_sectors) {
         ide_atapi_cmd_error(s, ILLEGAL_REQUEST, ASC_LOGICAL_BLOCK_OOR);
         return;
@@ -1256,6 +1262,10 @@ static void cmd_read_toc_pma_atip(IDEState *s, uint8_t* buf)
 static void cmd_read_cdvd_capacity(IDEState *s, uint8_t* buf)
 {
     uint64_t total_sectors = s->nb_sectors >> 2;
+
+#ifdef XBOX
+    total_sectors = xdvd_get_sector_cnt(&s->xdvd_security, total_sectors);
+#endif
 
     /* NOTE: it is really the number of sectors minus 1 */
     stl_be_p(buf, total_sectors - 1);
--- a/hw/ide/atapi.c
+++ b/hw/ide/atapi.c
@@ -37,13 +37,13 @@
 #ifdef XBOX
 #include "ui/xemu-settings.h"
 #include "hw/xbox/xdvd/xdvd.h"
-#endif
 
-// #define DEBUG
+#define DEBUG
 #ifdef DEBUG
 # define XBOX_DPRINTF(format, ...)     printf(format, ## __VA_ARGS__)
 #else
 # define XBOX_DPRINTF(format, ...)     do { } while (0)
+#endif
 #endif
 
 static void ide_atapi_cmd_read_dma_cb(void *opaque, int ret);
@@ -925,6 +925,9 @@ static void cmd_mode_select_cb(void *opaque, int ret)
     ide_set_inactive(s, false);
 
     int max_len = lduw_be_p(buf) + 2;
+    // FIXME: It is understood that a original xbox drive requires a full unlock challenge-response sequence before authenticating the disk,
+    // however for simplicity we authenticate after the first challenge-response. This code still correctly answers all responses sent by the xbox.
+    // Testing shows that this appears to be sufficient.
     if (max_len == XDVD_SECURITY_PAGE_LEN && buf[8] == MODE_PAGE_XBOX_SECURITY)
     {
         memcpy(&s->xdvd_security, buf, XDVD_SECURITY_PAGE_LEN);
@@ -1063,8 +1066,8 @@ static void cmd_mode_sense(IDEState *s, uint8_t *buf)
             memcpy(buf, &s->xdvd_security, sizeof(s->xdvd_security));
 
             ide_atapi_cmd_reply(s, sizeof(XBOX_DVD_SECURITY), max_len);
+            break;
 #endif
-        break;
         default:
             goto error_cmd;
         }
@@ -1340,9 +1343,10 @@ static void cmd_read_dvd_structure(IDEState *s, uint8_t* buf)
         }
     }
 
+#ifdef XBOX
      // Still some important info we don't want to lose so we disable clearing
      // on Xbox for now
-#ifndef XBOX
+#else
     memset(buf, 0, max_len > IDE_DMA_BUF_SECTORS * BDRV_SECTOR_SIZE + 4 ?
            IDE_DMA_BUF_SECTORS * BDRV_SECTOR_SIZE + 4 : max_len);
 #endif
--- a/hw/ide/atapi.c
+++ b/hw/ide/atapi.c
@@ -75,11 +75,27 @@ static inline int media_present(IDEState *s)
 /* XXX: DVDs that could fit on a CD will be reported as a CD */
 static inline int media_is_dvd(IDEState *s)
 {
+#ifdef XBOX
+    // We need to prevent small XISOs from being reported as CD-ROMs
+    // If the DVD security path is set, we can assume it is an XISO.
+    const char *dvd_security_path = g_config.sys.files.dvd_security_path;
+    if (strlen(dvd_security_path) > 0) {
+        return media_present(s);
+    }
+#endif
     return (media_present(s) && s->nb_sectors > CD_MAX_SECTORS);
 }
 
 static inline int media_is_cd(IDEState *s)
 {
+#ifdef XBOX
+    // We need to prevent small XISOs from being reported as CD-ROMs
+    // If the DVD security path is set, we can assume it is an XISO.
+    const char *dvd_security_path = g_config.sys.files.dvd_security_path;
+    if (strlen(dvd_security_path) > 0) {
+        return 0;
+    }
+#endif
     return (media_present(s) && s->nb_sectors <= CD_MAX_SECTORS);
 }
 
--- a/hw/ide/atapi.c
+++ b/hw/ide/atapi.c
@@ -857,6 +857,47 @@ static void cmd_get_configuration(IDEState *s, uint8_t *buf)
     ide_atapi_cmd_reply(s, len, max_len);
 }
 
+#ifdef XBOX
+static void cmd_mode_select_cb(void *opaque, int ret)
+{
+    IDEState *s = opaque;
+    uint8_t *buf = s->io_buffer;
+
+    if (ret < 0) {
+        block_acct_failed(blk_get_stats(s->blk), &s->acct);
+        ide_set_inactive(s, false);
+        return;
+    }
+
+    if (s->bus->dma->ops->rw_buf(s->bus->dma, 0))
+    {
+        s->status = READY_STAT | SEEK_STAT;
+        s->nsector = (s->nsector & ~7) | ATAPI_INT_REASON_IO | ATAPI_INT_REASON_CD;
+        ide_bus_set_irq(s->bus);
+    }
+    block_acct_done(blk_get_stats(s->blk), &s->acct);
+    ide_set_inactive(s, false);
+}
+
+static void cmd_mode_select(IDEState *s, uint8_t *buf)
+{
+    int parameter_list_len = lduw_be_p(buf + 7);
+
+    // Does MODE SELECT contain more data? Read it out
+    if (parameter_list_len > 0)
+    {
+        s->lba = -1;
+        s->packet_transfer_size = parameter_list_len;
+        s->io_buffer_size = parameter_list_len;
+        s->elementary_transfer_size = 0;
+
+        block_acct_start(blk_get_stats(s->blk), &s->acct, parameter_list_len, BLOCK_ACCT_WRITE);
+        s->status = READY_STAT | SEEK_STAT | DRQ_STAT;
+        ide_start_dma(s, cmd_mode_select_cb);
+    }
+}
+#endif
+
 static void cmd_mode_sense(IDEState *s, uint8_t *buf)
 {
     int action, code;
@@ -1293,6 +1334,9 @@ static const struct AtapiCmd {
     [ 0x46 ] = { cmd_get_configuration,             ALLOW_UA },
     [ 0x4a ] = { cmd_get_event_status_notification, ALLOW_UA },
     [ 0x51 ] = { cmd_read_disc_information,         CHECK_READY },
+#ifdef XBOX
+    [ 0x55 ] = { cmd_mode_select,/* (10) */         0 },
+#endif
     [ 0x5a ] = { cmd_mode_sense, /* (10) */         0 },
     [ 0xa8 ] = { cmd_read, /* (12) */               CHECK_READY },
     [ 0xad ] = { cmd_read_dvd_structure,            CHECK_READY },
--- a/hw/ide/atapi.c
+++ b/hw/ide/atapi.c
@@ -34,6 +34,18 @@
 #define ATAPI_SECTOR_BITS (2 + BDRV_SECTOR_BITS)
 #define ATAPI_SECTOR_SIZE (1 << ATAPI_SECTOR_BITS)
 
+#ifdef XBOX
+#include "ui/xemu-settings.h"
+#include "hw/xbox/xdvd/xdvd.h"
+#endif
+
+// #define DEBUG
+#ifdef DEBUG
+# define XBOX_DPRINTF(format, ...)     printf(format, ## __VA_ARGS__)
+#else
+# define XBOX_DPRINTF(format, ...)     do { } while (0)
+#endif
+
 static void ide_atapi_cmd_read_dma_cb(void *opaque, int ret);
 
 static void padstr8(uint8_t *buf, int buf_size, const char *src)
@@ -487,6 +499,24 @@ static int ide_dvd_read_structure(IDEState *s, int format,
             {
                 int layer = packet[6];
                 uint64_t total_sectors;
+#ifdef XBOX
+                int block_number = ldl_be_p(buf + 2);
+
+                // Challenge table is located on disk runout, on Xbox these locations are always set to these for this command it looks like
+                if (layer == XDVD_STRUCTURE_LAYER && block_number == XDVD_STRUCTURE_BLOCK_NUMBER)
+                {
+                    if (xdvd_get_encrypted_challenge_table(s->xdvd_challenges_encrypted))
+                    {
+                        xdvd_get_decrypted_responses(s->xdvd_challenges_encrypted, s->xdvd_challenges_decrypted);
+                        memcpy(buf, s->xdvd_challenges_encrypted, XDVD_STRUCTURE_LEN);
+                        return XDVD_STRUCTURE_LEN;
+                    }
+                }
+
+                // We deferred clearing this earlier so do it now
+                int max_len = lduw_be_p(buf + 8);
+                memset(buf, 0, MIN(max_len, IDE_DMA_BUF_SECTORS * BDRV_SECTOR_SIZE + 4));
+ #endif
 
                 if (layer != 0)
                     return -ASC_INV_FIELD_IN_CMD_PACKET;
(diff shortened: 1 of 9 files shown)


hw/ide/atapi.c
```diff
@@ -61,11 +73,27 @@ static inline int media_present(IDEState *s)
 /* XXX: DVDs that could fit on a CD will be reported as a CD */
 static inline int media_is_dvd(IDEState *s)
 {
+#ifdef XBOX
+    // We need to prevent small XISOs from being reported as CD-ROMs
+    // If the DVD security path is set, we can assume it is an XISO.
+    const char *dvd_security_path = g_config.sys.files.dvd_security_path;
+    if (strlen(dvd_security_path) > 0) {
```
mborgerson, line 82: Should you also check `media_present` here, like in original expression?

hw/ide/atapi.c
```diff
@@ -855,6 +902,56 @@ static void cmd_get_configuration(IDEState *s, uint8_t *buf)
...
+        memcpy(&s->xdvd_security, buf, DVD_SECURITY_PAGE_LEN);
+        XBOX_DPRINTF("Authenticated: %d, Partition: %d\n", s->xdvd_security.page.Authenticated, s->xdvd_security.page.Partition);
+        s->xdvd_security.page.Authenticated = 1;
+        ide_atapi_cmd_ok(s);
+    }
+}
+
+static void cmd_mode_select(IDEState *s, uint8_t *buf)
+{
+    int parameter_list_len = parameter_list_len = lduw_be_p(buf + 7);
```
mborgerson: Redundant init

hw/xbox/xdvd/WjCryptLib_Rc4.c
```diff
@@ -0,0 +1,152 @@
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
```
mborgerson: We have an RC4 and SHA1 implementation already for EEPROM generation stuff. Preferably we would use that. It will need to be refactored to a common place but it shouldn't be tough

hw/ide/atapi.c
```diff
@@ -61,11 +73,27 @@ static inline int media_present(IDEState *s)
 /* XXX: DVDs that could fit on a CD will be reported as a CD */
 static inline int media_is_dvd(IDEState *s)
 {
+#ifdef XBOX
+    // We need to prevent small XISOs from being reported as CD-ROMs
+    // If the DVD security path is set, we can assume it is an XISO.
+    const char *dvd_security_path = g_config.sys.files.dvd_security_path;
```
mborgerson, line 81: We should probably make this a qdev property, like disc file path, media type, etc. so it can play nicely with other systems

hw/ide/atapi.c
```diff
@@ -485,6 +513,25 @@ static int ide_dvd_read_structure(IDEState *s, int format,
             {
                 int layer = packet[6];
                 uint64_t total_sectors;
+#ifdef XBOX
+                int block_number = ldl_be_p(buf + 2);
+
+                // Challenge table is located on disk runout, on Xbox these locations are always set to these for this command it looks like
+                if (layer == 0xFE && block_number == 0xFF02FDFF)
```
mborgerson: nit: Please name these constants

hw/ide/atapi.c
```diff
@@ -485,6 +513,25 @@ static int ide_dvd_read_structure(IDEState *s, int format,
...
+                    {
+                        xdvd_get_decrypted_responses(s->xdvd_challenges_encrypted, s->xdvd_challenges_decrypted);
+                        memcpy(buf, s->xdvd_challenges_encrypted, XDVD_STRUCTURE_LEN);
+                        return XDVD_STRUCTURE_LEN;
+                    }
+                }
+
+                // We deferred clearing this earlier so do it now
+                int max_len = lduw_be_p(buf + 8);
+                memset(buf, 0, max_len > IDE_DMA_BUF_SECTORS * BDRV_SECTOR_SIZE + 4 ?
```
mborgerson: nit: `MIN(max_len, IDE_DMA_BUF_SECTORS...)`

hw/xbox/xdvd/xdvd.c
```diff
@@ -0,0 +1,183 @@
+#include <assert.h>
```
mborgerson, line 20: Please include copyright header

hw/xbox/xdvd/xdvd.h
```diff
@@ -0,0 +1,92 @@
+#ifndef HW_XDVD_H
```
mborgerson, line 20: Copyright header

hw/xbox/xdvd/xdvd.h
```diff
@@ -0,0 +1,92 @@
...
+
+// Standard SCSI Mode Sense/Select header
+typedef struct _MODE_PARAMETER_HEADER10 {
+    uint8_t ModeDataLength[2];
+    uint8_t MediumType;
+    uint8_t DeviceSpecificParameter;
+    uint8_t Reserved[2];
+    uint8_t BlockDescriptorLength[2];
+} __attribute__((packed)) MODE_PARAMETER_HEADER10, *PMODE_PARAMETER_HEADER10;
+_Static_assert(sizeof(MODE_PARAMETER_HEADER10) == 8,
```
mborgerson: Instead of using `_Static_assert` directly, let's use `QEMU_BUILD_BUG_MSG`. Note it does invert the condition.

hw/xbox/xdvd/xdvd.c
```diff
@@ -0,0 +1,183 @@
...
+    static const XBOX_DVD_SECURITY s = {
+        .header.ModeDataLength[0] = 0,
+        .header.ModeDataLength[1] = sizeof(XBOX_DVD_SECURITY) - 2,
+        .page.PageCode = MODE_PAGE_XBOX_SECURITY,
+        .page.PageLength = sizeof(XBOX_DVD_SECURITY_PAGE) - 2,
+        .page.Unk1 = 1,
+        .page.BookTypeAndVersion = 0xD1,
+        .page.Unk2 = 1,
+    };
+    memcpy(xdvd_security, &s, sizeof(XBOX_DVD_SECURITY));
```
mborgerson: nit: Use `sizeof(s)` instead of replicating the type name

include/hw/ide/internal.h
```diff
@@ -457,6 +461,11 @@ struct IDEState {
     uint8_t *smart_selftest_data;
     /* AHCI */
     int ncq_queues;
+#ifdef XBOX
+    uint8_t xdvd_challenges_encrypted[XDVD_STRUCTURE_LEN];
```
mborgerson: Do any of these fields need to be saved in vm state to support snapshots?

hw/xbox/xdvd/xdvd.c
```diff
@@ -0,0 +1,183 @@
...
+};
+
+bool xdvd_is_redump(uint64_t total_sectors)
+{
+    return total_sectors == XDVD_REDUMP_SECTOR_CNT;
```
mborgerson: Is it possible for this to misclassify if a non-redump image happens to be this same size?

hw/xbox/xdvd/xdvd.c
```diff
@@ -0,0 +1,183 @@
...
+    int file_size;
+    fseek(f, 0, SEEK_END);
+    file_size = ftell(f);
+    fseek(f, 0, SEEK_SET);
+
+    // If the security structure is read from the DVD drive it will be 1636 bytes long
+    // If it is read from the DVD disk (via Kreon etc) it will be 2048 bytes long
+    if (file_size == XDVD_STRUCTURE_LEN) {
+        fread(xdvd_challenge_table_encrypted, 1, file_size, f);
+    } else if (file_size == 2048) {
```
mborgerson: Please name this constant

hw/xbox/xdvd/xdvd.c
```diff
@@ -0,0 +1,183 @@
...
+            lba += XGD1_LSEEK_OFFSET / ATAPI_SECTOR_SIZE;
+        }
+    }
+    return lba;
+}
+
+bool xdvd_get_encrypted_challenge_table(uint8_t *xdvd_challenge_table_encrypted)
+{
+    bool status = true;
+    // Check if we have already read it from the file
```
mborgerson: Where is this checked?

mborgerson (commented):
Wow, nice work! Left some initial comments.

hw/xbox/xdvd/xdvd.c
Ryzee119: Youre right it would be technically possible; although I dont think any retail xbox title in XISO format could possibly have this size but maybe some homebrew could have.

Any suggestions? I presume there is some magic bytes in the redump style image somewhere, Id need to research it

antangelo: You should be able to check the volume descriptor magic fields at the appropriate offset: https://xboxdevwiki.net/XDVDFS#Volume_descriptor

hw/xbox/xdvd/xdvd.c
Ryzee119: Done

hw/xbox/xdvd/xdvd.h
Ryzee119: Done

hw/ide/atapi.c
Ryzee119: Fixed

hw/ide/atapi.c
Ryzee119: Ive called them XDVD_STRUCTURE_LAYER and XDVD_STRUCTURE_BLOCK_NUMBER, defined in xdvd.h

hw/xbox/xdvd/xdvd.c
Ryzee119: Done

hw/xbox/xdvd/xdvd.c
Ryzee119: Ive called it  XDVD_SECURITY_SECTOR_LEN

hw/xbox/xdvd/xdvd.c
Ryzee119: Removed the comment. Left over from old code

hw/ide/atapi.c
Ryzee119: Fixed

mborgerson: `&& 1` is always true. Can be simplified to:

```c
if (strlen(g_config.sys.files.dvd_security_path)) {
    return media_present(s);
}
```

hw/ide/atapi.c
```diff
@@ -61,11 +73,27 @@ static inline int media_present(IDEState *s)
...
 }
 
 static inline int media_is_cd(IDEState *s)
 {
+#ifdef XBOX
+    // We need to prevent small XISOs from being reported as CD-ROMs
+    // If the DVD security path is set, we can assume it is an XISO.
+    const char *dvd_security_path = g_config.sys.files.dvd_security_path;
+    if (strlen(dvd_security_path) > 0) {
+        return media_present(s) && 0;
```
mborgerson: `&& 0` is always false

hw/ide/atapi.c
```diff
@@ -1206,8 +1339,11 @@ static void cmd_read_dvd_structure(IDEState *s, uint8_t* buf)
         }
     }
 
-    memset(buf, 0, max_len > IDE_DMA_BUF_SECTORS * BDRV_SECTOR_SIZE + 4 ?
-           IDE_DMA_BUF_SECTORS * BDRV_SECTOR_SIZE + 4 : max_len);
+     // Still some important info we don't want to lose so we disable clearing
+     // on Xbox for now
+#ifndef XBOX
+    memset(buf, 0, MIN(max_len, IDE_DMA_BUF_SECTORS * BDRV_SECTOR_SIZE + 4));
```
mborgerson: Please don't change code that we are just `#ifndef`ing away. For a bit of context, we don't actually expect to build with `XBOX` not defined, these mostly just help me when merging in QEMU updates and if the code also changes it makes it worse. That said, I do like the refactor on the new spot you added this.

hw/xbox/xdvd/WjCryptLib_Rc4.c
Ryzee119: I've rebased on top of https://github.com/xemu-project/xemu/pull/1660 so hopefully this is resolved now

hw/ide/atapi.c
Ryzee119: Fixed. I just had it to make it painfully explicit that was being forced when present but happy with this

hw/ide/atapi.c
Ryzee119: Oops did the wrong spot. Reinstated the ifndef section but applied the refactor on the new spot

hw/ide/atapi.c
```diff
@@ -1206,8 +1338,12 @@ static void cmd_read_dvd_structure(IDEState *s, uint8_t* buf)
         }
     }
 
+     // Still some important info we don't want to lose so we disable clearing
```
PatrickvL, line 1347: Since this is one of the few "not defined" conditionals, it's perhaps nicer to change it into an `#ifdef XBOX` then the comment lines, and then an `#else` line before the code that has to be skipped under an Xbox build.

hw/ide/atapi.c
```diff
@@ -32,6 +32,18 @@
 #define ATAPI_SECTOR_BITS (2 + BDRV_SECTOR_BITS)
 #define ATAPI_SECTOR_SIZE (1 << ATAPI_SECTOR_BITS)
 
+#ifdef XBOX
+#include "ui/xemu-settings.h"
+#include "hw/xbox/xdvd/xdvd.h"
+#endif
+
+// #define DEBUG
+#ifdef DEBUG
```
PatrickvL, line 42: This could be moved inside the above `#ifdef XBOX` conditional block.

hw/ide/atapi.c
```diff
@@ -941,6 +1037,32 @@ static void cmd_mode_sense(IDEState *s, uint8_t *buf)
...
+                         s->xdvd_security.page.ResponseValue);
+            XBOX_DPRINTF("Authenticated: %d, Partition: %d\n",
+                         s->xdvd_security.page.Authenticated,
+                         s->xdvd_security.page.Partition);
+
+            memcpy(buf, &s->xdvd_security, sizeof(s->xdvd_security));
+
+            ide_atapi_cmd_reply(s, sizeof(XBOX_DVD_SECURITY), max_len);
+#endif
+        break;
```
PatrickvL: This `break` must reside in the above conditional block, not after it.

hw/xbox/xdvd/xdvd.c
```diff
@@ -0,0 +1,195 @@
...
+uint32_t xdvd_get_challenge_response(const uint8_t *xdvd_challenge_table_decrypted,
+                            uint8_t challenge_id)
+{
+    int challengeEntryCount =
+        xdvd_challenge_table_decrypted[CR_TABLE_NUM_ENTRIES];
+    PXBOX_DVD_CHALLENGE challenges =
+        (PXBOX_DVD_CHALLENGE)(&xdvd_challenge_table_decrypted[CR_ENTRIES_OFFSET]);
+
+    for (int i = 0; i < challengeEntryCount; i++) {
+        if (challenges[i].type != 1)
```
mborgerson: * We should name this constant. Supposedly 1 means the entry is valid.
* nit: Braces around condition body, same with below

hw/xbox/xdvd/xdvd.c
```diff
@@ -0,0 +1,195 @@
...
+            xdvd_security->page.Partition == 1) {
+            lba += XGD1_LSEEK_OFFSET / ATAPI_SECTOR_SIZE;
+        }
+    }
+    return lba;
+}
+
+bool xdvd_get_encrypted_challenge_table(uint8_t *xdvd_challenge_table_encrypted)
+{
+    bool status = true;
```
PatrickvL: Either move this variable declaration down to reside along with the others below (since it's not referred to until after those), or (if this project still requires all variable declarations to reside at the start of a function) move all below variable declarations towards here.

PatrickvL (commented):
A few code review remarks

hw/xbox/xdvd/xdvd.c
```diff
@@ -0,0 +1,195 @@
...
+        .page.PageCode = MODE_PAGE_XBOX_SECURITY,
+        .page.PageLength = sizeof(XBOX_DVD_SECURITY_PAGE) - 2,
+        .page.Unk1 = 1,
+        .page.BookTypeAndVersion = 0xD1,
+        .page.Unk2 = 2,
+    };
+    memcpy(xdvd_security, &s, sizeof(s));
+};
+
+bool xdvd_is_redump(uint64_t total_sectors)
```
mborgerson: I'd like to avoid using the term "redump" in xemu code. Although we use the term informally to mean a complete disc backup, it really is just the name of the metadata archival project and not tied directly to the layout of the disc image. So instead of using the term redump here, let's use something like `xdvd_has_video_partition`.

hw/ide/atapi.c
```diff
@@ -855,6 +901,56 @@ static void cmd_get_configuration(IDEState *s, uint8_t *buf)
...
+        ide_set_irq(s->bus);
+    }
+    block_acct_done(blk_get_stats(s->blk), &s->acct);
+    ide_set_inactive(s, false);
+
+    int max_len = lduw_be_p(buf) + 2;
+    if (max_len == XDVD_SECURITY_PAGE_LEN && buf[8] == MODE_PAGE_XBOX_SECURITY)
+    {
+        memcpy(&s->xdvd_security, buf, XDVD_SECURITY_PAGE_LEN);
+        XBOX_DPRINTF("Authenticated: %d, Partition: %d\n", s->xdvd_security.page.Authenticated, s->xdvd_security.page.Partition);
```
mborgerson, line 934: My understanding is that the drive requires the full unlock challenge-response sequence to happen before it will unlock and we are not yet fully modeling this state machine. If this is correct and you don't plan to implement this, can you leave a comment so we can remember to come back to it?

hw/xbox/xdvd/xdvd.c
```diff
@@ -0,0 +1,195 @@
...
+    rc4_crypt(&rc4_ctx, &xdvd_challenge_table_decrypted[CR_ENTRIES_OFFSET],
+                   CR_ENTRIES_LEN);
+}
+
+// Given the already decrypted challenge table and the challenge ID sent by the
+// Xbox, return the required response dword
+uint32_t xdvd_get_challenge_response(const uint8_t *xdvd_challenge_table_decrypted,
+                            uint8_t challenge_id)
+{
+    int challengeEntryCount =
```
mborgerson: nit: `challenge_entry_count`

Triticum0:
Can Someone post a build of this pr. I think it needed for Def Jam - Fight for NY as I think game has some sort of copy protection which make if infinity load.

Steff833:
Strange. I cannot find this in Actions. This commit is 9 months old.  I would like to test this with Def Jam.

Ryzee119:
> Strange. I cannot find this in Actions. This commit is 9 months old. I would like to test this with Def Jam.

I just rebased this with current master latest build should be here https://github.com/xemu-project/xemu/actions/runs/15503131067

@Triticum0 
@Steff833

Steff833:
> > Strange. I cannot find this in Actions. This commit is 9 months old. I would like to test this with Def Jam.
> 
> I just rebased this with current master latest build should be here https://github.com/xemu-project/xemu/actions/runs/15503131067
> 
> @Triticum0 @Steff833

Thx but unfortunately your PR dosnt fix the infinite loading on Def Jam before the EA Games Logo appears. I had high hopes the game would finally boot. Dammit whats the problem with this game for so long? The Demo runs perfect. The Emulator is not crashing. It wont go further.

Triticum0:
You managed to get ss.bin working having issue on Linux i going test it on windows wmwhen I get the chance most not should how to use terminal but retry on windows hopefully I have more luck

PR #1660, opened 2024-06-12
Move SHA1 and RC4 functions from eeprom specific file to utils

Ryzee119:
As suggested in https://github.com/xemu-project/xemu/pull/1659, instead of pulling in another RC4/SHA1 implementation use the one already provided in the EEPROM generation code.

To do this it needs to be moved to a more accessible location.

To ensure I didnt break anything I generated an eeprom with xemu and loaded it up with https://github.com/Ernegien/XboxEepromEditor and made sure the hashes matched up

### DIFF
--- a/hw/xbox/eeprom_generation.c
+++ b/hw/xbox/eeprom_generation.c
@@ -2,6 +2,7 @@
  * QEMU Xbox EEPROM Generation (MCPX Version 1.0)
  *
  * Copyright (c) 2020 Mike Davis
+ * Copyright (c) 2024 Ryan Wendland
  *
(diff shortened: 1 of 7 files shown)


util/sha1_rc4.c
```diff
@@ -0,0 +1,168 @@
+/*
```
mborgerson: Let's split these out to separate files, one for SHA1 and another for RC4

Ryzee119: Fixed

mborgerson:
Thanks!

Issue #1020, opened 2022-06-03
Meta : Z Fighting


Triticum0:
### Bug Description

Some Games experience z-fighting on object which does occur on hardware

 ### Games Affected
- [x] [Alias](https://xemu.app/titles/41430016/#Alias) (Video In the comments) Occurs on hardware https://youtu.be/7zg9VORgJsg?t=310
- [ ] [Aggressive Inline](https://xemu.app/titles/41430007/#Aggressive-Inline) (Video In the comments)
- [x] [Beyond Good & Evil](https://xemu.app/titles/55530018/#Beyond-Good-Evil) (Video In the comments) (Occurs on hardware)
- [ ] [Burnout 2](https://xemu.app/titles/41430019/#Burnout-2-Point-of-Impact) (#1161)
- [x] Carmen Sandiego: The Secret of the Stolen Drums (When picking up first puzzle piece) Occurs on hardware https://youtu.be/PpP_U4HxjLw?t=1182
- [ ] [Conker: Live & Reloaded](https://xemu.app/titles/4d530051/#Conker-Live-Reloaded) #575
- [x] https://xemu.app/titles/464c0002/#Crusty-Demons (Tree first mission)
- [ ] Ford vs Chevy
- [x] [Kao the Kangaroo: Round 2](https://xemu.app/titles/4a570003/#Kao-the-Kangaroo-Round-2) (Lighthouse Door)
![xemu-2022-08-06-00-47-58](https://user-images.githubusercontent.com/71127352/183224832-95b6df58-31ed-4296-8a73-5e392f793c07.png)
- [x] [King Arthur](https://xemu.app/titles/4b4e0024/#King-Arthur)(Rock on the first level has some Z fighting)
![xemu-2022-08-06-01-08-52](https://user-images.githubusercontent.com/71127352/183226197-039efc18-503c-4b04-b319-0004dff7a8be.png)

- 
- [ ] [NCAA Football 2005](https://xemu.app/titles/45410051/#NCAA-Football-2005)
- [ ] [NCAA Football 06](https://xemu.app/titles/45410080/#NCAA-Football-06)
- [ ] Outrun 2  (Video In the comments)
- [ ] Outrun 2 coast to coast
- [x] [Robin Hood: Defender of the Crown](https://xemu.app/titles/4343000c/#Robin-Hood-Defender-of-the-Crown) Occurs on hardware https://youtu.be/gaBWP_Scnmk?t=382
[...]

PR #1707, opened 2024-07-23
nv2a: Clip range and see through fix

polymetal0:
Thanks to @coldhex for your work! I used part of the code you mention in https://github.com/xemu-project/xemu/issues/55#issuecomment-2241693861 , though I didn't implement any of the other fixes. As per the related commit:

> z_perspective is true seems to imply w-buffering and then the w-coordinate stored in the depth buffer should also be interpolated in a perspective-correct way. We do this by calculating w and setting gl_FragDepth in the fragment shader.
> 
> Since enabling polygon offset and setting values using glPolygonOffset won't have any effect when manually setting gl_FragDepth, we introduce the depthOffset variable to obtain similar behaviour (but the glPolygonOffset factor-argument is currently not emulated.)

I fixed the regressions by leaving the shader code for no z_perspective unaltered. Also used NV_PGRAPH_ZCLIPMIN and NV_PGRPAH_ZCLIPMAX as clip planes, thus fixing things such as shadows.

### DIFF
--- a/hw/xbox/nv2a/nv2a_int.h
+++ b/hw/xbox/nv2a/nv2a_int.h
@@ -417,6 +417,7 @@ typedef struct PGRAPHState {
 
     unsigned int surface_scale_factor;
     uint8_t *scale_buf;
+    float fragment_depth_offset;
 } PGRAPHState;
 
 typedef struct NV2AState {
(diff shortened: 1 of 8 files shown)


Triticum0:
If you want you want to make a test case so this can get merged, here a [nxdk_pgraph_tests](https://github.com/abaire/nxdk_pgraph_tests/tree/main/src/tests) seems to have some test cases related so your fix so you could modify them and confirm it works correctly on hardware and xemu so it can be merge as it should fix a lot of game as clip range affect many games #1020

polymetal0:
@Triticum0 sure, I have also been testing with the w_param tests and have been looking at the code. But I'm holding until the Vulkan renderer branch gets merged, since it will introduce fundamental changes in how the shader code is built. I'm implementing my changes locally as it goes.
Unfortunately I have no access to hardware for testing, so I would focus on fixing the w_param tests to have a better understanding of how the hardware works here.

Triticum0:
Also forgot to mention you need to fix #1362 as is another reason why clipping occurs on the emulator and not on hardware.

Fabxx:
this draft PR fixes the clipping issues on the first SC

immagine

PR #1732, opened 2024-08-03
HW rounding fix for all scales

polymetal0:
@abaire @antangelo I continued the work from #735 and #1362 by adding the scale factor to the shader function. From the tests I've done, the behavior is the same for all scales.

### DIFF
--- a/hw/xbox/nv2a/shaders.c
+++ b/hw/xbox/nv2a/shaders.c
@@ -859,6 +859,7 @@ GLSL_DEFINE(texMat3, GLSL_C_MAT4(NV_IGRAPH_XF_XFCTX_T3MAT))
     }
     mstring_append(header, "\n");
 
+    unsigned int scale = state->surface_scale_factor;
     mstring_append_fmt(header, "\n"
             "vec2 adjust_pixel_center(vec2 screen_pos, float w) {\n"
             "  if (w == 0.0 || isinf(w)) {\n"
@@ -875,9 +876,9 @@ GLSL_DEFINE(texMat3, GLSL_C_MAT4(NV_IGRAPH_XF_XFCTX_T3MAT))
             "  vec2 bias = vec2(0.002);\n"
             "  subpixel += mix(bias, -bias, round_down);\n"
 
-            "  return w * (pixel + subpixel);\n"
-            "}\n"
-            );
+            "  return w * (pixel + subpixel / %d);\n"
+            "}\n",
+            scale);
 
     MString *body = mstring_from_str("void main() {\n");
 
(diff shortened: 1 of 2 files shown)


antangelo:
I ran the pgraph tests against this and it doesn't look like the output is correct at or above 3x. Testing on BFBB has shadows clipping at or above 2x. If I understand the approach correctly (and I need to spend some more time thinking about it either way), this is essentially flooring the vertex coordinate, applying the scale and then attempting to correct rounding using the unscaled subpixel. At 10x a subpixel offset of 0.5 will expand to 5 pixels error after scaling, which I think is the artifacting I'm seeing.

In any case, I think the more pressing issue to be resolved before any of these approaches can be landed is passing the `ProjAdjacentGeometry_0.5625` test case and fixing the regression in Tiger Woods that the test case is based on. The issue of it not working above 1x is less critical, and we can probably fix it later (if there even is a good fix-- I don't think the problem is well defined).

polymetal0:
@antangelo Thanks for the quick testing. Oh, shame...yeah, I noticed some weird wavy things at high scales. 

I implemented this while working on #1707 because depth testing messes something up and the W-param tests also suffer from rounding issues. I thought maybe the z-fighting/flickering could be solved like this, but nope. 

Still, this fixes the top and left margins issue in games like Fable and Halo CE.

polymetal0:
> In any case, I think the more pressing issue to be resolved before any of these approaches can be landed is passing the `ProjAdjacentGeometry_0.5625` test case and fixing the regression in Tiger Woods that the test case is based on. The issue of it not working above 1x is less critical, and we can probably fix it later (if there even is a good fix-- I don't think the problem is well defined).

In case it might be of help, setting the lessThan operator to lessThanEqual in the vsh code breaks the other 0.5625 tests, but I get this output for this particular case:

ProjAdjacentGeometry_0 5625

MasonT8198:
I've tried some alternate methods to alter the shift in the shader function and I've not found a happy balance that makes 1x look good or not break at higher res scales... just wondering if anyone else here has made any further progress

MasonT8198:
Just doing some more research on forums (https://discussions.unity.com/t/half-a-pixel-visual-offset-between-directx-and-opengl/433187/5) and some DirectX 8 technical documentation from nVidia

Are we sure that this is the right offset for the difference between DX8.1 and GL/VK?

polymetal0:
> Just doing some more research on forums (https://discussions.unity.com/t/half-a-pixel-visual-offset-between-directx-and-opengl/433187/5) and some DirectX 8 technical documentation from nVidia
> 
> Are we sure that this is the right offset for the difference between DX8.1 and GL/VK?

Afaik it's confirmed that the pixel center coordinates are 0.5625 for nv2a. There are a few approaches to solve this that work, but everything I tried breaks at higher scales, most likely because when multiplying the original pixel by surface factor the offset is also multiplied but it cannot be corrected by just dividing.

I remember someone suggesting offsetting the whole viewport or only rounding the left and bottom margins could do the trick. I hope this helps, I haven't made enough research and pretty much came to a dead end here.

Triticum0:
Closed in favor of #2183

mborgerson:
As @Triticum0 has mentioned, we have ultimately gone with another approach to fix the underlying issue. Thank you for your contribution

PR #1737, opened 2024-08-06
scripts/tracetool: Fix SyntaxWarning invalid escape sequence

sertonix:


abaire (commented):
Python changes lgtm.

The pci.c change should probably be upstreamed to qemu instead of patched here to minimize complexity of future qemu updates.

sertonix:
It looks like the usage of `basename` has been removed from the code since creating this MR. I dropped the pci.c change now since it isn't needed anymore.

Ref https://github.com/xemu-project/xemu/commit/5456b1867d10826e3857fda417b59a669fa5cce9 https://github.com/xemu-project/xemu/commit/7e63b3113889b531cbb56b09d69cc182608bb78e https://github.com/xemu-project/xemu/commit/213ae3ffda463c0503e39e0cf827511b5298c314

mborgerson:
Thanks for the patch. It looks like this was fixed upstream a while back. I rebased this PR and GitHub auto-closed it because the diff was eliminated.

PR #1774, opened 2024-09-18
Update apu.c

drewseibert:
This line throws assertion error often


mborgerson:
Thanks for the PR. Assertions help us confirm understanding of the hardware, so let's not remove them without more study.

PR #1787, opened 2024-11-14
Scale line thickness

faha223:
I've added a checkbox to the Video settings to enable scaling the line width to match the render scale. This helps maintain the art direction in Dragon's Lair 3D, but it also makes the HUD legible at higher resolutions in Steel Battalion: Line of Contact, and makes the UI look more correct at higher resolutions in Steel Battalion Line of Contact and in Steel Battalion

Original at 1x scale
xemu-2024-11-14-18-10-48

4x scale with default line width
xemu-2024-11-14-18-10-43

4x scale with 4x line width
xemu-2024-11-14-18-10-38


mborgerson:
When would someone *not* want to scale up the lines? Do we really need it to be configurable?

faha223:
I don't know for certain that there are no games that look worse with this enabled, so I made it configurable just in case. 

The only games that I know for certain use Line primitives are the Steel Battalion games and Dragon's Lair 3D.

mborgerson:
It makes sense to me to just scale the lines up accordingly like we do with everything else. So let's simplify this and drop the configuration unless we find a compelling reason to keep it

Spidy123222:
> When would someone _not_ want to scale up the lines? Do we really need it to be configurable?

I think some would want to keep the line thickness if it looks weird for some games. but some might prefer it smaller (higher res). I propose including a hacks section to disable this function or others and name this toggle “keep line thickness”. This would also make it clear if hacks is enabled that you cannot do a compatibility report. I have that function in my overclocking pull request.

mborgerson:
@Spidy123222 

> I think some would want to keep the line thickness if it looks weird for some games.

Please provide an example case. Which game does it look weird in?

Spidy123222:
> @Spidy123222
> 
> > I think some would want to keep the line thickness if it looks weird for some games.
> 
> Please provide an example case. Which game does it look weird in?

I think it looks weird in the photos shown above but I do see why some would like it. im used to the original thickness.

hw/xbox/nv2a/pgraph.c
```diff
@@ -3956,8 +3958,13 @@ unsigned int nv2a_get_surface_scale_factor(void)
 
 static void pgraph_reload_surface_scale_factor(NV2AState *d)
 {
-    int factor = g_config.display.quality.surface_scale;
-    d->pgraph.surface_scale_factor = factor < 1 ? 1 : factor;
+    d->pgraph.surface_scale_factor = MAX(1, g_config.display.quality.surface_scale);
+}
+
+static void pgraph_reload_line_width(NV2AState *d)
```
mborgerson: Let's drop this new function and just stick the call to `glLineWidth` [here](https://github.com/xemu-project/xemu/blob/8707d2aa2626063cb67b5ea20382584a0848dce7/hw/xbox/nv2a/pgraph.c#L3057) for now with other state setup

ui/xui/main-menu.cc
```diff
@@ -465,7 +465,7 @@ void MainMenuDisplayView::Draw()
                      "Increase surface scaling factor for higher quality")) {
         nv2a_set_surface_scale_factor(rendering_scale+1);
     }
-
+
```
mborgerson: Remove

hw/xbox/nv2a/pgraph.c
```diff
@@ -3053,6 +3053,8 @@ DEF_METHOD(NV097, SET_BEGIN_END)
             glDisable(GL_DITHER);
         }
 
+        float lineWidth = (float)MAX(1, g_config.display.quality.surface_scale);
+        glLineWidth(lineWidth);
```
mborgerson: Let's simplify `glLineWidth(pg->surface_scale_factor)`

hw/xbox/nv2a/pgraph.c
```diff
@@ -3915,8 +3917,8 @@ void nv2a_set_surface_scale_factor(unsigned int scale)
 {
     NV2AState *d = g_nv2a;
 
-    g_config.display.quality.surface_scale = scale < 1 ? 1 : scale;
-
+    g_config.display.quality.surface_scale = MAX(1, scale);
```
mborgerson: Unrelated changes

hw/xbox/nv2a/pgraph.c
```diff
@@ -3956,8 +3958,7 @@ unsigned int nv2a_get_surface_scale_factor(void)
 
 static void pgraph_reload_surface_scale_factor(NV2AState *d)
 {
-    int factor = g_config.display.quality.surface_scale;
-    d->pgraph.surface_scale_factor = factor < 1 ? 1 : factor;
+    d->pgraph.surface_scale_factor = MAX(1, g_config.display.quality.surface_scale);
```
mborgerson: Unrelated changes

mborgerson:
Thanks!

PR #1789, opened 2024-11-15
Update build.yml

erfansvsh:
For Nintendo Switch


mborgerson:
There are currently no Switch builds of xemu (and this PR doesn't make one), but you can run xemu on Linux on Switch (e.g. L4T/Switchroot) if you really want to

PR #1792, opened 2024-11-17
NV2A: Check for supported line widths before setting the line width

faha223:
Added checks to GL_ALIASED_LINE_WIDTH_RANGE and and GL_SMOOTH_LINE_WIDTH_RANGE to ensure that the line width we set is within the supported range. I also moved the glLineWidth call down a few lines so it would be called after we know the supported line width range.


hw/xbox/nv2a/pgraph.c
```diff
@@ -3062,9 +3063,12 @@ DEF_METHOD(NV097, SET_BEGIN_END)
         if (!anti_aliasing && pg->regs[NV_PGRAPH_SETUPRASTER] &
                                   NV_PGRAPH_SETUPRASTER_LINESMOOTHENABLE) {
             glEnable(GL_LINE_SMOOTH);
+            glGetFloatv(GL_SMOOTH_LINE_WIDTH_RANGE, supportedLineWidthRange);
```
antangelo: Would it be better to load both line width ranges into separate arrays once during init and then just select between them based on the above condition?

hw/xbox/nv2a/pgraph.c
```diff
@@ -381,6 +381,8 @@ static const SurfaceFormatInfo kelvin_surface_zeta_fixed_format_map[] = {
         {4, GL_DEPTH24_STENCIL8, GL_DEPTH_STENCIL, GL_UNSIGNED_INT_24_8, GL_DEPTH_STENCIL_ATTACHMENT},
 };
 
+static GLfloat supportedLineWidthRange[2] = { 0.0f, 0.0f };
```
antangelo: If my other comment does not apply, then I don't think this needs to be a static global

antangelo:
The commit/PR title should be of the form `<subsystem>: One line description of the change` (where in this case the subsystem is `nv2a`)

PR #1802, opened 2024-12-04
UI: Device emulation - Xbox Controller S

faha223:
This PR adds the emulation of just the Xbox Controller S. It's the simplest of the devices I'm adding support for because it uses the same XID Input and Output report structures as the Xbox Controller

This PR exists mainly to establish the added UI control for picking which device the user intends to emulate on a given controller port

I also split out the Xbox Gamepad specific structures from the shared XID code in preparation for adding support for other XID devices. I'm going to create PRs for each of those individually in addition to this one. There will be some shared code between the PRs, mostly UI stuff.


ui/xemu-input.c
```diff
@@ -119,6 +128,37 @@ static const char **peripheral_params_settings_map[4][2] = {
 
 static int sdl_kbd_scancode_map[25];
 
+static const char *get_bound_driver(int port)
+{
+    assert(port >= 0 && port <= 3);
+    const char *driver = NULL;
+    switch (port) {
```
antangelo: Can we use the `port_index_to_driver_settings_key_map` lookup table here?

ui/xemu-input.c
```diff
@@ -119,6 +128,37 @@ static const char **peripheral_params_settings_map[4][2] = {
...
+        return DRIVER_DUKE; // Shouldn't be possible
+    if (strlen(driver) == 0)
+        return DRIVER_DUKE;
+    if (strcmp(driver, DRIVER_DUKE) == 0)
+        return DRIVER_DUKE;
+    if (strcmp(driver, DRIVER_S) == 0)
+        return DRIVER_S;
+
+    // Shouldn't be possible
+    return DRIVER_DUKE;
```
antangelo, line 147: This can be reached if the config is manually edited, or if someone jumps between a feature branch that implements a new driver without changing it first.
(Picking the Duke driver is probably a good default here anyway, but the assumption that it's unreachable is not correct)

ui/xemu-input.h
```diff
@@ -81,7 +87,7 @@ typedef struct ControllerState {
     // Input state
     uint16_t buttons;
     int16_t  axis[CONTROLLER_AXIS__COUNT];
-
+
```
antangelo: nit: Extra spacing here

ui/xui/gl-helpers.cc
```diff
@@ -33,7 +34,7 @@
 #include "ui/shader/xemu-logo-frag.h"
 
 Fbo *controller_fbo, *xmu_fbo, *logo_fbo;
-GLuint g_controller_tex, g_logo_tex, g_icon_tex, g_xmu_tex;
+GLuint g_controller_tex, g_controller_s_tex, g_logo_tex, g_icon_tex, g_xmu_tex;
```
antangelo: Should we rename `g_controller_tex` to `g_controller_duke_tex`?

ui/xui/gl-helpers.cc
```diff
@@ -464,7 +467,7 @@ static void RenderMeter(DecalShader *s, float x, float y, float width,
     RenderDecal(s, x, y, width * p, height, 0, 0, 1, 1, 0, 0, color_fg);
 }
 
-void RenderController(float frame_x, float frame_y, uint32_t primary_color,
+static void RenderController_Duke(float frame_x, float frame_y, uint32_t primary_color,
```
antangelo: nit: This doesn't technically follow the naming convention, maybe something like `RenderDukeController` or `RenderControllerS?

ui/xui/main-menu.cc
```diff
@@ -158,6 +158,49 @@ void MainMenuInputView::Draw()
...
+    // Render device driver combo
+    //
+
+    // List available device drivers
+    const char *driver = bound_drivers[active];
+
+    if (strcmp(driver, DRIVER_DUKE) == 0)
+        driver = DRIVER_DUKE_DISPLAY_NAME;
+    else if (strcmp(driver, DRIVER_S) == 0)
+        driver = DRIVER_S_DISPLAY_NAME;
```
antangelo, line 171: What if driver doesn't compare to either of these (e.g. in one of the cases I described above)?

ui/xemu-input.c
faha223: Yeah, I'm not sure why I didn't do that here already

ui/xui/gl-helpers.cc
faha223: I think that's a good idea

ui/xui/main-menu.cc
faha223: bound_drivers is initialized via get_bound_driver which defaults to DRIVER_DUKE when the value read from the config file is not recognized.

bound_drivers can also be changed via the dropdown, but at that point it should only be possible to assign known good values because they're pulled from available_drivers.

I probably should add an assert(false) here if it doesn't match a known good value

faha223: If somehow the value isn't a known good value, then no value in the dropdown will be selected but the user should still be able to assign a value to it

data/controller_mask_s.png
mborgerson, line 1: Can you please add the original source svg for this in case we need to modify it?

It looks like I did not do so myself originally, so I will add the old duke one too.

faha223: I didn't create it as an SVG. I created it in GIMP using the Paths tool and the "Stroke Path" function. I'll have to recreate it as an SVG but I'm not sure how to do that just yet

mborgerson: Ah don't worry about spending extra time on it. If you just want to add the original gimp file we can do it later

faha223: I didn't see your comment until just now. I just finished recreating it as an SVG

mborgerson:
Thanks!

PR #1803, opened 2024-12-06
UI: Device emulation - Steel Battalion Controller

faha223:
This PR adds the emulation of the Steel Battalion Controller. This controller uses different Input and Output XID reports than the Xbox Controller S. I split out the XID Gamepad-specific code from the XID shared code, and added an XID header. The new xid-gamepad.c and xid-steel-battalion.c both reference that header but they have their own code for handling USB endpoint transfers.

Since the Steel Battalion controller doesn't have an internal USB Hub, it's not possible to insert XMUs or other peripherals in the expansion slots. For this reason, rebinding XMUs is disabled and the Expansion Slots section is not rendered when the Steel Battalion controller driver is selected.

This PR shares some changes with https://github.com/xemu-project/xemu/pull/1802


faha223:
In the most recent commit, I changed the keybindings for the Steel Battalion controller to use the mouse for the aiming lever. I mapped the X and Y axes of the stick to the mouse position within the window, and I mapped the main weapon, lock on, and sub weapon controls to the mouse left, mouse right, and mouse x/y buttons (the thumb buttons). 

As a result I felt it was necessary to remove the right mouse button binding to show the popup menu because it was interfering with using the right mouse button in-game

ui/xemu-input.c
```diff
@@ -422,75 +538,340 @@ void xemu_input_update_controllers(void)
...
 
-    for (int i = 0; i < 15; i++) {
-        state->buttons |= kbd[sdl_kbd_scancode_map[i]] << i;
-    }
+    if (state->bound < 0)
+        return;
+
+    const char *bound_driver = get_bound_driver(state->bound);
+    if (strcmp(bound_driver, DRIVER_STEEL_BATTALION) == 0) {
+        state->sbc.buttons = 0;
```
Florin9doi: already present on L542

faha223: Thanks!

blakeplusplus:
Since xemu is still not able to run Steel Battalion past the intro FMV (thus no gameplay), I can only give limited feedback based on what I see. Never played Steel Battalion for real, so I also can't give feedback on how the keybinds might actually work in a real scenario.

- F5, F6, F7, and F8 are bound to Snapshot loading. This could cause accidental snapshot loading during gameplay. I'm not sure exactly how implementation of the keys themselves work, but is there a way you could do combos of keys? EG Shift + 5 instead of F5
- F12 does screenshots. See above
- While in the input display, it'd be good to have a button you can click to lock inputs to the testing window only temporarily. Keys like ESC you cannot press and see that key working within the visualized controller
- Middle mouse click currently doesn't do anything, while the mouse thumb buttons do. Perhaps this could be swapped for increased compatibility with mice that don't have thumb buttons?
- Currently after you select the Steel Battalion Controller, you can select the input device to be your keyboard. However, even though the SBC utilizes the mouse for joystick movement and certain buttons, there is no mention or selection of a mouse as an input device. Perhaps another dropdown should be added, or just clarification within the settings menu itself?
- Double left clicking within Xemu will cause it to go full-screen. I don't know how that should be resolved; whether this PR should account for it or whether an option should be put in somewhere to disable this functionality.
--
- Possible foot pedal support is perhaps outside the scope of this PR, but foot pedals are a lot more common than the SB controller, since foot pedals are often seen in racing setups. They're also relatively cheap to buy on their own.

faha223:
I think adding support for binding multiple physical devices to simulate the Steel Battalion controller is the way to go. That said, it would require a whole new section to the config file for custom bindings, and there would need to be a way to assign bindings within the UI. I'll have to give it some thought. It's a good idea but it's a lot of work. Especially for two niche titles, one of which doesn't get past the menus, while the other runs very, very poorly at the moment.

PR #1806, opened 2024-12-11
Tests/chihiro stuffs

GXTX:
Just doing some quick bring-ups of old stuff.

Maybe an interested party can continue and actually hookup the block driver to a device. :)

The `DriveInfo` struct changed at some point and doesn't allow specifying it's `BlockDriver`

`BlockDriver`'s themselves also changed and no longer accept member functions 
```
    .bdrv_read = memory_read,
    .bdrv_write = memory_write,
```
Rather, are expecting some AIO, or CO? functions, `bdrv_co_preadv`, `bdrv_aoi_preadv`.
I didn't dig into what the difference between the 2 are and just hooked up the memory_read/write functions do the CO type.

### DIFF
--- a/block/blkmemory.c
+++ b/block/blkmemory.c
@@ -1,0 +1,152 @@
+/*
+ * MemoryRegion backed block driver
+ *
+ * Copyright (c) 2013 espes
+ *
+ * Based on "Add an in-memory block device" patch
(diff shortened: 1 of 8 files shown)


GXTX:
```
-device lpcsega
-drive index=1,media=disk,format=memory
```

PR #1866, opened 2025-01-13
meson: try some system libs first

abouvier:
`volk`, `VMA` and `SPIRV-Reflect` could be unbundled too ;)


subprojects/SPIRV-Reflect.wrap
```diff
@@ -2,3 +2,4 @@
 url=https://github.com/KhronosGroup/SPIRV-Reflect
 revision=vulkan-sdk-1.3.296.0
 depth=1
+method=cmake
```
mborgerson: * Is this necessary? Since meson.build already uses the CMake module to handle this wrap, I'm not sure why this change is required.
* `method` looks to be introduced with Meson v1.3.0, but meson.build requires meson_version: '>=1.1.0'.  If it really is required, then we need to bump the minimum Meson version.

ui/thirdparty/meson.build
```diff
@@ -29,14 +29,20 @@ endif
...
+  httplib_deps = [httplib_proj.get_variable('cpp_httplib_dep')]
+  if host_os == 'windows'
+    httplib_deps += [crypt32]
+  endif
+  httplib = declare_dependency(dependencies: httplib_deps)
 endif
-httplib = declare_dependency(dependencies: httplib_deps)
 
-libfatx = static_library('fatx', sources: 'fatx/fatx.c')
+libfatx = static_library('fatx', sources: files('fatx/fatx.c') + genh)
```
mborgerson, line 37: Why does libfatx now depend on genh?

subprojects/SPIRV-Reflect.wrap
abouvier: `meson subprojects download` exits with an error code when cmake subprojects exists:
```
$ meson subprojects download
(...)
Download nv2a_vsh_cpu...
  -> Subproject exists but has no meson.build file.
(...)
Download SPIRV-Reflect...
  -> Subproject exists but has no meson.build file.
(...)
WARNING: Please check logs above as command failed in some subprojects which could have been left in conflict state: SPIRV-Reflect, VulkanMemoryAllocator, glslang, nv2a_vsh_cpu, volk
$ echo $?
5
```

ui/thirdparty/meson.build
abouvier: Because otherwise you get this error:
```
[13/1482] Linking static target thirdparty/libvma.a
samu: job failed: cc -Iui/thirdparty/libfatx.a.p -Iui/thirdparty -I../xemu/ui/thirdparty -I/usr/include/glib-2.0 -I/usr/lib/glib-2.0/include -I/usr/include/sysprof-6 -fdiagnostics-color=auto -Wall -Winvalid-pch -std=gnu11 -Wempty-body -Wendif-labels -Wexpansion-to-defined -Wformat-security -Wformat-y2k -Wignored-qualifiers -Wimplicit-fallthrough=2 -Winit-self -Wmissing-format-attribute -Wmissing-prototypes -Wnested-externs -Wold-style-declaration -Wold-style-definition -Wredundant-decls -Wshadow=local -Wstrict-prototypes -Wtype-limits -Wwrite-strings -Wno-missing-include-dirs -Wno-psabi -Wno-shift-negative-value -isystem /build/xemu-git/src/xemu/linux-headers -isystem linux-headers -iquote . -iquote /build/xemu-git/src/xemu -iquote /build/xemu-git/src/xemu/include -iquote /build/xemu-git/src/xemu/host/include/x86_64 -iquote /build/xemu-git/src/xemu/host/include/generic -iquote /build/xemu-git/src/xemu/tcg/i386 -pthread -mcx16 -msse2 -D_GNU_SOURCE -D_FILE_OFFSET_BITS=64 -D_LARGEFILE_SOURCE -fno-strict-aliasing -fno-common -fwrapv -ftrivial-auto-var-init=zero -fzero-call-used-regs=used-gpr -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fexceptions -Wp,-D_FORTIFY_SOURCE=3 -Wformat -Werror=format-security -fstack-clash-protection -fcf-protection -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -flto=auto -DXBOX=1 -fPIE -MD -MQ ui/thirdparty/libfatx.a.p/fatx_fatx.c.o -MF ui/thirdparty/libfatx.a.p/fatx_fatx.c.o.d -o ui/thirdparty/libfatx.a.p/fatx_fatx.c.o -c ../xemu/ui/thirdparty/fatx/fatx.c
In file included from /build/xemu-git/src/xemu/include/qemu/osdep.h:38,
                 from ../xemu/ui/thirdparty/fatx/fatx.h:4,
                 from ../xemu/ui/thirdparty/fatx/fatx.c:1:
/build/xemu-git/src/xemu/include/exec/poison.h:7:10: fatal error: config-poison.h: No such file or directory
    7 | #include "config-poison.h"
      |          ^~~~~~~~~~~~~~~~~
```

mborgerson:
Thanks for the patch! I've rebased, removed some changes, and fixed a couple of other things.

abouvier:
Why did you force `pkg-config` for `glslang`? It is only discoverable by cmake on Arch Linux for example.

mborgerson:
It's used elsewhere, but actually I don't remember if it's required. Would need to double check.

I'm not sure why the arch distribution of glslang doesn't support pkg-config, frankly I think it should.

Anyway, feel free to send a PR and we'll find out.

abouvier:
> It's used elsewhere, but actually I don't remember if it's required. Would need to double check.

Without specifying any method, both pkg-config and cmake are tried.

> I'm not sure why the arch distribution of glslang doesn't support pkg-config, frankly I think it should.

Because glslang [does not provide any pc file](https://github.com/KhronosGroup/glslang/issues/1715) officially and Arch Linux doesn't add unnecessary patches.

> Anyway, feel free to send a PR and we'll find out.

I will try ;)

mborgerson:
> > It's used elsewhere, but actually I don't remember if it's required. Would need to double check.
> 
> Without specifying any method, both pkg-config and cmake are tried.

You are right! Thanks

> 
> > I'm not sure why the arch distribution of glslang doesn't support pkg-config, frankly I think it should.
> 
> Because glslang [does not provide any pc file](https://github.com/KhronosGroup/glslang/issues/1715) officially and Arch Linux doesn't add unnecessary patches.
> 

Ah. I see Ubuntu packaging has included the pkg-config compatibility. Maybe upstream will support it eventually. I disagree that it's unnecessary, but happy that meson solves this problem for our purposes.

> > Anyway, feel free to send a PR and we'll find out.
> 
> I will try ;)

Thanks!

PR #1872, opened 2025-01-15
nv2a: Vulkan line width

faha223:
Set LineWidth to SurfaceScaleFactor to match the line width behavior of the OpenGL renderer

### DIFF
--- a/hw/xbox/nv2a/pgraph/vk/draw.c
+++ b/hw/xbox/nv2a/pgraph/vk/draw.c
@@ -951,7 +951,7 @@ static void create_pipeline(PGRAPHState *pg)
     int num_dynamic_states = 2;
 
     snode->has_dynamic_line_width =
-        (r->physical_device_features.wideLines == VK_TRUE)
+        (r->enabled_physical_device_features.wideLines == VK_TRUE)
         && (r->shader_binding->state.polygon_front_mode == POLY_MODE_LINE ||
             r->shader_binding->state.primitive_mode == PRIM_TYPE_LINES ||
             r->shader_binding->state.primitive_mode == PRIM_TYPE_LINE_LOOP ||
--- a/hw/xbox/nv2a/pgraph/vk/draw.c
+++ b/hw/xbox/nv2a/pgraph/vk/draw.c
@@ -810,13 +810,21 @@ static void create_pipeline(PGRAPHState *pg)
         // FIXME: Handle in shader?
     }
 
+    float lineWidth = 1.0f;
+    if(r->physical_device_features.wideLines == VK_TRUE)
+    {
+        lineWidth = MIN(r->device_props.limits.lineWidthRange[1],
+                    MAX(r->device_props.limits.lineWidthRange[0],
+                        (float)pg->surface_scale_factor));
+    }
+
     VkPipelineRasterizationStateCreateInfo rasterizer = {
         .sType = VK_STRUCTURE_TYPE_PIPELINE_RASTERIZATION_STATE_CREATE_INFO,
         .depthClampEnable = VK_TRUE,
         .rasterizerDiscardEnable = VK_FALSE,
         .polygonMode = pgraph_polygon_mode_vk_map[r->shader_binding->state
                                                       .polygon_front_mode],
-        .lineWidth = 1.0f,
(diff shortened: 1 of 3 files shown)


hw/xbox/nv2a/pgraph/vk/display.c
```diff
@@ -413,7 +413,7 @@ static void create_display_pipeline(PGRAPHState *pg)
         .depthClampEnable = VK_FALSE,
         .rasterizerDiscardEnable = VK_FALSE,
         .polygonMode = VK_POLYGON_MODE_FILL,
-        .lineWidth = 1.0f,
+        .lineWidth = (float)pg->surface_scale_factor,
```
mborgerson: Like for GL, before using a value other than 1, we must check that the device supports wideLines and the supported value ranges. Check out VkPhysicalDeviceFeatures/VkPhysicalDeviceLimits

faha223: Thanks! I removed this change as well as the one in create_clear_pipeline in draw.c because it became apparent that these functions aren't being called when surface scale factor changes and that doesn't seem to affect the rendered image.

In the remaining change in create_pipeline in draw.c I've added logic to clamp the line width to a value within the range defined in r->device_props.limits.lineWidthRange

mborgerson: Yeah, display and clear pipelines shouldn't be drawing lines. Only draw pipeline needs the line width adjustment

hw/xbox/nv2a/pgraph/vk/draw.c
```diff
@@ -809,14 +809,18 @@ static void create_pipeline(PGRAPHState *pg)
     } else {
         // FIXME: Handle in shader?
     }
+    
+    float lineWidth = MIN(r->device_props.limits.lineWidthRange[1],
```
mborgerson: Also need to check for `wideLines` physical device feature (see vk/instance.c), and tolerate device not supporting it

faha223: Thanks, I added a check for the wideLines feature before attempting to set it. If the wideLines feature is not supported by the physical device then it leaves it at 1.0f

hw/xbox/nv2a/pgraph/vk/draw.c
```diff
@@ -809,14 +809,25 @@ static void create_pipeline(PGRAPHState *pg)
     } else {
         // FIXME: Handle in shader?
     }
+    
+    VkPhysicalDeviceFeatures available_features;
+    vkGetPhysicalDeviceFeatures(r->physical_device, &available_features);
```
mborgerson: Should not call this here (relatively hot spot), see vk/instance.c for where we query device features

hw/xbox/nv2a/pgraph/vk/draw.c
```diff
@@ -945,8 +956,8 @@ static void create_pipeline(PGRAPHState *pg)
         .blendConstants[3] = blend_constant[3],
     };
 
-    VkDynamicState dynamic_states[2] = { VK_DYNAMIC_STATE_VIEWPORT,
```
mborgerson, line 948: Unrelated change

hw/xbox/nv2a/pgraph/vk/draw.c
faha223: I've removed this call. In its place I added the VkPhysicalDeviceFeatures object to the PGRAPHVkState struct so that I can use the stored result here. Is that change ok?

hw/xbox/nv2a/pgraph/vk/draw.c
faha223: I've removed this change

ShizCalev:
Fixes https://github.com/xemu-project/xemu/issues/1939

ShizCalev (approved):
Implementation is correct. :)

ShizCalev:
@faha223 You seem to have a bad merge, you're readding an old submodule to the repo; `subprojects/cpp-httplib`

Just delete that folder and you should be good to go.

mborgerson:
Thanks for the patch! I squashed+rebased the branch and added some improvements.

PR #1873, opened 2025-01-16
UI: Add Xbox Live Communicator to UI

faha223:
Added UI for attaching XBLC to expansion port
Refactored XBLC to use SDL2 Audio instead of QEMU Audio so that audio devices can be selected


faha223:
It still needs a bit more testing, and I need to recreate the quick-and-dirty XBLC image as a proper SVG

MasonT8198:
Looks good so far!

Just checking, does this support XBLC with an XMU attached (port 1/port 2) etc

faha223:
> Looks good so far!
> 
> Just checking, does this support XBLC with an XMU attached (port 1/port 2) etc

Yes. Each controller has 2 Expansion slots. The XBLC can only go in Slot A but you can have an XBLC in Slot A and an XMU in Slot B, just like on a real controller

hw/xbox/xblc.c
```diff
@@ -164,85 +165,229 @@ static void usb_xblc_handle_reset(USBDevice *dev)
...
-    while (avail > 0 && !fifo8_is_full(&s->in.fifo)) {
-        max_len   = MIN(sizeof(s->in.packet), fifo8_num_free(&s->in.fifo));
-        processed = AUD_read(s->in.voice, s->in.packet, max_len);
-        avail    -= processed;
-        fifo8_push_all(&s->in.fifo, s->in.packet, processed);
-        if (processed < max_len) return;
+    USBXBLCState *s = (USBXBLCState *)userdata;
+    
+    s->in.average_volume = s->in.volume * 
+        calc_average_amplitude((int16_t*)stream, len / 2) / 128;
```
mborgerson: * If dividing by two to calculate number of samples from number of bytes, can you make it more explicit, e.g. (`size_t num_samples = len / sizeof(int16_t);`
* I assume `/128` is dividing through by `SDL_MIX_MAXVOLUME`?

hw/xbox/xblc.c
```diff
@@ -164,85 +165,229 @@ static void usb_xblc_handle_reset(USBDevice *dev)
...
-    // Write speaker data into audio backend
-    while (avail > 0 && !fifo8_is_empty(&s->out.fifo)) {
-        max_len   = MIN(fifo8_num_used(&s->out.fifo), avail);
-        data      = fifo8_pop_bufptr(&s->out.fifo, max_len, &max_len);
-        processed = AUD_write(s->out.voice, (void *)data, max_len);
-        avail    -= processed;
-        if (processed < max_len) return;
+static int calc_average_amplitude(const int16_t *samples, int len) {
+    int max = 0;
+    for(int i = 0; i < len / 2; i++) {
```
mborgerson: Why divide by 2 here? If `samples` is given as a `int16_t`, `len` should be the number of samples not bytes.

hw/xbox/xblc.c
```diff
@@ -164,85 +165,229 @@ static void usb_xblc_handle_reset(USBDevice *dev)
...
+}
 
-    // Write speaker data into audio backend
-    while (avail > 0 && !fifo8_is_empty(&s->out.fifo)) {
-        max_len   = MIN(fifo8_num_used(&s->out.fifo), avail);
-        data      = fifo8_pop_bufptr(&s->out.fifo, max_len, &max_len);
-        processed = AUD_write(s->out.voice, (void *)data, max_len);
-        avail    -= processed;
-        if (processed < max_len) return;
+static int calc_average_amplitude(const int16_t *samples, int len) {
```
mborgerson: This function doesn't calculate the average?

hw/xbox/xblc.c
```diff
@@ -164,85 +165,229 @@ static void usb_xblc_handle_reset(USBDevice *dev)
...
+    USBXBLCState *s = (USBXBLCState *)userdata;
+    
+    s->in.average_volume = s->in.volume * 
+        calc_average_amplitude((int16_t*)stream, len / 2) / 128;
+
+    // Don't try to put more into the queue than will fit
+    uint32_t max_len = MIN(len, fifo8_num_free(&s->in.fifo));
+    if (max_len > 0) {
+        if(s->in.volume < SDL_MIX_MAXVOLUME) {
+            uint8_t *buffer = g_malloc(max_len);
```
mborgerson: We should avoid memory allocation in this very frequently called function

hw/xbox/xblc.c
```diff
@@ -164,85 +165,229 @@ static void usb_xblc_handle_reset(USBDevice *dev)
...
+        channel->device_name = g_strdup(device_name);
+
+    fifo8_reset(&channel->fifo);
+    if(capture)
+        s->in.average_volume = 0;
+
+    SDL_AudioSpec desired_spec;
+    desired_spec.channels = 1;
+    desired_spec.freq = s->sample_rate;
+    desired_spec.format = AUDIO_S16LSB;
```
mborgerson: ~~What if the actual format differs from the desired format?~~ SDL apparently does conversion for you.

hw/xbox/xblc.c
```diff
@@ -164,85 +165,229 @@ static void usb_xblc_handle_reset(USBDevice *dev)
...
+    return s->out.volume;
+}
+
+int xblc_audio_stream_get_input_volume(void *dev)
+{
+    USBXBLCState *s = (USBXBLCState*)dev;
+    return s->in.volume;
+}
+
+void xblc_audio_stream_set_output_volume(void *dev, int volume)
```
mborgerson: You have to dig a bit to figure out that volume should be given in [0, SDL_MIX_MAXVOLUME]. We should  document it. Alternatively, a float between [0-1] would be reasonable

hw/xbox/xblc.c
```diff
@@ -164,85 +165,229 @@ static void usb_xblc_handle_reset(USBDevice *dev)
...
+}
+
+static bool should_init_stream(const XBLCStream *stream, const char *requested_device_name)
+{
+    // If the voice has not been initialized, initialize it
+    if (stream->voice == 0)
+        return true;
+    
+    // If one of the names is null and the other is not, initialize it
+    if ((stream->device_name == NULL) ^ (requested_device_name == NULL))
```
mborgerson: What does it mean when they are both `NULL`? Can this happen?

ui/xemu-input.c
```diff
@@ -489,10 +578,16 @@ void xemu_input_bind(int index, ControllerState *state, int save)
         // Unbind any XMUs
```
mborgerson: Comment out of date

ui/xemu-input.c
```diff
@@ -489,10 +578,16 @@ void xemu_input_bind(int index, ControllerState *state, int save)
...
         for (int i = 0; i < 2; i++) {
             if (bound_controllers[index]->peripherals[i]) {
-                // If this was an XMU, unbind the XMU
-                if (bound_controllers[index]->peripheral_types[i] ==
-                    PERIPHERAL_XMU)
-                    xemu_input_unbind_xmu(index, i);
+                // If a peripheral was bound, unbind the peripheral
+                switch (bound_controllers[index]->peripheral_types[i]) {
```
mborgerson: How about just testing with PERIPHERAL_NONE?

ui/xemu-input.c
```diff
@@ -359,6 +366,88 @@ void xemu_input_process_sdl_events(const SDL_Event *event)
...
+
+    const char *output_device_label = xblc->output_device_name == NULL ?
+        default_device_name : xblc->output_device_name;
+    const char *input_device_label = xblc->input_device_name == NULL ?
+        default_device_name : xblc->input_device_name;
+
+    return g_strdup_printf("%d|%d|%s|%s", (int)(200 * xblc->output_device_volume), (int)(200 * xblc->input_device_volume), output_device_label, input_device_label);
+}
+
+XblcState *xemu_input_deserialize_xblc_settings(const char *param)
```
mborgerson: This custom config parsing is tricky and I'd strongly like to avoid it. I think we can simplify it to an array of structures with genconfig and avoid parsing altogether. We can also enhance genconfig if need be

ui/xui/main-menu.cc
```diff
@@ -277,171 +278,359 @@ void MainMenuInputView::Draw()
...
+static void DrawVolumeControlSlider(int active, XblcState *xblc, int is_capture)
+{
+    float max_width = ImGui::GetColumnWidth() - (10 * g_viewport_mgr.m_scale);
+    xblc->input_device_volume = xblc_audio_stream_get_input_volume(xblc->dev) / 128.0f;
+    xblc->output_device_volume = xblc_audio_stream_get_output_volume(xblc->dev) / 128.0f;
+    float *ui_volume = (is_capture == 0) ? &xblc->output_device_volume : &xblc->input_device_volume;
+    float original_volume = *ui_volume;
+    const char *label = (is_capture == 0) ? "Speaker" : "Microphone";
+    char description[32];
+    sprintf(description, "Volume: %.2f%%", 100 * original_volume);
```
mborgerson: Don't need so much precision

ui/xui/main-menu.cc
```diff
@@ -277,171 +278,359 @@ void MainMenuInputView::Draw()
...
+    const char *combo_label = (is_capture == 0) ? "###Speaker" : "###Microphone";
+    // ImGui::Text("%s", label_text);
+    ImGui::SetNextItemWidth(max_width);
+    if(ImGui::BeginCombo(combo_label, selected_device, ImGuiComboFlags_NoArrowButton)) {
+
+        int numOutputDevices = SDL_GetNumAudioDevices(is_capture);
+        for(int device_index = -1; device_index < numOutputDevices; device_index++) {
+            const char *device_name = default_device_name;
+            if(device_index >= 0)
+                device_name = SDL_GetAudioDeviceName(device_index, is_capture);
```
mborgerson: Probably want to cache these device names and not query them on every frame

mborgerson (commented):
Thanks for the work. Some initial feedback

hw/xbox/xblc.c
faha223: If `stream->device_name == NULL` is true then the stream is currently initialized to the default device
If `requested_device_name == NULL` is true then the function is being called to check if it should initialize the default audio device

If both are true then that means it's checking if the stream should be initialized but the stream is already initialized to the default device. This happens when the user changes the mic or speaker away from default while the other is still set to default.

hw/xbox/xblc.c
faha223: I've changed this to use a function scope buffer of a fixed size, but would it be better to use a static global scope buffer?

BarracaDoSeuZe:
i was trying to make an Portable .ini Setting for Xemu, and i found a problem while trying to automatically configure the XBLC into the Controller. if you set "[input.peripherals.port1] peripheral_type_0 = 2" without any "[input.bindings]" set in, the emulator will hard crash with any game that actually uses the XBLC. can this be somewhat improved?

MasonT8198:
Xemu already supports portable mode.

faha223:
> i was trying to make an Portable .ini Setting for Xemu, and i found a problem while trying to automatically configure the XBLC into the Controller. if you set "[input.peripherals.port1] peripheral_type_0 = 2" without any "[input.bindings]" set in, the emulator will hard crash with any game that actually uses the XBLC. can this be somewhat improved?

It doesn't make sense to have an [input.peripherals.port1] section in your ini if there is no controller connected to port 1. You can't attach an XBLC to the Console if the controller it connects through does not exist.

It shouldn't cause a crash the emulator though. I'm going to investigate this and see if I can fix it. Thanks for finding this!

BarracaDoSeuZe:
> It doesn't make sense to have an [input.peripherals.port1] section in your ini if there is no controller connected to port 1. You can't attach an XBLC to the Console if the controller it connects through does not exist.

This was a thing i was testing, while using the Auto Bind controller option, the emulator would automatically setup the xbcl into the Controller Port ( since i want everything to organize itself automatically, without me having to configure anything ). at least doing this actually works, the only problem is the crash .-

faha223:
> > It doesn't make sense to have an [input.peripherals.port1] section in your ini if there is no controller connected to port 1. You can't attach an XBLC to the Console if the controller it connects through does not exist.
> 
> This was a thing i was testing, while using the Auto Bind controller option, the emulator would automatically setup the xbcl into the Controller Port ( since i want everything to organize itself automatically, without me having to configure anything ). at least doing this actually works, the only problem is the crash .-

I've made a change that I think should fix this issue. I was able to easily replicate it with Halo 2. After the build is finished, can you check if the issue you're seeing still exists?

BarracaDoSeuZe:
> I've made a change that I think should fix this issue. I was able to easily replicate it with Halo 2. After the build is finished, can you check if the issue you're seeing still exists?

Tested it here, and now it does not Crash on Startup ;). but i encountered a weird glitch, but I don't know if that would catch your attention.
After letting the emulator automatically pickup the Controller, it doens't seem to know what Audio Device he should be getting. i thought it would use the "Default" option, but it get this random text.
xblcglitch

faha223:
> > I've made a change that I think should fix this issue. I was able to easily replicate it with Halo 2. After the build is finished, can you check if the issue you're seeing still exists?
> 
> Tested it here, and now it does not Crash on Startup ;). but i encountered a weird glitch, but I don't know if that would catch your attention. After letting the emulator automatically pickup the Controller, it doens't seem to know what Audio Device he should be getting. i thought it would use the "Default" option, but it get this random text. xblcglitch

I saw this exactly once, but I haven't been able to reproduce it

faha223:
> > I've made a change that I think should fix this issue. I was able to easily replicate it with Halo 2. After the build is finished, can you check if the issue you're seeing still exists?
> 
> Tested it here, and now it does not Crash on Startup ;). but i encountered a weird glitch, but I don't know if that would catch your attention. After letting the emulator automatically pickup the Controller, it doens't seem to know what Audio Device he should be getting. i thought it would use the "Default" option, but it get this random text. xblcglitch

I found the cause of this bug, I think. I was caching the value returned by SDL_GetAudioDeviceName without making my own copy

ui/xui/main-menu.cc
```diff
@@ -572,6 +414,369 @@ void MainMenuInputView::Draw()
...
+
+    ImGui::SetCursorPosX(
+        ImGui::GetCursorPosX() +
+        (int)((ImGui::GetColumnWidth() -
+                xmu_w * g_viewport_mgr.m_scale -
+                2 * port_padding * g_viewport_mgr.m_scale) /
+                2));
+
+    xmu_fbo->Target();
+    ImTextureID id = (ImTextureID)(intptr_t)xmu_fbo->Texture();
```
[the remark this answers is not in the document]
faha223: This value is used in the call to ImGui::Image 20 lines later

mborgerson:
Now that SDL3 is merged, we'll have to migrate the SDL changes to SDL3, which fortunately has improved audio support. I can help with this. May want to just separate out the UI changes from the XBLC changes so we can solve them separately. What do you think?

faha223:
> Now that SDL3 is merged, we'll have to migrate the SDL changes to SDL3, which fortunately has improved audio support. I can help with this. May want to just separate out the UI changes from the XBLC changes so we can solve them separately. What do you think?

I think separating the xblc changes from the ui changes is a good idea. It just means the ui PR will have to wait til the xblc PR is merged. I'm almost done converting the xblc code to SDL3 already, so I'll just copy those changes into a new branch and make a PR once I've confirmed that it's good.

PR #1875, opened 2025-01-18
UI: Device Emulation - Light Gun

faha223:
This PR adds the emulation of just the Light Gun. It can be emulated using the Keyboard and Mouse, which should make it possible to use any modern PC light gun as well

I also split out the Xbox Gamepad specific structures from the shared XID code in preparation for adding support for other XID devices. This shares code with the Steel Battalion controller PR as they both use mouse input for Keyboard and Mouse controls.

I'm making this a Draft because there are some things I'd like to redo the Light Gun image and make it an SVG


Nicktamer:
Hi, any news about this PR?
Any merging soon into master?

faha223:
As of right now, the holdup is mostly just that I'm not completely satisfied with the Light Gun Mask texture. A lot of the buttons are on the other side and they're not visible

Triticum0:
Can this be updated to also support fps game with keyboard and mouse which use controller inputs. Just a question if it possible.

faha223:
I believe that would be out of scope for this PR

Triticum0:
OK thanks.  Just wanted to know. I Remember testing games, before I had a controller and fps on a keyboard only was a nightmare.

PR #1886, opened 2025-01-26
nv2a: Speed up swizzle even more

mborgerson:
Multiversion [un]swizzle to optimize for common bpp. ~3x speedup (now ~7.7GiB/s on my machine).

### DIFF
--- a/hw/xbox/nv2a/pgraph/swizzle.c
+++ b/hw/xbox/nv2a/pgraph/swizzle.c
@@ -4,6 +4,7 @@
  * Copyright (c) 2015 Jannik Vogel
  * Copyright (c) 2013 espes
  * Copyright (c) 2007-2010 The Nouveau Project.
+ * Copyright (c) 2025 Matt Borgerson
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
@@ -68,7 +69,7 @@ static void generate_swizzle_masks(unsigned int width,
     *mask_z = z;
 }
 
-void swizzle_box(
+static inline void swizzle_box_internal(
     const uint8_t *src_buf,
     unsigned int width,
     unsigned int height,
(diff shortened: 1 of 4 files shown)


ehw:
Not sure if you want some testing yet. I regret not testing @NZJenkins's PR 1608 before, but the performance benefits look pretty promising.

The Dead or Alive 1 port's weird performance issues are gone with these changes. The game on my 5800X could only do about 40fps tops at all times during some matches, now it's a solid 60fps at all times. Even at 10x IR with Vulkan. :)


v.0.8.8
image

v.0.8.14
image

This PR:
image


I can test some more games with performance problems (not related to Vulkan) that I'm aware of if needed. :)

Amazing work from the both of you!

hw/xbox/nv2a/pgraph/swizzle.c
```diff
@@ -149,24 +150,35 @@ void unswizzle_box(
...
+/* Multiversioned to optimize for common bytes_per_pixel */         \
+#define C(m, bpp)                                                   \
+    m##_internal(src_buf, width, height, depth, dst_buf, row_pitch, \
+                 slice_pitch, bpp)
+#define MULTIVERSION(m)                                                     \
+    void m(const uint8_t *src_buf, unsigned int width, unsigned int height, \
+           unsigned int depth, uint8_t *dst_buf, unsigned int row_pitch,    \
+           unsigned int slice_pitch, unsigned int bytes_per_pixel)          \
+    {                                                                       \
+        switch (bytes_per_pixel) {                                          \
```
GXTX, line 162: Which is used more often? We can micro optimize by putting the most common on the top, especially considering how "hot" this path is. :^)

Triticum0:
Improved PGR performance significantly on opengl
v.0.8.8
Screenshot_20250305_231437

v.0.8.14
PGR 0 8 14

Masters
pgr 0 8 20

Vulkun performs is better but 20 fps worse avgs

PR #1889, opened 2025-01-28
build.yml: Fix PPA Deployment

MasonT8198:
Due to some really long multi-year issue/regression you must either manually patch python on Ubuntu 22.04 or force pip to grab the deps to setup the venv, 99% sure this will fix it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -450,6 +450,9 @@ jobs:
 
         # Ensure subprojects are uploaded
         rm src/subprojects/.gitignore
+    - name: Install dependencies
+      run: |
+        python3 -m pip install --upgrade build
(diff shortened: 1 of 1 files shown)


MasonT8198:
https://bugs.launchpad.net/ubuntu/+source/python-build/+bug/1992108

.github/workflows/build.yml
```diff
@@ -450,6 +450,9 @@ jobs:
 
         # Ensure subprojects are uploaded
         rm src/subprojects/.gitignore
+    - name: Install dependencies
+      run: |
+        python3 -m pip install --upgrade build
```
mborgerson, line 455: This changes a build step on GitHub actions, but won't affect PPA build (which happens on launchpad)

Issue #587, opened 2021-12-04
"`NV_PGRAPH_TEXFILTER_SIGNED` not supported" - leads to incorrect looking textures


Triticum0:
### Bug Description

The following games are known to be affected by the following assert block:

https://github.com/mborgerson/xemu/blob/aa7acc8f64e64bb517a0bd0e5e41672671e61df1/hw/xbox/nv2a/pgraph.c#L5786

That code handles asserts for NV_PGRAPH_TEXFILTER0_ASIGNED, NV_PGRAPH_TEXFILTER0_RSIGNED, NV_PGRAPH_TEXFILTER0_GSIGNED and NV_PGRAPH_TEXFILTER0_BSIGNED which are not supported.

Only the first triggered assert is categorized

A-Signed (possibly also R,G,B):

    Pirates: The Legend of Black Kat

R-Signed (possibly also G,B):

    American McGee Presents: Scrapland
    Bass Pro Shops: Trophy Bass 2007
    Ninja Gaiden Black
    Power Drome
    SWAT: Global Strike Team
    Urban Chaos: Riot Response
    Midnight Club 3: DUB Edition

G-Signed (possibly also B):

    Crimson Skies: High Road to Revenge (Also triggers B-Signed)
    Stake: Fortune Fighters
    Tork: Prehistoric Punk
    Wings of War
    Buffy the vampire slayer

Affected but unknown channels:

- [Dead or Alive 1 Ultimate](https://xemu.app/titles/54430006/#Dead-or-Alive-1-Ultimate)
- [Farcry Instincts](https://xemu.app/titles/55530008/#Far-Cry-Instincts)
- [The Spongebob Squarepants Movie](https://xemu.app/titles/54510027/)
- [Halo 2](https://xemu.app/titles/4d530064/#Halo-2)
- [Blinx 2 - Masters of Time & Space](https://xemu.app/titles/4d530065/#Blinx-2-Battle-of-Time-Space-Blinx-2-Masters-of-Time-Space)
- [Blazing Angels: Squadrons of WWII](https://xemu.app/titles/55530057/#Blazing-Angels-Squadrons-of-WWII)
- [Links 2004](https://xemu.app/titles/4d53005d/#Links-2004)
- [Malice](https://xemu.app/titles/4253000a/#Malice)
- [Heroes of the Pacific](https://xemu.app/titles/45430004/#Heroes-of-the-Pacific)
- [Fable](https://xemu.app/titles/4d53000d/#Fable)
[...]

PR #1892, opened 2025-01-30
add support for some signed texture formats

oltolm:
Even though we can not support the general case, we can easily support the case when all components are signed. See #587.

### DIFF
--- a/hw/xbox/nv2a/pgraph/gl/constants.h
+++ b/hw/xbox/nv2a/pgraph/gl/constants.h
@@ -222,7 +222,7 @@ static const ColorFormatInfo kelvin_color_format_gl_map[66] = {
          {GL_RED, GL_RED, GL_RED, GL_GREEN}},
 
     [NV097_SET_TEXTURE_FORMAT_COLOR_SZ_R6G5B5] =
-        {2, false, GL_RGB8_SNORM, GL_RGB, GL_BYTE}, /* FIXME: This might be signed */
+        {2, false, GL_RGB8, GL_RGB, GL_UNSIGNED_BYTE},
     [NV097_SET_TEXTURE_FORMAT_COLOR_SZ_G8B8] =
         {2, false, GL_RG8, GL_RG, GL_UNSIGNED_BYTE,
          {GL_RED, GL_GREEN, GL_RED, GL_GREEN}},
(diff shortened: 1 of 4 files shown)


oltolm:
@abaire Sorry for the ping, but can you take a look?

mborgerson:
FYI while glancing at @coldhex's development branch in regards to #1895 and #1896, I noticed they also had been [working on signed texture](https://github.com/xemu-project/xemu/commit/6cad12d74c89983ab4e6b3eda8c0858f306c6b56#diff-ef21fd8a996afe41274a2441bdf4e7331bad001100cbfa03dc716bf6ac06f44b) support. I haven't reviewed their code yet, but we should coordinate here to avoid duplicate efforts.

oltolm:
Sure, I am all for it.

oltolm:
I think this PR is useful regardless of other PRs because it solves the problem in the specific case when all components are signed. It does not conflict with other more general solutions and can coexist with them.

oltolm:
@coldhex Can you take a look?

hw/xbox/nv2a/pgraph/gl/constants.h
```diff
@@ -319,4 +330,12 @@ static const SurfaceFormatInfo kelvin_surface_zeta_fixed_format_gl_map[] = {
         {4, GL_DEPTH24_STENCIL8, GL_DEPTH_STENCIL, GL_UNSIGNED_INT_24_8, GL_DEPTH_STENCIL_ATTACHMENT},
 };
 
+static const uint32_t kelvin_signed_format_mask_gl_map[66] = {
+    [NV097_SET_TEXTURE_FORMAT_COLOR_SZ_R6G5B5] =
+        NV_PGRAPH_TEXFILTER0_RSIGNED | NV_PGRAPH_TEXFILTER0_GSIGNED | NV_PGRAPH_TEXFILTER0_BSIGNED,
```
coldhex, line 335: R might usually be unsigned. See also the texture data conversion done in, which should change depending on if signs are set or not:
https://github.com/xemu-project/xemu/blob/a00820746f7a6e3322a9ee2c471138d3791c2f32/hw/xbox/nv2a/pgraph/texture.c#L380-L397

hw/xbox/nv2a/pgraph/gl/surface.c
```diff
@@ -280,11 +281,15 @@ static void render_surface_to_texture_slow(NV2AState *d,
...
                                            TextureShape *texture_shape,
-                                           int texture_unit)
+                                           int texture_unit,
+                                           uint32_t filter)
 {
     PGRAPHState *pg = &d->pgraph;
 
-    const ColorFormatInfo *f = &kelvin_color_format_gl_map[texture_shape->color_format];
+    ColorFormatInfo f = kelvin_color_format_gl_map[texture_shape->color_format];
+    if (filter & kelvin_signed_format_mask_gl_map[texture_shape->color_format])
```
coldhex: Doesn't this only test if one of the sign bits match, not all?

coldhex:
I also considered a solution similar to this PR, but then went with implementing a general solution in my dev branch. While this PR should improve water effect in e.g. Halo 2, I have no idea if it might introduce artifacts in other games. To be safe, even if B8G8 or R8B8 texture are used, all sign bits R, G, B, A should be set if SNORM texture is used (because for e.g. B8G8 texture, hardware uses data from B and G channels also for R and A channels, but the sign setting is independent for each channel.) However, requiring all channels RGBA to have the sign bits wouldn't fix Halo 2 water, because Halo 2 doesn't set them all, and therefore is really not an option.

(And as far as my dev branch solution is concerced, it works fine, but I am still considering implementing signed texture handling in a different way. At some point I'm going to try a solution that does bilinear (or trilinear in case of mip-mapping) texture filtering manually in the fragment shader. This would avoid the need to adjust texture channel data sign bits before uploading to GPU.)

hw/xbox/nv2a/pgraph/gl/surface.c
oltolm: Yes, I fixed it.

hw/xbox/nv2a/pgraph/gl/constants.h
oltolm: I don't understand what this code does, I can't do it.

oltolm:
I don't know if makes any game worse, but it fixes the water in Ninja Gaiden Black.

PR #1894, opened 2025-02-03
Add 720x480 resolution to display settings

DCFUKSURMOM:
This small change adds the 720x480 resolution to the display settings. This isn't a huge deal for games but it is supported display mode for the original Xbox, which is currently missing from the emulator. This also fixes an annoying mismatch between Xbox Linux running at 720x480 and Xemu only having 640x480, which causes both the aspect ratio to be wrong and the mouse to be unaligned. I have to admit that the Xbox Linux thing is the whole reason I did this in the first place to be honest.
Screenshot of the added option in the settings:
image
Screenshot of the now properly aligned mouse cursor:
image
Screenshot of the misaligned cursor at 640x480:
image

### DIFF
--- a/ui/xui/main-menu.cc
+++ b/ui/xui/main-menu.cc
@@ -525,6 +525,7 @@ void MainMenuDisplayView::Draw()
     if (ChevronCombo("Window size", &g_config.display.window.startup_size,
                      "Last Used\0"
                      "640x480\0"
+                     "720x480\0"
                      "1280x720\0"
                      "1280x800\0"
                      "1280x960\0"
--- a/config_spec.yml
+++ b/config_spec.yml
@@ -151,7 +151,7 @@ display:
     fullscreen_exclusive: bool
     startup_size:
       type: enum
-      values: [last_used, 640x480, 1280x720, 1280x800, 1280x960, 1920x1080, 2560x1440, 2560x1600, 2560x1920, 3840x2160]
+      values: [last_used, 640x480, 720x480, 1280x720, 1280x800, 1280x960, 1920x1080, 2560x1440, 2560x1600, 2560x1920, 3840x2160]
       default: 1280x960
     last_width:
       type: integer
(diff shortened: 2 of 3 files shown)


ui/xui/main-menu.cc
```diff
@@ -525,6 +525,7 @@ void MainMenuDisplayView::Draw()
     if (ChevronCombo("Window size", &g_config.display.window.startup_size,
                      "Last Used\0"
                      "640x480\0"
+                     "720x480\0"
```
mborgerson, line 528: You'll also need to update https://github.com/xemu-project/xemu/blob/ac781ea8d1dcd7e67de62ea7ecaca60566e5e138/config_spec.yml#L154

mborgerson (changes requested):
See comment

ui/xui/main-menu.cc
DCFUKSURMOM: Ok I'll get to that some time later today, thanks for pointing it out

DCFUKSURMOM: This completely slipped my mind, I've had a crazy day (real life drama, crazy neighbors) and I just remembered as I was laying down. I'll try to get it done tomorrow (it's currently 10:33PM here)

mborgerson: No rush, thanks

DCFUKSURMOM: Well I accidentally did it as another pr (was trying to add it to this one but I'm half asleep still), but I edited the file as requested.

DCFUKSURMOM: Ok that time it actually added the commit to this pr like I meant to do the first time, I still don't know why it didn't do it the first time. GitHub being GitHub I guess. The file has been edited and added to this PR as requested.

DCFUKSURMOM: Sorry about the delayed responses, I've had a crazy week and on top of that it decided to not notify me about some of the comments at all.

mborgerson (approved):

mborgerson:
Thanks!

Issue #1773, opened 2024-09-06
Doom 3 audio


AlfredoDore:
### Bug Description

In Doom 3 there is an audio buzzing problem where there is a buzzing sound, even overlapping the speeches

### Expected Behavior

hum free audio

### xemu Version

07132

### System Information

Windows 11
xemu_version: 0.7.132
xemu_branch: master
xemu_commit: 8707d2aa2626063cb67b5ea20382584a0848dce7
xemu_date: Wed Aug 21 05:38:37 UTC 2024
xemu_settings_get_base_path: base path: C:\Users\alfre\AppData\Roaming\xemu\xemu\
xemu_settings_get_path: config path: C:\Users\alfre\AppData\Roaming\xemu\xemu\xemu.toml
CPU: AMD Ryzen 7 2700X Eight-Core Processor         
OS_Version:  23H2
GL_VENDOR: ATI Technologies Inc.
GL_RENDERER: Radeon RX 580 Series
GL_VERSION: 4.0.0 Core Profile Context 24.7.1.240618
GL_SHADING_LANGUAGE_VERSION: 4.60
audio: Device ac97: audiodev default parameter is deprecated, please specify audiodev=snd0

### Additional Context

xemu_version: 0.7.132
xemu_branch: master
xemu_commit: 8707d2aa2626063cb67b5ea20382584a0848dce7
xemu_date: Wed Aug 21 05:38:37 UTC 2024
xemu_settings_get_base_path: base path: C:\Users\alfre\AppData\Roaming\xemu\xemu\
xemu_settings_get_path: config path: C:\Users\alfre\AppData\Roaming\xemu\xemu\xemu.toml
CPU: AMD Ryzen 7 2700X Eight-Core Processor         
OS_Version:  23H2
GL_VENDOR: ATI Technologies Inc.
GL_RENDERER: Radeon RX 580 Series
GL_VERSION: 4.0.0 Core Profile Context 24.7.1.240618
GL_SHADING_LANGUAGE_VERSION: 4.60
audio: Device ac97: audiodev default parameter is deprecated, please specify audiodev=snd0

Triticum0:
@AlfredoDore Please retest as couldn't reproduce

Triticum0:
Tested the game for around 2 hour and only heard poping noise didn't hear any buzzing or overlapping speech

PR #1907, opened 2025-02-15
mcpx: Implement APU multipass voice processing

coldhex:
Voice Processor (VP) multipass feature allows configuring lists of voices that are first mixed (in order) into a designated mixbin which is then used as a sample source when processing voices with multipass flag set to true in NV_PAVS_VOICE_CFG_FMT. Setting correct voice order in lists is the responsibility of the game/application and in practice is handled by the DirectSound library. The multipass mixbin is hardcoded to 31 in DirectSound, but hardware would allow other bins.

This implementation also adds additional info to audio debug UI to see what the source and destination voices involved are. The info is only shown when DSP processing is off, i.e. "VP Only" (MON_VP) is selected. This is because storing the voice numbers requires additional digging which is required for MON_VP anyway and therefore is free. The multipass feature itself works fine with DSP (i.e. GP and EP) enabled, only the additional debug info is not shown.

The main motivation was to fix Doom 3 audio which uses multipass voice processing almost exclusively. Xemu 0.8.16 is very noisy and positional audio doesn't work. Here are videos to compare:

Xemu 0.8.16:

https://github.com/user-attachments/assets/749cd124-c62e-48d9-834a-5af0d8c27a28

Xbox hardware:

https://github.com/user-attachments/assets/38ff254c-7641-41ff-825b-62ff04191142

PR:

https://github.com/user-attachments/assets/f13a8244-0ff2-4d3f-b912-bda783449cca

### DIFF
--- a/hw/xbox/mcpx/apu.c
+++ b/hw/xbox/mcpx/apu.c
@@ -218,7 +218,7 @@ static int voice_resample(MCPXAPUState *d, uint16_t v, float samples[][2],
 static void voice_reset_filters(MCPXAPUState *d, uint16_t v);
 static void voice_process(MCPXAPUState *d,
                           float mixbins[NUM_MIXBINS][NUM_SAMPLES_PER_FRAME],
-                          uint16_t v);
+                          uint16_t v, int voice_list);
 static int voice_get_samples(MCPXAPUState *d, uint32_t v, float samples[][2],
                              int num_samples_requested);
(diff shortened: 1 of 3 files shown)


coldhex:
Issue #1773 is probably related.

socialenigma-og:
Positional audio works in:

BLiNX: The Time Sweeper;
Breakdown;
Half-Life 2.

mborgerson:
Thanks!

PR #1915, opened 2025-02-21
compat.cc: Revise Compatibility Rating System

### LINKED ISSUE #176: [QOL Improvement] Compat System Revision
Taking a look at xemu's current compat rating system, a large sum of the titles land in the "Playable Category" which isn't incredibly telling when it comes to how "Playable" a title is, I'd like to revise the project's current rating system to have an "Almost Perfect" or similar naming scheme for a segment of playable, that isn't "Perfect" but is "damn near close". I've played around 100 games on the emulator at this point and I can definitely say there are titles that are very very very close to perfect, with rather minor graphical or sound glitches that with our current rating system would land in playable, however it performs amazingly.

- Incredible Hulk: Ultimate Destruction is such a title, I have more that aren't coming to mind, however this game runs amazingly, only lands in playable because of skybox flickering, and a slight graphical bug in the pause menu, cutscenes. however it runs at around 0-1 mspf, and with an uncapped framerate will run at near 400 fps on my gt710 system on w11 21H2.

I don't mind to do all the Backend PRs required to implement this.

MasonT8198:
Depends on https://github.com/xemu-project/xemu-website/pull/177 , fixes https://github.com/xemu-project/xemu-website/issues/176

### DIFF
--- a/ui/xui/compat.cc
+++ b/ui/xui/compat.cc
@@ -57,14 +57,16 @@ void CompatibilityReporter::Draw()
         "Intro",
         "Starts",
         "Playable",
+        "Almost Perfect",
         "Perfect",
     };
 
     const char *playability_descriptions[] = {
         "This title crashes very soon after launching, or displays nothing at all.",
         "This title displays an intro sequence, but fails to make it to gameplay.",
-        "This title starts, but may crash or have significant issues.",
-        "This title is playable, but may have minor issues.",
+        "This title starts, but may crash, have significant issues, or be unplayable.",
+        "This title runs but has noticeable issues, graphical, audio, or performance issues that may affect gameplay.",
+        "This title is playable with very minor graphical, audio, or performance issues.",
         "This title is playable from start to finish with no noticable issues."
     };
 


ui/xui/compat.cc
```diff
@@ -57,14 +57,16 @@ void CompatibilityReporter::Draw()
...
+        "Almost Perfect",
         "Perfect",
     };
 
     const char *playability_descriptions[] = {
         "This title crashes very soon after launching, or displays nothing at all.",
         "This title displays an intro sequence, but fails to make it to gameplay.",
-        "This title starts, but may crash or have significant issues.",
-        "This title is playable, but may have minor issues.",
+        "This title starts, but may crash, have significant issues, or be unplayable.",
```
abaire, line 67: Rather than "starts", maybe "makes it past any intro sequences"? 

I can imagine that the difference between "intro" and "starts" could be subtle for folks who don't use English as their primary language. I wonder if there's much value in differentiating between "it shows a movie but you can't play" and "it gets to the main menu but you can't play"?

ui/xui/compat.cc
```diff
@@ -57,14 +57,16 @@ void CompatibilityReporter::Draw()
...
+        "This title runs but has noticeable issues, graphical, audio, or performance issues that may affect gameplay.",
```
abaire, line 68: nit, I think you can remove the first "issues," and just say "noticeable graphical, audio, or performance issues".

I wonder if "significant" might be better than "noticeable", since the new "Almost Perfect" would also mean that at least one issue was noticed.

ui/xui/compat.cc
```diff
@@ -57,14 +57,16 @@ void CompatibilityReporter::Draw()
...
+        "This title is playable with very minor graphical, audio, or performance issues.",
         "This title is playable from start to finish with no noticable issues."
```
abaire, line 70: Should we try to clarify that "start to finish" means that the game can be fully completed with no issues? I.e., if you choose to finish after 10 minutes, you should not submit a "Perfect" report.

MasonT8198: I agree, this was just a baseline of the general idea I had, comments/criticisms are encouraged!

I do think that perfect, and possibly (almost perfect) do need a higher barrier of playtime to be classified as that. Almost perfect being maybe not 100% completion but more than 10-15 minutes of game time. Pretty hard to quantify how playable a game is in that span of time.

ui/xui/compat.cc
MasonT8198: I've honestly never really been super partial to "intro" and "starts" as they seem pretty synonymous with one another in general.

I'd possibly like to possibly rework those... but I will take your comments under advisement

ui/xui/compat.cc
MasonT8198: Yeah, didn't notice the doubling of words there. Will fix.

MasonT8198: Also, yes. Makes sense on the second part, "Almost Perfect" would delineate very minor issues, like flickering elements such as a skybox, or just a slight audio crackle, or whatnot that honestly doesn't impede gameplay or performance, just is a minor annoyance or deviation from how the game plays on hardware.

MasonT8198:
I think before doing something concrete I'd like to get Matt's perspective, it may honestly be worth revisiting intro/starts like mentioned previously due to how similar they are rating wise.

It'd be nice if we considered possibly finding a way to store asserts to report in compat reports... but given previous discussions today I could just see forgoing this for just sending reports out and possibly webhook creating github issues for asserts with titles that have those issues. But that's just me spitballing

ui/xui/compat.cc
Calinou: "In-game" is commonly used for other emulators' compatibility ratings, and is the equivalent to "Starts" for Xemu. It should be understood as "It goes in-game, but is not particularly playable due to performance issues or bugs that prevent progression".

MasonT8198:
I'm going to close this as I don't see the compat system being changed any time soon, however maybe something to consider in the future as less games are listed as broken

PR #1960, opened 2025-03-06
meson: Trim unreferenced functions, data

mborgerson:
Trim ~4MiB

- [ ] Trim more
- [ ] Fix build
   - [ ] macOS *pass the -dead_strip option to ld. You should also pass the -gfull option to GCC* [ref](https://developer.apple.com/library/archive/documentation/Performance/Conceptual/CodeFootprint/Articles/CompilerOptions.html)
- [ ] Compare to LTO

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -353,6 +353,10 @@ qemu_common_flags = [
 qemu_cflags = []
 qemu_ldflags = []
 
+# Trim unreferenced functions, data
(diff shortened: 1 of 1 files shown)


mborgerson:
Doesn't appear to have significant savings compared (or in addition to) to LTO. Closed for now.

PR #1966, opened 2025-03-08
meson: fix system glslang detection

abouvier:
Continuation of https://github.com/xemu-project/xemu/pull/1866


mborgerson:
Thank you!

Issue #1322, opened 2023-01-07
Sign release executables


mborgerson:
### Feature Request

Begin signing release executables.

- [x] macOS (#1972)
- [ ] Windows
- [ ] Linux AppImage

### Alternatives

Do nothing

PR #1972, opened 2025-03-09
ci: Start signing macOS releases

mborgerson:
#1322


eduard-co:
Would be a great thing.

PR #1998, opened 2025-03-12
nv2a: Fix screen clip coordinate calculation when reusing surfaces with different clip bounds

### LINKED ISSUE #1986: Halo 2: Broken split screen
### Title

https://xemu.app/titles/4d530064/#Halo-2

### Bug Description

Player 2 screen has missing geometry.

Image

### Expected Behavior

Should appear similar to Player 1's screen.

### xemu Version

v0.8.33

### System Information

```
CPU:          
OS Platform:  Linux
OS Version:   Ubuntu 24.04.1 LTS
GPU Model:    NVIDIA GeForce RTX 4090/PCIe/SSE2
Driver:       4.0.0 NVIDIA 535.183.01
```

### Additional Context

_No response_

lostromb:
Factor in the difference between surface and clip size when calculating clip coords.

This should be considered a hack. I have no justification for this change, and no hardware tests to support it. But it does fix splitscreen rendering in Halo 2 and Phantom Dust without regressing any other game I've tested.

Master branch:
image

This PR:
image

!fixes #1986

### DIFF
--- a/hw/xbox/nv2a/pgraph/gl/draw.c
+++ b/hw/xbox/nv2a/pgraph/gl/draw.c
@@ -332,10 +332,8 @@ void pgraph_gl_draw_begin(NV2AState *d)
 
     /* Surface clip */
     /* FIXME: Consider moving to PSH w/ window clip */
-    unsigned int xmin = (pg->surface_shape.clip_x - pg->surface_binding_dim.clip_x) +
-                        (pg->surface_binding_dim.width - pg->surface_binding_dim.clip_width),
-                 ymin = (pg->surface_shape.clip_y - pg->surface_binding_dim.clip_y) +
-                        (pg->surface_binding_dim.height - pg->surface_binding_dim.clip_height);
+    unsigned int xmin = pg->surface_shape.clip_x - pg->surface_binding_dim.clip_x,
+                 ymin = pg->surface_shape.clip_y - pg->surface_binding_dim.clip_y;
     unsigned int xmax = xmin + pg->surface_shape.clip_width - 1,
                  ymax = ymin + pg->surface_shape.clip_height - 1;
 
--- a/hw/xbox/nv2a/pgraph/gl/draw.c
+++ b/hw/xbox/nv2a/pgraph/gl/draw.c
@@ -332,8 +332,8 @@ void pgraph_gl_draw_begin(NV2AState *d)
 
     /* Surface clip */
     /* FIXME: Consider moving to PSH w/ window clip */
-    unsigned int xmin = pg->surface_shape.clip_x - pg->surface_binding_dim.clip_x,
-                 ymin = pg->surface_shape.clip_y - pg->surface_binding_dim.clip_y;
+    unsigned int xmin = pg->surface_shape.clip_x,
+                 ymin = pg->surface_shape.clip_y;
     unsigned int xmax = xmin + pg->surface_shape.clip_width - 1,
                  ymax = ymin + pg->surface_shape.clip_height - 1;
 
--- a/hw/xbox/nv2a/pgraph/gl/draw.c
+++ b/hw/xbox/nv2a/pgraph/gl/draw.c
@@ -332,8 +332,10 @@ void pgraph_gl_draw_begin(NV2AState *d)
 
     /* Surface clip */
     /* FIXME: Consider moving to PSH w/ window clip */
-    unsigned int xmin = pg->surface_shape.clip_x - pg->surface_binding_dim.clip_x,
-                 ymin = pg->surface_shape.clip_y - pg->surface_binding_dim.clip_y;
+    unsigned int xmin = (pg->surface_shape.clip_x - pg->surface_binding_dim.clip_x) +
+                        (pg->surface_binding_dim.width - pg->surface_binding_dim.clip_width),
+                 ymin = (pg->surface_shape.clip_y - pg->surface_binding_dim.clip_y) +
+                        (pg->surface_binding_dim.height - pg->surface_binding_dim.clip_height);
     unsigned int xmax = xmin + pg->surface_shape.clip_width - 1,
                  ymax = ymin + pg->surface_shape.clip_height - 1;
 
--- a/hw/xbox/nv2a/pgraph/gl/surface.c
+++ b/hw/xbox/nv2a/pgraph/gl/surface.c
@@ -1152,7 +1152,6 @@ static void update_surface_part(NV2AState *d, bool upload, bool color)
                 pg->surface_binding_dim.height = found->height;
                 pg->surface_binding_dim.clip_y = found->shape.clip_y;
                 pg->surface_binding_dim.clip_height = found->shape.clip_height;
-
                 found->upload_pending |= mem_dirty;
                 pg->surface_zeta.buffer_dirty |= color;
                 should_create = false;
--- a/hw/xbox/nv2a/pgraph/gl/surface.c
+++ b/hw/xbox/nv2a/pgraph/gl/surface.c
@@ -1148,9 +1148,13 @@ static void update_surface_part(NV2AState *d, bool upload, bool color)
                 /* FIXME: Refactor */
                 pg->surface_binding_dim.width = found->width;
                 pg->surface_binding_dim.clip_x = found->shape.clip_x - found->width + found->shape.clip_width;
+                if (pg->surface_binding_dim.clip_x < 0)
+                    pg->surface_binding_dim.clip_x = 0;
                 pg->surface_binding_dim.clip_width = found->shape.clip_width;
                 pg->surface_binding_dim.height = found->height;
                 pg->surface_binding_dim.clip_y = found->shape.clip_y - found->height + found->shape.clip_height;
+                if (pg->surface_binding_dim.clip_y < 0)
+                    pg->surface_binding_dim.clip_y = 0;
                 pg->surface_binding_dim.clip_height = found->shape.clip_height;
                 found->upload_pending |= mem_dirty;
                 pg->surface_zeta.buffer_dirty |= color;
@@ -1187,9 +1191,13 @@ static void update_surface_part(NV2AState *d, bool upload, bool color)
             /* FIXME: Refactor */
             pg->surface_binding_dim.width = entry.width;
             pg->surface_binding_dim.clip_x = entry.shape.clip_x - entry.width + entry.shape.clip_width;
+            if (pg->surface_binding_dim.clip_x < 0)
+                pg->surface_binding_dim.clip_x = 0;
             pg->surface_binding_dim.clip_width = entry.shape.clip_width;
             pg->surface_binding_dim.height = entry.height;
             pg->surface_binding_dim.clip_y = entry.shape.clip_y - entry.height + entry.shape.clip_height;
+            if (pg->surface_binding_dim.clip_y < 0)
+                pg->surface_binding_dim.clip_y = 0;
             pg->surface_binding_dim.clip_height = entry.shape.clip_height;
 
             if (color && r->zeta_binding && (r->zeta_binding->width != entry.width || r->zeta_binding->height != entry.height)) {
--- a/hw/xbox/nv2a/pgraph/gl/surface.c
+++ b/hw/xbox/nv2a/pgraph/gl/surface.c
@@ -1147,10 +1147,10 @@ static void update_surface_part(NV2AState *d, bool upload, bool color)
             if (is_compatible) {
                 /* FIXME: Refactor */
                 pg->surface_binding_dim.width = found->width;
-                pg->surface_binding_dim.clip_x = found->shape.clip_x;
+                pg->surface_binding_dim.clip_x = found->shape.clip_x - found->width + found->shape.clip_width;
                 pg->surface_binding_dim.clip_width = found->shape.clip_width;
                 pg->surface_binding_dim.height = found->height;
-                pg->surface_binding_dim.clip_y = found->shape.clip_y;
+                pg->surface_binding_dim.clip_y = found->shape.clip_y - found->height + found->shape.clip_height;
                 pg->surface_binding_dim.clip_height = found->shape.clip_height;
                 found->upload_pending |= mem_dirty;
                 pg->surface_zeta.buffer_dirty |= color;
@@ -1186,10 +1186,10 @@ static void update_surface_part(NV2AState *d, bool upload, bool color)
 
             /* FIXME: Refactor */
             pg->surface_binding_dim.width = entry.width;
-            pg->surface_binding_dim.clip_x = entry.shape.clip_x;
+            pg->surface_binding_dim.clip_x = entry.shape.clip_x - entry.width + entry.shape.clip_width;
             pg->surface_binding_dim.clip_width = entry.shape.clip_width;
             pg->surface_binding_dim.height = entry.height;
-            pg->surface_binding_dim.clip_y = entry.shape.clip_y;
+            pg->surface_binding_dim.clip_y = entry.shape.clip_y - entry.height + entry.shape.clip_height;
             pg->surface_binding_dim.clip_height = entry.shape.clip_height;
 
             if (color && r->zeta_binding && (r->zeta_binding->width != entry.width || r->zeta_binding->height != entry.height)) {
--- a/hw/xbox/nv2a/pgraph/gl/surface.c
+++ b/hw/xbox/nv2a/pgraph/gl/surface.c
@@ -1147,15 +1147,12 @@ static void update_surface_part(NV2AState *d, bool upload, bool color)
             if (is_compatible) {
                 /* FIXME: Refactor */
                 pg->surface_binding_dim.width = found->width;
-                pg->surface_binding_dim.clip_x = found->shape.clip_x - found->width + found->shape.clip_width;
-                if (pg->surface_binding_dim.clip_x < 0)
-                    pg->surface_binding_dim.clip_x = 0;
+                pg->surface_binding_dim.clip_x = found->shape.clip_x;
                 pg->surface_binding_dim.clip_width = found->shape.clip_width;
                 pg->surface_binding_dim.height = found->height;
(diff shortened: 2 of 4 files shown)


MasonT8198:
Can you run this against abaires test cases to see if this breaks anything? I feel like it probably will

lostromb:
The only hardware I have to test with is an 8-year-old Quadro P400. I get a segfault about 10 seconds into the suite. Also the first time I ever saw the Xemu codebase was this morning so I really don't know any of this.

mborgerson:
Thanks for the debugging effort and PR!

If you want to help build some surface tests (or check for existing ones, maybe some exist since last I checked) it will help build lasting support for the PR and reduce risk of future regression. Once upon a time, Halo 2 split screen was functional and regressed to the current state. Having those tests might have prevented it.

I'm aware some tests cause xemu to crash, we will fix them in due time. But until then, you can disable those tests that are crashing on xemu master branch.

abaire:
> I'm aware some tests cause xemu to crash, we will fix them in due time. But until then, you can disable those tests that are crashing on xemu master branch.

Or you can try out https://github.com/abaire/xemu-dev_pgraph_test_results which will proceed past the crashes and generate a page with all of the diffs caused by this PR. Instructions are in the README. 

So far I think I'm the only one who has used it, so there may be rough edges to work out; feel free to let me know if you run into trouble.

lostromb:
Thank you for that. It looks like there's still a separate problem with depth buffer loss of precision (behaves worse on AMD?) that's breaking the top screen in 2-player mode.... I thought this was caused by my change but apparently not, it's because the second lighting pass uses `glDepthTest(GL_EQUAL)`
image.

I'll try and dig into that next and see what's the difference between the draw calls in this configuration.

I was able to get Win64 crossbuild working so I can test on a better graphics card. I can run individual pgraph tests now, but I don't actually see any coverage for scissor testing / clip rectangles in particular? There's cases for stencil buffers and viewports, but I don't think either of those apply directly to the changes I'm making.

abaire:
> ... I don't actually see any coverage for scissor testing / clip rectangles in particular? There's cases for stencil buffers and viewports, but I don't think either of those apply directly to the changes I'm making.

Yeah, the test coverage is nowhere exhaustive. Most of these tests are just related to some specific bug/feature I (or a contributor) was working on at the time. As Matt suggested above, it'd be excellent if you could add tests that reproduce the issue you're fixing along with this PR. You can ping me here or on Discord if you run into issues; I've recently started adding some documentation for the test infra but it's still sparse.

lostromb:
This PR also fixes Phantom Dust splitscreen, which might be notable as that's the only other splitscreen game I've found that uses clip coordinates that are not just (screen size divided by 2).

Note how each player's view spills over to the left a bit on master, this is corrected in the PR.
image

lostromb:
Other games are unremarkable, no regressions I've been able to find.

4x4 EVO
image

Aggressive Inline
image

Halo CE
image

Halo 2
image

MX Superfly
image

Project Gotham
image

mborgerson:
Do you plan to create test cases for this change?

lostromb:
I would like to, of course. Haven't been able to build the pgraph test project yet. The cases themselves should be very straightforward - basically copies of the viewport tests except setting clip registers instead of viewport bounds. The game tests yesterday confirmed that this might actually be a more legitimate fix than the hack I first though it was. Still just kinda poking around with stuff and learning the tools.

I think the trickiest part of setting up the pgraph test is replicating the discrepancy between clip size and render target size that cause the issue in the first place. I believe Phantom Dust may be the best scenario to create a test case from just because of its unique splitscreen arrangement.

hw/xbox/nv2a/pgraph/gl/draw.c
```diff
@@ -332,8 +332,10 @@ void pgraph_gl_draw_begin(NV2AState *d)
 
     /* Surface clip */
     /* FIXME: Consider moving to PSH w/ window clip */
-    unsigned int xmin = pg->surface_shape.clip_x - pg->surface_binding_dim.clip_x,
-                 ymin = pg->surface_shape.clip_y - pg->surface_binding_dim.clip_y;
+    unsigned int xmin = (pg->surface_shape.clip_x - pg->surface_binding_dim.clip_x) +
```
mborgerson: I think this should simplify to `xmin = pg->surface_shape.clip_x;`, since clip offset is included in surface dims now. Similarly for ymin.

https://github.com/xemu-project/xemu/blob/fac9ae83c7f673f5991ab03217226f0b97b3578b/hw/xbox/nv2a/pgraph/gl/surface.c#L1046-L1047

mborgerson:
> I would like to, of course.

Great, thanks! I'll await the surface tests cases you are creating before merging the change then. In the meantime I'll move this PR into draft state.

lostromb:
Turns out there's already a little bit of coverage - the final two test cases under Surface Clip. I didn't notice the few red pixels on the side in these specific cases.

Master
image

PR
image

MasonT8198:
> Turns out there's already a little bit of coverage - the final two test cases under Surface Clip. I didn't notice the few red pixels on the side in these specific cases.
> 
> Master image
> 
> PR image

Abaire has a repository with golden results, compare to that just to see what the difference is

lostromb:
> > I'm aware some tests cause xemu to crash, we will fix them in due time. But until then, you can disable those tests that are crashing on xemu master branch.
> 
> Or you can try out https://github.com/abaire/xemu-dev_pgraph_test_results which will proceed past the crashes and generate a page with all of the diffs caused by this PR. Instructions are in the README.
> 
> So far I think I'm the only one who has used it, so there may be rough edges to work out; feel free to let me know if you run into trouble.

I don't know enough python to figure this out, unfortunately...
```
./.venv/bin/pip -r requirements.txt
ERROR: Could not find a version that satisfies the requirement nxdk-pgraph-test-runner>=0.1.12 (from versions: none)
ERROR: No matching distribution found for nxdk-pgraph-test-runner>=0.1.12
```

> 
> Abaire has a repository with golden results, compare to that just to see what the difference is

It's.... [inconclusive???](https://github.com/abaire/nxdk_pgraph_tests_golden_results/blob/main/results/Surface_clip/x16y8_w512h384.png?raw=true)
x16y8_w512h384

abaire:
What version of Python are you using (and what platform are you on)?


The test being inconclusive just means a better test probably to be written, I don't recall what I was working on when I wrote those tests, but I probably wasn't doing a very deep dive on the behavior.

According to https://abaire.github.io/xemu-nxdk_pgraph_tests_results/compare/xemu-0.8.34-master-a143f66ce48ef37ccb15bbdd4daed3c6df2eaee9/Darwin_arm64/gl_Apple_Apple_M3_Max%3Agslv_4.10/Surface_clip.html there are some existing differences on master (Darwin/OpenGL) vs Xbox so maybe the test is good enough but could be made more obvious?

lostromb:
> According to https://abaire.github.io/xemu-nxdk_pgraph_tests_results/compare/xemu-0.8.34-master-a143f66ce48ef37ccb15bbdd4daed3c6df2eaee9/Darwin_arm64/gl_Apple_Apple_M3_Max%3Agslv_4.10/Surface_clip.html there are some existing differences on master (Darwin/OpenGL) vs Xbox so maybe the test is good enough but could be made more obvious?

Yeah, I think if you just added one more case with like `x=160,y=80,w=300,h=200` then the difference would become super obvious.
The Halo 2 behavior should be `x=0,y=240,w=640,h=240` (and for 4-way, `x=320,y=240,w=320,h=240`) so you could add those too

mborgerson:
> Yeah, I think if you just added one more case with like `x=160,y=80,w=300,h=200` then the difference would become super obvious.
> The Halo 2 behavior should be `x=0,y=240,w=640,h=240` (and for 4-way, `x=320,y=240,w=320,h=240`) so you could add those too

@lostromb You can add the necessary test cases and send a PR with them to https://github.com/abaire/nxdk_pgraph_tests

abaire:
> @lostromb You can add the necessary test cases and send a PR with them to https://github.com/abaire/nxdk_pgraph_tests

Normally I would've said the same thing, but I need to add the new doc comments to those tests anyway so I've already incorporated the additional regions (and added tests for 32bpp surfaces). I'm waiting on a HW regression, but you can hopefully test the ISO from  https://github.com/abaire/nxdk_pgraph_tests/pull/200 (or wait an hour or so until I merge the change)

MasonT8198:
Abaire do you mind grabbing the golden on your devkit and posting it? My
xboxen are in storage and I have yet to go grab them (as in I forgot where
I put the key to my storage building on my back lawn)

lostromb:
> > @lostromb You can add the necessary test cases and send a PR with them to https://github.com/abaire/nxdk_pgraph_tests
> 
> Normally I would've said the same thing, but I need to add the new doc comments to those tests anyway so I've already incorporated the additional regions (and added tests for 32bpp surfaces). I'm waiting on a HW regression, but you can hopefully test the ISO from [abaire/nxdk_pgraph_tests#200](https://github.com/abaire/nxdk_pgraph_tests/pull/200) (or wait an hour or so until I merge the change)

Thank you much for helping out 👍. The difference in test outputs should be very obvious now with the new cases. I eagerly await the hardware parity results.

Master branch
image

PR
image

In the meantime I tested every other splitscreen game I could find and couldn't tell any differences in any of them.

abaire:
The results are already up https://abaire.github.io/nxdk_pgraph_tests_golden_results/results/Surface_clip/index.html and the PR output does not look correct to me. 

It'd be good to figure out why the dev test script is failing for you so you can run a full regression. What Python version are you using and what OS?

lostromb:
Okay, the new pgraph tests are great because they confirm my hunch that this PR's "fix" is still kind of a hack that's actually just canceling out a miscalculation which happened earlier. The fact that the clip boundaries change based on the surface format was the main tip-off. With format RGB565 the clip gets messed up, but with RGBA8888 it works as expected.

The root cause, I believe, is in surface.c when we're trying to look around to see if we can reuse an existing surface rather than recreating it. When we find a match, inherit the properties from the found match, including clip bounds, and reuse that surface instead of recreating it. The problem arises when we've got our own set of clip bounds that don't match with the clip bounds set in the found surface.

I assume based on #2005 that the ideal scenario would be to allow surface reuse even for when clip bounds don't match rather than always recreate the surface, so I think it's just a matter of correctly reconciling those bounds.

> It'd be good to figure out why the dev test script is failing for you so you can run a full regression. What Python version are you using and what OS?

Let's discuss at https://github.com/abaire/xemu-dev_pgraph_test_results/issues/9

lostromb:
I think the fixes are in the correct place now, games look fine, running the pgraph clip tests manually all look fine to me.
Batch test results PR is here https://github.com/abaire/xemu-dev_pgraph_test_results/pull/10

abaire:
> I think the fixes are in the correct place now, games look fine, running the pgraph clip tests manually all look fine to me. Batch test results PR is here [abaire/xemu-dev_pgraph_test_results#10](https://github.com/abaire/xemu-dev_pgraph_test_results/pull/10)

Unfortunately the GH action has some issues running on PRs not authored by the repo owner so it didn't generate the comparison page. I think if you create the PR against your own repo's main branch it'll run there and you can provide a link to the results page (it'll look something like https://abaire.github.io/xemu-dev_pgraph_test_results/feat_support_alpha_from_specular/index.html )

I updated the instructions on the dev repo to hopefully clarify for future users; sorry for the inconvenience!

lostromb:
> Unfortunately the GH action has some issues running on PRs not authored by the repo owner so it didn't generate the comparison page. I think if you create the PR against your own repo's main branch it'll run there and you can provide a link to the results page (it'll look something like https://abaire.github.io/xemu-dev_pgraph_test_results/feat_support_alpha_from_specular/index.html )

???? https://github.com/lostromb/xemu-dev_pgraph_test_results/pull/1

In other news, this change feels incredible
https://www.youtube.com/watch?v=NrDpNfDCsuM

abaire:
It looks like it didn't run the actions for some reason? https://github.com/lostromb/xemu-dev_pgraph_test_results/actions shows nothing.

You may also need to enable Pages in the settings for the repo, I'm not sure if that gets cloned in templates or not. I assume the settings page will be at https://github.com/lostromb/xemu-dev_pgraph_test_results/settings/pages The source should be set to "GitHub Actions". (Please let me know if you do so I can add that to the instructions as well)

[edit] You may also need to mess with the Actions settings: https://github.com/lostromb/xemu-dev_pgraph_test_results/settings/actions 
 + Allow all actions and reusable workflows
 + Read repository contents and packages permissions

lostromb:
Okay, so I had to
- Manually enable workflows in the forked repository
- Manually setup Github Pages to use workflows as you described
- Manually create the wiki repo by creating a dummy page
- Must have a branch called "master", so just "main" won't work (Possibly?)

With all that, the github pages [still don't deploy](https://github.com/lostromb/xemu-dev_pgraph_test_results/actions/runs/13869285118/job/38813989917?pr=3), but I did get [a wiki page](https://github.com/lostromb/xemu-dev_pgraph_test_results/wiki/screen_clip_fix) that is half broken. The comparison output still seems to hardcode the repository path as "abaire/whateverreponame" so it's trying to pull the results from the PR and never finding them.

For now it seems like it would be easier to pull [my results branch](https://github.com/lostromb/xemu-dev_pgraph_test_results/tree/screen_clip_fix) into your repository and just run the actions from there.

abaire:
Thanks for testing!

I assume it's not literally "abaire/whateverreponame" and depending on what "whateverreponame" actually is, it may be correct. The golden images are all accessed in other repos under my account.

I removed wiki generation entirely as it's not very useful anyway. I also updated the instructions to try to capture all the first time setup steps and added some debug logging to try to track down why the site didn't generate.

You're welcome to grab the updates and retry (probably just need the [`build.yml`](https://github.com/abaire/xemu-dev_pgraph_test_results/blob/main/.github/workflows/build.yml). I'll see if I can do some testing with another account to work out the remaining kinks.

[Edit] I added instructions to the README on the need for a one-time setup of the `github-pages` environment, otherwise I was able to create a page successfully in an alternate account. I suspect part of the issue is that the original instructions suggested forking the repo instead of creating a clone via the "Use this template" button; it may be worth deleting your current repo and starting again from the template (you can just copy over the "results" dir from your branch to save time re-running tests).

lostromb:
Okay, I also just don't have the hardware configuration to run an adequate test for this PR right now, unless I could get the test harness working on Windows. Or I could reformat another spare storage box to Linux and throw a GTX 1550 in there.... This might take a while.

abaire:
> Okay, I also just don't have the hardware configuration to run an adequate test for this PR right now, unless I could get the test harness working on Windows. Or I could reformat another spare storage box to Linux and throw a GTX 1550 in there.... This might take a while.

In theory it should run on Windows (the underlying packages do, since that's what I use to gen the xemu golden results). I'm traveling for a bit and won't have access to a Windows machine but happy to provide suggestions if you run into problems there. In my case I have Python3.11 installed and run the test suite from cygwin.

That said, you've already generated the diffs, so all of the actual processing (the comparisons vs goldens, site generation) should be happening via the GitHub actions, so I think all you need to get the diffs working is git. 
If you're still working on the xemu PR you should be able to use the pre-built pgraph test ISO without having to spend time getting that building. You can also create a JSON config file to prevent autorun and skip all the tests besides the "Surface clip" and throw it on the E: to speed up your debug cycle (instructions in the repo, you may be the first person besides me to try them so let me know if you try and hit trouble so I can improve them).

lostromb:
Okay, well I fixed up the wiki page well enough that you can at least view the diffs in a somewhat readable way. [Here are the surface clip results.](https://github.com/lostromb/xemu-dev_pgraph_test_results/wiki/screen_clip_fix#surface_clip). But I would assume all of my other differences are a result of doing the test on an old Quadro that can't even run the Vulkan renderer.

**In all surface clip tests the PR matches hardware**, except for what looks like a separate issue where the clipping interacts with the test name overlay itself (maybe xemu doesn't apply clipping rects to blittable areas or something)

### PR vs HW
image

MasonT8198:
Weird that Swath Width is invisible via PR

abaire:
> **In all surface clip tests the PR matches hardware**, except for what looks like a separate issue where the clipping interacts with the test name overlay itself (maybe xemu doesn't apply clipping rects to blittable areas or something)

Yeah, confirmed that this is an issue at master. I'm adding a test specifically for this case and fixing the text rendering in the rest of the surface clip tests.

> Weird that Swath Width is invisible via PR

Looks like it's not just Swath Width. It'd be good to check master on the same hardware to verify that this is an issue unrelated to the PR.

lostromb:
[Got a proper test result using a newer Nvidia card on Linux](https://lostromb.github.io/pgraph_test_results_2/site/screen_clip_test_gtx1660/index.html), but it doesn't appear any better than the last one. Don't know why swath width, or even basic cases like `Quads-inlinebuf-ps` don't render anything, but it's consistent on different test machines as well (Windows AMD, Linux Nvidia...)

abaire:
Just to clarify, are you testing across machines with an unchanged build at master, or just with the changes in this branch?

lostromb:
> Just to clarify, are you testing across machines with an unchanged build at master, or just with the changes in this branch?

I'm moving this branch's dev build to different machines and manually looking at individual test results since I couldn't get the automated script to run in Cygwin.

I added an ugly-looking patch to the code to avoid negative clips, these are the diffs now. Outside of surface clip tests, there are 11 deltas vs. HW golden
https://lostromb.github.io/pgraph_test_results_2/site/screen_clip_test_7dbbef1/

hw/xbox/nv2a/pgraph/gl/draw.c
lostromb: I should have listened to this in the first place because it's exactly correct

lostromb:
Okay the correct fix is actually a lot simpler than I was making it out to be. 4 lines changed. @mborgerson this should be ready for review.

[Pgraph test results](https://lostromb.github.io/pgraph_test_results_2/site/screen_clip_test_4df526b/)

mborgerson:
Thanks!

PR #2007, opened 2025-03-14
build.yml: Omit scripts path on push

MasonT8198:
As far as I'm aware we shouldn't be generating a new rel on any push in regards to modifying a script.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,7 @@ on:
       - '.github/**'
       - '!.github/workflows/build.yml'
       - 'README.md'
+      - 'scripts/**'
       - 'ubuntu-win64-cross/**'
   pull_request:
     paths-ignore:


mborgerson:
We do depend on some of the scripts for release assets, eg https://github.com/xemu-project/xemu/blob/master/scripts/gen-license.py

MasonT8198:
> We do depend on some of the scripts for release assets, eg https://github.com/xemu-project/xemu/blob/master/scripts/gen-license.py

Do we need to generate a new release when that script is modified? Or would it be safe to just wait for another PR merge do gen a release, afaik most of the time a script is edited, we do not need to do another release, it could just wait until the next update to the application

mborgerson:
This is just one example. We have other scripts that manage macOS library downloads, for example. Neither script nor code changes strictly necessitate a release

MasonT8198:
I see what you're saying, closing

PR #2011, opened 2025-03-14
build.yml: Cache Meson Dependencies

MasonT8198:


.github/workflows/build.yml
```diff
@@ -23,6 +23,11 @@ jobs:
       uses: actions/checkout@v4
       with:
         fetch-depth: 0
+    - name: Cache Meson dependencies
+      uses: actions/cache@v4
+      with:
+        path: /path/to/meson/cache
```
mborgerson, line 29: * What Meson dependencies are you trying to cache?
* Doesn't look like a valid path

MasonT8198:
lol I left my example path in when I was trying to find ways to improve
speed on the CI

mborgerson:
You might shave ~10 seconds by caching subprojects, but cache has a cost of ~5 seconds or so. Apart from that I'm not sure what's worth caching for meson, so I'll close this PR. You're welcome to play around with it and open another PR if you achieve some worthwhile savings

MasonT8198:
@mborgerson I think in the end a self-hosted option would be nice to have, saving random breakages due to github updating things quasi-randomly with MacOS, and an Ubuntu self hosted machine

Rather easy to do, just 2-3 vms at the end of the day, given how paltry the amount of cores and memory github gives to their runners are.

2013 Mac Pro Trashcans are good to run off of, or just something else that's rather basic.

I've seen a 80% reduction in comp time on my hardware running all 3, just as an alternative, because once you set it up you can nix most of the comp steps in the build.yml, just commenting them out for use later.

I think it is at least worth considering. I'm not sure how much more opt can be done with the current setup.

PR #2012, opened 2025-03-14
Migrate xemu to SDL3

MasonT8198:
This is a draft, and nowhere near done, or the most current version I have of what I've been working on

Imgui/Implot will need reconfigured to support SDL3 with xemu via the xemu-project subproject repos

Audio subsystem requires a complete rewrite to be in line with SDL3... unfortunately.

and many, many other things


MasonT8198:
I've put a lot of effort into getting this to properly compile, SDL3 has deprecated **a lot of functions** _and nerfed a crapload more_, If someone else wants to finish this up, I welcome it, or if Matt decides to, it gets about 80% of the way done with comp, and gets through almost all of the SDL related segments, I just feel like I'm going around in circles trying to nail out these last few bugs.

MacOS builds to roughly the same extent as Ubuntu, it never works via actions for me, even with the latest master sync.

Windows requires the cross comp toolchain to be updated to support SDL3 using the same build instructions that I used, Most of the fullscreen subsystem has to be written, and a lot of the SDL Color Formats no longer support the amount of args that we are feeding into them, scancodes have been completely changed and are very convoluted, and honestly I'm surprised I got as far as I got!

Sorry I couldn't get further.

MasonT8198:
I'm going to move to close this for now, If I get further I may choose to bring this back, feel free to continue on what I've done to finish this up.

PR #2026, opened 2025-03-17
nv2a/gl/vk: Add an anisotropic filtering option to the graphics settings menu

### LINKED ISSUE #409: Anisotropic filtering option
### Feature Request
**Anisotropic filtering** option would be very useful as it can greatly improve how games look and it'd be nicer if you could use it from xemu itself rather than having to force it from GPU control panels 

Example of forced AF:
No AF
image

16x AF
image

Calinou:
This is available in both Vulkan and OpenGL renderers. Note that the Vulkan renderer requires a restart for changes to be effective, while changes in OpenGL are applied immediately.

- This closes https://github.com/xemu-project/xemu/issues/409.

## Preview

RalliSport Challenge 2 at 1x resolution scale (focus on the road surface):

Disabled | 16× AF
-|-
Image | Image

## TODO

- [ ] Make changes in Vulkan effective without requiring a restart (or switching renderers back and forth).
  - I'm not sure how to proceed here, since I'm not familiar with xemu's codebase.
- [ ] Check if the `reload` functions are really needed and merge them with the setter if there's no real need to have them.

### DIFF
--- a/config_spec.yml
+++ b/config_spec.yml
@@ -146,6 +146,9 @@ display:
     surface_scale:
       type: integer
       default: 1
+    anisotropic_filter_level:
+      type: integer
+      default: 1
   window:
     fullscreen_on_startup: bool
     fullscreen_exclusive: bool
--- a/hw/xbox/nv2a/nv2a.h
+++ b/hw/xbox/nv2a/nv2a.h
@@ -27,6 +27,8 @@ int nv2a_get_framebuffer_surface(void);
 void nv2a_release_framebuffer_surface(void);
 void nv2a_set_surface_scale_factor(unsigned int scale);
 unsigned int nv2a_get_surface_scale_factor(void);
+void nv2a_set_anisotropic_filter_level(unsigned int level_po2);
+unsigned int nv2a_get_anisotropic_filter_level(void);
 const uint8_t *nv2a_get_dac_palette(void);
 int nv2a_get_screen_off(void);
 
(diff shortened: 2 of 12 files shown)


MattPonton:
I don't know if 16x is required if it adds a performance hit. I haven't experienced any, but just saying it as a word of caution. It seems 8x does the job in emulating original hardware on my testings. Perhaps we can add it to the nvidia profile over making a unique settings option for it?

Dead or Alive 3 as an example

**OG Hardware, captured through Xbox HDMI through an XRG1 device to OBS**:
Screenshot_2025-07-03_23-28-50

XEMU screenshot Nvidia Profile Aniso **Application Controlled**:
xemu-2025-07-03-23-36-54

XEMU screenshot Nvidia Profile Aniso **2x**:
xemu-2025-07-03-23-37-48

XEMU screenshot Nvidia Profile Aniso **4x**:
xemu-2025-07-03-23-39-22

XEMU screenshot Nvidia Profile Aniso **8x**:
xemu-2025-07-03-23-40-35

XEMU screenshot Nvidia Profile Aniso **16x**:
xemu-2025-07-03-23-41-37


As far as I can tell, 8x gives the same visual clarity as 16x, so I don't think we would need to set the profile to max even though I have no performance difference.

Calinou:
> Perhaps we can add it to the nvidia profile over making a unique settings option for it?

The NVIDIA profile won't work on AMD/Intel GPUs, nor on Linux or macOS. (NVIDIA application profiles do exist on Linux, but NVAPI integration is Windows-only in most projects I know.)

> As far as I can tell, 8x gives the same visual clarity as 16x, so I don't think we would need to set the profile to max even though I have no performance difference.

I think 8x is a better default for modern games, which are likely to be memory bandwidth-bound on integrated GPUs. However, this is very unlikely in emulation where the bottleneck is typically on the CPU instead. Therefore, you should basically always be able to use 16x with no noticeable performance loss.

Issue #1887, opened 2025-01-26
Feature Request: Fast Forward / Turbo (Unlimited FPS)


ehw:
### Feature Request

It would be fun if Xemu supported a way to fast forward or basically output all the frames as fast as the host machine can output. This is a good way to benchmark performance in games in which the game is already running at the full internal framerate of a game, but you want to see how fast a machine can handle a given scenario at its absolute limit. It can also provide a way to fast forward games so you can get to certain scenarios faster, like skipping boring unskippable cutscenes and maybe even for debugging. I personally use this for testing the performance impact of various settings, benchmarking new hardware, and for getting through some boring sections of games quicker.

### Alternatives

Maybe an actual frame limiter? I don't really understand the use cases for having a frame limiter, but I know various emulators have always supported one throughout history. The only area I personally have used a frame limiter is to ensure the output framerate is evenly divided into the refresh rate to prevent micro stuttering. But this isn't a problem with a variable refresh rate monitor as far as I know...

### Additional Context

RPCS3 and Ryujinx are sometimes used to benchmark new CPU releases. Since emulators are traditionally mostly CPU bound, emulation efficiency can be best measured if the frame limit were uncapped. It would be neat to use Xemu to benchmark games using different hardware configurations, drivers, software settings, etc especially over time. :) 

Maybe one day we'll see Xemu on a benchmarking site like TechPowerUp?

https://www.techpowerup.com/review/amd-ryzen-7-9800x3d/8.html

PR #2028, opened 2025-03-17
UX: Add CPU overclock and display rate override settings

Calinou:
- Rebase of https://github.com/xemu-project/xemu/pull/1467, with an additional feature added (display rate override).

These settings can be used to improve framerate or set a new framerate limit (different from the typical 30/60 FPS).

Many games work well with this, but using overrides will break games that are not designed around FPS-independent physics. Therefore, submitting compatibility reports is disabled if either of the overrides are active.

Thanks to @Spidy123222 for the original code :slightly_smiling_face: I've kept the commit author intact and added myself as a co-author instead.

## Preview

Screenshot_20250317_041241

RalliSport Challenge 2 was my main motivation for adding the display rate override.[^1] With the CPU overclock from https://github.com/xemu-project/xemu/pull/1467, this game will speed up or slow down (audio playback is intact). With the display rate override, correct game speed is preserved, the actual presented frame rate is increased and audio playback remains intact. The in-game timer on the HUD only updates at 60 FPS, but the actual game world updates at the presented framerate.

[^1]: As a bonus, the display rate override also works on the Xbox intro logo.

> [!NOTE]
>
> This is a 120 FPS video, so make sure to watch it on a 120 Hz+ display. Alternatively, play the video at 50% speed by right-clicking it and choosing a speed option.

https://github.com/user-attachments/assets/103ff1a6-ffc4-4969-b6b6-ab714512643c

In a sense, the behavior of this setting is quite similar to what https://github.com/dolphin-emu/dolphin/pull/11486 does for Dolphin. It changes the display timings but does not affect the rest of the console.

### Additional notes on CPU clock override

While working on this PR, I attempted using very high CPU clock overrides which leads to an interesting behavior, at least on RalliSport Challenge 2:

https://github.com/user-attachments/assets/8d3ecb98-5f8e-43ed-a490-d6b8d65896c6

Notice how game time increases faster and faster as the clock speed increases, but it will go back to normal speed at around 650% of the original CPU clock speed. However, if you keep wiggling the slider back and forth around that value, game time will advance much faster than usual. Figuring this out could be interesting to implement https://github.com/xemu-project/xemu/issues/1887, so that we can make the CPU run as fast as possible while disregarding accurate game speed (for fast-forwarding).

### DIFF
--- a/ui/xui/main-menu.cc
+++ b/ui/xui/main-menu.cc
@@ -66,6 +66,30 @@ void MainMenuGeneralView::Draw()
     Toggle("Cache shaders to disk", &g_config.perf.cache_shaders,
            "Reduce stutter in games by caching previously generated shaders");
 
+    SectionTitle("Tweaks");
+    Toggle("Emulated CPU clock override", &g_config.perf.override_clockspeed,
+           "Override default CPU clock speed (can break games)");
+
+    char cpu_clock_buf[32];
+    snprintf(cpu_clock_buf, sizeof(cpu_clock_buf), "Clock speed: %d%% (%.2f MHz)", (int)(g_config.perf.cpu_clockspeed_scale * 100), (733333333 * g_config.perf.cpu_clockspeed_scale) / 1000000);
+    Slider("Virtual CPU clock", &g_config.perf.cpu_clockspeed_scale, cpu_clock_buf, 0.25f, 4.f, 0.01f);
+
+    if (fabs(g_config.perf.cpu_clockspeed_scale - 1.f) <= 0.0099f) {
+        g_config.perf.cpu_clockspeed_scale = 1;
+    }
+
+    Toggle("Display rate override", &g_config.perf.override_display_rate,
+        "Override default presentation frame rate (can break games)");
+
+    char display_rate_buf[35];
+    snprintf(display_rate_buf, sizeof(display_rate_buf), "Display rate: %d%% (%d / %d FPS)", (int)(g_config.perf.display_rate_scale * 100), (int)(30 * g_config.perf.display_rate_scale), (int)(60 * g_config.perf.display_rate_scale));
+    // Set slider increment so that 60 FPS games can be adjusted at a per-FPS level in terms of precision
+    Slider("Display rate", &g_config.perf.display_rate_scale, display_rate_buf, 0.3333333f, 4.f, 0.0166667f);
+
+    if (fabs(g_config.perf.display_rate_scale - 1.f) <= 0.0099f) {
+        g_config.perf.display_rate_scale = 1;
+    }
+
     SectionTitle("Miscellaneous");
     Toggle("Skip startup animation", &g_config.general.skip_boot_anim,
            "Skip the full Xbox boot animation sequence");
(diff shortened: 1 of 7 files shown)


ui/xui/main-menu.cc
```diff
@@ -66,6 +66,30 @@ void MainMenuGeneralView::Draw()
...
+    Toggle("Emulated CPU clock override", &g_config.perf.override_clockspeed,
+           "Override default CPU clock speed (can break games)");
+
+    char cpu_clock_buf[32];
+    snprintf(cpu_clock_buf, sizeof(cpu_clock_buf), "Clock speed: %d%% (%.2f MHz)", (int)(g_config.perf.cpu_clockspeed_scale * 100), (733333333 * g_config.perf.cpu_clockspeed_scale) / 1000000);
+    Slider("Virtual CPU clock", &g_config.perf.cpu_clockspeed_scale, cpu_clock_buf, 0.25f, 4.f, 0.01f);
+
+    if (fabs(g_config.perf.cpu_clockspeed_scale - 1.f) <= 0.0099f) {
+        g_config.perf.cpu_clockspeed_scale = 1;
+    }
```
Calinou, line 79: Regarding this (and the other `fabs()` below), see conversation in https://github.com/xemu-project/xemu/pull/1467#discussion_r1269421234.

Personally, I'm in favor of removing both and increasing the step intervals so that it's easier to choose a value with the mouse. This would reduce flexibility in the values you can choose, but I don't think there's a lot of use cases for CPU clock increments below 5% anyway.

For the display rate override, the setting could be changed to store an integer value that defaults to `60` (so `120` would be equivalent to `2.0` currently). This value can then be used to calculate the actual sleep interval. This means we'd no longer be dealing with floats in the UI for the display rate override.

FZXO11:
The game looks so smooth this is amazing.

Gareth17andMinnie14:
Perfect work there guys, it looks like the emulator might be able to stable itself with a CPU & GPU together like if you were using an actual computer to play games that has a cpu and gpu which seems very clever

blackhand1001:
Can someone approve the workflow for this Pull request so I can do some testing. It don't have my build environment configured at the moment.

blackhand1001:
Ok I was able build this from source. It seems like the display override is working but I don't think the CPU overclock is working. Even at 25% I am not getting any slowdown and at higher speeds no FPS improvements in games that are cpu limited.

blackhand1001:
@Calinou Can you take a look at this. Maybe theres a part of the CPU code that wasn't added to your commit.

Calinou:
> @Calinou Can you take a look at this. Maybe theres a part of the CPU code that wasn't added to your commit.

Which games did you test this with? I've compared my PR with #1467 and the CPU-side changes are virtually the same. (The only difference is that the other PR only multiplies the integer clock speed with a float if the override is enabled, which may be a good thing to incorporate in my PR to avoid any precision issues.)

blackhand1001:
I tried Tony hawks Pro Skater 2x, 007 Nightfire, 007 Agent Under Fire etc. The Car levels on Nightfire have framerate drops at the stock clock speed.

Spidy123222:
@Calinou Thanks for the PR change. I kind of just stopped after trying because I wasn't getting much input on what needs changed at end of my commits. It's my first C project attempt also.

I would like to note the original code does also change up how sliders work so it's more customizable. If you updated to recent Xemu and if they add new sliders. It's going to also have to be updated. 

I have recently proposed maybe having a hacks section to help differentiate to add more hacky features than normal features like this or possibly to do patches for the future.

from what I am gathering from your description above. Changing frame rate will be more like vi overclock. Wouldn’t it be appropriate to do a similar name?

After some testing. it does decrease performance in general. with or without the toggles enabled adding a stutter in Burnout 3: Takedown and Conkers live and reloaded.

With changes:  

https://github.com/user-attachments/assets/db440e5f-44ed-4a36-af6a-7fe0e30f28bd

Current Stable:

https://github.com/user-attachments/assets/144bf0b5-569b-4107-8040-5b33abd2ce8d


This is consistent in the whole game. stock speed and modified. Using a m2 pro MacBook Pro on latest macOS beta.

Spidy123222:
> Ok I was able build this from source. It seems like the display override is working but I don't think the CPU overclock is working. Even at 25% I am not getting any slowdown and at higher speeds no FPS improvements in games that are cpu limited.

This might be related to the display changes. Conkers live and reloaded has issues currently on it but not on Stable nor my original PR.

https://github.com/user-attachments/assets/3498c85c-6fe6-4c2c-8094-42ccd207dea9

changing clock changes the animation speed so speed issues im not sure what's changed for that.

PGHJA991id12:
Hi. May somebody please post a compiled build of this?

Triticum0:
@Calinou could you rebase this

Calinou:
> @Calinou could you rebase this

I started looking into rebasing, but https://github.com/xemu-project/xemu/pull/2691 has changed code modified by this PR, so it's going to be nontrivial. I can't look into it right now, maybe this weekend.

PR #2050, opened 2025-03-21
vl: Always pass dvd_path to the driver and handle errors

### LINKED ISSUE #1964: Improve dvd_drive path checking
### Feature Request

vl.c enforces that the dvd_path exists locally

https://github.com/xemu-project/xemu/blob/1d662cb49ed6c6b8c0b5546ddb7d4160a459e470/system/vl.c#L2975

But now that we are shipping releases with libcurl, dvd_path can now also be a URL. Unfortunately, any connection/server/path issues will terminate xemu immediately. So we need to pass the path to the driver and let it fail gracefully with an error message.

### Alternatives

_No response_

### Additional Context

_No response_

wilkovatch:
Closes #1964.

Side note: loading images via URLs works on Linux, but not on Windows, where it fails with the following messages:
```
fd=1672 is not a socket, AIO implementation is missing
fd=1672 is not a socket, AIO implementation is missing
curl: Connection timed out after 5002 milliseconds
warning: Could not read image for determining its format: Input/output error
```
(EDIT: looks like an upstream issue, the same happens on QEMU)
(EDIT 2: It's caused by [this](https://github.com/xemu-project/xemu/commit/f5fd677ae7cf7cfb07b12adbfd479c460ddc3ac5) and the next 3 commits, a partial revert (for the non-socket case only) solves the issue, but that's outside the scope of this PR)

### DIFF
--- a/system/vl.c
+++ b/system/vl.c
@@ -651,8 +651,20 @@ static int cleanup_add_fd(void *opaque, QemuOpts *opts, Error **errp)
 static int drive_init_func(void *opaque, QemuOpts *opts, Error **errp)
 {
     BlockInterfaceType *block_default_type = opaque;
+    const char *file_path = qemu_opt_get(opts, "file");
+    bool is_cdrom_with_file = !strcmp(qemu_opt_get(opts, "media"), "cdrom") && strlen(file_path) > 0;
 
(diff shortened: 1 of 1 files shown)


mborgerson:
Thanks!

Issue #784, opened 2022-03-17
Ninja gaiden black particle effects glitch


ko81e24wy:
### Title

https://xemu.app/titles/5443000d/#Ninja-Gaiden-Black

### Bug Description

particle effects is thiner than actual xbox,and it depends on res-scale level
![5555555](https://user-images.githubusercontent.com/38323456/158748636-ab816596-c74a-4a2c-88ee-2a507be6e4b4.jpg)
![444444](https://user-images.githubusercontent.com/38323456/158748663-cf5ad407-dd42-4c0a-89ca-50bbf0844373.jpg)
![66666666](https://user-images.githubusercontent.com/38323456/158768496-9f9e12bc-d5ad-4d19-8689-f177d9f2739a.jpg)

### Expected Behavior

expect to display correctly

### xemu Version

Version: 0.6.2-82-ghttps://github.com/mborgerson/xemu/commit/2ff5f23235387a296fc9a5f11943b4dd29bd2837235
Branch: master
Commit: 2ff5f23
Date: Thu Mar 17 04:24:58 UTC 2022

### System Information

OS:Windows 11
CPU:Intel I9 9900K
GPU:NVIDIA GTX 1080
GPU Driver:NVIDIA 515.36

mborgerson:
According to https://github.com/xemu-project/xemu/pull/2079#issuecomment-2771248754 this was fixed on master sometime prior to v0.8.46.

Issue #701, opened 2022-02-02
nv2a 0x97 -> 0x03b8 (3D_SEPARATE_SPECULAR_ENABLE) unhandled


dracc:
### Title

https://xemu.app/titles/4d53006e
https://xemu.app/titles/4d5300d1
https://xemu.app/titles/5553005e

### Bug Description

According to the debug output, 0x97 -> 0x03B8 is unhandled.
In the [NV20 nouveau sources](https://gitlab.freedesktop.org/mesa/mesa/-/blob/184a690fca700aa9ef2ae2e58b0f4bd86562aa6d/src/mesa/drivers/dri/nouveau/nv20_3d.xml.h#L747), this instruction is `3D_SEPARATE_SPECULAR_ENABLE`, and the values it can take matches what is received in the function calls.

### Expected Behavior

The NV2A instruction 0x03B8 should be handled.

### xemu Version

0.6.2-55-gcaf84884c4

### System Information

N/A

Fabxx:
Adding this title: https://xemu.app/titles/5553005e/#Tom-Clancy-s-Splinter-Cell-Double-Agent

`nv2a_pgraph_method_unhandled 0: 0x97 -> 0x03b8 0x1`

Triticum0:
Adding this title: https://xemu.app/titles/53430007/#Rolling

`nv2a_pgraph_method_unhandled 0 : 0x97 -> 0x03b8 0x0`

Triticum0:
Add this title : https://xemu.app/titles/534300fd/#Lara-Croft-Tomb-Raider-Legend
`nv2a_pgraph_method_unhandled 0 : 0x97 -> 0x03b8 0x0`

abaire:
This one looks like it's already known as `NV097_SET_SPECULAR_ENABLE`, I already had it in my test suite likely from observing https://github.com/fgsfdsfgs/pbgl/blob/06e70a368bc107c229c701e476e4198da928f0b8/src/push.h#L16 or https://github.com/mborgerson/xsm64/blame/b0ca582f0640ddd400d0503a282913fa2c3900b3/src/pc/gfx/gfx_xbox.c#L34

abaire:
- Added tests in https://github.com/abaire/nxdk_pgraph_tests/pull/178 (e.g., https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/tests/lighting_control_tests.cpp) 
[...]

PR #2079, opened 2025-03-31
Specular lighting fixes

### LINKED ISSUE #697: nv2a 0x97 -> 0x0294 (glLightModel?) unhandled
### Title
https://xemu.app/titles/5553005e

### Bug Description

According to the debug output, 0x97 -> 0x0294 is unhandled.
In the [NV20 nouveau sources](https://gitlab.freedesktop.org/mesa/mesa/-/blob/184a690fca700aa9ef2ae2e58b0f4bd86562aa6d/src/mesa/drivers/dri/nouveau/nv20_3d.xml.h#L751), this instruction is `3D_LIGHT_MODEL`, and the values it can take matches what is received in the function calls.

### Expected Behavior

The NV2A instruction 0x0294 should be handled.

### xemu Version

0.6.2-55-gcaf84884c4

### System Information

N/A

### Additional Context

_No response_

### LINKED ISSUE #783: Ninja gaiden black specular and bloom effects of skin and armor are missing
### Title

https://xemu.app/titles/5443000d/#Ninja-Gaiden-Black

### Bug Description

specular and bloom effects of skin and armor are missing

### Expected Behavior

expect to display correctly

### xemu Version

Version: 0.6.2-82-g2ff5f23235
Branch:  master
Commit:  2ff5f23235387a296fc9a5f11943b4dd29bd2837
Date:    Thu Mar 17 04:24:58 UTC 2022

### System Information

OS:Windows 11
CPU:Intel I9 9900K
GPU:NVIDIA GTX 1080
GPU Driver:NVIDIA 515.36

### Additional Context

_No response_

### LINKED ISSUE #838: Sonic riders no shine/specular
### Title

https://xemu.app/titles/53450087/#Sonic-Riders

### Bug Description

shine on character models not present
https://cdn.discordapp.com/attachments/680488810429153315/958175321658646528/unknown.png
https://cdn.discordapp.com/attachments/680488810429153315/958185880558899210/unknown.png

### Expected Behavior

192 168 1 139-image17
192 168 1 139-image20

### xemu Version

0.6.2 90

### System Information

OS windows 10
CPU i9 9900kf
GPU 1070

### Additional Context

_No response_

### LINKED ISSUE #1288: Dead or Alive 3: missing specular effects on character models
### Title

https://xemu.app/titles/4d53002d/#Dead-or-Alive-3

NTSC: TC-001

### Bug Description

Specular effects are missing on characters. See attached screenshot for example.

1. Launch game
2. Start any game mode
3. Select a characters
4. Expected specular effects are missing

xemu

### Expected Behavior

Specular effects should be visible, check skin and leather
other

### xemu Version

 0.7.71

### System Information

CPU: Intel(R) Core(TM) i5-10300H
OS: Windows 11
GPU: NVIDIA GeForce GTX 1650 Ti
Driver: 4.0.0 NVIDIA 511.79

### Additional Context

_No response_

abaire:
Various interdependent fixes for specular lighting.

The results generally don't exactly match HW but are much closer than the current implementation.

[Test results](https://abaire.github.io/xemu-dev_pgraph_test_results/feat_spec_params/index.html)

* Fixes #783 
* Fixes #838 
* Fixes #1288 
* Fixes #697
* Related #701


abaire:
@Triticum0 any chance you could spot check this against games with known lighting issues? I am hopeful that it'll improve highlights in games using the fixed function pipeline.

re4thewin:
this pr fixes [838](https://github.com/xemu-project/xemu/issues/838)

abaire:
> this pr fixes [838](https://github.com/xemu-project/xemu/issues/838)

Cool, thanks for testing!

Triticum0:
Not got at my computer for a while sorry. Will do some testing once I get back to my pc in few days

ko81e24wy:
Also fixes https://github.com/xemu-project/xemu/issues/784, hard to say its same as XBOX,but much better than before.

abaire:
Cool, thanks for testing!

The reconstruction of the specular power factor isn't perfect and there are still some issues with ambient/emissive handling in certain cases so I expect that there will be some delta vs HW with this fix.

mborgerson:
> Also fixes #784, hard to say its same as XBOX,but much better than before.

@ko81e24wy Can you confirm the particle effect scaling described in #784 is actually broken on master (v0.8.46) and not broken in this PR? I'd expect this PR not to have fixed that issue.

Also, are those screenshots from a real Xbox, or from Xbox 360?

ko81e24wy:
> > Also fixes #784, hard to say its same as XBOX,but much better than before.
> 
> @ko81e24wy Can you confirm the particle effect scaling described in #784 is actually broken on master (v0.8.46) and not broken in this PR? I'd expect this PR not to have fixed that issue.
> 
> Also, are those screenshots from a real Xbox, or from Xbox 360?

Its from XSX,I will do a comparison after work.

ko81e24wy:
MASTER 3X RES
MASTER3X
PR 3X RES
PR3X
MASTER 1X RES
MASTER1X
PR 1X RES
PR1X
@mborgerson ,I have test them，you are right,the particle effect was fixed in master.

mborgerson:
Thank you for the tests @ko81e24wy

ko81e24wy:
I test NGB and DOA,this PR has stutter in game when master is smooth.

hw/xbox/nv2a/pgraph/pgraph.c
```diff
@@ -1791,6 +1805,42 @@ DEF_METHOD_INC(NV097, SET_FOG_PLANE)
...
+    // It does not appear possible for a DirectX-generated value to be positive, so while this differs from hardware
+    // behavior, it may be irrelevant in practice.
+    float invert = 1.f;
+    if (c3 > 0.0f) {
+        invert = -1.f;
+        c3 *= invert;
+    }
+
+    c3 -= SPECULAR_POWER_CONSTANT_COEFFICIENT;
+    float ret = SPECULAR_POWER_LOG_CONSTANT / log2(c3 / SPECULAR_POWER_COEFFICIENT_A);
```
mborgerson: 1. Cross checked against [this implementation](https://github.com/dracc/xgu/blob/db3172d8c983629f0dc971092981846da22438ae/xgux.h#L279). The power estimated by this function appears to have very high error for `pow<1.0`. Can you confirm?
2. It would be nice to determine the actual specular power formula hardware uses based on these parameters, but we can save it for future work.
3. It's also possible to calculate the exact input value given the observed original lookup table, but this curve fit approach is probably fine if accuracy is within a few percent given (2) anyway.

hw/xbox/nv2a/pgraph/pgraph.c
```diff
@@ -1791,6 +1805,42 @@ DEF_METHOD_INC(NV097, SET_FOG_PLANE)
...
+    assert(!isnan(ret) && "Failed to reconstruct specular power factor");
+    return ret * invert;
+}
+
+DEF_METHOD_INC(NV097, SET_SPECULAR_PARAMS)
+{
+    int slot = (method - NV097_SET_SPECULAR_PARAMS) / 4;
+    NV2A_DPRINTF("NV097_SET_SPECULAR_PARAMS[%d] 0x%X\n", slot, parameter);
```
mborgerson: Unnecessary debug prints? We should see this with nv2a_pgraph_method traces

hw/xbox/nv2a/pgraph/glsl/vsh-ff.c
```diff
@@ -345,52 +357,83 @@ GLSL_DEFINE(materialEmissionColor, GLSL_LTCTXA(NV_IGRAPH_XF_LTCTXA_CM_COL) ".xyz
...
+                "    float pf;\n"
+                "    if (nDotVP == 0.0) {\n"
+                "      pf = 0.0;\n"
+                "    } else {\n"
+                "      pf = pow(nDotHV, %f);\n"
+                "    }\n"
+                "    vec3 lightAmbient = lightAmbientColor(%d) * attenuation;\n"
+                "    vec3 lightDiffuse = lightDiffuseColor(%d) * attenuation * nDotVP;\n"
+                "    vec3 lightSpecular = lightSpecularColor(%d) * attenuation * pf;\n",
+                state->specular_power, i, i, i);
```
mborgerson: 1. Hardcoding specular power value in shader source can cause creation of many extra shaders. I assume this is the cause of the stuttering mentioned [here](https://github.com/xemu-project/xemu/pull/2079#issuecomment-2772404844). We probably want to make it a uniform.
2. If not made to be a uniform, then it also needs to be added to the dirty register checklists

hw/xbox/nv2a/pgraph/pgraph.c
```diff
@@ -1791,6 +1805,42 @@ DEF_METHOD_INC(NV097, SET_FOG_PLANE)
...
+    if (slot == 3) {
+        pg->specular_power = reconstruct_specular_power_from_c3(parameter);
```
mborgerson: It would be nice to find where this is actually stored in device state

abaire: I assume it's not stored anywhere; presumably the 6 coefficients are stored since it wouldn't make sense to pass them only to have the HW reconstruct the original power value from DX/GL.

If someone reverses the actual implementation (the values look suspiciously similar to the ones used for `NV097_SET_LIGHT_SPOT_FALLOFF` which is also unimplemented) we could work out the registers with that?

mborgerson: > presumably the 6 coefficients are stored 

This is what I'm referring to

hw/xbox/nv2a/pgraph/pgraph.c
abaire: Interesting, given that it's disjoint in the LUT func I'm not surprised. It may not be possible to do a trivial curve fit, will look into it.

hw/xbox/nv2a/pgraph/pgraph.c
abaire: Unfortunately it doesn't seem to be a straightforward one. Unless the values are being transformed before storage (I'm asuming this is unlikely since they're floats rather than bitvectors) there doesn't seem to be any viable sets in the diff.

This is particularly annoying since it looks like the power can't be inferred from c3 alone; for vals < 1 c2 is pretty close on its own, but c2 rolls over around 1561.

I may just stash the values in `PGRAPHState` for now (since I need to accumulate them for curve fitting) and leave finding the correct locations to a future PR, if that's OK.

hw/xbox/nv2a/pgraph/pgraph.c
abaire: Oops, I wrote this on an old branch before they were removed from the other handlers. Removed.

hw/xbox/nv2a/pgraph/glsl/vsh-ff.c
abaire: Good call, changed to a uniform.

hw/xbox/nv2a/pgraph/pgraph.c
abaire: I ended up segmenting the curve and using a couple functions to keep percent error < 3. Things get wacky around power 2500 as params start moving in the opposite direction, but I would assume in practice power is typically < 1000 (and the same wackiness can be observed on HW).

The end results are still not quite correct; it looks like the localeye variants are quite close but the infinite ones are a bit overbright. I'm inclined to look into it in a followup since this PR adds a bunch of other functionality.

ko81e24wy:
I test the latest build of this PR,crash when loading in game.
log file
[xemu.zip](https://github.com/user-attachments/files/19681158/xemu.zip)

abaire:
@ko81e24wy oops! I was setting up the uniform in the infinite light handler and that game has 3 infinite lights. Should be fixed now.

hw/xbox/nv2a/pgraph/pgraph.h
```diff
@@ -197,6 +197,11 @@ typedef struct PGRAPHState {
     float light_local_position[NV2A_MAX_LIGHTS][3];
     float light_local_attenuation[NV2A_MAX_LIGHTS][3];
 
+    float specular_params[6];
```
abaire, line 200: @mborgerson should these be added to `VMStateDescription`? 

It might be useful to document the decision factor for adding pgraph state to snapshots (maybe with CI enforcement/reminders at some point, if it's easy to automate)

ko81e24wy:
> @ko81e24wy oops! I was setting up the uniform in the infinite light handler and that game has 3 infinite lights. Should be fixed now.

It works now.

hw/xbox/nv2a/pgraph/pgraph.c
abaire: Ended up fixing the overbrightness from the point light -it seems that localeye off does not modify the VP vector in the way I'd assumed from the GL spec.

ko81e24wy:
Performance is better,the cutsecne after boss fight stage1 in chapter 15 has no more frame drops like before.

hw/xbox/nv2a/pgraph/pgraph.h
mborgerson: This is one reason to prefer PGRAPHState.regs over new fields in PGRAPHState. If they are stored in regs, they are already being stored in VMStateDescription. If not, then they need to be added specifically to the VMStateDescription.

The decision factor to include it in VMStateDescription is basically this: if the data are part of device state that cannot be calculated, then they must be in VMStateDescription for faithful state loading. Imagine if the game sets specular params once on startup and never does so again. If a snapshot is saved, then loaded later in a new instance of xemu, the restored state won't have correct specular params.

hw/xbox/nv2a/pgraph/pgraph.c
mborgerson: > I may just stash the values in PGRAPHState for now (since I need to accumulate them for curve fitting) and leave finding the correct locations to a future PR, if that's OK.

No problem. We can iterate on it as we learn more. But as we discussed in your other comment, the new fields should be saved in the VMStateDescription

hw/xbox/nv2a/pgraph/pgraph.h
abaire: > This is one reason to prefer PGRAPHState.regs over new fields in PGRAPHState.

Agreed on preferring regs, in this case I'm just not seeing any change in an obvious way.

> ... if the data are part of device state that cannot be calculated, then they must be in VMStateDescription 

In that case I think there are a bunch of fields missing. E.g., `light_infinite_half_vector` and the other constants (likely more, I just happened to have an editor open looking at that line).

Would it make sense to either adopt a naming convention for vals that should (not?) be saved in vmstate or otherwise differentiate between recalculable and non fields? Then in theory we could have a check to make sure new fields update the state.

Or is the intent more nuanced since it's likely that many of the pgraph state fields will be set very frequently and it's acceptable to have a bad frame or two after a load?

hw/xbox/nv2a/pgraph/pgraph.c
mborgerson: On second thought, let's not bother saving the new field in the VMStateDescription in this PR. Other fields are not saved right now, and we'll have to come up with a better solution for them. Games likely refreshes them anyway, so it's not a huge issue.

hw/xbox/nv2a/pgraph/pgraph.h
mborgerson: Yeah, there are other fields missing. I mentioned in my [comment above](https://github.com/xemu-project/xemu/pull/2079#discussion_r2038643794) but I'll repeat here, I'm okay with deferring storing these new (and existing) fields VMStateDescription for now. Apologies for the misdirection.

hw/xbox/nv2a/pgraph/pgraph.c
```diff
@@ -1791,6 +1805,113 @@ DEF_METHOD_INC(NV097, SET_FOG_PLANE)
...
+
+  return reconstructed_power;
+}
+
+DEF_METHOD_INC(NV097, SET_SPECULAR_PARAMS)
+{
+    int slot = (method - NV097_SET_SPECULAR_PARAMS) / 4;
+    pg->specular_params[slot] = *(float *)&parameter;
+    if (slot == 5) {
+        pg->specular_power = reconstruct_specular_power(pg->specular_params);
```
mborgerson, line 1911: We have some work to do to make this more accurate, but I'm happy with iterating on it

mborgerson:
Thanks!

PR #2092, opened 2025-04-08
nv2a: Fix assert when setting fog gen mode to fog_x

abaire:
There is a mismatch between the values set into `CSV0_D` and the PGRAPH parameters leading to an assertion failure in the fixed function shader when selecting `NV097_SET_FOG_GEN_MODE_V_FOG_X` generation mode.

I verified that the `CSV0_D` value actually is 4 using abaire/nxdk_nv2a_register_finder, so the issue here is that the `VshFoggen` enumeration was incorrectly based on the `NV097_SET_FOG_GEN_MODE` parameters rather than the register vals.

Not sure if any games use `NV097_SET_FOG_GEN_MODE_V_FOG_X`, I was expanding the fog tests in the pgraph tester and ran into this.


mborgerson:
Thanks!

PR #2117, opened 2025-04-16
nv2a: Handle SET_FOG_COORD and SET_WEIGHT* commands

abaire:
I doubt it's used often in games, but realized that I use these in the pgraph tests and they were not being handled.

Improves many of the fog tests: https://abaire.github.io/xemu-dev_pgraph_test_results/handle_fog_and_weight_setters/index.html


hw/xbox/nv2a/nv2a_regs.h
```diff
@@ -1099,6 +1099,9 @@
 #   define NV097_SET_TEXCOORD3_4F                             0x00001620
 #   define NV097_SET_TEXCOORD3_2S                             0x00001610
 #   define NV097_SET_TEXCOORD3_4S                             0x00001630
+#   define NV097_SET_FOG_COORD                                0x00001698
+#   define NV097_SET_WEIGHT1F                                 0x0000169C
```
mborgerson, line 1103: How about *2F, *3F while we are at it?

abaire: I don't think they exist, or at least they're not in the nxdk reg list: https://github.com/abaire/nxdk/blob/8be82da51b5cc6fe3bf30e2fbb198c757d6b297c/lib/pbkit/nv_regs.h#L443

abaire: Ah, I see them in JFR's repo: https://github.com/JayFoxRox/xqemu-jfr/blob/db0ece37e860c12a7c0c7ccf4371bf308e8290fb/hw/xbox/nv2a_gpu_methods.h#L159

abaire: Added. Also added some pgraph tests to verify the behavior of non-specified components and corrected the implementation to match.

mborgerson:
Thanks!

PR #2118, opened 2025-04-16
build.sh: Add xemu_version fallback to macOS builds

abaire:
Fixes the macOS build action on forked repos (at least on mine)


mborgerson:
Thanks!

Issue #2175, opened 2025-05-14
HW rounding bug Breaks Tile rendering  Above 1x resolution 


Triticum0:
### Bug Description

Some game render water of terrain in squares tiles, issues with rounding offset makes it look like a grid and render  incorrectly.

### Games affected
Type:Water
[Kingdom Under Fire Hero's](https://xemu.app/titles/504c0006/) (#2162)
[The Elder Scrolls III: Morrowind: Game of the Year Edition](https://xemu.app/titles/42530005/#The-Elder-Scrolls-III-Morrowind-Game-of-the-Year-Edition) (#725)
[Splinter Cell](https://xemu.app/titles/5553000c/#Tom-Clancy-s-Splinter-Cell) (#951)
[Ty the Tasmanian Tiger
](https://xemu.app/titles/45410029/#TY-the-Tasmanian-Tiger) (#2334/#1981)

Types:Terrain
[MotorGP:Ultimate Racing Technology 3](https://xemu.app/titles/54510088/#MotoGP-Ultimate-Racing-Technology-3) (#1104)
MotorGP 2
[Secret Weapons over Normandy](https://xemu.app/titles/4c410006/#Secret-Weapons-over-Normandy) (#1115)

### Expected Behavior

All game should render correctly

### xemu Version

V0.8.58

### System Information

OS | Windows 10
CPU | AMD Ryzen 5 5700x Eight-Core Processoer
Graphics Device | NVIDIA GeForce RTX 3060 Ti/PCIe/SSE2
Graphics Driver | 4.0.0 NVIDIA 570.133.07
Render | Opengl

### Additional Context

All game should be fixed with coldhex pr implement screen coordinate rounding to 4 bit fractional precision other than MotorGP:Ultimate Racing Technology 3 that has other graphical issue with road texture

PR #2183, opened 2025-05-18
nv2a: Implement screen coordinate rounding to 4-bit fractional precision

coldhex:
This PR implements rounding of floating point screen coordinates x and y to 4-bit fractional precision. Rounding is towards zero (i.e. truncating). This is based on tests https://github.com/coldhex/nxdk_pgraph_tests/commit/96017220b1fb0be796176574df1e0c905f304494

(I suspect that NV2A rasterization has 4-bit sub-pixel precision and converts floating-point screen coordinates to a fixed-point representation.)

Previous work on this includes these PRs:
- https://github.com/xemu-project/xemu/pull/1732
- https://github.com/xemu-project/xemu/pull/1362
- https://github.com/xemu-project/xemu/pull/735

However, #1732 (at 1x) and #1362 don't match hardware (see TopLeftRaster results below). I didn't test #735, but quite likely would not match either.

To test and interpret vertex rounding results the triangle rasterization rule matters. NV2A rasterizes using the top-left triangle rasterization rule of Direct3D (which is a tie-breaking rule when pixel center falls precisely on a triangle edge.) On the other hand, both OpenGL and Vulkan leave the details of
such tie-breaking up to implementation, but GPUs generally just follow the same Direct3D top-left rule (but I haven't tested Apple silicon GPUs).

Considering OpenGL, while the top-left rule is what implementations follow when rendering to a framebuffer, things change when rendering to a framebuffer object (FBO). I have a gist demostrating this: https://gist.github.com/coldhex/114f251c536d8a362ef2cd2cf8ee0964 . When rendering to FBO the rule effectively changes to the bottom-left rule. I suspect this has something to do with bottom-left coordinate convention of OpenGL, forcing implementations to flip y-coordinates and render scenes upside-down to FBOs. This PR also contains an OpenGL-backend commit that flips y-coordinates thereby restoring the top-left rule. Vulkan-backend doesn't need a corresponding modification since Vulkan doesn't have the top-left/bottom-left discrepancy in the first place.

At 1x scale, this PR should fix many problems in games related to the 0.5 pixel offset problem caused by Direct3D (before version 10) coordinate convention. However, I'll list explicitly what this PR doesn't do:

1. Does not fix such problems for GPUs that don't follow the top-left rasterization rule (if there even are modern such GPUs.) It would be possible to add hacks to fix the situation in some cases. For example, if hardware truly followed e.g. the bottom-left rule in Vulkan, then subtracting a small implementation-dependent value from y-coordinates can change render result to look like top-left rule was followed. However, such a hack would also mean that interpolated values such as texture coordinates would be sampled slightly off pixel center and therefore would be slightly off how
NV2A samples. For OpenGL FBO rendering, instead of applying such a hack, I prefer the commit in this PR.

2. Does not fix such problems for scales higher than 1x. At 1x scale Xemu can emulate what the hardware does and the 0.5 pixel offset problem disappears as a result. However, at higher scales the correct rendering result depends on intentions of game developers. Direct3D adds 0.5 offset usually, but this depends on what game devs decided and e.g. Halo 2 doesn't apply the 0.5 offset always. For example, suppose 2x scaling was used and a game wants to render a rectangle with top-left and bottom-right coordinates (0.5, 0.5) and (256.5, 256.5). For 2x scale, consider these two choices: either (1.0, 1.0) and (513.0, 513.0) (which is what Xemu currently does), or (0.5, 0.5) and (512.5, 512.5). The first would be correct if the game did not apply the 0.5 offset and the rectangle was actually in motion and moving towards (0.0, 0.0) and we just happened to catch one frame of it in its path. The second would be correct if the game did apply the 0.5 offset and if the rectangle is supposed to be flush top-left of screen. So fixing the 0.5 offset problem generally requires some guess work. It's difficult to know if the game applied the 0.5 offset or not (because NV2A has programmable vertex shaders.) One possibility would be to add a UI toggle to enable subtracting an offset and let users experiment what looks right per-game (but a game may also be mixed in its use of the 0.5 offset.)

3. Does not fix NV2A floating-point arithmetic emulation. For example, floating-point multiplication result in NV2A vertex shader seems to round towards zero, while modern GPUs follow the IEEE-754 round-to-even rule. Also, NV2A floating point addition rounding is peculiar because the result of e.g. 16777000.0+(-1.0/32.0) rounds to 16776999.0, but adding anything even slightly larger than -1.0/32.0 rounds to 16777000.0. The reason I mention this is that the ProjAdjacentGeometry tests in nxdk_pgraph_tests are sensitive to floating-point rounding, especially ProjAdjacentGeometry_0.5625, the result of which even changed on Xbox hardware due to a math library change in nxdk_pgraph_tests commit [66b32a0b1feba32a0db7a95d6358e84f7a6246ad](https://github.com/abaire/nxdk_pgraph_tests/commit/66b32a0b1feba32a0db7a95d6358e84f7a6246ad). Differing results in ProjAdjacentGeometry tests are caused by inaccurate floating-point arithmetic emulation, not because of the 4-bit vertex rounding (which is what this PR implements.)

4. Does not fix multisampling anti-aliasing (MSAA) sampling points. The vertex rounding in this PR should be correct for MSAA too, but Xemu rasterization for MSAA uses supersampling (SSAA) and if similar test-cases as the TopLeftRaster for this PR was made for MSAA, the results wouldn't match NV2A. For MSAA 2x, NV2A seems to use pixel sample points (0.0, 0.0) and (0.5, 0.5), while Xemu effectively uses (0.25, 0.5) and (0.75, 0.5). For MSAA 4x, NV2A uses (0.0, 0.0), (0.5, 0.0), (0.0, 0.5)
and (0.5, 0.5), while Xemu effectively uses (0.25, 0.25), (0.75, 0.25), (0.25, 0.75) and (0.75, 0.75).

Here are the TopLeftRaster results (https://github.com/coldhex/nxdk_pgraph_tests/commit/96017220b1fb0be796176574df1e0c905f304494)
Tests were run with AMD Radeon RX 6600 Vulkan (but PR with OpenGL gives exactly the same result.)
top-left-raster

The match between HW and PR is almost pixel-perfect. The colors for grey (around 0x808080 have some 1-off RGB interpolation errors.

Truncation (instead of flooring) is somewhat difficult to see and needs to be inspected using an image editor. The top-most pixel row (line 0) contains three thin quads with slightly differing colors. In TopLeftRaster this is the code for these quads:

https://github.com/coldhex/nxdk_pgraph_tests/blob/96017220b1fb0be796176574df1e0c905f304494/src/tests/vertex_shader_rounding_tests.cpp#L1200-L1228

(Both Xbox and PR have interpolation inaccuracies that I did not expect for such a quad. Ideally the quad would have a constant color, but since quads are rendered using two triangles e.g. the right-most segment consists of both RGB 0x808080 and 0x7F7F7F.)

Here is Xemu 0.8.62 with OpenGL. Note how the bottom two horizontal lines swapped in thickness (due to FBO bottom-left rule) compared to the 0.8.62 with Vulkan above:
TopLeftRaster

Here is the previous #1732 (at 1x, OpenGL):
TopLeftRaster

### DIFF
--- a/hw/xbox/nv2a/pgraph/gl/draw.c
+++ b/hw/xbox/nv2a/pgraph/gl/draw.c
@@ -92,7 +92,6 @@ void pgraph_gl_clear_surface(NV2AState *d, uint32_t parameter)
                  scissor_height = ymax - ymin + 1;
     pgraph_apply_anti_aliasing_factor(pg, &xmin, &ymin);
     pgraph_apply_anti_aliasing_factor(pg, &scissor_width, &scissor_height);
-    ymin = pg->surface_binding_dim.height - (ymin + scissor_height);
 
     NV2A_DPRINTF("Translated clear rect to %d,%d - %d,%d\n", xmin, ymin,
                  xmin + scissor_width - 1, ymin + scissor_height - 1);
@@ -204,9 +203,10 @@ void pgraph_gl_draw_begin(NV2AState *d)
     }
 
     /* Front-face select */
+    /* Winding is reverse here because clip-space y-coordinates are inverted */
     glFrontFace(pgraph_reg_r(pg, NV_PGRAPH_SETUPRASTER)
                     & NV_PGRAPH_SETUPRASTER_FRONTFACE
-                        ? GL_CCW : GL_CW);
+                        ? GL_CW : GL_CCW);
 
     /* Polygon offset */
     /* FIXME: GL implementation-specific, maybe do this in VS? */
@@ -340,7 +340,6 @@ void pgraph_gl_draw_begin(NV2AState *d)
 
     pgraph_apply_anti_aliasing_factor(pg, &xmin, &ymin);
     pgraph_apply_anti_aliasing_factor(pg, &scissor_width, &scissor_height);
-    ymin = pg->surface_binding_dim.height - (ymin + scissor_height);
     pgraph_apply_scaling_factor(pg, &xmin, &ymin);
     pgraph_apply_scaling_factor(pg, &scissor_width, &scissor_height);
 
--- a/hw/xbox/nv2a/pgraph/gl/surface.c
+++ b/hw/xbox/nv2a/pgraph/gl/surface.c
@@ -137,11 +137,7 @@ static void init_render_to_texture(PGRAPHState *pg)
         "layout(location = 0) out vec4 out_Color;\n"
         "void main()\n"
         "{\n"
-        "    vec2 texCoord;\n"
-        "    texCoord.x = gl_FragCoord.x;\n"
-        "    texCoord.y = (surface_size.y - gl_FragCoord.y)\n"
-        "                 + (textureSize(tex,0).y - surface_size.y);\n"
-        "    texCoord /= textureSize(tex,0).xy;\n"
+        "    vec2 texCoord = gl_FragCoord.xy / textureSize(tex, 0).xy;\n"
         "    out_Color.rgba = texture(tex, texCoord);\n"
         "}\n";
 
@@ -298,7 +294,7 @@ static void render_surface_to_texture_slow(NV2AState *d,
     size_t bufsize = width * height * surface->fmt.bytes_per_pixel;
 
     uint8_t *buf = g_malloc(bufsize);
-    surface_download_to_buffer(d, surface, false, true, false, buf);
+    surface_download_to_buffer(d, surface, false, false, false, buf);
 
     width = texture_shape->width;
     height = texture_shape->height;
@@ -738,7 +734,7 @@ static void surface_download(NV2AState *d, SurfaceBinding *surface, bool force)
 
     nv2a_profile_inc_counter(NV2A_PROF_SURF_DOWNLOAD);
 
-    surface_download_to_buffer(d, surface, true, true, true,
+    surface_download_to_buffer(d, surface, true, false, true,
                                d->vram_ptr + surface->vram_addr);
 
     memory_region_set_client_dirty(d->vram, surface->vram_addr,
@@ -875,20 +871,26 @@ void pgraph_gl_upload_surface_data(NV2AState *d, SurfaceBinding *surface,
                        surface->fmt.bytes_per_pixel);
     }
 
-    /* FIXME: Replace this flip/scaling */
+    /* FIXME: Replace this scaling */
 
     // This is VRAM so we can't do this inplace!
-    uint8_t *flipped_buf = (uint8_t *)g_malloc(
-        surface->height * surface->width * surface->fmt.bytes_per_pixel);
-    unsigned int irow;
-    for (irow = 0; irow < surface->height; irow++) {
-        memcpy(&flipped_buf[surface->width * (surface->height - irow - 1)
-                                 * surface->fmt.bytes_per_pixel],
-               &buf[surface->pitch * irow],
-               surface->width * surface->fmt.bytes_per_pixel);
+    uint8_t *optimal_buf = buf;
+    unsigned int optimal_pitch = surface->width * surface->fmt.bytes_per_pixel;
+
+    if (surface->pitch != optimal_pitch) {
+        optimal_buf = (uint8_t *)g_malloc(surface->height * optimal_pitch);
+
+        uint8_t *src = buf;
+        uint8_t *dst = optimal_buf;
+        unsigned int irow;
+        for (irow = 0; irow < surface->height; irow++) {
+            memcpy(dst, src, optimal_pitch);
+            src += surface->pitch;
+            dst += optimal_pitch;
+        }
     }
 
-    uint8_t *gl_read_buf = flipped_buf;
+    uint8_t *gl_read_buf = optimal_buf;
     unsigned int width = surface->width, height = surface->height;
 
     if (pg->surface_scale_factor > 1) {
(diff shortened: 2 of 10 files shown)


Triticum0:
Should fix all of #2174, #2175, #2156 ,#1326, #1210, #1177, #1158, #1142 first issue, #1141, #1115, #951, #725, #2162, #1634, #1040 ,#1097, #988, #598, #979, #576, #544, #2030, #1935, #1765,  #1677, #1139 ,#866, #1739, #1680, #1222

Triticum0:
#2135, #2055, #1981, #1938, #1763, #1742, #1424, #998, #57, #1438 These will need to be retested

viniciusol263:
image
No bright corner issue anymore on Splinter Cell Chaos Theory and Double Agent anymore

abaire:
@Triticum0 have you checked Tiger Woods 2005 yet? I think that was the game that has failed for the previous PRs (which all introduced gaps due to vertex misalignment between fixed function and programmable pipes).

coldhex:
> @Triticum0 have you checked Tiger Woods 2005 yet? I think that was the game that has failed for the previous PRs (which all introduced gaps due to vertex misalignment between fixed function and programmable pipes).

That game has gaps on Xbox hardware too. It's not surprising since the fixed-function has scaling and translation multiplied into composite matrix and the programmable vertex shader does scaling and translation after perspective division. In animation, small floating-point inaccuracies are bound to occasionally end up on different sides of a rounding boundary.

hw/xbox/nv2a/pgraph/gl/surface.c
```diff
@@ -878,11 +874,12 @@ void pgraph_gl_upload_surface_data(NV2AState *d, SurfaceBinding *surface,
     /* FIXME: Replace this flip/scaling */
 
     // This is VRAM so we can't do this inplace!
+    // TODO: is this copy needed anymore? Needed because of surface->pitch?
```
abaire: If it's only a pitch fixup, maybe we should make it conditional on a mismatch and rename it? Now that it's not flipping anything it seems misleading to call it `flipped_buf`

hw/xbox/nv2a/pgraph/glsl/vsh-ff.c
```diff
@@ -471,15 +469,16 @@ GLSL_DEFINE(materialEmissionColor, GLSL_LTCTXA(NV_IGRAPH_XF_LTCTXA_CM_COL) ".xyz
...
     mstring_append(body,
-    "   oPos = tPosition * compositeMat;\n"
-    "   oPos.w = clampAwayZeroInf(oPos.w);\n"
-    "   oPos = invViewport * oPos;\n"
+    "  oPos = tPosition * compositeMat;\n"
+    "  oPos.z = oPos.z / clipRange.y;\n"
+    "  oPos.w = clampAwayZeroInf(oPos.w);\n"
+    "  oPos.xy /= oPos.w;\n"
+    "  oPos.xy += c[" stringify(NV_IGRAPH_XF_XFCTX_VPOFF) "].xy;\n"
+    "  oPos.xy = trunc(oPos.xy * 16.0f) / 16.0f;\n"
```
abaire: Worth adding a comment similar to the one in vsh-prog.c or adding a shared helper to do the 4-bit truncation?

abaire:
> That game has gaps on Xbox hardware too.

Interesting, I don't know that I ever actually tested it on HW but recall the reported issue sending me down a spiral of flailing at the PR trying to find a solution that would mesh. I think that's what stalled out #1362 as well.

Nice work!

hw/xbox/nv2a/pgraph/gl/surface.c
mborgerson: Yeah, for pitch. Agree with @abaire on rename and condition

mborgerson:
Fantastic!

> This PR also contains an OpenGL-backend commit that flips y-coordinates thereby restoring the top-left rule.

I had been meaning to do this to the GL backend for a while. I'm happy to see this not only simplifies things but fixes bugs in the process!

> I'll list explicitly what this PR doesn't do: [...]

Great points for future investigation.

hw/xbox/nv2a/pgraph/gl/draw.c
```diff
@@ -206,7 +205,7 @@ void pgraph_gl_draw_begin(NV2AState *d)
     /* Front-face select */
     glFrontFace(pgraph_reg_r(pg, NV_PGRAPH_SETUPRASTER)
                     & NV_PGRAPH_SETUPRASTER_FRONTFACE
-                        ? GL_CCW : GL_CW);
+                        ? GL_CW : GL_CCW);
```
mborgerson, line 209: May want to comment on winding order reversal due to y-inversion

hw/xbox/nv2a/pgraph/gl/surface.c
coldhex: I renamed flipped_buf and made the copy conditional. Branch rebased.

hw/xbox/nv2a/pgraph/glsl/vsh-ff.c
coldhex: Yes, that's nicer. Added common roundScreenCoords function. Branch rebased.

hw/xbox/nv2a/pgraph/gl/draw.c
coldhex: Comment added. Branch rebased.

coldhex:
Thanks for the comments @abaire @mborgerson!

About Tiger Woods 2005, here is Tiger Woods 2005 and PR (note the three white dots):
pr-tiger-woods-2005

Here Xbox HW:
xboxhw-tiger-woods-2005

My Xbox composite video capture is interlaced and blurry, but the three white dots are distinguishable in the same spots (left and about middle in height). I've checked PR using Renderdoc and the dots appear along the seam of fixed-function and programmable shaders, with one coordinate being rounded different (the 1.0/16.0 jump). There's no reason to expect that the gaps would always match that well between PR and HW because of floating-point calculation differences, but in this case they match quite well.

mborgerson:
> About Tiger Woods 2005, here is Tiger Woods 2005 and PR (note the three white dots):

It's always nice to see xemu being accurate enough to reproduce game bugs

mborgerson:
Thanks for the patch! I especially appreciate the test cases and detailed PR description.

PR #2203, opened 2025-05-25
cli: Allow overriding skip startup animation

ShiralynDev:
Allows skipping the startup animation with `-short-animation` which is very neat for me when developing games but don't want to override my current setting for when I play games.

I was unsure what to name the arg but as the internals seems to call it "short-animation", I figured it would make the most sense.

### DIFF
--- a/system/vl.c
+++ b/system/vl.c
@@ -2902,9 +2902,19 @@ void qemu_init(int argc, char **argv)
         "none",
     }[g_config.sys.avpack];
 
+    // Allow overriding the skip startup animation setting from command line
+    bool short_animation = g_config.general.skip_boot_anim;
+    for (int i = 1; i < argc; i++) {
+        if (argv[i] && strcmp(argv[i], "-short-animation") == 0) {
+            argv[i] = NULL;
+            short_animation = true;
+            break;
+        }
+    }
+
     fake_argv[fake_argc++] = g_strdup_printf("xbox%s%s%s,avpack=%s",
         (bootrom_arg != NULL) ? bootrom_arg : "",
-        g_config.general.skip_boot_anim ? ",short-animation=on" : "",
+        short_animation ? ",short-animation=on" : "",
         ",kernel-irqchip=off",
         avpack_str
         );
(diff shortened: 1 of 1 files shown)


system/vl.c
```diff
@@ -2902,9 +2902,19 @@ void qemu_init(int argc, char **argv)
         "none",
     }[g_config.sys.avpack];
 
+    // Allow overriding the skip startup animation setting from command line
+    bool short_animation = g_config.general.skip_boot_anim;
+    for (int i = 1; i < argc; i++) {
+        if (argv[i] && strcmp(argv[i], "-short-animation") == 0) {
```
abaire: I think naming it `-skip_boot_anim` to match the toml (similar to `dvd_path` behavior) might be easier to keep track of. That way the toml file serves as somewhat discoverable documentation and eventually may be prime for a refactor to allow command-line overrides of any/most of the keys via some common function + table instead of having to scatter bespoke `strcmp` loops everywhere.

ShiralynDev: I was also thinking about `-skip_boot_anim` but it feels a bit wired to mix abbreviations like that as that too would differ from the toml. Would `-skip_boot_animation` be good or is it too long? I hear what you are saying about a refactor but im going to leave it up for someone else to do as I still am very new to the codebase and mostly code C++ instead of C and I don't want to mess up.

ShiralynDev: Sorry about that differ from toml comment, I somehow missed that it is exactly what the toml is.

abaire (approved):

mborgerson:
You can do this already by appending the following:
* To skip intro animation sequence: `-machine xbox,short-animation=on`
* To play intro animation sequence: `-machine xbox,short-animation=off`

ShiralynDev:
> You can do this already by appending the following:
> 
> * To skip intro animation sequence: `-machine xbox,short-animation=on`
> * To play intro animation sequence: `-machine xbox,short-animation=off`

I could not find this is the documentation on the website, is it part of the qemu launch args?

mborgerson:
> I could not find this is the documentation on the website,

It's not documented on the website.

> is it part of the qemu launch args?

You can add this to your xemu.exe launch arguments and it will be appended to the QEMU launch arguments.

ShiralynDev:
Should probably be added to documentation. But will close this PR

PR #2218, opened 2025-05-31
meson: Use optimization=3

mborgerson:


mborgerson:
Will handle this in distribution instead.

PR #2238, opened 2025-06-08
ui: Add .xiso to load disc file filter

### LINKED ISSUE #2237: Update open filter to allow files with .xiso extension
### Feature Request

On Discord we decided that the use of .iso creates enough user confusion that we should switch tools over to use a ".xiso" extension. This will require modifying xemu to allow .xiso extensions in the load dialog as well as .iso.

See https://github.com/xemu-project/xemu/blob/e02e41ccaaceffcad816cfb554dd1195767e952d/ui/xui/actions.cc#L37

### Alternatives

_No response_

### Additional Context

_No response_

abaire:
Fixes #2237

### DIFF
--- a/ui/xui/actions.cc
+++ b/ui/xui/actions.cc
@@ -39,7 +39,7 @@ void ActionLoadDisc(void)
     Error *err = NULL;
 
     const char *iso_file_filters =
-        "xiso Files\0*.xiso;*.iso\0All Files\0*.*\0";
+        "Disc Image Files (*.iso, *.xiso)\0*.iso;*.xiso\0All Files\0*.*\0";
     const char *new_disc_path =
         PausedFileOpen(NOC_FILE_DIALOG_OPEN, iso_file_filters,
                        g_config.sys.files.dvd_path, NULL);
--- a/ui/xui/actions.cc
+++ b/ui/xui/actions.cc
@@ -38,7 +38,8 @@ void ActionLoadDisc(void)
 {
     Error *err = NULL;
 
-    const char *iso_file_filters = ".iso Files\0*.iso\0All Files\0*.*\0";
+    const char *iso_file_filters =
+        "xiso Files\0*.xiso;*.iso\0All Files\0*.*\0";
     const char *new_disc_path =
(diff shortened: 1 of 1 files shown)


ui/xui/actions.cc
```diff
@@ -38,7 +38,8 @@ void ActionLoadDisc(void)
 {
     Error *err = NULL;
 
-    const char *iso_file_filters = ".iso Files\0*.iso\0All Files\0*.*\0";
+    const char *iso_file_filters =
+        ".xiso Files\0*.xiso\0.iso (xiso) Files\0*.iso\0All Files\0*.*\0";
```
abaire: I split .xiso and .iso into distinct options under the assumption that doing "xiso Files": ".xiso;.iso" would still lead users to try to open non-xiso isos.

mborgerson: Hmm.. I agree with this in principle but, because both extract-xiso and xdvdfs currently produce `*.iso` files, I'm concerned that filtering `*.xiso` and `*.iso` files independently at this point might increase user confusion; in this case, the user may have to change the file extensions manually, which could be difficult for novice users if file extensions are hidden, or always have to adjust the dialog filter to show their `*.iso` files. It might be worth exploring integration of xdvdfs into xemu to detect common failure cases. I think for now to smooth the transition to `*.xiso` extension support, we might want to filter them both in by default.

abaire: Agreed, the best solution would be to detect that the input is invalid and bomb out with a message. Updated to support both under the "xiso Files" moniker

antangelo: I don't think xdvdfs (the library) is required to check for XGD files being loaded; it should be sufficient to check if the [xdvdfs magic string](https://xboxdevwiki.net/XDVDFS#Volume_descriptor) exists at the various [XGD (or xiso) offsets](https://github.com/antangelo/xdvdfs/blob/main/xdvdfs-core/src/blockdev.rs#L14-L20). It might be useful in detecting other failure modes (like reading the xbe certificate media flags, or comparing checksums of data inside the image), but checking for XGDs should be simple enough to replicate.

mborgerson:
Thanks!

Issue #2191, opened 2025-05-21
Star Wars: Battlefront II: Apparent Z-fighting on water/sky


ArtemRU33:
### Title

https://xemu.app/titles/4c41001a/#Star-Wars-Battlefront-II

### Bug Description

Z-fighting on water, skyboxes. Most noticable on Naboo map.

Though I did use a modded game, virtually all vanilla maps have the same issue.

![Image](https://github.com/user-attachments/assets/ac821bfd-294f-4e92-989f-4c3d40e1e993)

https://files.catbox.moe/wjpz6c.mp4

### Expected Behavior

From v0.8.29

https://files.catbox.moe/kyk0sk.mp4

### xemu Version

v0.8.66

Last working in: v0.8.29

### System Information

OS: Windows 11
CPU: Intel® Core™ i7-10870H
GPU: Nvidia RTX 3060
GPU Driver: 576.02 Game Ready Driver

Triticum0:
Should be fixed on coldhex dev build.

ArtemRU33:
yes i just found out in the discord, that was the coldhex build someone mentioned, i still need it in main version for stability purpose

Edit: and also coop link

mborgerson:
@ArtemRU33:

In the future, please make the issue title/body more descriptive of the actual problem(s), and upload your photos/videos to GitHub by dragging them into the page so readers don't need to go to external hosting sites for the media. I've edited the issue text for clarity.

@Triticum0:

GitHub users, repos, and branches tend to move around. Please be more specific when referencing other builds for investigation, like with a permalink to the specific patch that you've tested.

Triticum0:
Its depth buffer precision improvements and polygon offset slope factor commit which should fix this issue. Coldhex mentioned a similar issue happening with shenmue 2 causing lots of z-fighting #2049. And tested other games which had that regression and for the most part are all fixed by this pr.

Triticum0:
also effect.
https://xemu.app/titles/4541001d/#NCAA-Football-2003

Triticum0:
@ArtemRU33 should be fixed

Triticum0:
Fixed might also need Signed texture support as missing transparency.

<img width="640" height="480" alt="Image" src="https://github.com/user-attachments/assets/9adcad78-82be-4379-9a38-c03540f84b17" />

Issue #1155, opened 2022-07-02
Big Bumpin' Menu FMV only render once you been In-game


Triticum0:
### Title

https://xemu.app/titles/4d5300d2/#Big-Bumpin

### Bug Description

The menu is missing until the demo screen plays then the menu becomes visible this also happens if you finish a match and return to the menu.

Xemu:
![xemu-2022-07-02-20-40-17](https://user-images.githubusercontent.com/71127352/177014266-bc4f0abb-3d1a-49a0-9650-eb9982863346.png)

### Expected Behavior

Should appear on the first boot
https://www.youtube.com/watch?v=qjy9tC-mz_8

### xemu Version

Version: 0.7.56
Branch: master
Commit: https://github.com/mborgerson/xemu/commit/d06a0c22f1617d635bde11d73223b24cd803b966
Date: Sat Jul 2 03:53:35 PM UTC 2022

### System Information

<html><body>

Field | Value
-- | --
OS | Windows 10
CPU | AMD Ryzen 5 2600 Six-Core Processor
Graphics Device | NVIDIA GeForce RTX 3060 Ti/PCIe/SSE2
Graphics Driver | 4.0.0 NVIDIA 512.95

</body>
</html>

Triticum0:
Seems to be fixed on cold-hex dev build

abaire:
@Triticum0 is this one still valid now that coldhex's PR is merged?

Triticum0:
Seem to be a bit laggy but it fixed

Issue #2228, opened 2025-06-02
Burnout Revenge: HUD elements are not visible


antoniodesousa:
### Title

https://xemu.app/titles/45410076/#burnout-revenge

NTSC EA-118 1.01

### Bug Description

HUD elements are not visible in several parts of the game. The issue was first introduced on the v0.8.30 release.

Steps to reproduce:
1 - Load the game
2 - Skip the intro
3 - Several sections in the title screen, menus, etc; are not visible.

![Image](https://github.com/user-attachments/assets/64e30717-0d2e-4d2b-a9c6-f4de171a65b0)

### Expected Behavior

HUD elements should be visible across the game.

![Image](https://github.com/user-attachments/assets/f38bc8e3-8e23-4084-9429-7c2a277cc5f8)

### xemu Version

Version: 0.8.30
Branch: master
Commit: 3eb22b6b81c121f6fe266884b01a9cd3e77e279d
Date: Sat Mar  8 21:58:22 UTC 2025

Last working version:
Version: 0.8.29
Branch: master
Commit: 45078ef51f6e357a05e0efda8692bac01937115b
Date: Sat Mar  8 11:03:58 UTC 2025

### System Information

CPU: AMD Ryzen 9 6900HX with Radeon Graphics        
OS Platform: Windows
OS Version: 24H2
Manufacturer: ATI Technologies Inc.
GPU Model: AMD Radeon RX 6600M
Driver: 4.0.0 Core Profile Context 25.5.2.250527
Shader: 4.60

### Additional Context

The culprit commit of the issue:
https://github.com/xemu-project/xemu/commit/798ad308198e85e9506af96ff641cd2bd43aaa66

Triticum0:
Test with coldhex dev build. should be fixed with depth buffer precision improvements and polygon offset slope factor commit. He just hasn't created a pr for the issue yet.

antoniodesousa:
I just tested coldhex dev build and it doesn't have the issue.

Triticum0:
Retest should be fixed by #2240

antoniodesousa:
Fixed on Xemu v0.8.109!

Issue #2213, opened 2025-05-30
Beat Down: Vertical lines and crooked borders in text windows


PlatypusFNC:
### Title

https://xemu.app/titles/43430015/#Beat-Down-Fists-of-Vengeance

### Bug Description

Text windows during gameplay have vertical lines at the bottom and have crooked borders (note the left and right borders of the dialogue window in the screenshot).

Xemu 0.8.67
![Image](https://github.com/user-attachments/assets/8c1edb45-2eee-4cc3-b50d-aa53da2dfd7a)

### Expected Behavior

![Image](https://github.com/user-attachments/assets/866e5b1c-1d74-4c05-a41d-9a4ccb9efd6b)

### xemu Version

0.8.67

### System Information

CPU:          Intel(R) Core(TM) i5-9400F CPU @ 2.90GHz
OS Platform:  Windows
OS Version:   Windows 8.1 Pro
Manufacturer: NVIDIA Corporation
GPU Model:    GeForce GTX 1050 Ti/PCIe/SSE2
Driver:       4.0.0 NVIDIA 442.59
Shader:       4.00 NVIDIA via Cg compiler

Triticum0:
A regression caused by #2183

coldhex:
This regression doesn't occur on Xemu 0.8.71, Linux, Mesa 25.1.1, with neither Intel UHD 770 nor AMD Radeon RX 6600 (also not with the llvmpipe software renderer.)

I inspected rendering of the text window in Renderdoc. The game draws the window border using several small quads around the border using a 64x128 texture atlas with nearest-neighbor sampling. The left edges of the quads have x-coordinates at half-pixel coordinates while the texture coordinates are at exactly integer coordinates (u=0, 32 or 64, and v = 0, 32, 64, 96 or 128). This means texture sampling occurs at exactly half-way between the centers of two neighboring texels. According to OpenGL/Vulkan specs the chosen texel should then be consistently the one on the right. Looking at the screenshot in the bug description it appears that the GeForce GPU nevertheless often chooses the left texel which belongs to the sub-image on the left of the correct one in the texture atlas, which would explain the artifacts. The crooked right border appears as if the chosen texel even changes at the vertical middle of those quads, even though the x-coordinates and texture u-coordinates are constant along the quad's left and right edge (and y and v are constant along top and bottom edges).

[...]

PR #2240, opened 2025-06-08
nv2a: Depth buffer precision improvements and polygon offset slope factor

### LINKED ISSUE #472: Super Monkey Ball Deluxe: cutscenes not rendering
### Title

https://xemu.app/titles/53450038

### Bug Description

When starting the game on storymode the cutscenes before the levels do not render.

Steps to reproduce the issue:

1. Launch game
2. Select Main Game
3. Select Story Mode
4. Choose a save slot and a name
5. Expected cutscene is not shown (hitting start will skip to playable game)

RSewlV7

### Expected Behavior

Should show a cutscene, below is the cutscene that should be showing.

-180-Super-Monkey-Ball-Deluxe-Xbox-longplay-YouTube

### xemu Version

0.6.1-12-ga8c73197da

### System Information

os_platform: Windows
os_version: Windows
cpu: Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz
gl_vendor: NVIDIA Corporation
gl_renderer: NVIDIA GeForce RTX 2080 Ti/PCIe/SSE2
gl_version: 4.0.0 NVIDIA 510.10
gl_shading_language_version: 4.00 NVIDIA via Cg compiler

### Additional Context

_No response_

### LINKED ISSUE #2041: The Chronicles of Riddick: Z fighting flickering
### Title

https://xemu.app/titles/56550026/#The-Chronicles-of-Riddick-Escape-from-Butcher-Bay

VU-038 NTSC v1.00

### Bug Description

Blood spatters on wall have an issue flickering if you place the player's angle to the side

Steps to reproduce:
Play the intro until you kill Johns and drop down the hatch. Kill the mechanic and make sure you hit/ kill him close to a wall so that blood can splatter on it. Then walk around and inspect the blood from different angles

https://github.com/user-attachments/assets/d53cae6f-13b7-44b4-bf24-32e4a3e2e7da

### Expected Behavior

Blood splatter should not flicker

https://youtu.be/Xu5BH4j0K1A?t=148

### xemu Version

0.8.39

### System Information

OS: Windows 11
CPU: 13th Gen Intel(R) Core(TM) i9-13980HX
GPU: NVIDIA GeForce RTX 4080 Laptop GPU/PCIe/SSE2
GPU Driver: 4.0.0 NVIDIA 566.14

### Additional Context

_No response_

### LINKED ISSUE #2049: Shenmue II: Flickering sky
### Title

https://xemu.app/titles/4d530032/#Shenmue-II

NTSC MS-050 v1.00

### Bug Description

Sky seems to be flickering weird patterns. Tested this with xemu version 0.8.29 and didn't have the issue on that build. Issue came up on version 0.8.30. Issue is still present on xemu version 0.8.40

xemu 0.8.29 (sky looks normal)

https://github.com/user-attachments/assets/71e1b098-1493-44c8-8b66-f93215e07079

xemu 0.8.30 (sky bugging out)

https://github.com/user-attachments/assets/f569f923-3faf-4132-8824-7887be7d3200

xemu 0.8.40
https://github.com/xemu-project/xemu/issues/988#issuecomment-2741538584

### Expected Behavior

Sky should not be flickering, should look like how xemu 0.8.29 looked

### xemu Version

0.8.29 (good sky)
0.8.30 (bad sky)
0.8.40 (bad sky)

### System Information

OS: Windows 11
CPU: 13th Gen Intel(R) Core(TM) i9-13980HX
GPU: NVIDIA GeForce RTX 4080 Laptop GPU/PCIe/SSE2
GPU Driver: 4.0.0 NVIDIA 566.14

### Additional Context

Issue seems to have come up on build 0.8.30

coldhex:
1. Use barycentric coordinates to interpolate depth values. Linux, Mesa and AMD Radeon RX 6600 with Vulkan driver currently has quite poor interpolation precision, see video and tests below. This PR replaces GPU interpolation with a manual one. Also note that current Xemu w-buffer interpolation uses gl_FragCoord.w which can't reproduce all w-values, e.g. 1.0f/16777046.0f equals 1.0f/16777047.0f in 32-bit floats and so inversion can't reproduce both 16777046.0f and 16777047.0f. Also, depth value differences are used in interpolation, which has the desired property that a triangle with the same z-value on all vertices will result in exactly that same z-value when interpolated. At least the game Shenmue II sky rendering relies on this. Fixes #2049. Fixes #472.

2. Computes polygon depth bias slope for both z-buffering and w-buffering. These are computed by taking the max and abs of partial derivatives of either of the functions z=z(x,y) or w=w(x,y), where x,y,z,w are screen-space coordinates. This matches Xbox hardware for z-buffering where the partial derivatives are constants over any fixed triangle. However, for w-buffering the partial derivatives vary over any fixed triangle, but Xbox appears to compute just a single depth slope at the first visible pixel (where "first" means something like first in top-left order) and uses that over the whole triangle. The way the PR computes the partial derivatives is by using the chain-rule, e.g. dw/dx = -w^2 * d(1/w)/dx. This is useful since 1/w is linear in screen-space and therefore d(1/w)/dx is constant over any fixed triangle. However, finding out the w-value for the first visible pixel of a triangle is difficult in OpenGL/Vulkan and is not done here. Instead we calculate depth slope per-pixel. Fixes #2041.

3. Since the depth buffer calculation modifications above require modifying geometry shader code and outputs, this PR also fixes some flat shading and polygon line mode bugs since keeping them as is with the new modifications would have been at least as much work.

Relevant additions to nxdk_pgraph_tests are here:
- Depth values: https://github.com/coldhex/nxdk_pgraph_tests/commit/7e6ca18c2b0707939224ac4f9450a91d44423713
- Shade model: https://github.com/coldhex/nxdk_pgraph_tests/commit/5bcdc6729da7b06febd08324a61b1486ca816d9d

I tested using Radeon RX 6600. The following video shows the depth precision issue (which is different from the depth slope z-fighting issue and is also fixed by PR) with Xemu 0.8.67  and Chronicles of Riddick (where depth value inaccuracy causes problems with shadow volume calculations):

https://github.com/user-attachments/assets/8e218f02-72d9-40b9-94a1-da8a254f81b6

Radeon OpenGL is a little better than Vulkan. (Btw, this inaccuracy issue doesn't occur with Intel UHD 770.) I reproduced one of the triangles in an nxdk_pgraph_test 24-bit integer W-buffering test. The following is with Vulkan:

WBuf24D_Riddick_V1_ZB0_ZS0

OpenGL and Xemu 0.8.67 produce depth values closer to Xbox HW values, but not close enough as the video above shows. That was Radeon, but I also tested Xemu 0.8.67 on Intel UHD 770 and it matches Xbox HW in this test, so Intel's hardware interpolation seems to be of better quality than AMDs.

Regarding the new depth slope implementation for W-buffering, here are tests for the case of 24-bit integer W-buffer. These use slope factor 65536.0, which is much larger than games would typically use (e.g. Riddick uses 1.0), but it makes it possible to see how the implementation in PR doesn't produce exactly the correct values. Note how Xbox HW appears to compute the slope offset value only once at first visible pixel. The quad is exactly the same among ClipF tests and exactly the same among ClipW tests, only window clipping changes (on the red line):

WBuf24D_ClipF-150-032_V1_ZB0_ZS1
WBuf24D_ClipF-150-128_V1_ZB0_ZS1
WBuf24D_ClipF-150-224_V1_ZB0_ZS1
WBuf24D_ClipW-159-000_V1_ZB0_ZS1
WBuf24D_ClipW-261-000_V1_ZB0_ZS1
WBuf24D_ClipW-363-000_V1_ZB0_ZS1

(Also note in ClipW tests how the last W-value 0x30CB5 doesn't change. This is because Xbox draws the quad as two triangles and the last W-value belongs to the second triangle which doesn't need clipping and therefore its depth slope is unaffected.)

To further show curious Xbox HW behaviour, the following tests draw the same triangle in 1-pixel offsets. TriH tests draws the triangles from left to right with 1-pixel vertical drops. TriV tests draw from top to bottom with 1-pixel shifts to right. The cycling with period 4 is due to the 1-pixel shifts only.

WBuf24D_TriH_V1_ZB0_ZS0
WBuf24D_TriH_V1_ZB0_ZS1
WBuf24D_TriV_V1_ZB0_ZS0
WBuf24D_TriV_V1_ZB0_ZS1

From this, I'm guessing NV2A rounds the first pixel to some point in a 4x4 pixel block and samples the depth slope once for the whole triangle. Sampling once in a such a manner is rather difficult to implement in OpenGL/Vulkan graphics pipeline so PR doesn't even try and implements slope calculation per-pixel instead.

For Z-buffering, depth slope calculation in this PR is about the same as Xemu 0.8.67. However, depth offset improves since, for whatever reason, Mesa adds the offset doubled:

ZBuf24D_FloorQuad_V1_ZB0_ZS0
ZBuf24D_FloorQuad_V1_ZB0_ZS1
ZBuf24D_FloorQuad_V1_ZB1_ZS0

Depth offset for lines is fixed in this PR. This depends on if V_PGRAPH_SETUPRASTER_POFFSETLINEENABLE is set. Xemu 0.8.67 applies the offset if any POFFSET is enabled, so PR improves on that, e.g.:

WBuf24D_LineStrip_V1_ZB0_ZS0
WBuf24D_LineStrip_V1_ZB1_ZS0

For the flat shading and polygon line mode changes, here are test cases where PR makes improvements: (I include some test cases where also Xemu 0.8.67 matches HW)

ProgLM_TriFan_Flat_First
ProgLM_TriFan_Flat_Last
ProgLM_Tri_Flat_First
ProgLM_Tri_Flat_Last
ProgLM_TriStrip_Flat_First
ProgLM_TriStrip_Flat_Last
ProgLM_TriStrip_Smooth_First
ProgLM_TriStrip_Smooth_Last
Prog_TriFan_Flat_First
Prog_TriFan_Flat_Last
Prog_Tri_Flat_First
Prog_Tri_Flat_Last
Prog_TriStrip_Flat_First
Prog_TriStrip_Flat_Last


Triticum0:
Should fix. #2204 #2191 #2228 #2038  . Also fixed madden series but never created an issue for it. Thought this would solve the remaining z-fighting issues but seem to be unrelated. #1020

Triticum0:
Also a couple games regressed from your previous pr, #2213. Also marvel vs capcom, metal slug 5 where also effected. In my extensive testing I never came across any issues unfortunately before pr was merged but hopefully you know now. Let me know if you want me to open an issues for it. Lastly keep up the great work.

coldhex:
@Triticum0 Yes, a screenshot of Metal Slug 5 would be nice since the Beat Down regression doesn't happen with my hardware.

Triticum0:
Here you go. My guess it nivida exclusive I need a build renderdoc debuh build of lastest realese. once I do I see what I can do.

Metal slug 5
xemu-2025-06-09-22-20-17

MvsC  2
xemu-2025-06-09-22-22-56

Edit: Tested with your suggestion and seem to fix the issue.

mborgerson:
Very cool, thanks! It may take me a few days to get time to review.

Triticum0:
@coldhex If you have time might be worth look at the remain z-fighting issuses I know matt has [Alias](https://xemu.app/titles/41430016/#Alias) so you could start there or outrun 2 as z-fighting easily noticeable and  could be related to this pr https://github.com/xemu-project/xemu/issues/1020

coldhex:
Thanks for the screenshots. Metal Slug 5 doesn't have such artifacts on Radeon (or Intel UHD 770). (On Beat Down issue, I commented that I now noticed there is an artifact present on Radeon, but less than on GeForce.) Looking at MS5 in renderdoc, it too renders the "Mission ..." text using rectangles at half-pixel coordinates and nearest-neighbor texture atlases, like Beat Down. So they have the same cause.

OutRun 2 has z-fighting on distant buildings even on Xbox HW. Seems to be caused by the usual low precision of z-buffering at distances far from the camera. This PR may (or may not) reduce the z-fighting to be closer to HW, but in general PR (or any Xemu version) does not produce exact HW z-buffer values. There may be small differences. Additionally, different floating-point rounding mode in NV2A and in modern GPUs cause differences in vertex program output already. In one nxdk pgraph test (one of texture shadow comparator), there is a +3 difference on vertex program z-coordinate output caused only by floating-point rounding mode (round-towards-zero vs round-to-nearest). To obtain exactly the same z-buffer values floating-point arithmetic would have to be exactly the same as in NV2A and also exactly the same interpolation algorithm and order of calculations would need to be followed (whatever those are.)

antoniodesousa:
I tested Metal Slug 5 and Beat Down on my AMD RX 6600 GPU and I don't have those lines.

hw/xbox/nv2a/pgraph/glsl/geom.c
```diff
@@ -23,136 +23,207 @@
...
-                   "  emit_vertex(1, 2);\n"
-                   "  emit_vertex(2, 2);\n"
+            body = "  mat4 pz = calc_triz(0, 1, 2);\n"
+                   "  emit_vertex(0, pz);\n"
+                   "  emit_vertex(1, pz);\n"
+                   "  emit_vertex(2, pz);\n"
+                   "  EndPrimitive();\n";
+        } else if (polygon_mode == POLY_MODE_LINE) {
+            need_linez = true;
+            /* FIXME: input here is lines and not triangles so we cannot
```
mborgerson, line 277: How do you think we should handle this?

hw/xbox/nv2a/pgraph/vk/draw.c
```diff
@@ -793,12 +789,10 @@ static void create_pipeline(PGRAPHState *pg)
     VkPipelineRasterizationProvokingVertexStateCreateInfoEXT provoking_state;
 
     if (r->provoking_vertex_extension_enabled) {
+        // TODO: remove use of provoking vertex extension since we just want
```
mborgerson: Any reason not to do it now?

mborgerson (commented):
This was a fun review--thanks! Once again, thanks also for providing test cases and comparisons against Xbox/current xemu. I took the liberty of rebasing the patch set onto current master and fixing the conflicts. Left a couple of questions in review before merge

Triticum0:
Should fix #2191

hw/xbox/nv2a/pgraph/glsl/geom.c
coldhex: I think n-polygon in line mode should be changed to use a triangle fan instead of a line strip (or line loop in OpenGL.)
https://github.com/xemu-project/xemu/blob/1343d5873f9b9de251f41d7a8c99d37d1741e640/hw/xbox/nv2a/pgraph/vk/draw.c#L83-L87

There isn't any built-in way for a geometry shader to detect the last triangle in a triangle fan, but one possibility is to add a new boolean vertex attribute such that it is true only for the last vertex of a triangle fan. The geometry shader could then detect the last triangle using that boolean and complete the boundary line loop.

hw/xbox/nv2a/pgraph/vk/draw.c
coldhex: Setting last vertex convention is actually necessary, because Vulkan default is first vertex (unlike OpenGL where default is last vertex.) I fixed the comment.

I also added a new fixme comment to the geometry shader. I'll explain this in more detail here. Suppose a triangle strip's vertex data are denoted by the following IDs: 1 2 3 4 5 6. When the last vertex convention is used, the sequence of four input triangles received by the geometry shader could be reasonably assumed to be:

Case A:
123
324
345
546

The last vertex for these triangles matches with the last vertex convention for the strip. (The second and fourth triangles had two vertices swapped to get correct winding order.) On the other hand if the first vertex convention was used instead, a more reasonable triangle vertex order would be:

Case B:
123
243
345
465

With this ordering, strip first vertex convention and triangle first vertex convention still match.

At least AMD Radeon RX 6600 and Nvidia GeForce GTX 1650 behave as above. However, Intel UHD 770 and Intel HD Graphics 630 always use Case A ordering regardless of the vertex convention setting. In the PR geometry shader, Case A ordering is assumed, so flat shading works for all these GPUs. However, I have now tested with the much older Intel HD Graphics 4000 (integrated in Ivy Bridge CPUs) and it always uses Case B ordering regardless of the vertex convention setting and therefore PR gets the flat shading colors wrong for every other triangle in strips. Same problem of implementation dependence also exists in Xemu master in line mode flat shading and n-polygon flat shading (but for e.g. triangle strips with fill mode, master doesn't use the geometry shader at all.)

OpenGL and Vulkan only specify winding order, but absolute vertex order is implementation dependent. OpenGL specs leave the reader wondering if the order is specified or not, but Vulkan docs mention this explicitly:

https://github.com/KhronosGroup/Vulkan-Docs/blob/481dd347dae6cac4bf4ef8dcfae41789ab39f650/chapters/geometry.adoc
"Vertices may be in a different absolute order than specified by the topology, but must adhere to the specified winding order."

Since there isn't a built-in way to detect how GPUs order input triangle vertices in geometry shaders, one possible fix would be to add an integer vertex attribute (along with texture coordinates etc.) giving the original vertex index. (The boolean vertex attribute for a triangle fan's last vertex could be combined with this attribute such that a negative index indicates the last vertex.)

coldhex:
Another thing about the PR: since this uses geometry shaders for triangles, triangle strips and fans, it adds to the stop-the-world pauses when shaders are compiled. Since geometry shader code generation depends only on a relatively few variables such as primitive type and polygon mode, there aren't all that many geometry shader code variations. However, currently geometry shaders are compiled every time there is a new vertex or fragment shader. It might be worth to cache geometry shaders separately to avoid recompiling the same geometry shader code.

Triticum0:
Fixes the z-fighting in madden series as well fixes z-fighting in Pariah look at the gun.
https://www.youtube.com/watch?v=9ki3JeHnRKA&list=PLmPVsOuxTsQJpSGHGa1YFSWRqnUohZSrt&index=3

mborgerson:
> It might be worth to cache geometry shaders separately to avoid recompiling the same geometry shader code.

Agreed. ~~Vertex, geometry, and fragment shaders should probably be cached independently. We'll defer it for now~~ Done in #2324

hw/xbox/nv2a/pgraph/vk/draw.c
mborgerson: > I have now tested with the much older Intel HD Graphics 4000 (integrated in Ivy Bridge CPUs) and it always uses Case B ordering regardless of the vertex convention setting and therefore PR gets the flat shading colors wrong for every other triangle in strips.

Thank you for testing with a variety of hardware, much appreciated! I'm a bit concerned about the impact for merging as-is with this caveat in mind

> one possible fix would be to add an integer vertex attribute (along with texture coordinates etc.) giving the original vertex index

Would you be willing to do this for this pr?

coldhex: >>   one possible fix would be to add an integer vertex attribute (along with texture coordinates etc.) giving the original vertex index

> Would you be willing to do this for this pr?

Yes, I can do that. It would involve these changes:

1. Original vertex index can't be a vertex attribute per se, because the maximum is 16 and those are already all taken for NV2A. With OpenGL 4.0, Buffer Textures can be used and indexed by the built-in gl_VertexID variable in shader. With Vulkan, either Uniform Texel Buffer or Storage Buffer.
2. The functions pgraph_gl_flush_draw and pgraph_vk_flush_draw need to be modified, especially for Inline Elements, since the draw command indexes a vertex array and the vertex index available to shader must be basically the index of the index array instead of the index into the vertex array. This involves unindexing and copying vertices into a new array so that datum of each vertex is separate. This also needs to be checked for Draw Arrays since at least in principle the start&count ranges could overlap. The Buffer Texture needs to be updated for each draw command.

Simpler would be to not use a Buffer Texture and only do vertex array copies similar to point 2 above. Then gl_VertexID in shader could be used by itself for flat shading vertex ordering, but it would be insufficient for closing line loops and n-polygon boundary loops, since gl_VertexID alone cannot be used to detect which primitive (e.g. triangle in a triangle fan) is last.

mborgerson: > because the maximum is 16

On what hardware? I think GL specifies a minimum of 16, but it is often higher? We may still be able to use this approach.

---

As an aside, I've been considering alternative approaches to the current geometry shader solution, mostly because MoltenVK doesn't support them and it would be nice to be able to start using Vulkan on macOS. The straightforward solution there is rewriting primitive on the host, which aligns with the copy approach you've described.

mborgerson: On second thought, if it's just ancient Intel HD4000 that's affected by this, I'm open to deferring this robustness improvement to the future if this effort to correct these primitive cases is going to be substantial

Triticum0:
@coldhex while I was testing some games I came across a where black triangle would cover most of the screen seem to stumble across it on multiple games and think I track it down to regression starting in PR: Perspective-correct interpolation for w-buffering it shows like this.

Bionicle - The Game
xemu-2025-06-29-00-23-50

Carmen Sandiego: The Secret of the Stolen Drums 2262
Image

ultimate beach soccer
xemu-2025-06-30-01-39-39


Only found out it was a regression today as wasn't very reproducible as only show up randomly when loading into the level and fixed itself when exiting levels. Only Bionicle - The Game was reproduceable so could only test now I found it good test case.

No rush only metioning it as this pr doesn't fix the issue. Also simiar issue in [Counter Terrorist Special Forces](https://xemu.app/titles/48500002/#Counter-Terrorist-Special-Forces-Fire-for-Effect) don't know if it also related

hw/xbox/nv2a/pgraph/vk/draw.c
coldhex: > On what hardware? I think GL specifies a minimum of 16, but it is often higher? We may still be able to use this approach.

Radeon OpenGL has 32, but Intel integrated and GeForce are still at 16 at least with Mesa on hardware what I tested. Online searching gives results of users posting GeForce OpenGL parameters and it always seems to be 16. GeForce GTX 1650 has 32 with Vulkan.

Anyway, I rebased the PR and added a new commit which implements point 2 above (and this fixes HD 4000). It doesn't need a new vertex attribute and just uses gl_VertexID (and gl_VertexIndex for Vulkan). The new vertex attribute would be needed for closing n-polygon line mode boundary. I initially thought the same approach could be used for plain line loops. That would almost work, but not for the special line loop having only two vertices since triangle fan requires at least 3. So fixing Vulkan line loop needs writing an additional vertex for the drawing command anyway and if that is done in the future then might as well add an additional vertex for n-polygon line mode too.

I'll mark this PR draft for a while. Copying inline elements to a new array seems to work fine and at least Halo 2 didn't show any noticable framerate drop for me, but I'll test this more and run all the nxdk_pgraph_tests. I don't know in what CPU Intel changed triangle vertex order to what it is in UHD 770 (and 630 is the same), but who knows if they might go back to HD 4000 style ordering in the future.

coldhex:
> @coldhex while I was testing some games I came across a where black triangle would cover most of the screen seem to stumble across it on multiple games and think I track it down to regression starting in PR: Perspective-correct interpolation for w-buffering it shows like this.

Bionicle and Counter Terrorist Special Forces seem to give garbled input data to a vertex shader. For the latter game, when exporting vertex data bytes to a file from Renderdoc, the data even contained an English word in ascii in it even though it is supposed to be coordinates and RGB values. Bionicle seems to have corrupt data too and possibly always has, but the regression might be caused by a bug in this if-condition:
https://github.com/xemu-project/xemu/blob/80f7efaba567a5c25c923e8df30ed5db17198f7a/hw/xbox/nv2a/pgraph/glsl/vsh.c#L214-L221

If the floating-point number t is a positive subnormal number (from garbled passthrough input data w-coordinate), then t>0.0 is false and so is the second condition and therefore t gets clamped using the negative range and sign is flipped. Flushing subnormals to zero is what GPUs do and this can be taken into account by replacing the if-condition with e.g. "floatBitsToInt(t) >= 0". I'll do a PR about this later. The garbled input data is a separate issue. I don't know if that is Xemu or bug in game itself.

Triticum0:
Seem the bug with ultimate beach soccer fixed with this pr.

hw/xbox/nv2a/pgraph/vk/draw.c
coldhex: I rebased this PR and reworked the last commit. (The old solution is still available in the branch https://github.com/coldhex/xemu/tree/deprecated/pr/depth-precision)

Instead of modifying and copying vertex data, this now tests and detects geometry shader vertex order during initialization. Since (likely all) GPUs are consistent with their input triangle vertex ordering in geometry shaders, this seems to me sufficient and the better solution. (The previous solution would work even if the vertex ordering varied randomly at runtime.)

The detected vertex ordering is printed during initialization. For Radeon it prints "0, 0, 0, 0" indicating that no vertex reordering is needed. For Intel it prints "0, 0, 0, 2" indicating that triangle fan vertices need reordering.

Triticum0:
Here Shader winding for 30 series of graphic cards.
VK geometry shader winding: 0, 0, 0, 0
GL geometry shader winding: 0, 0, 0, 0

Test Vega 8 on  integrated graphics 7730U and also same as above.

Also got this on 3060 TI when looking at the log
Could not determine triangle rotation, got color: R=0, G=129, B=190
Could not determine triangle strip0 rotation, got color: R=0, G=171, B=169
Could not determine triangle strip1 rotation, got color: R=84, G=171, B=86
Unexpected inconsistency in triangle fan winding, got colors: R=0, G=169, B=171 and R=84, G=86, B=171

Test was done just on xbox boot logo.

coldhex:
> 
> Also got this on 3060 TI when looking at the log Could not determine triangle rotation, got color: R=0, G=129, B=190 Could not determine triangle strip0 rotation, got color: R=0, G=171, B=169 Could not determine triangle strip1 rotation, got color: R=84, G=171, B=86 Unexpected inconsistency in triangle fan winding, got colors: R=0, G=169, B=171 and R=84, G=86, B=171

Thanks for testing! Which one is that, OpenGL or Vulkan? Those color values indicate that something is wrong the rendering. It's just supposed to render some triangles to an offscreen image and those are RGB values are invalid.

Triticum0:
Opengl

coldhex:
I added error code checking and printing for OpenGL and also a glFinish call (which should be unnecessary), but nothing in particular sticks out. @Triticum0 could you try it again and see what error it gives?

Triticum0:
The error code the same
 Could not determine triangle rotation, got color: R=0, G=129, B=190
Could not determine triangle strip0 rotation, got color: R=0, G=171, B=169
Could not determine triangle strip1 rotation, got color: R=84, G=171, B=86
Unexpected inconsistency in triangle fan winding, got colors: R=0, G=169, B=171 and R=84, G=86, B=171
GL geometry shader winding: 0, 0, 0, 0

coldhex:
Ok, thanks. Odd, those RGB values are what would be expected if the test geometry shader just passed through colors per vertex. I'll try to figure it out.

coldhex:
@Triticum0 Could you test with the latest commit? I added redundant computation to the test geometry shader which may fix this if it is a driver bug.

Triticum0:
@coldhex it does fix the driver bug. no longer show up in the logs

Triticum0:
here the log
[xemu.log](https://github.com/user-attachments/files/21141315/xemu.log)

coldhex:
@Triticum0 Cool, thanks!

I applied the same fix also to the Vulkan geometry shader tester, just in case. I also updated related code comments. Furthermore, now a fresh OpenGL context is created for the tester so that it always runs with clean state.

Triticum0:
Great, hopefully get merged soon, as had few game was going retest once it get merged. Worried when you put it back to draft after all you hard work. Thanks.

coldhex:
I added another commit to work around a Radeon OpenGL bug in Mesa. Geometry shader now emits separate triangles and line segments so that OpenGL/Vulkan implementation's provoking vertex choice doesn't matter for output primitives.

I found this bug in Mesa and created a gist that reproduces it on my system:
https://gist.github.com/coldhex/1744021258fe0af686dbca5b2e5550c4

Expected output is that the right rectangle should have the same colors as the left one, but with Radeon the output is:
radeon-geom-shader-bug

Triticum0:
Fixes #2394 and #2393

hw/xbox/nv2a/pgraph/vk/draw.c
mborgerson: > this now tests and detects geometry shader vertex order during initialization

Great! I like this approach more. Thanks for your patience, will try to review the new revision soon.

Triticum0:
Fixes #1155

abaire:
Likely will fix #2479

antoniodesousa:
Could you rebase the PR @coldhex? Thanks.

coldhex:
Rebased

mborgerson:
Very cool, thanks!

hw/xbox/nv2a/pgraph/vk/draw.c
Triticum0: Did this every get reviewed?

Rastafabisch:
This pull requests introduces a [severe performance](https://github.com/xemu-project/xemu/issues/2506) regression on macOS.

PR #2266, opened 2025-06-17
PPA fix attempt

thrimbor:
I didn't set up my own PPA to test, so I can't guarantee this fully fixes the PPA, but the issue seems to be with the meson wrapper for xxhash which we can bypass by installing the system's version first.
I also added the `static` parameter to avoid a new runtime dependency on the libxxhash package.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -22,6 +22,7 @@ Build-Depends: debhelper (>= 11),
  libslirp-dev,
  libvulkan-dev,
  libcurl4-gnutls-dev,
+ libxxhash-dev,
 Standards-Version: 3.9.8
 Homepage: https://xemu.app
 XS-Debian-Vcs-Browser: https://github.com/mborgerson/xemu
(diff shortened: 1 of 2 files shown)


MasonT8198:
Seems like a reasonable fix, we could probably do just xxhash, seems like it's all their libs https://launchpad.net/ubuntu/plucky/+source/xxhash based on their packaging... another potential problem based on some of the other errors in the build log I attributed to the possible extensive amount of arches in the control list?

I'm not even entirely positive if xemu runs on all of the arches, and pretty sure there are at least a few that definitely won't, fwiw

Edit: possibly wrong on the arches thing I haven't set up my system to confirm every arch

mborgerson:
Thanks for the PR! I checked out the build issue on launchpad:

`../subprojects/xxHash-0.8.3/meson.build:31:3: ERROR: Tried to access unknown option '32-endjmp'.`

It looks like the issue is meson_options.txt being filtered out of the release upload to PPA branch due to xxhash's .gitignore.

To maintain control of this small dependency, I think we should prefer to keep the wrap for this library instead of depending on the distribution package.

I'll close this in favor of fixing the .gitignore issue in https://github.com/xemu-project/xemu/pull/2299

PR #2267, opened 2025-06-17
Add files via upload

LuaXII:
## Summary
This PR enhances the existing auto-updater to support multiple platforms and package formats, improving reliability and cross-platform compatibility.

### 🚀 New Features
- **macOS Support**: Added ZIP archive extraction for macOS updates
- **Linux tar.gz Support**: Added libarchive-based extraction for regular Linux installations  
- **AppImage Support**: Added runtime detection and direct AppImage replacement
- **Dynamic Format Detection**: Automatically detects execution context (AppImage vs regular install)

### 🛠 Improvements
- **Enhanced Error Handling**: Added comprehensive error checking with platform-specific error messages
- **Better Logging**: Improved debug output with consistent formatting
- **Resource Management**: Better cleanup of file handles and archive objects
- **Version Parsing**: Enhanced version string handling with whitespace trimming
- **Progress Reporting**: Fixed progress calculation and cancellation handling

### 🔧 Technical Details
- Conditional compilation using `HAVE_LIBARCHIVE` for graceful degradation
- Proper include path fixes for build system compatibility
- Cross-platform utility functions for file operations
- Thread-safe progress callbacks

## Testing
- [ ] Windows x86_64 ZIP updates
- [ ] Windows aarch64 ZIP updates  
- [ ] macOS x86_64 ZIP updates
- [ ] macOS aarch64 ZIP updates
- [ ] Linux x86_64 tar.gz updates (with libarchive)
- [ ] Linux aarch64 tar.gz updates (with libarchive)
- [ ] Linux AppImage updates
- [ ] Error handling and cancellation

## Compatibility
- Maintains backward compatibility with existing auto-update configuration
- No breaking changes to public API
- Graceful fallback when optional dependencies unavailable

## Related Issues
Fixes #[issue_number] (if applicable)

## How to Test
1. Build on each target platform
2. Test update mechanism with mock server
3. Verify AppImage detection works correctly
4. Test graceful fallback when libarchive missing

### DIFF
--- a/ui/xui/update.cc
+++ b/ui/xui/update.cc
@@ -1,5 +1,5 @@
 //
-// xemu User Interface
+// xemu User Interface - Cross-Platform Auto-Update
 //
 // Copyright (C) 2020-2025 Matt Borgerson
 //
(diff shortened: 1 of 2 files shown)


ui/xui/update.hh
```diff
@@ -42,54 +45,71 @@ typedef enum {
     UPDATER_UPDATE_SUCCESSFUL
 } UpdateStatus;
 
+// Update format types for different platforms
```
abaire: Looks like this is only used as an implementation detail in update.cc and is already defined there. Is there a reason to leak it to the rest of the code?

ui/xui/update.hh
```diff
@@ -42,54 +45,71 @@ typedef enum {
...
 protected:
     int progress_cb(http_progress_cb_info *progress_info);
 
 public:
     Updater();
+    
+    // Status getters
     UpdateStatus get_status() { return m_status; }
     UpdateAvailability get_update_availability() { return m_update_availability; }
+    std::string get_update_version() { return m_latest_version; }
```
abaire: There's no reason to create a copy of the member, return a const ref instead.

ui/xui/update.hh
```diff
@@ -42,54 +45,71 @@ typedef enum {
...
-    UpdateStatus        m_status;
     UpdaterCallback     m_on_complete;
 
 protected:
     int progress_cb(http_progress_cb_info *progress_info);
 
 public:
     Updater();
+    
+    // Status getters
```
abaire: I'll defer to Matt, but generally I don't think xemu uses a lot of code section comments. Personally I'd argue that the divisions are self-evident and just adding a newline between them would give the same amount of info.

If there is a desire to keep them, in my opinion "Status getter" and "state checker" are very similar and could be combined.

ui/xui/update.hh
```diff
@@ -42,54 +45,71 @@ typedef enum {
...
-    std::string get_update_version() { return m_latest_version; }
+    
+    // Actions
     void cancel() { m_should_cancel = true; }
     void update();
-    void update_internal();
     void check_for_update(UpdaterCallback on_complete = nullptr);
+    void restart_to_updated(void);
+    
+    // Internal methods (public for thread access)
```
abaire: Not something you introduced, but do these actually need to be public? Looks like they're used from the worker funcs which are already privileged members of the class?

ui/xui/update.hh
```diff
@@ -42,54 +45,71 @@ typedef enum {
...
+    
     AutoUpdateWindow();
     void CheckForUpdates();
     void Draw();
 };
 
 extern AutoUpdateWindow update_window;
-#endif //_ WIN32
+
+#endif // Platform detection
```
abaire: Did `clang-format` remove the newline at EOF?

ui/xui/update.cc
```diff
@@ -17,36 +17,425 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 //
 #include "common.hh"
-#include "qemu/http.h"
 #include "update.hh"
 #include "viewport-manager.hh"
+
+extern "C" {
```
abaire: At least some of these already handle inclusion from C++ internally, is this actually needed?

ui/xui/update.cc
```diff
@@ -17,36 +17,425 @@
...
+// Update format detection
+typedef enum {
+    UPDATE_FORMAT_ZIP,
+    UPDATE_FORMAT_TAR_GZ,
+    UPDATE_FORMAT_APPIMAGE
+} UpdateFormat;
+
+#if defined(_WIN32)
+    #if defined(__x86_64__)
+        const char *download_url = "https://github.com/xemu-project/xemu/releases/latest/download/xemu-win-x86_64-release.zip";
```
abaire: Are any of these useful outside of this file? Can they all be made `static`?

ui/xui/update.cc
```diff
@@ -17,36 +17,425 @@
...
+        const char *executable_name = "xemu.exe";
+        const char *backup_name = "xemu-previous.exe";
+        const UpdateFormat update_format = UPDATE_FORMAT_ZIP;
+    #elif defined(__aarch64__)
+        const char *download_url = "https://github.com/xemu-project/xemu/releases/latest/download/xemu-win-aarch64-release.zip";
+        const char *executable_name = "xemu.exe";
+        const char *backup_name = "xemu-previous.exe";
+        const UpdateFormat update_format = UPDATE_FORMAT_ZIP;
+    #else
+        #error Unknown Windows architecture
```
abaire: In the hypothetical case that someone attempts to build for an unsupported architecture, do we really want to prevent them from building just because they can't get auto updates?

It seems like it'd be better to spit out a warning message and omit the functionality instead.

ui/xui/update.cc
```diff
@@ -17,36 +17,425 @@
...
+typedef enum {
+    UPDATE_FORMAT_ZIP,
+    UPDATE_FORMAT_TAR_GZ,
+    UPDATE_FORMAT_APPIMAGE
+} UpdateFormat;
+
+#if defined(_WIN32)
+    #if defined(__x86_64__)
+        const char *download_url = "https://github.com/xemu-project/xemu/releases/latest/download/xemu-win-x86_64-release.zip";
+        const char *executable_name = "xemu.exe";
```
abaire: These don't seem to be architecture dependent and a bunch of lines could be saved by moving them down (or up) into the os-level check.

ui/xui/update.cc
```diff
@@ -281,9 +686,40 @@ extern char **gArgv;
...
+    if (update_format == UPDATE_FORMAT_APPIMAGE) {
+        // For AppImage, restart using the APPIMAGE environment variable path
+        const char *appimage_path = getenv("APPIMAGE");
+        if (appimage_path) {
+            DPRINTF("Restarting AppImage: %s\n", appimage_path);
+            execv(appimage_path, gArgv);
+            DPRINTF("AppImage restart failed: %s\n", strerror(errno));
+            exit(1);
+        } else {
+            DPRINTF("APPIMAGE environment variable not found\n");
```
abaire: I'd suggest giving the user instructions on what to do next (I guess just manually restart the app?)

ui/xui/update.cc
```diff
@@ -281,9 +686,40 @@ extern char **gArgv;
...
+            exit(1);
+        }
+    }
+#endif
+    
+    // Regular restart for non-AppImage cases
+    const char *base_path = SDL_GetBasePath();
+    if (!base_path) {
+        DPRINTF("Failed to get base path for restart\n");
```
abaire: Same here, "Please restart xemu" or something like that will hopefully prevent people having to go to Discord to ask. If we never expect this to happen, we could also suggest that they report a bug.

ui/xui/update.cc
```diff
@@ -199,10 +611,10 @@ int Updater::progress_cb(http_progress_cb_info *info)
     if (info->dltotal == 0) {
         m_update_percentage = 0;
     } else {
-        m_update_percentage = info->dlnow*100/info->dltotal;
+        m_update_percentage = (int)((info->dlnow * 100) / info->dltotal);
```
abaire: Aren't these already integer fields?

ui/xui/update.cc
```diff
@@ -159,11 +550,23 @@ void Updater::check_for_update_internal()
...
     } else if (res != 200) {
+        DPRINTF("Version check failed with HTTP status %d\n", res);
         m_status = UPDATER_ERROR;
         goto finished;
     }
 
-    m_latest_version = std::string((const char *)data->data, data->len);
+    // Ensure null-termination for version string
+    g_byte_array_append(data, (const guint8*)"\0", 1);
+    m_latest_version = std::string((const char *)data->data);
```
abaire: nit (and not introduced by you) - C-style cast in C++ code?

ui/xui/update.cc
```diff
@@ -159,11 +550,23 @@ void Updater::check_for_update_internal()
...
+    
+    // Trim whitespace from version string
+    size_t start = m_latest_version.find_first_not_of(" \t\r\n");
+    size_t end = m_latest_version.find_last_not_of(" \t\r\n");
+    if (start != std::string::npos && end != std::string::npos) {
```
abaire: If `find_first_not_of` found something, isn't `find_last_not_of` with the same search string guaranteed to find at least the same character?

ui/xui/update.cc
```diff
@@ -151,6 +540,8 @@ void *Updater::checker_thread_worker_func(void *updater)
 
 void Updater::check_for_update_internal()
 {
+    m_status = UPDATER_CHECKING_FOR_UPDATE;
```
abaire: I know the original issue isn't your code, but I think this still leaves a race condition of `check_for_update` is called in rapid succession. Should we have the status changed before the thread is created so that a subsequent call to `check_for_update` doesn't stomp the first call's thread? It still wouldn't work in the case that the check is invoked from multiple threads, but I think the updater is only ever called from one thread.

Same applies to the `update` func

ui/xui/update.cc
```diff
@@ -17,36 +17,425 @@
...
+            }
+            g_free(dir_path);
+            continue;
+        }
+
+        char *dst_path = g_strdup_printf("%s%s", base_path, fstat.m_filename);
+        DPRINTF("extracting %s to %s\n", fstat.m_filename, dst_path);
+
+        // Handle executable backup
+        if (!strcmp(fstat.m_filename, executable_name)) {
```
abaire: Will this work for signed macOS app bundles?

abaire (commented):
Did a first pass, some high level notes:

1) I think a change is required to main.cc to allow the updater to be called on non win32 platforms. E.g., https://github.com/xemu-project/xemu/blob/4b4bef8bb91ffc26c7cf5001fcc262b842837f54/ui/xui/main.cc#L153

2) The "backup then attempt to overwrite the original" strategy seems backwards to me. I think it'd be safer if we extract everything into a temp directory (verifying that it extracts correctly), then move the original install to a backup, then move/rename the temp data to the original. Also selectively backing up only the xemu binary doesn't feel very future proofed, but I don't know if we need to care right now.

3) Super nitpicky, but this adds a lot of comments that seem fairly self-explanatory and in general I think xemu tends towards reserving comments for things that are likely to be unusually confusing to future maintainers.

ui/xui/update.cc
LuaXII: Yes it should work, just needs thorough testing with different MacOS mechanics

LuaXII:
I read all your comments and made some changes to it

MasonT8198:
Nit: In the past revisions have been required for commit structures that that don't conform to xemu's uniformity style 

ex: update.cc: added MacOS auto-updater

Commits such as that, as well as the title etc

This can be fixed via `git rebase`

mborgerson:
It seems like this PR was generated by a LLM... While it's possible there are some valuable changes here, they are in a sea of other miscellaneous changes with little description of change intent. Because the author has indicated they have not and do not plan to test this PR at all, I don't plan to review it.

PR #2283, opened 2025-06-21
nvnet: Return actual MII status on MMIO reads

Ryzee119:
Since the neat changes in https://github.com/xemu-project/xemu/commit/43c68c22a0d621b40a344e3de0e1fc8b9454c391 I checking link state changes with nxdk.

I was finding the MII status register was just returning zeros even though we are setting the link change bit. This change fixes it.


mborgerson:
Oops! Good catch, thanks!

PR #2291, opened 2025-06-22
nv2a: Implement SET_LINE_WIDTH

### LINKED ISSUE #698: nv2a 0x97 -> 0x0380 (glLineWidth?) unhandled
### Title

### Bug Description

According to the debug output, 0x97 -> 0x0380 is unhandled.
In the [NV20 nouveau sources](https://gitlab.freedesktop.org/mesa/mesa/-/blob/184a690fca700aa9ef2ae2e58b0f4bd86562aa6d/src/mesa/drivers/dri/nouveau/nv20_3d.xml.h#L327), this instruction is `3D_LINE_WIDTH`.

Do note that `glLineWidth` which seems to be the OpenGL counterpart has ([allegedly](https://stackoverflow.com/a/34867791)) always been an optional part of the spec.

### Expected Behavior

The NV2A instruction 0x0380 should be handled.

### xemu Version

0.6.2-55-gcaf84884c4

### System Information

N/A

### Additional Context

_No response_

mborgerson:
Straightforward implementation of line width support. Line widths are clamped to device limits. In the future we'll replace device dependent line-width rendering with a custom rasterization pass, but this is step in the right direction.

FIXME:
- [ ] Log when the device doesn't support full range

Closes #698


hw/xbox/nv2a/pgraph/pgraph.c
```diff
@@ -1532,6 +1523,14 @@ DEF_METHOD(NV097, SET_SHADE_MODE)
     }
 }
 
+DEF_METHOD(NV097, SET_LINE_WIDTH)
+{
+    if (parameter < 0x200) {
```
mborgerson, line 1528: Define a constant

abaire:
Relevant tests: 
* https://github.com/abaire/nxdk_pgraph_tests/blob/94e94c332b089321759c284b86ca995b36fa0ddf/src/tests/line_width_tests.cpp#L22
* https://github.com/abaire/nxdk_pgraph_tests/blob/94e94c332b089321759c284b86ca995b36fa0ddf/src/tests/zpass_pixel_count_tests.cpp#L419

PR #2294, opened 2025-06-23
nv2a/vk: Only include palette in texture key when necessary

mborgerson:
Fixes degraded performance due to garbage palette details polluting the texture cache.


medievil1:
confirmed ogl and vulkan now have identical performance in CT3
on wndows

PR #2295, opened 2025-06-23
nv2a: Shader gen/use refactoring

mborgerson:
Some extensive refactoring to clean up shader gen and use:
   * Reduce shader cache pollution
   * Separates uniform vsh/psh state construction
   * Unifies uniform definitions and update paths


hw/xbox/nv2a/pgraph/shaders.c
```diff
@@ -107,14 +115,35 @@ ShaderState pgraph_get_shader_state(PGRAPHState *pg)
             pgraph_reg_r(pg, NV_PGRAPH_CSV0_C), NV_PGRAPH_CSV0_C_DIFFUSE);
```
abaire: Super nit: Maybe move `state.fixed_function = fixed_function;` from line 96 below the `specular_enable` assignment so there's nothing in between? Or maybe move the specular enable down to line 143 so it's near the other specular state values?

I wonder if it would make things more obvious if we moved all of the fixed function state setters into a helper function (inlinable)? The list is long enough that it is easy to miss this nuance.

Similarly, I wonder if creating a `FixedFunctionState` substruct similar to what we do with `PshState` would simplify keeping things separated?

hw/xbox/nv2a/pgraph/shaders.c
```diff
@@ -107,14 +115,35 @@ ShaderState pgraph_get_shader_state(PGRAPHState *pg)
...
 
     state.separate_specular = GET_MASK(
         pgraph_reg_r(pg, NV_PGRAPH_CSV0_C), NV_PGRAPH_CSV0_C_SEPARATE_SPECULAR);
     state.ignore_specular_alpha = !GET_MASK(
         pgraph_reg_r(pg, NV_PGRAPH_CSV0_C), NV_PGRAPH_CSV0_C_ALPHA_FROM_MATERIAL_SPECULAR);
-    state.local_eye = GET_MASK(
-        pgraph_reg_r(pg, NV_PGRAPH_CSV0_C), NV_PGRAPH_CSV0_C_LOCALEYE);
 
     state.specular_power = pg->specular_power;
     state.specular_power_back = pg->specular_power_back;
```
abaire: Similar to above, maybe we should move the "if (vertex_program)" block on line 185 closer to the "vertex program stuff" on line 152? I also think organization via a helper struct could help here.

abaire (commented):
Good catch; I wonder if we can do some minor reorganization and enforce correctness with automation in the future.

hw/xbox/nv2a/pgraph/shaders.h
```diff
@@ -112,5 +112,6 @@ typedef struct ShaderState {
 typedef struct PGRAPHState PGRAPHState;
```
abaire: If we don't move to a helper struct for fixed-function specific, I think it'd be good to reorganize `ShaderState` to keep all of the fixed-function specific members near each other, ideally with a `/* fixed function */` banner similar to the `/* vertex program */` one that already exists on line 94.

abaire: Also, maybe a doccomment for the struct explicitly covering the "no uniforms" bit?

```
/**
 * Defines the persistent state of the GLSL shaders.
 *
 * This struct should only contain data used to generate GLSL shaders (i.e.,
 * it should not contain any data passed as uniforms to shaders). Data in this
 * struct is used as the cache key for the shader program.
 */
```

and something similar for PshState


```
/**
 * Defines the persistent state of the pixel shader / color combiner.
 *
 * This struct should only contain data used to generate new GLSL shaders (i.e.,
 * it should not contain data passed as uniforms to shaders). Data in this
 * struct is used as part of the cache key for the shader program.
 */
```

hw/xbox/nv2a/pgraph/shaders.c
mborgerson: > I wonder if it would make things more obvious if we moved all of the fixed function state setters into a helper function (inlinable)? ...  creating a FixedFunctionState substruct

Agreed! I had planned to do this

hw/xbox/nv2a/pgraph/shaders.h
mborgerson: Good ideas to improve clarity

> If we don't move to a helper struct ...  ideally with a /* fixed function */ banner similar to the /* vertex program */ one

 I'll prefer the structural separation here for clarity 

> doccomment for the struct explicitly covering the "no uniforms" bit?

I'll consider structural separation and naming improvements first and add the suggested comments if it's still unclear

hw/xbox/nv2a/pgraph/gl/renderer.h
```diff
@@ -96,40 +96,40 @@ typedef struct ShaderBinding {
...
-    GLint clip_region_loc[8];
-
-    GLint point_params_loc[8];
-    GLint material_alpha_loc;
-
-    GLint color_key_loc[4];
-    GLint color_key_mask_loc[4];
+    struct {
+        struct {
+            GLint alpha_ref;
```
abaire: Super nit: The move to the structs is excellent but I worry that dropping the "_loc" suffix on the fields makes it harder to differentiate the bindings from the actual data when grep'ing, which often uses the same name (it's less trivial now to exclude things based on the extra `_`).

hw/xbox/nv2a/pgraph/gl/shaders.c
```diff
@@ -129,89 +130,107 @@ static void update_shader_constant_locations(ShaderBinding *binding)
...
-        binding->vsh_constant_loc[i] = glGetUniformLocation(binding->gl_program, tmp);
+        binding->uniform_locs.vsh.vsh_constant[i] =
+            glGetUniformLocation(binding->gl_program, tmp);
     }
-    binding->surface_size_loc = glGetUniformLocation(binding->gl_program, "surfaceSize");
-    binding->clip_range_loc = glGetUniformLocation(binding->gl_program, "clipRange");
-    binding->depth_offset_loc = glGetUniformLocation(binding->gl_program, "depthOffset");
-    binding->fog_color_loc = glGetUniformLocation(binding->gl_program, "fogColor");
-    binding->fog_param_loc = glGetUniformLocation(binding->gl_program, "fogParam");
+    binding->uniform_locs.psh.surface_size =
```
abaire: nit: these are fragment shader uniforms, move above line 158 or remove the sectional comments?

hw/xbox/nv2a/pgraph/gl/shaders.c
```diff
@@ -129,89 +130,107 @@ static void update_shader_constant_locations(ShaderBinding *binding)
...
+        binding->uniform_locs.vsh.specular_power =
             glGetUniformLocation(binding->gl_program, "specularPower");
     } else {
-        binding->material_alpha_loc = -1;
-        binding->specular_power_loc = -1;
+        binding->uniform_locs.vsh.material_alpha = -1;
+        binding->uniform_locs.vsh.specular_power = -1;
     }
 
     for (int i = 0; i < 4; ++i) {
```
abaire: Same nit with this one; once these comment-delimited sections get long it becomes easy to lose track of them and append new things to the bottom of the function. I think if it's important to keep them separated it'd be easier if they were in separate functions and if it's not important, the comment blocks are unnecessary (and now a bit misleading).

hw/xbox/nv2a/pgraph/glsl/vsh.c
```diff
@@ -156,23 +157,19 @@ MString *pgraph_gen_vsh_glsl(const ShaderState *state, bool prefix_outputs)
...
-        assert(false);
+        pgraph_gen_vsh_prog_glsl(
+            VSH_VERSION_XVS, (uint32_t *)state->programmable.program_data,
+            state->programmable.program_length, header, body);
     }
 
-
     /* Fog */
 
     if (state->fog_enable) {
```
abaire: nit: while you're in here, maybe invert this and move the else condition first since it's short and pushed off the page by the fog gen code?

hw/xbox/nv2a/pgraph/shaders.c
```diff
@@ -41,190 +41,239 @@ static uint32_t get_colorkey_mask(unsigned int color_format)
...
-                                   NV_PGRAPH_CSV0_D_MODE) == 2;
+    assert(i < NV2A_MAX_TEXTURES);
+    uint32_t fmt = pgraph_reg_r(pg, NV_PGRAPH_TEXFMT0 + i*4);
+    unsigned int color_format = GET_MASK(fmt, NV_PGRAPH_TEXFMT0_COLOR);
+    return get_colorkey_mask(color_format);
+}
 
-    bool fixed_function = GET_MASK(pgraph_reg_r(pg, NV_PGRAPH_CSV0_D),
-                                   NV_PGRAPH_CSV0_D_MODE) == 0;
+static FixedFunctionVshState get_fixed_function_vsh_state(PGRAPHState *pg)
```
abaire: nit: Does the struct copy get optimized out reliably? Should this take a pointer to a FFVS struct and populate it rather than construct & return?

(Same question for the other construction funcs)

hw/xbox/nv2a/pgraph/shaders.c
```diff
@@ -41,190 +41,239 @@ static uint32_t get_colorkey_mask(unsigned int color_format)
...
+    memset(&ff, 0, sizeof(ff));
+
+    ff.skinning = (enum VshSkinning)GET_MASK(pgraph_reg_r(pg, NV_PGRAPH_CSV0_D),
+                                             NV_PGRAPH_CSV0_D_SKIN);
+    ff.lighting =
+        GET_MASK(pgraph_reg_r(pg, NV_PGRAPH_CSV0_C), NV_PGRAPH_CSV0_C_LIGHTING);
+    ff.normalization = pgraph_reg_r(pg, NV_PGRAPH_CSV0_C) &
+                       NV_PGRAPH_CSV0_C_NORMALIZATION_ENABLE;
+
+    /* color material */
```
abaire: nit: This feels obvious, similar to the comments you removed in psh.h (I think this applies to all of the sectional comments here)

hw/xbox/nv2a/pgraph/shaders.c
```diff
@@ -235,25 +284,27 @@ ShaderState pgraph_get_shader_state(PGRAPHState *pg)
...
+            color_format ==
+                NV097_SET_TEXTURE_FORMAT_COLOR_LU_IMAGE_DEPTH_X8_Y24_FLOAT;
 
         uint32_t border_source =
             GET_MASK(tex_fmt, NV_PGRAPH_TEXFMT0_BORDER_SOURCE);
         bool cubemap = GET_MASK(tex_fmt, NV_PGRAPH_TEXFMT0_CUBEMAPENABLE);
-        state.psh.border_logical_size[i][0] = 0.0f;
-        state.psh.border_logical_size[i][1] = 0.0f;
-        state.psh.border_logical_size[i][2] = 0.0f;
+        psh.border_logical_size[i][0] = 0.0f;
```
abaire: nit: aren't these already 0'd by the memset?

hw/xbox/nv2a/pgraph/shaders.c
```diff
@@ -41,190 +41,239 @@ static uint32_t get_colorkey_mask(unsigned int color_format)
...
+}
 
-    bool fixed_function = GET_MASK(pgraph_reg_r(pg, NV_PGRAPH_CSV0_D),
-                                   NV_PGRAPH_CSV0_D_MODE) == 0;
+static FixedFunctionVshState get_fixed_function_vsh_state(PGRAPHState *pg)
+{
+    FixedFunctionVshState ff;
+
+    // We will hash it, so make sure any padding is zeroed
+    memset(&ff, 0, sizeof(ff));
```
abaire: Assuming we do construct in place, we could presumably drop these inner memsets since these are all static functions called after the full struct is cleared?

hw/xbox/nv2a/pgraph/vk/shaders.c
```diff
@@ -244,84 +248,88 @@ static void update_shader_constant_locations(ShaderBinding *binding)
...
     }
 
     for (int i = 0; i < NV2A_MAX_TEXTURES; i++) {
         snprintf(tmp, sizeof(tmp), "texScale%d", i);
-        binding->tex_scale_loc[i] =
+        binding->uniform_locs.psh.tex_scale[i] =
             uniform_index(&binding->fragment->uniforms, tmp);
     }
 
     /* lookup vertex shader uniforms */
```
abaire: Same comment as on the GL side re: these section comments being incorrect/misleading due to intermixed use of vsh and psh below.

hw/xbox/nv2a/pgraph/vk/shaders.c
```diff
@@ -482,98 +501,105 @@ static void shader_update_constants(PGRAPHState *pg, ShaderBinding *binding,
...
         int alpha_ref = GET_MASK(pgraph_reg_r(pg, NV_PGRAPH_CONTROL_0),
-                                   NV_PGRAPH_CONTROL_0_ALPHAREF);
-        uniform1i(&binding->fragment->uniforms, binding->alpha_ref_loc,
-                         alpha_ref);
+                                 NV_PGRAPH_CONTROL_0_ALPHAREF);
+        uniform1i(&binding->fragment->uniforms,
+                  binding->uniform_locs.psh.alpha_ref, alpha_ref);
     }
-    if (binding->color_key_loc != -1) {
+    if (binding->uniform_locs.vsh.color_key != -1) {
```
abaire: nit: It feels odd to me to have the mask and key set far apart, should we move this closer to line 581?

abaire (commented):
This will make it much easier to keep the states isolated, big upgrade :)

hw/xbox/nv2a/pgraph/shaders.c
mborgerson: Refactored FF / programmable stuff into separate functions for better logical grouping

hw/xbox/nv2a/pgraph/shaders.h
mborgerson: Implemented structural improvements to resolve this

hw/xbox/nv2a/pgraph/gl/renderer.h
mborgerson: Separated them logically for clarity

hw/xbox/nv2a/pgraph/gl/shaders.c
mborgerson: This is logically separated and automated now so comment no longer applies

hw/xbox/nv2a/pgraph/gl/shaders.c
mborgerson: See above response

hw/xbox/nv2a/pgraph/shaders.c
mborgerson: Yeah, compilers should optimize returns of struct by value, but I've modified this to take a destination pointer to elide redundant padding clears anyway

hw/xbox/nv2a/pgraph/shaders.c
mborgerson: Agreed, comment was removed

hw/xbox/nv2a/pgraph/shaders.c
mborgerson: Should be, but now that I've modified this function to take a destination pointer (see review comment above) I think we should keep this for consistency

hw/xbox/nv2a/pgraph/shaders.c
mborgerson: In this version (before I have moved it to take a destination pointer) we don't have a guarentee the padding of any structures are cleared so this was necessary. But I've elided these memsets by adjusting the functions to init state by pointer

hw/xbox/nv2a/pgraph/vk/shaders.c
mborgerson: Should be resolved now by restructuring

hw/xbox/nv2a/pgraph/vk/shaders.c
mborgerson: Agreed, fixed

hw/xbox/nv2a/pgraph/glsl/vsh.c
mborgerson: Agreed, fixed

hw/xbox/nv2a/pgraph/glsl/common.h
```diff
@@ -21,8 +22,69 @@
...
+
+typedef int ivec4[4];
+typedef float mat2[2 * 2];
+typedef unsigned int uint;
+typedef float vec2[2];
+typedef float vec3[3];
+typedef float vec4[4];
+
+#define UNIFORM_ELEMENT_TYPE_X(DECL) \
+    DECL(float)                      \
```
abaire, line 36: nit: Any reason not to just pre-define all of them (maybe excepting the `double` types since I think we've previously determined that they have serious perf implications on some systems)? 

It'd potentially save a contributor in the future having to jump through several levels of preprocessor indirection to realize that e.g., "bool" or "uvec4" is missing here.

mborgerson: I'd like to avoid defining types that are not actually used. This pattern generates type enumeration and the associated string representations to avoid maintaining the lists in multiple places. If contributor tries to use an undefined type, compilation should fail and this will help us avoid the case when a contributor updates one structure but not the other. In any case, I expect novel type introduction to be a rare event.

Issue #1622, opened 2024-03-28
Dinosaur Hunting - broken water environment mapping (was Fog, dirt particles, and some water appear black)


ButtercupKitten:
### Title

https://xemu.app/titles/4d530069/#Dinosaur-Hunting-Ushinawareta-Daichi

### Bug Description

Fog effects appear black causing some sections with heavy fog use to be nearly unplayable. Some water effects also appear black and so do dirt particles.
![image](https://github.com/xemu-project/xemu/assets/165239401/fa94efba-085a-4d37-ae14-c2731ae3ae59)
![image](https://github.com/xemu-project/xemu/assets/165239401/8743a5e8-79a0-42a4-8293-10b1e32d78aa)
![image](https://github.com/xemu-project/xemu/assets/165239401/462f7e41-1b12-4664-ad84-e107fa6c6222)
![image](https://github.com/xemu-project/xemu/assets/165239401/f6501d31-75b1-4c67-b3f4-f52eba7b3050)

### Expected Behavior

Fog and water are not supposed to appear pitch black and neither is dust
![image](https://github.com/xemu-project/xemu/assets/165239401/4115dfb8-541d-4ec6-ac44-3602d42112a9)
![image](https://github.com/xemu-project/xemu/assets/165239401/ea8f3079-f89a-424c-8a80-329177576999)

### xemu Version

0.7.120

### System Information

CPU:          Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz
OS Platform:  Windows
OS Version:    22H2
Manufacturer: NVIDIA Corporation
GPU Model:    NVIDIA GeForce RTX 3060/PCIe/SSE2
Driver:       4.0.0 NVIDIA 551.86
Shader:       4.00 NVIDIA via Cg compiler

Triticum0:
~~Test with https://github.com/xemu-project/xemu/pull/1780 hopefully should fix the issue~~ Has no affect

Triticum0:
Here a renderdoc https://drive.google.com/file/d/1vfBnTYfvfEvf2lHBU886P7aI-E_u4TIe/view?usp=sharing

Triticum0:
Here a trace

[xemu.zip](https://github.com/user-attachments/files/22414947/xemu.zip)

abaire:
[...]

PR #2298, opened 2025-06-23
nv2a: Clamp vertex fog factor to non-exceptional float range

### LINKED ISSUE #2270: Marvel Nemesis: Rise of the Imperfects: Top of the world stage is almost black
### Title

https://xemu.app/titles/4541038a/#marvel-nemesis-rise-of-the-imperfects

NTSC EA-906 v1.00

### Bug Description

Almost everything in the Top of the World stage render black.

Steps to reproduce:
1 - Start a new game in story mode
2 - Keep playing until you unlock Elektra's story
3 - Play Elektra's Top of the World stage

Image
 

### Expected Behavior

The Top of the World stage should render fine.

Image

### xemu Version

Version: 0.8.73
Branch: master
Commit: 381bcfcf635cb8546383aea946945fef09613293
Date: Tue Jun 17 19:01:23 UTC 2025

### System Information

CPU: AMD Ryzen 9 6900HX with Radeon Graphics        
OS Platform: Windows
OS Version: 24H2
Manufacturer: ATI Technologies Inc.
GPU Model: AMD Radeon RX 6600M
Driver: 4.0.0 Core Profile Context 25.6.1.250522
Shader: 4.60

### Additional Context

Renderdoc of the Top of the World stage:

Vulkan
[Marvel Nemesis - Rise of the Imperfects.zip](https://github.com/user-attachments/files/20784829/Marvel.Nemesis.-.Rise.of.the.Imperfects.zip)

OpenGL
[Marvel Nemesis - Rise of the Imperfects.zip](https://github.com/user-attachments/files/20791756/Marvel.Nemesis.-.Rise.of.the.Imperfects.zip)

abaire:
This addresses HW/driver-dependent handling of  `inf` and `-inf` values. The final clamping to [0,1] is already handled in the pixel shader after interpolation, this clamp range matches observed HW behavior in the fog tests.

Fixes #2270 on systems other than macOS (where the problem does not exist)

[Test results](https://abaire.github.io/xemu-dev_pgraph_test_results/fix_2270_clamp_vtx_fog_output/index.html)

### DIFF
--- a/hw/xbox/nv2a/pgraph/glsl/vsh.c
+++ b/hw/xbox/nv2a/pgraph/glsl/vsh.c
@@ -54,6 +54,8 @@ MString *pgraph_gen_vsh_glsl(const ShaderState *state, bool prefix_outputs)
         GLSL_DEFINE(texMat3, GLSL_C_MAT4(NV_IGRAPH_XF_XFCTX_T3MAT))
 
         "\n"
+        "#define FLOAT_MAX uintBitsToFloat(0x7F7FFFFFu)\n"
+        "\n"
         "vec4 oPos = vec4(0.0,0.0,0.0,1.0);\n"
         "vec4 oD0 = vec4(0.0,0.0,0.0,1.0);\n"
         "vec4 oD1 = vec4(0.0,0.0,0.0,1.0);\n"
@@ -244,7 +246,13 @@ MString *pgraph_gen_vsh_glsl(const ShaderState *state, bool prefix_outputs)
             break;
         }
 
-        mstring_append(body, "  oFog = NaNToOne(vec4(fogFactor));\n");
+        /* Fog is clamped to min/max normal float values here to match HW
+         * interpolation. It is then clamped to [0,1] in the pixel shader.
+         */
+        // clang-format off
+        mstring_append(body,
+                       "  oFog = clamp(NaNToOne(vec4(fogFactor)), -FLOAT_MAX, FLOAT_MAX);\n");
+        // clang-format on
     } else {
         /* FIXME: Is the fog still calculated / passed somehow?!
          */


medievil1:
hmmm.. I thought it was only clamped in PS
and then only if fogenable was turned off
I did a lot of fog stuff on cxbx
```

       // fogging
	const float fogDepth      =   xIn.iFog.x; // Don't abs this value! Test-case : DolphinClassic xdk sampl
	const int   fogTableMode  =   FOGINFO.x;
	const float fogDensity    =   FOGINFO.y;
	const float fogStart      =   FOGINFO.z;
	const float fogEnd        =   FOGINFO.w;

	const int FOG_TABLE_NONE    = 0;
	const int FOG_TABLE_EXP     = 1;
	const int FOG_TABLE_EXP2    = 2;
	const int FOG_TABLE_LINEAR  = 3;

	float fogFactor;
	
	if(FOGENABLE == 0){
	    fogFactor = 1;
	}
	else{
	if(fogTableMode == FOG_TABLE_NONE) 
		fogFactor = fogDepth;  
	if(fogTableMode == FOG_TABLE_EXP) 
		fogFactor = 1 / exp(fogDepth * fogDensity); // 1 / e^(d * density)
	if(fogTableMode == FOG_TABLE_EXP2) 
		fogFactor = 1 / exp(pow(fogDepth * fogDensity, 2)); // 1 / e^((d * density)^2)
	if(fogTableMode == FOG_TABLE_LINEAR) 
		fogFactor = (fogEnd - fogDepth) / (fogEnd - fogStart);
	}

```
	My understanding is fogfactor really just controls the fog alpha (of the current fog color) 0 or 1 would be not transparent or transparent with no in between possibilities
	Not sure why they would have the more complex calculations if they could have just made it less than .5 = 0 greater = 1
	
	 gonna make me revisit my previous stuff and do some experiments...lol

medievil1:
ahhhh NEVERMIND...lol.. I see you found the more specific issue....

abaire:
@antoniodesousa could you check #2270 again once this change finishes building? I modified the clamping range in the vertex shader to match the HW interpolation.

@medievil1 the fog disabled path is handled just below this change and similarly just passes `1`. You're correct that the final clamp to 0-1 is [done in the pixel shader](https://github.com/xemu-project/xemu/blob/1fa242bd005467874bb169247c57450d7b3ddd30/hw/xbox/nv2a/pgraph/glsl/psh.c#L941). The issue this PR attempts to address is that it appears that exceptional values get munged between the vertex and fragment stages, so it's not safe to allow an `inf` to escape the vertex shader.

medievil1:
> @antoniodesousa could you check #2270 again once this change finishes building? I modified the clamping range in the vertex shader to match the HW interpolation.
> 
> @medievil1 the fog disabled path is handled just below this change and similarly just passes `1`. You're correct that the final clamp to 0-1 is [done in the pixel shader](https://github.com/xemu-project/xemu/blob/1fa242bd005467874bb169247c57450d7b3ddd30/hw/xbox/nv2a/pgraph/glsl/psh.c#L941). The issue this PR attempts to address is that it appears that exceptional values get munged between the vertex and fragment stages, so it's not safe to allow an `inf` to escape the vertex shader.

yes, You updated and changed to handling INF while I was typing :P

antoniodesousa:
> @antoniodesousa could you check https://github.com/xemu-project/xemu/issues/2270 again once this change finishes building? I modified the clamping range in the vertex shader to match the HW interpolation.

It's still fixed!

xemu-2025-06-23-22-22-14

Triticum0:
Doesn't fixed https://github.com/xemu-project/xemu/issues/1622 will add a renderdoc later

hw/xbox/nv2a/pgraph/glsl/vsh.c
```diff
@@ -244,7 +244,13 @@ MString *pgraph_gen_vsh_glsl(const ShaderState *state, bool prefix_outputs)
             break;
         }
 
-        mstring_append(body, "  oFog = NaNToOne(vec4(fogFactor));\n");
+        /* Fog is clamped to min/max normal float values here to match HW
+         * interpolation. It is then clamped to [0,1] in the pixel shader.
+         */
+        // clang-format off
+        mstring_append(body,
+                       "  oFog = clamp(NaNToOne(vec4(fogFactor)), -3.4028234663852886e+38, 3.4028234663852886e+38);\n");
```
mborgerson: These long constants are a bit cumbersome. How about we define them as constants, and use something like  `uintBitsToFloat(0x7F7FFFFFu) /* +FLT_MAX */` for clarity?

abaire: That was my initial inclination too, but I was not sure if the bits to float conversion has a runtime cost or if it'd be optimized out. My thought was that having the conversion done once at xemu compile time was simplest. We could keep the literal value but give it a symbolic name either in C or GLSL, or we could just not prematurely optimize :)

abaire: Done

hw/xbox/nv2a/pgraph/glsl/vsh.c
```diff
@@ -53,6 +53,9 @@ MString *pgraph_gen_vsh_glsl(const ShaderState *state, bool prefix_outputs)
         GLSL_DEFINE(texMat2, GLSL_C_MAT4(NV_IGRAPH_XF_XFCTX_T2MAT))
         GLSL_DEFINE(texMat3, GLSL_C_MAT4(NV_IGRAPH_XF_XFCTX_T3MAT))
 
+        "\n"
+        "#define FLOAT_MIN uintBitsToFloat(0xFF7FFFFFu)\n"
```
mborgerson: nit: `FLOAT_MIN` or `FLT_MIN` in [other contexts](https://learn.microsoft.com/en-us/cpp/cpp/floating-limits?view=msvc-170) means smallest positive number representable. We should probably rename this to something like `FLOAT_MAX_NEG` or use `-FLOAT_MAX` below

hw/xbox/nv2a/pgraph/glsl/vsh.c
```diff
@@ -244,7 +247,13 @@ MString *pgraph_gen_vsh_glsl(const ShaderState *state, bool prefix_outputs)
             break;
         }
 
-        mstring_append(body, "  oFog = NaNToOne(vec4(fogFactor));\n");
+        /* Fog is clamped to min/max normal float values here to match HW
```
mborgerson, line 249: nit: I'm not sure how much value this comment adds over just reading the code

hw/xbox/nv2a/pgraph/glsl/vsh.c
abaire: Done

hw/xbox/nv2a/pgraph/glsl/vsh.c
abaire: Normally I would agree with dropping comments, but in this particular case I worry somebody might decide to optimize away what looks like an invalid clamp based on the GL spec or an unnecessary one based on knowledge that the fragment shader is going to re-clamp anyway.

I guess this could be changed to something like "Exceptional fog factor values are fixed prior to interpolation, the final GL clamp to [0, 1] is done in the pixel shader." to reduce a bit of the redundancy?

mborgerson:
Thanks!

PR #2303, opened 2025-06-24
use jammy instead of noble to support LTS and bookworm among other systems

thelamer:
Ran into this today trying to run on current Debian Bookworm which is `Debian GLIBC 2.36-9+deb12u10` , building on Jammy greatly expands compatibility.


mborgerson:
Thanks!

PR #2319, opened 2025-07-01
ui: Add render scale and backend to compat reports

abaire:


ui/xui/compat.cc
```diff
@@ -91,6 +87,23 @@ void CompatibilityReporter::Draw()
         description[0] = '\x00';
         report.compat_comments = description;
 
+        char render_scale_buffer[8];
+        snprintf(render_scale_buffer, sizeof(render_scale_buffer), "%dx",
+                 nv2a_get_surface_scale_factor() - 1);
```
mborgerson: Maybe copied from other UI code? The `-1` elsewhere is for translating to lookup table, but here this will result in `0x` being printed

ui/xui/reporting.hh
```diff
@@ -41,15 +41,17 @@ public:
 	std::string compat_rating;
 	std::string compat_comments;
 	std::string xbe_headers;
+        std::string rendering_scale;
```
mborgerson: Unusual indentation?

ui/xui/reporting.cc
```diff
@@ -57,6 +49,8 @@ const std::string &CompatibilityReport::GetSerializedReport()
 		{"compat_rating", compat_rating},
 		{"compat_comments", compat_comments},
 		{"xbe_headers", xbe_headers},
+                {"rendering_scale", rendering_scale},
```
mborgerson: * Indentation doesn't match surrounding
* I'll have to update the backend to support saving these fields

ui/xui/compat.cc
abaire: Yup, copied from UI code. Fixed.

ui/xui/reporting.cc
abaire: Weird, once again this looks fine in the GH UI for me. Looks like the file is using tabs. What do you have your tabs set to so I can catch these before they're a PR issue (at least until we run clang-format and get rid of them)?

Screenshot 2025-07-01 at 18 30 13

ui/xui/reporting.hh
abaire: Fixed, tabs once again.

ui/xui/reporting.cc
mborgerson: Huh, weird. GitHub on my end was showing a 1 space tab. Locally I usually use tab=4 spaces. Yeah, we'll clang-format and convert the tabs to spaces at some point.

MasonT8198 (approved):
Took another look at this and it appears to work correctly rebased to correct on my end with the example compat report working correctly (had to do some build hackery to get the reporter to appear since it's unofficial, but it does look good!)

Issue #1011, opened 2022-06-02
Framerate caps on xemu doesn't work correctly


Triticum0:
### Bug Description

On Some games, the Frames caps don't hit a consistent 30. they can go past 30 sometimes and mostly hover all over the place making the frame time really inconsistent.

### Games Affect 
https://xemu.app/titles/54430009/#Gungriffon-Allied-Strike
https://xemu.app/titles/544d0009/#SeaBlade

### Expected Behavior

Frame time should be more stable and not go over the Xbox frame cap

### xemu Version

Version: 0.7.25
Branch: master
Commit: https://github.com/mborgerson/xemu/commit/7d6da227ba3b224647e35d2385419a3bec25a3b3
Date: Tue May 31 08:20:36 PM UTC 2022

### System Information

Field | Value
-- | --
OS | Windows 10
CPU | AMD Ryzen 5 2600 Six-Core Processo
Graphics Device | NVIDIA GeForce RTX 3060 Ti/PCIe/SSE2
Graphics Driver | 4.0.0 NVIDIA 512.95

Issue #2399, opened 2025-08-23
Adjustable framerate cap


Avid29:
### Feature Request

This would serve largely as a debug tool or soft patch solution to timing issues.

For example, I ensured that the playback issues in #2017 can be resolved by lowering the framerate, and this would also presumably give a work around for issues resolving in #1099, which previously required changing the Windows display refresh rate and enabling VSync. 

Currently it is technically capped at 60hz.

### Alternatives

Like mentioned above, one could set the framerate using VSync and changing their Windows refresh rate, but this is high inconvenient and has many side effects.

Alternatively, all timing in the emulator could be adjusted to match the original hardware. Much easier said than done.

### Additional Context

I've already implemented this in a [personal branch](https://github.com/Avid29/xemu/tree/fps_limiter). Here's a video:

https://github.com/user-attachments/assets/653de648-34cc-46c7-947a-db2fcb40a502

ChaseDerick:
This looks interesting, just to double check this doesn't bypass any hardset FPS limits in games I assume right? Also looking at the options under your github, I would like it if it could be more varied in terms of FPS choices (0,5,10,15,20,25,30,45,60,90,120,144) or possibly a custom option might be cool to see.

abaire:
Capping unnecessarily high refresh in the UI layer makes sense (i.e., there's no reason to update menus and redraw the guest framebuffer at 120hz for games that generally run at 30-60) but the existing code should already be locking UI updates to around 60fps.

Any games that are running too fast should be fixed at the guest level, since they wouldn't have run that fast on original HW. The guest framerate is supposed to be independent of the host/UI refresh, since you don't want a game that is tanking to prevent menu interactions to kill it. I believe most of the contributions over the last few years have focused more on fixing missing functionality rather than performance optimization, when feature additions start to wind down there will be more focus on optimizing things which should improve many games that run slower than they should; there's still a lot of headroom for improvement.

I think this request duplicates #1738 and #1887 (and #2167 to a certain degree)
I think the triggering issue of titles needing a limit is probably covered by #1013 and various other title-specific issues where things run too fast/too slow.

Avid29:
@abaire Yeah... I didn't do my homework looking into existing issues before creating this one. Just wanted something to explicitly pair with the PR I'm gonna create.

The code I replaced capped the refresh rate at 60hz. This is only a cap, if the game enforces its own refresh rate with sleeps or spins this has no ability to override it.

[...]

Issue #2017, opened 2025-03-15
Oddworld: Munch's Oddyssee: Sped up video cutscenes


KreeseXD:
### Title

https://xemu.app/titles/4d530001/#Oddworld-Munch-s-Oddysee

NTSC MS-001 v1.00

### Bug Description

The game's video cutscenes seem to be running at a really fast pace, not native. Attached is a video of the intro, but all cutscenes that are basically videos run into this issue. Audio seems to be sped up a bit too, not sure if they are interlinked together

https://github.com/user-attachments/assets/3fd5217b-85cf-49af-9060-cc8b4a40b384

### Expected Behavior

Videos should be running at native speed and not sped up
The first 32 seconds of this video show the speed it should be playing at: 
https://youtu.be/WM35GjwU2Ks?si=80fg_I9buKBkqzga

### xemu Version

0.8.35

### System Information

OS: Windows 11
CPU: 13th Gen Intel(R) Core(TM) i9-13980HX
GPU: NVIDIA GeForce RTX 4080 Laptop GPU/PCIe/SSE2
GPU Driver: 4.0.0 NVIDIA 566.14

KreeseXD:
Maybe related to this https://github.com/xemu-project/xemu/issues/1099

I can't find a way to force my laptop display to output to 30hz to test

Avid29:
@KreeseXD I've just made a [branch on my fork](https://github.com/Avid29/xemu/tree/fps_limiter) that allows you to specify an FPS cap. I can confirm that the framerate directly effects the playback rate

https://github.com/user-attachments/assets/aad545d7-73b9-4593-b568-33e07a959793

Issue #1099, opened 2022-06-25
Meta: Timing Issues


Triticum0:
### Bug Description

Some games have timing issues making them run too fast or too slow

### Games Affected
~Antz-Extreme-Racing~ (Too fast)
~https://xemu.app/titles/4d530010/#Blood-Wake~ (Too fast)
~The-Da-Vinci-Code~ (Too fast cutscences)
https://xemu.app/titles/41540001/#Touge-R (Too Slow)
~Dance-UK~ (Too fast)
https://xemu.app/titles/4b4e0006/#ESPN-NBA-2Night-2002 (Too Slow, Half speed)
https://xemu.app/titles/544d0001/#Shrek (Too fast)
~https://xemu.app/titles/42570001/#Robot-Wars-Extreme-Destruction~ (Opening Intro's Too fast)

### Expected Behavior

Should play at their intended speed

### xemu Version

Version: 0.7.49
Branch: master
Commit: https://github.com/mborgerson/xemu/commit/b3abb982e82edbbb142a2c275b893e7f62b10661
Date: Sat Jun 25 06:50:26 UTC 2022

### System Information

Field | Value
-- | --
OS | Windows 10
CPU | AMD Ryzen 5 2600 Six-Core Processo
Graphics Device | NVIDIA GeForce RTX 3060 Ti/PCIe/SSE2
Graphics Driver | 4.0.0 NVIDIA 512.95

### Additional Context

True Crime: New York City cut-scene issue was issue with the dump.

Triticum0:
Note: [Touge-R](https://xemu.app/titles/41540001/#Touge-R) Seem to be fixed by using the real-time DSP option in the menu.
Also fixes V-rally 3 half speed issue when real time dsp turned on.

MA-121:
A workaround for Bloodwake to set the refresh rate of your monitor to 30Hz and enable VSYNC in Xemu. I haven't tested it for the other games, but it works for Bloodwake since the physics are tied to the framerate.

e.g. On Windows 11 it's under Settings->System->Display->Advanced Display->Choose a refresh rate->30Hz

To enable VSYNC on Xemu go to Machine->Settings->Display->Vertical refresh sync

mrdc:
Dead or Alive 3 is also affected on macOS arm64 - slow motion.
https://xemu.app/titles/4d53002d#Dead-or-Alive-3

PR #2400, opened 2025-08-24
Adjustable display refresh rate (in the UI layer)

Avid29:
This PR replaces the current 60Hz refresh rate cap in the UI layer with a variable cap specified by the config.

This PR is paired to issue #2399 but also provides a hot patch to many of the over-speeding issues. This provides a definitive work-around for #2017 that can be used while watching cut scenes. It also may provide a solution to some of the issues experienced in #1099.

Here are screenshots of the UI changes to add this control.

Options in MenuBar: (If you look at the fps graph you can see I was changing it too)
image

Options in Settings: (The custom FPS slider is hidden when the FPS is set to anything besides custom)
image


Triticum0:
only fixes Blood Wake other game are unaffected. Note only tested Antz Extreme Racing

Triticum0:
Would fix #1011

Avid29:
FYI, there's a force push because I revised the commit names, applied clang formatting, and merged a few commits that accomplished very similar things

Calinou:
Out of curiosity, how does this compare to the display rate override setting from https://github.com/xemu-project/xemu/pull/2028?

Avid29:
> Out of curiosity, how does this compare to the display rate override setting from #2028?

@Calinou I was unaware of that code what I wrote my version. I did look over it afterwards though. 

The display rate adjustment is performed in the same region of code. 

#2028 uses the current default of 60fps as a baseline and introduces an adjusting coefficient to scale the frame expiration duration on each iteration. My approach replaces the constant expiration time of `16666666` with the variable duration of `min_frame_duration`, which is calculated as `(float)NANOSECONDS_PER_SECOND / frame_rate`. The framerate setting is stored as an integer.

They also chose to store the setting as a reciprocal of the coefficient and calculate both the reciprocal and multiplication of the time scaling on each frame. This isn't really expensive, but it is unnecessary and weird. Sorry to be snobby about meaningless optimization and semi-arbitrary code quality.

PR #2433, opened 2025-09-16
nv2a: Fix fixed point NV_PVIDEO_POINT_IN handling

### LINKED ISSUE #2432: PVIDEO handling of `NV_PVIDEO_POINT_IN` is incorrect (should be interpreted as fixed point)
### Bug Description

Current xemu treats the values passed via `NV_PVIDEO_POINT_IN` as whole texel offsets into the overlay buffer. In fact, these values are x.4 and y.3 fixed point values. In cases where in offset is used, this will result in significantly larger shifts when compared to HW.

See https://github.com/abaire/nxdk_pgraph_tests/blob/c047da43bb6b6d8b011310204cae6eb6e77b2c5a/src/tests/pvideo_tests.cpp#L1063

### Expected Behavior

Behavior should match HW. In the case of the PVIDEO tests, the checkerboard pattern should be advanced 2 or 1 boxes rather than the current behavior which shifts many many more.

HW results:
https://youtube.com/clip/UgkxbWRtAyBgfevUoZeSp1OgpZdp2ilgwBHH?si=_dJsdGNG6d186-sa

### xemu Version

N/A affects all xemu versions.

### System Information

N/A affects all platforms.

### Additional Context

_No response_

abaire:
Tests: https://github.com/abaire/nxdk_pgraph_tests/blob/c047da43bb6b6d8b011310204cae6eb6e77b2c5a/src/tests/pvideo_tests.cpp#L1063

Fixes #2432

### DIFF
--- a/hw/xbox/nv2a/pgraph/gl/display.c
+++ b/hw/xbox/nv2a/pgraph/gl/display.c
@@ -269,7 +269,7 @@ static void render_display_pvideo_overlay(NV2AState *d)
                  GL_UNSIGNED_BYTE, tex_rgba);
     g_free(tex_rgba);
     glUniform1i(r->disp_rndr.pvideo_tex_loc, 1);
-    glUniform2f(r->disp_rndr.pvideo_in_pos_loc, in_s, in_t);
+    glUniform2f(r->disp_rndr.pvideo_in_pos_loc, in_s / 16.f, in_t / 8.f);
     glUniform4f(r->disp_rndr.pvideo_pos_loc,
                 out_x, out_y, out_width, out_height);
     glUniform3f(r->disp_rndr.pvideo_scale_loc,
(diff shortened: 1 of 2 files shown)


abaire:
Note: I don't have a VK capable platform, so this is only tested with the GL renderer.

antoniodesousa:
Do you know what games are affected by this? Are games where the FMV aspect ratio is wrong? I know that Fatal Frame is one of them.

abaire:
Nope, quite possibly no games are affected. I would guess this functionality would be most useful for stimulating a scrolling effect, so maybe titles where a background scrolls 8-16 times faster than it should. 

There are issues with input size and ratio handling that could explain aspect issues, they're tracked separately.

mborgerson:
Thanks!

PR #2472, opened 2025-10-02
nv2a: Handle 2D textures given to 3D texture modes

### LINKED ISSUE #1622: Dinosaur Hunting - broken water environment mapping (was Fog, dirt particles, and some water appear black)

abaire:
Hardware transparently supports passing 2D textures to operations that are intended to operate on cubemaps. This change introduces a remapping function to mimic the hardware behavior in cases where the texture sampler does not match the expectations of the texture mode.

Fixes #1622

Tests: https://github.com/abaire/nxdk_pgraph_tests/blob/4a1fde90854f318a4725178c9c6a7f1fcd016d05/src/tests/texture_2d_as_cubemap_tests.cpp#L79
HW results: https://abaire.github.io/nxdk_pgraph_tests_golden_results/results/Texture_2D_as_cubemap/index.html

### DIFF
--- a/hw/xbox/nv2a/pgraph/glsl/psh.c
+++ b/hw/xbox/nv2a/pgraph/glsl/psh.c
@@ -134,6 +134,7 @@ void pgraph_glsl_set_psh_state(PGRAPHState *pg, PshState *state)
         uint32_t border_source =
             GET_MASK(tex_fmt, NV_PGRAPH_TEXFMT0_BORDER_SOURCE);
         bool cubemap = GET_MASK(tex_fmt, NV_PGRAPH_TEXFMT0_CUBEMAPENABLE);
+        state->tex_cubemap[i] = cubemap;
         state->border_logical_size[i][0] = 0.0f;
         state->border_logical_size[i][1] = 0.0f;
         state->border_logical_size[i][2] = 0.0f;
@@ -607,13 +608,13 @@ static const char *get_sampler_type(struct PixelShader *ps, enum PS_TEXTUREMODES
         return NULL;
 
     case PS_TEXTUREMODES_PROJECT2D:
-        if (state->dim_tex[i] == 2) {
+        if (dim == 2) {
             if (state->tex_x8y24[i] && ps->opts.vulkan) {
                 return "usampler2D";
             }
             return sampler2D;
         }
-        if (state->dim_tex[i] == 3) return sampler3D;
+        if (dim == 3) return sampler3D;
         assert(!"Unhandled texture dimensions");
         return NULL;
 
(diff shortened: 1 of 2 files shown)


abaire:
@Triticum0 can you see if this fixes #1622 when you have a chance? It at least passes the tests I wrote based on the behavior in that bug.

Triticum0:
I tested it, seem to renderer the water texture correctly so you can see it now but black blotches  on water still rendered on top. Still not Internet so can't help debug. On masters water texture doesn't render at all so you did fix one bug. Main issue still remains have tested other fog in level and all still remain the same, pr only fixes cubemap Reflection. When though most of first level and only cubemap water Reflection on the map. Guessing fog issue properly just fog gen on calculation carryover

abaire:
Thanks for testing! 

> ... Main issue still remains have tested other fog in level and all still remain the same, pr only fixes cubemap Reflection.

I think we should split the bug up into separate issues since it seems there are a few failures going on. I think we should look at the shaders before we can decide whether it's another register carryover issue or something novel.

I believe the renderdoc attached to the existing issue demonstrates the water reflection mapping, so maybe we make #1622 just about the water and create a new issue for the fog problem (with a new RenderDoc capture that demonstrates the problem)?

Triticum0:
Good idea but don't know when my Internet will be probably sometime Monday ask around incase anyone one can create issue for you or send you a better renderedoc only used that for the issues as quickest to get to from a fresh save. Did see another minor issue but save it for another time. Great work as always.

antoniodesousa:
I can get a renderdoc of that game. You want me to use this PR? Also, you need a pgraph event trace?

abaire:
> I can get a renderdoc of that game. You want me to use this PR? Also, you need a pgraph event trace?

That'd be very much appreciated, thanks!

Ideally two renderdocs:

1 with this PR that shows the water issue from the bug, ideally wherever this is so it's the same as the other capture: ideally wherever this is

Then another that shows the fog issue. That one doesn't need this PR and ideally would have the PGRAPH trace along with the RenderDoc capture (if you use my `work` branch it'll be close to master and will label the draw events in the RDC to match the pgraph trace, which makes life much easier).

If you could post them on #1622 rather than here we can make sure they don't get lost; I don't expect to fix both issues in one PR.

antoniodesousa:
Oh yeah, I forgot that renderdoc is not enabled by default in other builds except your work branch... well, I guess master it is.

abaire:
> Oh yeah, I forgot that renderdoc is not enabled by default in other builds except your work branch... well, I guess master it is.

The artifact from https://github.com/abaire/xemu/pull/31 will give you this PR on top of my work branch for the first capture. I want to see if the black blotches are a problem with this change or if they're another issue.

antoniodesousa:
> If you could post them on https://github.com/xemu-project/xemu/issues/1622 rather than here we can make sure they don't get lost; I don't expect to fix both issues in one PR.

Done! This game has several issues. The PR only seems to fix the water reflection. The fog and smoke effects are still broken in some specific cases.

abaire:
Cool, that's aligned with what I expected based on @Triticum0's feedback, thanks for confirming that they're separate bugs.

I added a response to your comment, I think something was wrong with the PR/water env mapping capture, but I think there's enough info to say that this change seems to fix the issue. I'll split #1622 into one or two new issues once I can confirm whether the smoke and fog issues are distinct or not.

mborgerson:
Thanks!

PR #2490, opened 2025-10-25
nv2a: Fix macOS crash due to glo context create off main thread

### LINKED ISSUE #2489: Commit 36e4772996d212c46f4c2d5d210d10ae6aaaf405 breaks xemu on macOS
### Bug Description

The latest xemu release dies creating an SDL window

Seems that this is caused by commit 36e4772996d212c46f4c2d5d210d10ae6aaaf405 from #2240, syncing to b2ded13f6b91e52b93f4fc8741269c6c8ca18397 works. 

Specifically it is the context creation triggered by the new `pgraph_gl_determine_gpu_properties` that is dying as it is called off the main thread:

`glo_context_create: Failed to create window NSWindow should only be instantiated on the main thread!`

### Expected Behavior

xemu should still run on macOS without a fatal SDL error.

### xemu Version

Master branch build at v0.8.109

### System Information

M3 MacBook Pro
macOS 15.7.1

### Additional Context

_No response_

abaire:
Fixes #2489

### DIFF
--- a/hw/xbox/nv2a/pgraph/gl/gpuprops.c
+++ b/hw/xbox/nv2a/pgraph/gl/gpuprops.c
@@ -338,8 +338,6 @@ void pgraph_gl_determine_gpu_properties(void)
     const int width = 640;
     const int height = 480;
 
-    GloContext *g_context = glo_context_create();
-
     uint8_t *pixels = render_geom_shader_triangles(width, height);
     determine_triangle_winding_order(pixels, width, height,
                                      &pgraph_gl_gpu_properties);
@@ -350,8 +348,6 @@ void pgraph_gl_determine_gpu_properties(void)
             pgraph_gl_gpu_properties.geom_shader_winding.tri_strip0,
             pgraph_gl_gpu_properties.geom_shader_winding.tri_strip1,
             pgraph_gl_gpu_properties.geom_shader_winding.tri_fan);
-
-    glo_context_destroy(g_context);
 }
 
 GPUProperties *pgraph_gl_get_gpu_properties(void)
--- a/hw/xbox/nv2a/pgraph/gl/gpuprops.c
+++ b/hw/xbox/nv2a/pgraph/gl/gpuprops.c
@@ -333,7 +333,7 @@ static void determine_triangle_winding_order(uint8_t *pixels, int width,
     props->geom_shader_winding.tri_fan = (fan_rot + 2) % 3;
 }
 
-void pgraph_gl_determine_gpu_properties(NV2AState *d)
+void pgraph_gl_determine_gpu_properties(void)
 {
     const int width = 640;
     const int height = 480;
--- a/hw/xbox/nv2a/pgraph/gl/gpuprops.c
+++ b/hw/xbox/nv2a/pgraph/gl/gpuprops.c
@@ -339,7 +339,6 @@ void pgraph_gl_determine_gpu_properties(void)
     const int height = 480;
 
     GloContext *g_context = glo_context_create();
-    glo_set_current(g_context);
 
     uint8_t *pixels = render_geom_shader_triangles(width, height);
     determine_triangle_winding_order(pixels, width, height,
@@ -353,7 +352,6 @@ void pgraph_gl_determine_gpu_properties(void)
             pgraph_gl_gpu_properties.geom_shader_winding.tri_fan);
 
     glo_context_destroy(g_context);
-    glo_set_current(g_nv2a_context_render);
 }
 
 GPUProperties *pgraph_gl_get_gpu_properties(void)
--- a/hw/xbox/nv2a/pgraph/gl/renderer.c
+++ b/hw/xbox/nv2a/pgraph/gl/renderer.c
@@ -31,6 +31,8 @@ static void early_context_init(void)
 {
     g_nv2a_context_render = glo_context_create();
     g_nv2a_context_display = glo_context_create();
+
+    pgraph_gl_determine_gpu_properties();
 }
 
 static void pgraph_gl_init(NV2AState *d, Error **errp)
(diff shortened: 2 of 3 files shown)


abaire:
Not sure if we should do the same in the `HAVE_EXTERNAL_MEMORY` guarded sections of the vk path or not. Doesn't feel as urgent since VK isn't supported on macOS at the moment anyway.

Edit: I went ahead and did it under the assumption that it won't hurt anything and might prevent confusion if moltenvk ends up working for xemu.

hw/xbox/nv2a/pgraph/vk/renderer.c
```diff
@@ -63,7 +65,16 @@ static void pgraph_vk_init(NV2AState *d, Error **errp)
     pgraph_vk_update_vertex_ram_buffer(&d->pgraph, 0, d->vram_ptr,
                                    memory_region_size(d->vram));
 
+#if HAVE_EXTERNAL_MEMORY
```
mborgerson: `pgraph_vk_determine_gpu_properties` doesn't use any GL API, so we should not need to provide/manage an additional GL context for it

hw/xbox/nv2a/pgraph/gl/gpuprops.c
```diff
@@ -333,13 +333,13 @@ static void determine_triangle_winding_order(uint8_t *pixels, int width,
     props->geom_shader_winding.tri_fan = (fan_rot + 2) % 3;
 }
 
-void pgraph_gl_determine_gpu_properties(NV2AState *d)
+void pgraph_gl_determine_gpu_properties(NV2AState *d, GloContext **context)
```
mborgerson: Context management feels a little awkward here (e.g. created elsewhere, destroyed/cleared here, global unused after startup). Since this function doesn't actually need NV2AState, how about we just call this from `early_context_init` to ensure it happens on the main thread and avoid the extra management?

hw/xbox/nv2a/pgraph/vk/renderer.c
abaire: Removed

hw/xbox/nv2a/pgraph/gl/gpuprops.c
abaire: Done. The asymmetry makes me worry that somebody may "fix" this in the future, should we add a comment indicating that it has to be called from early_context_init?

hw/xbox/nv2a/pgraph/gl/gpuprops.c
mborgerson: Let's drop this call to `glo_set_current`

hw/xbox/nv2a/pgraph/gl/renderer.c
```diff
@@ -31,6 +31,8 @@ static void early_context_init(void)
 {
     g_nv2a_context_render = glo_context_create();
     g_nv2a_context_display = glo_context_create();
+
+    pgraph_gl_determine_gpu_properties();
```
mborgerson, line 40: Let's move this up to be the first thing called in the function

hw/xbox/nv2a/pgraph/gl/gpuprops.c
mborgerson: Good idea, sure. Not necessarily from `early_context_init`, but rather from the main thread

hw/xbox/nv2a/pgraph/gl/gpuprops.c
abaire: Done

hw/xbox/nv2a/pgraph/gl/renderer.c
abaire: Unfortunately moving this before the other contexts are created leads to a red outlined Xbox controller being visible on the screen; something must be bleeding over? I haven't started looking into why; I thought the whole point of glo was that it rendered to an offscreen/invisible window so I'm surprised the presented behavior would change.

EDIT: It also persists after startup, preventing any actual title data from being displayed.
Screenshot 2025-10-27 at 13 59 45



Leaving it last and letting the context be NULL after the destroy seems to work, at least on macOS. Out of caution, I'm setting it to the display context, which looks like it should match the state created by the previous glo_context_create.

mborgerson: The GL contexts we create are "shared", so we can access textures initialized in one context (i.e. rendering context) from another context (i.e. display context). I suspect creating a context and destroying it left the thread with no context to share with subsequently created contexts and the display thread was now rendering with identifiers from an unshared context (in this case, selecting the controller interface texture). Let's just keep the call to pgraph_gl_determine_gpu_properties at the end in this case

mborgerson: Hmm.. This may also interfere with subsequent renderers which also initialize "shared" contexts with the main display context (e.g. breaking Vulkan renderer if it is initalized next). To avoid this interference we'll want to preserve the original context set in the call to pgraph_gl_determine_gpu_properties

abaire: In that case, should we just move context management out of `pgraph_gl_determine_gpu_properties` entirely and have it be owned by the caller?

abaire: It also looks like we don't do anything special to clean up the glo contexts, I guess we could just create the context in `pgraph_gl_determine_gpu_properties` and leak it? Maybe add a static var to assert that `pgraph_gl_determine_gpu_properties` is only called once.

abaire: Moved the context management out of `pgraph_gl_determine_gpu_properties`. 

I think as long as we create any other shared context prior to the gpu props context we can still destroy it and reset to the previous value.

If there was a way to get the current `GloContext` inside of the gpu_props method it'd make sense to do everything there (though w/ the main thread req), but having it hardcoded to reset to `g_nv2a_context_display` feels a bit fragile/opaque to me.

mborgerson: Works for me!

mborgerson:
Thanks! Made some slight adjustment to commit message

PR #2518, opened 2025-11-05
nv2a: Add VK physical device to config

abaire:
Just doing the bare minimum here to unblock anybody with unfortunate device ordering (like the VK machine I briefly had access to for testing). 

Eventually it'd be good to add a settings UI for this and maybe switch to device IDs instead of names; I'm not sure what gets displayed in a system with two identical GPUs. I could have it dump the device ID when it enumerates available devices so that users could cut and paste that into the config, if preferred.


mborgerson:
Thanks!

PR #2528, opened 2025-11-08
nv2a: Prevent GL assert on END without BEGIN

abaire:
Fixes an assert case in debug mode by ignoring end's without begins (similar to begin without end).


mborgerson:
Thanks!

PR #2547, opened 2025-11-16
ui/xui: Add "Recent" under Machine to keep track of recently loaded disc images

### LINKED ISSUE #2544: Recent files tab
### Feature Request

Add a recent files tab to the top-bar. Would save a bit of time searching for iso files.

### Alternatives

_No response_

### Additional Context

Potentially add a "recent files" (xiso files) to the top-bar for xemu.

ehw:
Disclaimer: I had help in implementing this with minimax ai.

Adds an item under Machine in the menu bar at the top that lists the last 10 recently opened disc images (let me know if you want more, could go a little higher). If a disc image is deleted, it's greyed out so the end user can't click on it and open it by mistake. An option to clear history is added to the bottom if the user wants to remove all the entries without editing their xemu.toml file. Entries are saved in the xemu.toml file.

Tested under Windows, needs testing under macOS/Linux just to make sure it all works properly.

Fixes #2544 .

image


config_spec.yml
```diff
@@ -19,6 +19,12 @@ general:
       f7: string
       f8: string
     filter_current_game: bool
+  history:
+    discs:
+      type: array
+      items:
+        type: string
+      max_items: 10
```
abaire: This seems error prone; we wouldn't want to allow the value here to get out of sync with the value enforced in the code. 

Can we just let the code clean up any excess values?

ui/xui/actions.cc
```diff
@@ -57,6 +57,41 @@ void ActionLoadDiscFile(const char *file_path)
...
+                    for (int j = i; j > 0; j--) {
+                        g_config.general.history.discs[j] = g_config.general.history.discs[j-1];
+                    }
+                    g_config.general.history.discs[0] = g_strdup(file_path);
+                    break;
+                }
+            }
+
+            if (i == g_config.general.history.discs_count) {
+                if (g_config.general.history.discs_count >= 10) {
```
abaire: Use a symbolic constant rather than a magic val (particular one that needs to be kept in sync with the following line and the YAML)

If the `discs_count` was somehow larger than 10 you'd probably want to free the excess here.

ui/xui/actions.cc
```diff
@@ -57,6 +57,41 @@ void ActionLoadDiscFile(const char *file_path)
...
+                    g_config.general.history.discs_count++;
+                }
+
+                for (int i = g_config.general.history.discs_count - 1; i > 0; i--) {
+                    g_config.general.history.discs[i] = g_config.general.history.discs[i-1];
+                }
+                g_config.general.history.discs[0] = g_strdup(file_path);
+            }
+        }
+        xemu_settings_save();
```
abaire: I don't think we do explicit saving anywhere else in xemu, it's always done `atexit`.

ui/xui/actions.cc
```diff
@@ -57,6 +57,41 @@ void ActionLoadDiscFile(const char *file_path)
...
+            }
+
+            int i;
+            for (i = 0; i < g_config.general.history.discs_count; i++) {
+                if (g_strcmp0(g_config.general.history.discs[i], file_path) == 0) {
+                    for (int j = i; j > 0; j--) {
+                        g_config.general.history.discs[j] = g_config.general.history.discs[j-1];
+                    }
+                    g_config.general.history.discs[0] = g_strdup(file_path);
+                    break;
```
abaire: Assuming you drop the save, return early here and you can get rid of some indentation below.

ui/xui/actions.cc
```diff
@@ -112,3 +147,31 @@ void ActionLoadSnapshotChecked(const char *name)
...
 }
+
+void ActionLoadDiscFromHistory(int index)
+{
+    if (index < 0 || index >= g_config.general.history.discs_count) {
+        return;
+    }
+
+    const char *file_path = g_config.general.history.discs[index];
+    if (!file_path || !file_path[0]) {
```
abaire: Should this assert? Seems like it'd be a programming error to call this on an invalid index?

ui/xui/actions.cc
```diff
@@ -112,3 +147,31 @@ void ActionLoadSnapshotChecked(const char *name)
...
+        return;
+    }
+
+    if (qemu_access(file_path, F_OK) == -1) {
+        return;
```
abaire, line 173: Should the invalid entry be removed?

config_spec.yml
ehw: Removed!

ui/xui/actions.cc
ehw: Good call. I'm not sure if 10 might be enough either, so it might be something that could be changed down the road if need be. I set one up and set it for 10 for now.

ui/xui/actions.cc
ehw: I always wondered about that, since I know if Xemu crashes it never bothers saving pretty much anything during run time. I can understand the logic in case a setting change is causing Xemu to crash which would allow it to revert to a previous state which was working. I'll remove it if that's what we do.

ui/xui/actions.cc
ehw: Added a return.

ui/xui/actions.cc
ehw: Added.

ui/xui/actions.cc
ehw: I added something in the event of an invalid file path.

mborgerson:
This PR currently adds actions for managing a recent list, but doesn't actually have the mentioned UI changes? Maybe you did not stage the full set of changes?

ehw:
> This PR currently adds actions for managing a recent list, but doesn't actually have the mentioned UI changes? Maybe you did not stage the full set of changes?

My bad, I forgot to commit the changes for the menu bar.

ui/xui/menubar.cc
```diff
@@ -138,6 +138,39 @@ void ShowMainMenu()
...
+                            ImGui::MenuItem(filename, NULL, false, false);
+                        }
+                        
+                        g_free((void *)filename);
+                    }
+                    
+                    ImGui::Separator();
+                }
+                
+                if (ImGui::MenuItem("Clear History")) {
```
abaire: Is it worth showing "clear" if there are no entries?

abaire: Should we have this say "Clear Recent" so it matches the name of the parent item?

ui/xui/menubar.cc
```diff
@@ -138,6 +138,39 @@ void ShowMainMenu()
...
+                    ActionClearDiscHistory();
+                    xemu_settings_save();
```
abaire: Same comment as before about proactive saving. 

I don't think it's necessarily a bad thing since I believe an assert will bypass the atexit handler, but we don't do it anywhere else and should probably be consistent one way or the other.

ui/xui/menubar.cc
```diff
@@ -138,6 +138,39 @@ void ShowMainMenu()
             if (ImGui::MenuItem("Eject Disc", SHORTCUT_MENU_TEXT(E))) ActionEjectDisc();
             if (ImGui::MenuItem("Load Disc...", SHORTCUT_MENU_TEXT(O))) ActionLoadDisc();
 
+            if (ImGui::BeginMenu("Recent")) {
+                bool has_recent = g_config.general.history.discs && g_config.general.history.discs_count > 0;
```
abaire: Inline since this is only used once?

Is there a case where discs_count would be > 0 but discs would be null?

ui/xui/menubar.cc
```diff
@@ -138,6 +138,39 @@ void ShowMainMenu()
...
+                
+                if (has_recent) {
+                    for (int i = 0; i < g_config.general.history.discs_count; i++) {
```
abaire: You could probably simplify this by pulling i out of the loop and using it to decide whether or not to add the separator (it'd let you drop the "has_recent" check.

ui/xui/menubar.cc
```diff
@@ -138,6 +138,39 @@ void ShowMainMenu()
...
+                        
+                        const char *filename = g_path_get_basename(disc_path);
+                        bool file_exists = qemu_access(disc_path, F_OK) != -1;
+                        
+                        if (file_exists) {
+                            if (ImGui::MenuItem(filename)) {
+                                ActionLoadDiscFromHistory(i);
+                            }
+                        } else {
+                            ImGui::MenuItem(filename, NULL, false, false);
```
abaire: Should we provide some mechanism to remove entries for missing files (besides editing the toml or clearing the whole list)?

Maybe allow them to be selected and pop a dialog indicating that the file doesn't exist and ask if it should be removed?

ui/xui/actions.cc
```diff
@@ -112,3 +151,42 @@ void ActionLoadSnapshotChecked(const char *name)
 {
     g_snapshot_mgr.LoadSnapshotChecked(name);
 }
+
+void ActionLoadDiscFromHistory(int index)
```
abaire: Maybe take an unsigned int since it'd be nonsensical to take a negative offset?

ui/xui/actions.cc
abaire: Where? There's an assert that the index is valid, but not that the contents of the array at a valid index are themselves valid.

ui/xui/menubar.cc
ehw: Renamed everything from History to Recent across the board for consistency.

ui/xui/menubar.cc
ehw: Should be taken care of now.

ui/xui/menubar.cc
ehw: I added a manual deletion by putting a red X on the left of every line, and I also added a setting to clear just the items that are missing.

ui/xui/actions.cc
ehw: Done.

ui/xui/menubar.cc
ehw: Done.

ehw:
Sorry for the delay in making these changes. 

I believe I addressed everything in the review and made some more changes:
- Renamed History to Recent for consistency
- Added a red X so that each entry can be manually deleted
- Made it so the "Recent Discs" becomes grey and unclickable if no discs are in the list
- Bumped the max disc in the list from 10 to 15
- Added something to clear just the missing discs from the list

Here's what the list looks like now:
image

Recent Discs grayed out with no discs in the list:
image

ui/xui/menubar.cc
```diff
@@ -142,6 +142,81 @@ void ShowMainMenu()
...
+            }
+            
+            bool has_valid_entries = (valid_count > 0);
+            if (ImGui::BeginMenu("Recent Discs", has_valid_entries)) {
+                for (int i = 0; i < g_config.general.recent.discs_count; i++) {
+                    const char *disc_path = g_config.general.recent.discs[i];
+                    if (!disc_path) continue;
+                    
+                    const char *filename = g_path_get_basename(disc_path);
+                    bool file_exists = qemu_access(disc_path, F_OK) != -1;
```
mborgerson, line 166: This gets called for every file in the list, for every frame that is rendered. It's also called above for every file. So each file is checked for existence ~120 times per second. I'm not sure the check is really even necessary, we can just error when someone selects a file that isn't present.

ui/xui/menubar.cc
```diff
@@ -142,6 +142,81 @@ void ShowMainMenu()
...
+                    
+                    if (file_exists) {
+                        ImGui::PushStyleColor(ImGuiCol_Text, IM_COL32(255, 0, 0, 255));
```
mborgerson, line 169: Lots of duplication here and below. I appreciate the thoughtfulness but I'm not sure why someone needs to curate their recent list so meticulously. Let's simplify this and just drop the ActionRemoveDiscFromRecent feature.

ui/xui/menubar.cc
```diff
@@ -142,6 +142,81 @@ void ShowMainMenu()
...
+                }
+                
+                if (g_config.general.recent.discs_count > 0) {
+                    ImGui::Separator();
+                    
+                    if (ImGui::MenuItem("Clear Recent Discs")) {
+                        ActionClearDiscRecent();
+                    }
+                    
+                    if (ImGui::MenuItem("Clear Missing Recent Discs")) {
```
mborgerson, line 212: Same argument above, let's just drop the missing disc handling

ui/xui/menubar.cc
```diff
@@ -142,6 +142,81 @@ void ShowMainMenu()
...
+                        ImGui::Text("%s", missing_filename);
+                        ImGui::PopStyleColor();
+                        
+                        g_free(missing_filename);
+                    }
+                    
+                    g_free((void *)filename);
+                }
+                
+                if (g_config.general.recent.discs_count > 0) {
```
mborgerson, line 205: This will always be true, because you've disabled the menu when there are none to show

ui/xui/actions.cc
```diff
@@ -112,3 +151,85 @@ void ActionLoadSnapshotChecked(const char *name)
 {
     g_snapshot_mgr.LoadSnapshotChecked(name);
 }
+
+void ActionLoadDiscFromRecent(unsigned int index)
+{
+    g_assert(index < g_config.general.recent.discs_count);
```
mborgerson, line 157: Not sure why we need this? We can just call ActionLoadDiscFile in the menu handler

ui/xui/actions.cc
```diff
@@ -57,6 +59,43 @@ void ActionLoadDiscFile(const char *file_path)
     if (err) {
         xemu_queue_error_message(error_get_pretty(err));
         error_free(err);
+    } else {
+        if (file_path && file_path[0]) {
+            if (!g_config.general.recent.discs) {
```
mborgerson, line 64: Let's simplify this logic

mborgerson (changes requested):
Appreciate the contribution, but in its current state I think it's too complex for a recent files list. Please see remarks inline.

ehw:
Due to some unfortunate circumstances I can no longer work on this for the time being. Feel free to do whatever.

mborgerson:
No need to close. Let's just move it to draft state for now and I (or someone) will address the requested changes and you will get the deserved credit for the contribution! Thanks for the work!

PR #2556, opened 2025-11-22
Add caption to metainfo file

j8r:
Required to pass Flatpak linter.

Noticed in https://github.com/flathub/app.xemu.xemu/pull/262#issuecomment-3564196086

### DIFF
--- a/xemu.metainfo.xml
+++ b/xemu.metainfo.xml
@@ -21,7 +21,7 @@
   </description>
 
   <screenshots>
-    <screenshot type="default"><image>https://xemu.app/screenshots/0.png</image></screenshot>
+    <screenshot type="default"><image>https://xemu.app/screenshots/0.png</image><caption>Xbox menu</caption></screenshot>
     <screenshot><image>https://xemu.app/screenshots/1.png</image></screenshot>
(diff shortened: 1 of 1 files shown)


xemu.metainfo.xml
```diff
@@ -21,7 +21,7 @@
   </description>
 
   <screenshots>
-    <screenshot type="default"><image>https://xemu.app/screenshots/0.png</image></screenshot>
+    <screenshot type="default"><image>https://xemu.app/screenshots/0.png</image><caption>Xbox menu</caption></screenshot>
     <screenshot><image>https://xemu.app/screenshots/1.png</image></screenshot>
```
abaire: Is the linter not going to warn on all the rest of them?

j8r: I was wondering too, that's right it is not: `flatpak run --command=flatpak-builder-lint org.flatpak.Builder --exceptions appstream xemu.metainfo.xml` returns one less "pedantic error" with than without.

mborgerson:
I captioned the remaining screenshots, ran a format pass, and adjusted your commit message slightly. Thanks!

PR #2558, opened 2025-11-22
Add nearest filtering method to menubar.

jyetcv:
Here is something to allow the user to change to nearest filtering without having to use third party programs.
This allows you to play games at 1x resolution without the blurring from scaling.
Hopefully this useful to someone.

nearest-filtering

Animations comparing linear and nearest filtering:
linear-nearest

linear-nearest-2


mborgerson:
Thanks!

Issue #1124, opened 2022-06-27
Def Jam: Fight for NY "`PTimer` upsupported" leading to freezing


Triticum0:
### Title

https://xemu.app/titles/45410049/#Def-Jam-Fight-for-NY

### Bug Description

The game never make it past the first loading screen properly and need one of the unimplemented commands to be playable

### Expected Behavior

Should make it to menu

### xemu Version

Version: 0.7.55
Branch: master
Commit: https://github.com/mborgerson/xemu/commit/db389b15086753cb7b72e05b24af205baa742bc3
Date: Sun Jun 26 05:58:57 UTC 2022

### System Information

<html><body>

Field | Value
-- | --
OS | Windows 10
CPU | AMD Ryzen 5 2600 Six-Core Processo
Graphics Device | NVIDIA GeForce RTX 3060 Ti/PCIe/SSE2
Graphics Driver | 4.0.0 NVIDIA 512.95

</body>
</html>

### Additional Context

[xemu.log](https://github.com/mborgerson/xemu/files/8995963/xemu.log)

CosmicStardus:
I'm not sure if this is known, but I made a discovery with this game that I posted in /r/xemu: The demo disc (NTSC-U region) to this game loads just fine in the latest version of xemu (0.7.118). Not only that, it plays great!

Could something simple be stopping the full game from loading?

papadkostas:
@CosmicStardus  could you share more? which exact iso you got?
I'm trying with xemu 0.7.119 usa and eu img and doesn't work, stays for ever in logo.

QuesoBlanco12:
> @CosmicStardus could you share more? which exact iso you got? I'm trying with xemu 0.7.119 usa and eu img and doesn't work, stays for ever in logo.

I can confirm that this issue does not exist in the "Demo" version of this ISO, and the game actually plays just fine.

![image](https://github.com/xemu-project/xemu/assets/76604675/5e3f592f-c4c8-4115-a0f3-c8fb802e0d9e)

If a dev is interested in investigating this, I am willing to provide a working xiso so that this can be investigated further. Just need to know what the best way would be to share it(don't want to break any rules or DMCA's, even if this is a demo disk, just playing it safe unless otherwise told).

flex50five:
Bumping this thread. Anyone been able to get this working with any new updates or bios?

MaouYumisu1:
As of the current build on USA Xemu 0.7.127 result is still the same as above.

[...]

PR #2568, opened 2025-11-26
nv2a: Implement PTIMER alarm interrupts

abaire:
Looks like the PTIMER alarm handling was started but never finished. This implements my best guess at what the HW is doing (it's hard to test with a nanosecond timer). It fixes the blocked loading for Def Jam: Fight for NY, which I determined to be caused by a event handle never being set due to the lack of an alarm interrupt in the same DPC that handles video blanking and other PMC interrupts. The PTIMER pending flag is known from the nxdk:
https://github.com/XboxDev/nxdk/blob/d0ae96acc926a6c3f6254b289a1a9a3c1db182f1/lib/pbkit/outer.h#L83

[See PTIMER tests in nxdk_low_level_nv2a_tests](https://github.com/abaire/nxdk_low_level_nv2a_tests)

Improves #1124 - it will now reliably get to the "Press start" screen but appears to still freeze later on due to a slightly different timing issue.

### DIFF
--- a/hw/xbox/nv2a/nv2a.c
+++ b/hw/xbox/nv2a/nv2a.c
@@ -560,7 +560,6 @@ static const VMStateDescription vmstate_nv2a = {
         VMSTATE_UINT32(ptimer.numerator, NV2AState),
         VMSTATE_UINT32(ptimer.denominator, NV2AState),
         VMSTATE_UINT32(ptimer.alarm_time, NV2AState),
-        VMSTATE_UINT64(ptimer.time_offset, NV2AState),
         VMSTATE_UINT32_ARRAY(pfb.regs, NV2AState, 0x1000),
         VMSTATE_UINT32(pcrtc.pending_interrupts, NV2AState),
         VMSTATE_UINT32(pcrtc.enabled_interrupts, NV2AState),
@@ -576,6 +575,7 @@ static const VMStateDescription vmstate_nv2a = {
         VMSTATE_UNUSED(1),
         VMSTATE_BOOL(pgraph.waiting_for_context_switch, NV2AState),
         VMSTATE_UINT32(ptimer.alarm_time_high, NV2AState),
+        VMSTATE_UINT64(ptimer.time_offset, NV2AState),
         VMSTATE_END_OF_LIST()
     },
 };
--- a/hw/xbox/nv2a/nv2a.c
+++ b/hw/xbox/nv2a/nv2a.c
@@ -46,7 +46,6 @@ void nv2a_update_irq(NV2AState *d)
     }
 
     /* PTIMER */
-    ptimer_process_alarm(d);
     if (d->ptimer.pending_interrupts & d->ptimer.enabled_interrupts) {
         d->pmc.pending_interrupts |= NV_PMC_INTR_0_PTIMER_PENDING;
     } else {
@@ -328,8 +327,7 @@ static void nv2a_reset(NV2AState *d)
     d->pmc.pending_interrupts = 0;
     d->pfifo.pending_interrupts = 0;
     d->ptimer.pending_interrupts = 0;
-    d->ptimer.alarm_time = 0xFFFFFFFF;
-    d->ptimer.time_offset = 0;
+    ptimer_reset(d);
     d->pcrtc.pending_interrupts = 0;
 
     for (int i = 0; i < 256; i++) {
@@ -459,6 +457,8 @@ static const VMStateDescription vmstate_nv2a = {
     .post_save = nv2a_post_save,
     .post_load = nv2a_post_load,
     .pre_load = nv2a_pre_load,
+    // NOTE: New fields should be appended, regardless of similarity to existing
+    //       fields.
     .fields = (VMStateField[]) {
         // FIXME: Split this up into subsections
         VMSTATE_PCI_DEVICE(parent_obj, NV2AState),
@@ -560,7 +560,6 @@ static const VMStateDescription vmstate_nv2a = {
         VMSTATE_UINT32(ptimer.numerator, NV2AState),
         VMSTATE_UINT32(ptimer.denominator, NV2AState),
         VMSTATE_UINT32(ptimer.alarm_time, NV2AState),
-        VMSTATE_UINT32(ptimer.alarm_time_high, NV2AState),
         VMSTATE_UINT64(ptimer.time_offset, NV2AState),
         VMSTATE_UINT32_ARRAY(pfb.regs, NV2AState, 0x1000),
         VMSTATE_UINT32(pcrtc.pending_interrupts, NV2AState),
@@ -576,6 +575,7 @@ static const VMStateDescription vmstate_nv2a = {
         VMSTATE_BOOL(pgraph.waiting_for_nop, NV2AState),
         VMSTATE_UNUSED(1),
         VMSTATE_BOOL(pgraph.waiting_for_context_switch, NV2AState),
+        VMSTATE_UINT32(ptimer.alarm_time_high, NV2AState),
         VMSTATE_END_OF_LIST()
     },
 };
@@ -627,5 +627,6 @@ void nv2a_init(PCIBus *bus, int devfn, MemoryRegion *ram)
     NV2AState *d = NV2A_DEVICE(dev);
     nv2a_init_memory(d, ram);
     nv2a_init_vga(d);
+    ptimer_init(d);
     qemu_add_vm_change_state_handler(nv2a_vm_state_change, d);
 }
--- a/hw/xbox/nv2a/nv2a.c
+++ b/hw/xbox/nv2a/nv2a.c
@@ -329,6 +329,7 @@ static void nv2a_reset(NV2AState *d)
     d->pfifo.pending_interrupts = 0;
     d->ptimer.pending_interrupts = 0;
     d->ptimer.alarm_time = 0xFFFFFFFF;
+    d->ptimer.time_offset = 0;
     d->pcrtc.pending_interrupts = 0;
 
     for (int i = 0; i < 256; i++) {
@@ -453,7 +454,7 @@ const VMStateDescription vmstate_nv2a_pgraph_vertex_attributes = {
 
 static const VMStateDescription vmstate_nv2a = {
     .name = "nv2a",
-    .version_id = 3,
+    .version_id = 4,
     .minimum_version_id = 1,
     .post_save = nv2a_post_save,
     .post_load = nv2a_post_load,
@@ -559,6 +560,8 @@ static const VMStateDescription vmstate_nv2a = {
         VMSTATE_UINT32(ptimer.numerator, NV2AState),
         VMSTATE_UINT32(ptimer.denominator, NV2AState),
         VMSTATE_UINT32(ptimer.alarm_time, NV2AState),
+        VMSTATE_UINT32(ptimer.alarm_time_high, NV2AState),
+        VMSTATE_UINT64(ptimer.time_offset, NV2AState),
         VMSTATE_UINT32_ARRAY(pfb.regs, NV2AState, 0x1000),
         VMSTATE_UINT32(pcrtc.pending_interrupts, NV2AState),
         VMSTATE_UINT32(pcrtc.enabled_interrupts, NV2AState),
--- a/hw/xbox/nv2a/nv2a.c
+++ b/hw/xbox/nv2a/nv2a.c
@@ -45,6 +45,14 @@ void nv2a_update_irq(NV2AState *d)
         d->pmc.pending_interrupts &= ~NV_PMC_INTR_0_PGRAPH;
     }
 
+    /* PTIMER */
+    ptimer_process_alarm(d);
+    if (d->ptimer.pending_interrupts & d->ptimer.enabled_interrupts) {
+        d->pmc.pending_interrupts |= NV_PMC_INTR_0_PTIMER_PENDING;
+    } else {
+        d->pmc.pending_interrupts &= ~NV_PMC_INTR_0_PTIMER_PENDING;
+    }
+
     if (d->pmc.pending_interrupts && d->pmc.enabled_interrupts) {
         trace_nv2a_irq(d->pmc.pending_interrupts);
         pci_irq_assert(PCI_DEVICE(d));
@@ -320,6 +328,7 @@ static void nv2a_reset(NV2AState *d)
     d->pmc.pending_interrupts = 0;
     d->pfifo.pending_interrupts = 0;
     d->ptimer.pending_interrupts = 0;
+    d->ptimer.alarm_time = 0xFFFFFFFF;
     d->pcrtc.pending_interrupts = 0;
 
     for (int i = 0; i < 256; i++) {
--- a/hw/xbox/nv2a/nv2a.c
+++ b/hw/xbox/nv2a/nv2a.c
@@ -47,9 +47,9 @@ void nv2a_update_irq(NV2AState *d)
 
     /* PTIMER */
     if (d->ptimer.pending_interrupts & d->ptimer.enabled_interrupts) {
-        d->pmc.pending_interrupts |= NV_PMC_INTR_0_PTIMER_PENDING;
+        d->pmc.pending_interrupts |= NV_PMC_INTR_0_PTIMER;
     } else {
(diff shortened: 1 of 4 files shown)


Triticum0:
I tested only other game that freezes during intro and is unrelated issue. [NHL Rivals 2004](https://xemu.app/titles/4d53004e/#NHL-Rivals-2004) Note: this game goes in-game on cxbx-reloaded. Also tested NFL Fever 2004 no change.

hw/xbox/nv2a/ptimer.c
```diff
@@ -31,6 +34,29 @@ static uint64_t ptimer_get_clock(NV2AState *d)
...
+    if (d->ptimer.alarm_time == 0xFFFFFFFF) {
+        return;
+    }
+    uint64_t now = ptimer_get_host_clock(d);
+    uint64_t guest_clock = ptimer_get_guest_clock(d, now);
+
+    uint32_t current_time_low = (guest_clock & CLOCK_LOW_MASK) << 5;
+    uint32_t current_time_high = (guest_clock >> 27) & CLOCK_HIGH_MASK;
+
+    if (d->ptimer.alarm_time <= current_time_low ||
```
PatrickvL: This check is wrongly succeeding when current time low bits is lower or equal to alarm time (low bits) but current high bits are less than alarm high bits. 

It seems wiser to reconstruct the full set of alarm bits and do a single comparison of that against the full bits of the current time.

PatrickvL (changes requested):
This looks like a bug

hw/xbox/nv2a/ptimer.c
abaire: I think it is correct due to the fact that there is no mechanism to specify the high bits of the alarm time.

* If current_time_low < alarm the first condition is false so behavior is governed by whether current_time_high is > alarm (the timer has wrapped).
* If current_time_low >= alarm the timer will fire (I'm assuming you meant that it wrongly succeeds if current time low bits is higher than alarm time?). If the high bits were controllable, this could indeed be erroneous.


I'm not sure what the HW behavior is if the alarm is set into the ambiguous past or if the caller is able to race the clock and set TIME_1 into the past before the alarm fires. 

If the values observed being set to the ALARM_0 weren't so close to the current time I would've expected it to be a delta into the future, in which case it would be unambiguous and using 64-bit would certainly make things clearer. 
The fact that it seems to really just be the low time bits makes me think that it's doing a 32-bit comparison with special affordance for the case where the value is very close to rollover (as implemented here). 

I've also seen values with the high bit set, so it's not a case of using a 31-bit value to catch rollover.


I suspect this may be a case where an assumption was made that the driver writer would do the right thing and that regular users either wouldn't know the register existed or wouldn't attempt ~nonsensical things like setting a nanosecond timer into the past.

Joshua-1248:
Def Jam - Fight For NY is actually able to make it to the menus with this particular commit with some brute-force resetting whenever it freezes during the intro. And much like the Demo version of the game, it makes it in-game with no issues, save for one; 

The game stutters with a 2-3 second freeze periodically throughout emulation, menus or in-game, so it's not exactly playable, but it does run full speed and can presumably be played from start to finish.

Tested with both OpenGL & Vulkan and both work, with Vulkan suffering less from the temporary freezes, though still not playable without interruptions.

xemu-2025-11-29-23-34-40

xemu-2025-11-29-23-53-14

xemu-2025-11-29-23-53-37

xemu-2025-11-29-23-53-53

hw/xbox/nv2a/ptimer.c
```diff
@@ -21,8 +21,11 @@
 
 #include "nv2a_int.h"
 
+#define CLOCK_HIGH_MASK 0x1fffffff
+#define CLOCK_LOW_MASK 0x7ffffff
+
 /* PTIMER - time measurement and time-based alarms */
-static uint64_t ptimer_get_clock(NV2AState *d)
+static uint64_t ptimer_get_host_clock(NV2AState *d)
```
mborgerson: This host/guest terminology here is a bit confusing. "Host" typically refers to the machine running xemu (and that is not the time domain source we are using--we are actually using the virtual clock here which stops when emulation is paused) and "guest" typically means the emulated machine state. Maybe host/guest &rarr; absolute/relative would be more clear in this case

hw/xbox/nv2a/nv2a.c
```diff
@@ -529,6 +539,8 @@ static const VMStateDescription vmstate_nv2a = {
         VMSTATE_UINT32(ptimer.numerator, NV2AState),
         VMSTATE_UINT32(ptimer.denominator, NV2AState),
         VMSTATE_UINT32(ptimer.alarm_time, NV2AState),
+        VMSTATE_UINT32(ptimer.alarm_time_high, NV2AState),
```
mborgerson: To be compatible with older vmstate versions, we should append new fields to the end

hw/xbox/nv2a/nv2a.c
```diff
@@ -45,6 +45,14 @@ void nv2a_update_irq(NV2AState *d)
         d->pmc.pending_interrupts &= ~NV_PMC_INTR_0_PGRAPH;
     }
 
+    /* PTIMER */
+    ptimer_process_alarm(d);
```
mborgerson: Even though this function is called regularly, alarm management should be independent from this interrupt management function. We should use `QEMUTimer` instead for time based event signaling

For reference: https://github.com/xemu-project/xemu/blob/9b7a43f489784876b6c1d848f2d862161241edfa/hw/xbox/mcpx/nvnet/nvnet.c#L669

mborgerson (commented):
Thanks! Some initial feedback

hw/xbox/nv2a/nv2a.c
abaire: Done

hw/xbox/nv2a/nv2a.c
abaire: Done, thanks for the pointer!

hw/xbox/nv2a/nv2a.c
abaire: Done, also added a note to hopefully ensure future additions are also appended.

hw/xbox/nv2a/ptimer.c
abaire: Switched to absolute/relative. I think it's still unusual enough to justify a bit of documentation.

hw/xbox/nv2a/nv2a.c
```diff
@@ -423,11 +431,13 @@ const VMStateDescription vmstate_nv2a_pgraph_vertex_attributes = {
 
 static const VMStateDescription vmstate_nv2a = {
     .name = "nv2a",
-    .version_id = 3,
+    .version_id = 4,
     .minimum_version_id = 1,
     .post_save = nv2a_post_save,
     .post_load = nv2a_post_load,
     .pre_load = nv2a_pre_load,
+    // NOTE: New fields should be appended, regardless of similarity to existing
```
mborgerson: * I don't think the comment is necessary
* `ptime.time_offset` addition not at end of list

abaire:
Looking at HW behavior, it seems that the alarm is not reset after firing, it's up to the application to clear it or disable the interrupt.

abaire:
Updated to match observed HW behavior in a trivial test that sets the alarm to an arbitrary value and increments a counter every time the interrupt fires. Hopefully this also fixes the freezes observed in DJFfNY (I've restarted dozens of times without my tracing plugin attached now and have been unable to repro).

hw/xbox/nv2a/ptimer.c
abaire: Turns out the alarm is not intended to be a one-shot (the game I was using to infer the behavior is just very spammy about updating the alarm time). I ended up writing a trivial test case by adding a counter to pbkit's DPC that is incremented on alarm interrupts and found that it fires every time the low 32 bits of the counter rolls over.

Triticum0:
Tested this build and freezing on intro fixed but freezing still occur when on start and first fight intro.
Only happen once and didn't occur after rebooting so not sure why it occurs.

retchless:
Also tested this with Def Jam Fight for NY. Game loads now, and navigating menus seems reliable. But it freezes randomly during cut scenes (seemingly at different points, so truly random). Game is still unplayable.

hw/xbox/nv2a/ptimer.c
```diff
@@ -49,11 +123,16 @@ uint64_t ptimer_read(void *opaque, hwaddr addr, unsigned int size)
...
-    case NV_PTIMER_TIME_1:
-        r = (ptimer_get_clock(d) >> 27) & 0x1fffffff;
+    case NV_PTIMER_TIME_0: {
+        uint64_t now = get_ptimer_clock(d, ptimer_get_absolute_clock(d));
+        r = now & 0xffffffff;
+    } break;
+    case NV_PTIMER_TIME_1: {
+        uint64_t now = get_ptimer_clock(d, ptimer_get_absolute_clock(d));
+        r = (now >> 32) & CLOCK_HIGH_MASK;
+    } break;
```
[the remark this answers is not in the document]
mborgerson: @gemini-code-assist: Latching could be the right approach, but we should review access patterns and confirm what hardware actually does. The proposed change might not be correct if `NV_PTIMER_TIME_1` is read before `NV_PTIMER_TIME_0`, for instance. I suspect hardware leaves this to be checked by software anyway and that the current solution is correct.

hw/xbox/nv2a/nv2a.c
mborgerson: Let's drop the comment. VMState is documented elsewhere

hw/xbox/nv2a/nv2a_regs.h
```diff
@@ -61,6 +61,7 @@
 #define NV_PMC_INTR_0                                    0x00000100
 #   define NV_PMC_INTR_0_PFIFO                                 (1 << 8)
 #   define NV_PMC_INTR_0_PGRAPH                               (1 << 12)
+#   define NV_PMC_INTR_0_PTIMER_PENDING                       (1 << 20)
```
mborgerson: To match other fields, drop `_PENDING` suffix

hw/xbox/nv2a/ptimer.c
```diff
@@ -21,14 +22,87 @@
...
+        return;
+    }
+
+    uint64_t now = get_ptimer_clock(d, ptimer_get_absolute_clock(d));
+    uint64_t alarm_time = (uint64_t)d->ptimer.alarm_time +
+                          ((uint64_t)d->ptimer.alarm_time_high << 32);
+
+    uint64_t diff_ns = 0;
+    if (alarm_time > now) {
+        diff_ns = ptimer_ticks_to_ns(d, (alarm_time - now) >> 5);
```
mborgerson: I'm a bit lost tracking the bit shifting. Is the left shifting in `get_ptimer_clock` just putting the clock in register format? We shift left in `get_ptimer_clock`, then shift right here to compute actual clock deltas. I wonder if this means we should really just do the shifting/masking at the register interface (e.g. in `ptimer_read`)

hw/xbox/nv2a/nv2a.c
abaire: Done, though I think I've broken this at least a couple times over the years by inserting or resizing things inline.

I wonder if the review bots read `AGENTS.md` and could be instructed to look for this specifically?

hw/xbox/nv2a/nv2a_regs.h
abaire: Done

hw/xbox/nv2a/ptimer.c
abaire: Yes, it's just moving in and out of the format used in the registers.

Aren't the read/write callbacks expected to act on data in HW format?

I wonder if some helper macros would be appropriate rather than messing with the guest-accessible callback behavior (but maybe I'm just not following your suggestion).

hw/xbox/nv2a/ptimer.c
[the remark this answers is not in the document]
abaire: Interesting response. My expectation is also that the HW isn't doing magic here and it really is left up to the SW to detect race-rollover if it actually cares to. When I was debugging the issue that precipitated this change I saw several places where the timer was being read in a tight loop, presumably to work around the potential race.

hw/xbox/nv2a/nv2a.c
mborgerson: I believe so! Not sure about the actual details. Maybe they can be instructed to read select pages in docs/ based on related subsystem under review.

abaire:
I've been continuing the investigation in #1124 - I think the remaining softlock is timing related rather than an issue with the PTIMER alarm code added in this PR. 

To summarize my understanding at the moment, the game tries to keep a consistent timer interval and does some checking using the CPU timestamp counter (`rdtsc`) and a delta versus the last alarm time to account for small variations in the rescheduling of the alarm. The deltas are generally very small, but in xemu the interrupt interval is much less stable than on HW and we eventually run into a case where the code responsible for rescheduling the PTIMER alarm runs after the alarm should've fired. This leads to a state that the title never bothered to handle, since it shouldn't ever happen in HW, and it fails to schedule the alarm and disables PTIMER alarm events generally, causing the logic to sit around forever waiting for an alarm that cannot ever come.

This will require a separate fix to address. At this point I've created [additional PTIMER tests](https://github.com/abaire/nxdk_low_level_nv2a_tests/blob/5c162be2ca46a5db96c7aecd2fdf88b7aeceef4c/src/ptimer_alarm_main.cpp#L1) and am relatively confident that the PTIMER behavior on its own matches hardware.

In the meantime you may be able to get the game to work by enabling the qemu `icount` feature. My ~30 minutes of testing showed that this prevents the freeze, but it causes a very noticeable performance degradation (at least in my debug build of xemu).

retchless:
@abaire Any more ideas on this one? Is it worth merging the work you've already done as it does partially improve things?

abaire:
> @abaire Any more ideas on this one? Is it worth merging the work you've already done as it does partially improve things?

If I didn't think it was worth merging it would be a draft and not a PR. 

My hope is that behavior will be improved when coupled with the changes in 134 to decouple guest framerate from the host, hopefully leading to more reliable on-time flips. I don't plan on spending any more time on this until it's merged and we can get feedback on what machines still run into the problem and how often.

hw/xbox/nv2a/nv2a.c
```diff
@@ -564,6 +572,8 @@ static const VMStateDescription vmstate_nv2a = {
         VMSTATE_BOOL(pgraph.waiting_for_nop, NV2AState),
         VMSTATE_UNUSED(1),
         VMSTATE_BOOL(pgraph.waiting_for_context_switch, NV2AState),
+        VMSTATE_UINT32(ptimer.alarm_time_high, NV2AState),
+        VMSTATE_UINT64(ptimer.time_offset, NV2AState),
```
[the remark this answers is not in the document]
abaire: Fair point, done.

hw/xbox/nv2a/ptimer.c
```diff
@@ -21,14 +22,87 @@
...
     return muldiv64(muldiv64(qemu_clock_get_ns(QEMU_CLOCK_VIRTUAL),
                              d->pramdac.core_clock_freq,
                              NANOSECONDS_PER_SECOND),
-                    d->ptimer.denominator,
-                    d->ptimer.numerator);
+                    d->ptimer.denominator, d->ptimer.numerator);
+}
+
+//! Returns the NV2A time adjusted by the last time register modification and
+//! shifted into range.
```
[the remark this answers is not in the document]
abaire: It looks like the additional lines are all hallucinated?

I changed the block comment, though I don't think xemu style is as strict wrt. modern C/C++ line comment use.

abaire:
I did some refactoring to try to make the shifting easier to follow and greatly expanded the test cases. The issue where the qemu timer could be infinitely rescheduled due to an alarm in the past is also fixed.

I've also changed the "Fixes" to just "Improves" - with this rebased on top of the guest/host decoupling stability seems to be slightly worse than it was before (at least on my M5 MacBook test machine). I think the remaining issue should be investigated and fixed in a followup as I'm somewhat confident it's an issue with how clocks are emulated rather than related to the PTIMER alarm implementation (see the comment thread in #1124 for details).

PR #2598, opened 2025-12-17
ui: Align shortcut handling with ImGui Ctrl/Cmd model

igorleonovich:
### Summary

This PR updates shortcut detection logic to match Dear ImGui’s current input model,
where Ctrl and Cmd are automatically swapped on macOS.

### Background

Starting with ImGui v1.92.5, the shortcut modifier handling was simplified:
Ctrl/Cmd mapping on macOS is now handled internally by ImGui, and
`ImGuiMod_Shortcut` was removed.

As a result, application code no longer needs to special-case macOS
or explicitly check for `KeySuper`.

### What changed

- Removed legacy macOS-specific shortcut branching
- Rely on `Ctrl` as the canonical cross-platform shortcut modifier
- Preserve existing behavior by rejecting Alt/Shift for plain shortcuts

### References

- Dear ImGui v1.92.5 release notes
- Commit: “Inputs: swap Ctrl and Cmd on macOS X, remove ImGuiMod_Shortcut”
  https://github.com/ocornut/imgui/commit/7747106

### DIFF
--- a/ui/xui/misc.hh
+++ b/ui/xui/misc.hh
@@ -55,12 +55,21 @@ std::string string_format( const std::string& format, Args ... args )
     return std::string( buf.get(), buf.get() + size - 1 ); // We don't want the '\0' inside
 }
 
-static inline bool IsShortcutKeyPressed(int scancode)
+static inline bool IsShortcutKeyPressed(ImGuiKey key)
 {
     ImGuiIO& io = ImGui::GetIO();
-    const bool is_osx = io.ConfigMacOSXBehaviors;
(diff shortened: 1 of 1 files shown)


ui/xui/misc.hh
```diff
@@ -55,12 +55,21 @@ std::string string_format( const std::string& format, Args ... args )
...
+
+    // Disallow Alt/Shift for plain shortcuts
+    if (io.KeyAlt || io.KeyShift) {
+        return false;
+    }
+
+    // ImGui input model (>= v1.92.5):
+    // On macOS Ctrl/Cmd are swapped internally by ImGui,
+    // so using Ctrl as the shortcut modifier is cross-platform idiomatic.
+    const bool shortcut_mod = io.KeyCtrl;
```
mborgerson: Let's drop the extra variable and comments

mborgerson:
Thanks!

PR #2619, opened 2025-12-26
ui: Replace xemu_open_web_browser with SDL_OpenURL

mborgerson:


ui/xemu-os-utils-linux.c
```diff
@@ -23,6 +23,7 @@
 #include <assert.h>
 #include <glib.h>
 #include <glib/gprintf.h>
+#include <SDL2/SDL.h>
```
mborgerson: Remove

PR #2637, opened 2026-01-02
meson: Update imgui

### LINKED ISSUE #1666: Retina Display/High-DPI scaling for ImGUI is missing on macOS
### Bug Description

Xemu simply doesn't support Retina displays ~~at all~~ for ImGUI elements, with its entire GUI looking barely any different irrespective of if NSHighResolutionCapable in Info.plist is enabled or disabled, in both cases it's being upscaled from half-resolution on each axis (except with NSHighResolutionCapable disabled the OS just pixel-doubles everything with a nearest-neighbour filter, whilst with the Xemu default of enabling NSHighResolutionCapable, the app itself just performs an ugly bilinear upscale it seems rather than actually becoming retina-aware, as that option should signify when used correctly).
The final bit of proof of the above is the fact that, if I were to switch my MacBook Air to run in the non-retina 2560x1664 video mode at 1x scale (which makes UI elements unusably tiny system-wide due to how pixel-dense the panel is), Xemu does end up taking advantage of the display's resolution and its UI looks perfectly crisp as a consequence. That effectively confirms that Xemu outright lacks proper Retina support for ImGUI elements as a result.

### Expected Behavior

Proper Retina support should be present when the OS is running in a 2x retina mode, i.e. the UI is rendered at 2x scale with sharp, crisp text and graphics shown in all places in the UI. ~~and the internal framebuffer should draw at a 1-1 pixel ratio such that they can take advantage of the display's full resolution (mostly applicable for when resolution scaling is in use to render Xbox titles closer to the native resolution of a Mac's Retina display, for example).~~

### xemu Version

0.7.126

### System Information

OS: macOS
System: MacBook Air M2 (2022)
Display: 2560x1664 Retina Display

### Additional Context

_No response_

schm1dtmac:
Integrates the macOS retina fixes from [imgui#1](https://github.com/xemu-project/imgui/pull/1).
Closes #1666

### DIFF
--- a/subprojects/imgui.wrap
+++ b/subprojects/imgui.wrap
@@ -1,4 +1,4 @@
 [wrap-git]
 url=https://github.com/xemu-project/imgui
-revision=7219d617a32b594f9a80b2356aec42e0e939e938
+revision=b911105fca3ca1b025706dd168e5798070f143a1
 depth=1


schm1dtmac:
CI artifact works great on arm64 and Intel (under Rosetta), all ImGUI elements scale correctly at a retina resolution, LGTM I'd say.

mborgerson:
Thanks!

PR #2669, opened 2026-01-14
ci: Fix Win version.rc compilation on forks

abaire:
Not sure how my initial "unofficial" PR worked in my branch a week ago but it doesn't with a master sync now. 

Looks like `xemu-version.sh` ends up passing "unofficial" to `version.rc` where it's treated as an integer. This change makes it `0` instead.

We could also probably switch from using `-` separators to `_` so the `cut` fallback logic catches its own `-0`.


abaire:
Possibly just a hiccup, but the macOS CI build failed to extract the source package initially: https://github.com/xemu-project/xemu/actions/runs/21013242595/job/60412981368

Retrying seems to have resolved it.

PR #2675, opened 2026-01-16
nv2a/pgraph: refactor blit into shared implementation with GL/VK wrap…

### LINKED ISSUE #2671: Refactor GL and VK blit.c code to a shared implementation
### Bug Description

Currently xemu has two practically identical handlers for the 0x9F blit functionality:

https://github.com/xemu-project/xemu/blob/master/hw/xbox/nv2a/pgraph/gl/blit.c
https://github.com/xemu-project/xemu/blob/master/hw/xbox/nv2a/pgraph/vk/blit.c

The only difference are a couple of calls into the renderer to deal with surface retrieval. These calls have identical signatures and could be passed into a common implementation as function pointers.

### Expected Behavior

To reduce maintenance burden, we should factor these into a single implementation in pgraph/blit.c (or similar). Eventually we may implement accelerated versions of the blits, but until that happens this will make it easier to maintain correctness between the renderers.

### xemu Version

n/a

### System Information

n/a

### Additional Context

See https://github.com/xemu-project/xemu/pull/2666/changes/BASE..b95148b8a5353118b4e0dba394963fb1f228867e#r2692942400 and associated conversation between `Matt` & `abaire` on Discord

KraftMacAndChee:
Fixes #2671


hw/xbox/nv2a/pgraph/gl/blit.c
```diff
@@ -1,223 +1,36 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "hw/xbox/nv2a/nv2a_int.h"
+#include "renderer.h"
+#include "hw/xbox/nv2a/pgraph/blit.h"
+
+// TODO: Optimize.
```
abaire: I think we can remove this comment now. There's not much that could be done to optimize the one function call and the comment doesn't give any detail on how to do renderer-specific optimization.

hw/xbox/nv2a/pgraph/blit.c
```diff
@@ -0,0 +1,176 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "hw/xbox/nv2a/nv2a_int.h"
+#include "hw/xbox/nv2a/pgraph/blit.h"
+
+
+// TODO: Optimize. Ideally this should all be done via OpenGL.
```
abaire: nit: Remove, now that this is a shared implementation it doesn't make sense to suggest GL.

hw/xbox/nv2a/pgraph/blit.c
```diff
@@ -0,0 +1,176 @@
...
+
+{
+    PGRAPHState *pg = &d->pgraph;
+    ContextSurfaces2DState *context_surfaces = &pg->context_surfaces_2d;
+    ImageBlitState *image_blit = &pg->image_blit;
+    BetaState *beta = &pg->beta;
+
+    ops->surface_update(d, false, true, true);
+
+    assert(context_surfaces->object_instance == image_blit->context_surfaces);
```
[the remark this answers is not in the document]
abaire: It probably does make sense to bail early on zero-sized draws. I don't think we perform any interesting side effects that wouldn't be covered elsewhere in the pipeline.

hw/xbox/nv2a/pgraph/blit.c
```diff
@@ -0,0 +1,176 @@
...
+    /* FIXME: What does hardware do in this case? */
+    assert(source_addr + source_offset + source_size <=
+           memory_region_size(d->vram));
+    assert(dest_addr + dest_offset + dest_size <= memory_region_size(d->vram));
+
+    uint8_t *source_row = source + source_offset;
+    uint8_t *dest_row = dest + dest_offset;
+
+    if (image_blit->operation == NV09F_SET_OPERATION_SRCCOPY) {
+        // NV2A_GL_DPRINTF(false, "NV09F_SET_OPERATION_SRCCOPY");
```
abaire: nit: While you're in here, we can probably lose these commented-out DPRINTF's as well.

hw/xbox/nv2a/pgraph/blit.h
```diff
@@ -0,0 +1,13 @@
+// Header for blit.c
```
abaire: I think we should bring over the comment from the original GL impl.

hw/xbox/nv2a/pgraph/blit.h
```diff
@@ -0,0 +1,13 @@
+// Header for blit.c
+
+#pragma once
```
abaire: I don't think we use `pragma once` outside of C++ files. Would be good to use old C-style guards like in https://github.com/xemu-project/xemu/blob/24087af22ab9df22a0fa214188266430c3cf1387/hw/xbox/nv2a/nv2a_regs.h#L21

hw/xbox/nv2a/pgraph/blit.h
```diff
@@ -0,0 +1,13 @@
+// Header for blit.c
+
+#pragma once
+
+#include "hw/xbox/nv2a/nv2a_int.h"   // NV2AState, hwaddr, SurfaceBinding
```
abaire: nit: I don't think we declare the types used from headers elsewhere and these can easily go stale without automation. I think we should remove them.

hw/xbox/nv2a/pgraph/blit.h
```diff
@@ -0,0 +1,13 @@
+// Header for blit.c
+
+#pragma once
+
+#include "hw/xbox/nv2a/nv2a_int.h"   // NV2AState, hwaddr, SurfaceBinding
+
+typedef struct PGRAPHSurfaceOps {
```
abaire: I wonder if we should put this in the PGRAPHRenderer so they're available everywhere?

https://github.com/xemu-project/xemu/blob/24087af22ab9df22a0fa214188266430c3cf1387/hw/xbox/nv2a/pgraph/pgraph.h#L108

abaire (commented):
Mostly nitpicky comments

mborgerson:
Looks like this change was made on top of be882c098ce9b61576de356ece630fd5a7f4ef4e and not 24087af22ab9df22a0fa214188266430c3cf1387 (which fixed a bug). Please rebase the patch

hw/xbox/nv2a/pgraph/gl/blit.c
```diff
@@ -1,223 +1,36 @@
...
+void pgraph_gl_image_blit(NV2AState *d)
+{
+    static const PGRAPHSurfaceOps ops = {
+        .surface_update = pgraph_gl_surface_update,
```
mborgerson: Let's make this even simpler and drop the extra indirection here.

* `surface_update` is already a renderer op
* Just add `surface_get`, add `surface_download_if_dirty` and remove `image_blit` in `PGRAPHRenderer::ops`
* Drop gl/blit.c, vk/blit.c, blit.h

mborgerson (changes requested):
Thanks for the patch. See comments about rebase and structure inline. Also don't worry about trying to fix pre-existing bugs with this patch. I've not evaluated the feedback of Gemini, but we should focus only on the unification in this patch. The delta between pgraph/blit.c, gl/blit.c, and vk/blit.c should be near 0. We can fix any bugs in the current implementation afterwards.

KraftMacAndChee:
> Thanks for the patch. See comments about rebase and structure inline. Also don't worry about trying to fix pre-existing bugs with this patch. I've not evaluated the feedback of Gemini, but we should focus only on the unification in this patch. The delta between pgraph/blit.c, gl/blit.c, and vk/blit.c should be near 0. We can fix any bugs in the current implementation afterwards.

I rebased it as you requested, I then worked to unify everything following your instruction and also removed gl/blit.c, vk/blit.c, and blit.h as requested. I spent a long time looking everything over and trying to double check myself and make sure I did everything correctly. However, I would appreciate you looking over it as I'm still getting my feet wet on this kind of stuff again.

hw/xbox/nv2a/pgraph/blit.c
```diff
@@ -0,0 +1,176 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "hw/xbox/nv2a/nv2a_int.h"
+#include "hw/xbox/nv2a/pgraph/blit.h"
+
+
+
```
abaire, line 58: Super nit: remove a newline or two. Looks like the convention is a single newline before the first function.

hw/xbox/nv2a/pgraph/pgraph.h
```diff
@@ -38,6 +38,8 @@ typedef struct NV2AState NV2AState;
 typedef struct PGRAPHNullState PGRAPHNullState;
 typedef struct PGRAPHGLState PGRAPHGLState;
 typedef struct PGRAPHVkState PGRAPHVkState;
+typedef struct SurfaceBinding SurfaceBinding;
+
```
abaire, line 42: nit: remove extra newline

abaire (commented):
This still looks like it has dropped all of the changes from https://github.com/xemu-project/xemu/commit/24087af22ab9df22a0fa214188266430c3cf1387

You can use https://github.com/abaire/xemu-dev_pgraph_test_results to look for other regressions (I haven't had a chance to update the goldens for the latest xemu yet, so it wouldn't catch the blit regression that this PR will cause).

KraftMacAndChee:
> This still looks like it has dropped all of the changes from [24087af](https://github.com/xemu-project/xemu/commit/24087af22ab9df22a0fa214188266430c3cf1387)
> 
> You can use https://github.com/abaire/xemu-dev_pgraph_test_results to look for other regressions (I haven't had a chance to update the goldens for the latest xemu yet, so it wouldn't catch the blit regression that this PR will cause).

Ah yes, you’re right. I wasn’t thinking and just rebased it without modifying anything. So it would be a regression. I’ll have to fix that when I get time.

hw/xbox/nv2a/pgraph/vk/renderer.c
```diff
@@ -19,7 +19,7 @@
 
 #include "hw/xbox/nv2a/nv2a_int.h"
 #include "renderer.h"
-
+#include "hw/xbox/nv2a/pgraph/surface.h"
```
mborgerson, line 22: Already included (via nv2a_int.h, which includes pgraph.h, which includes surface.h)

hw/xbox/nv2a/pgraph/blit.c
```diff
@@ -1,223 +1,223 @@
...
+{
+    for (unsigned int y = 0; y < height; y++) {
+        uint8_t *d = dest;
+        for (unsigned int x = 0; x < width_pixels; x++) {
+            d[x * 4 + 3] = alpha_val;
+        }
+        dest += dest_pitch;
+    }
+}
+void pgraph_common_image_blit(NV2AState *d)
```
mborgerson, line 70: * Fix whitespace
* Use name `pgraph_image_blit` (this was the former name, and since there is only one 'common' is redundant)

mborgerson (changes requested):

hw/xbox/nv2a/pgraph/pgraph.c
```diff
@@ -762,7 +762,7 @@ int pgraph_method(NV2AState *d, unsigned int subchannel,
             image_blit->height = parameter >> 16;
 
             if (image_blit->width && image_blit->height) {
-                d->pgraph.renderer->ops.image_blit(d);
+                pgraph_common_image_blit(d);
```
abaire, line 765: Should've pointed it out before, but "common" seems unnecessary if there are no renderer-spec. Maybe just "pgraph_image_blit"?

hw/xbox/nv2a/pgraph/blit.c
```diff
@@ -1,223 +1,223 @@
...
+
```
abaire, line 71: Remove newline

hw/xbox/nv2a/pgraph/blit.c
```diff
@@ -1,223 +1,223 @@
...
+            bytes_per_pixel = 4;
+            break;
+        default:
+            fprintf(stderr, "Unknown blit surface format: 0x%x\n",
+                    context_surfaces->color_format);
+            assert(false);
+            break;
+    }
+
+    hwaddr source_dma_len, dest_dma_len;
```
abaire, line 103: This is (I assume inadvertently) reverted from the split version in the current head, see line 104 and 111 in the diff.

hw/xbox/nv2a/pgraph/blit.c
```diff
@@ -1,223 +1,223 @@
...
+        uint8_t *src =
+            source_row + adjusted_height * context_surfaces->source_pitch;
+        uint8_t *dest =
+            dest_row + adjusted_height * context_surfaces->dest_pitch;
+
+        perform_blit(image_blit->operation, src, dest,
+                     leftover_bytes / bytes_per_pixel, 1, leftover_bytes,
+                     context_surfaces->source_pitch,
+                     context_surfaces->dest_pitch, beta);
+        }
```
abaire, line 187: This indentation is wrong.

Since this is a new file you can probably just run clang-format on it to avoid having to waste time with whitespace changes. I believe I clang-formatted the original blit file when I fixed the clipping bug.

hw/xbox/nv2a/pgraph/blit.c
```diff
@@ -1,223 +1,223 @@
...
+    if (clipped_dest_size < dest_size) {
+        adjusted_height = clipped_dest_size / context_surfaces->dest_pitch;
+        size_t consumed_bytes = adjusted_height * context_surfaces->dest_pitch;
+
+        leftover_bytes = clipped_dest_size - consumed_bytes;
+    }
+
+    SurfaceBinding *surf_dest = pg->renderer->ops.surface_get(d, dest_addr);
+    if (surf_dest) {
+         if (adjusted_height < surf_dest->height ||
```
abaire, line 155: This indentation looks wrong

hw/xbox/nv2a/pgraph/blit.c
abaire, line 70: Add the newline above this back

PR #2692, opened 2026-01-22
ui: fix crash bug when clicking config info

oltolm:
When I click in the text box "Config Information" xemu crashes. I fixed it.


mborgerson:
Thanks! Adjusted commit message a bit

PR #2698, opened 2026-01-24
xblc: Migrate to SDL3

faha223:
Changed the usb-xblc driver to use the SDL3 audio backend

To test this: add an xblc to your controller using the following command in the qemu monitor:
`device_add usb-xblc,port=1.3.2,id=xblc1`

You can also remove the xblc from your controller using the following command in the qemu monitor:
`device_del xblc1`

This will add the virtual XBLC device to expansion slot A of the controller in port 1

### DIFF
--- a/hw/xbox/xblc.c
+++ b/hw/xbox/xblc.c
@@ -160,20 +160,10 @@ static void usb_xblc_handle_reset(USBDevice *dev)
     DPRINTF("[XBLC] Reset\n");
 
     if (s->in.voice != NULL) {
-        SDL_LockAudioStream(s->in.voice);
+        SDL_ClearAudioStream(s->in.voice);
     }
     if (s->out.voice != NULL) {
-        SDL_LockAudioStream(s->out.voice);
-    }
-
-    SDL_ClearAudioStream(s->in.voice);
-    SDL_ClearAudioStream(s->out.voice);
-
-    if (s->in.voice != NULL) {
-        SDL_UnlockAudioStream(s->in.voice);
-    }
-    if (s->out.voice != NULL) {
-        SDL_UnlockAudioStream(s->out.voice);
+        SDL_ClearAudioStream(s->out.voice);
     }
 }
 
@@ -184,7 +174,6 @@ static void xblc_audio_channel_init(USBXBLCState *s, bool capture)
                                             SDL_AUDIO_DEVICE_DEFAULT_PLAYBACK;
 
     if (channel->voice != NULL) {
-        SDL_PauseAudioStreamDevice(channel->voice);
         SDL_DestroyAudioStream(channel->voice);
         channel->voice = NULL;
     }
@@ -280,18 +269,11 @@ static void usb_xblc_handle_data(USBDevice *dev, USBPacket *p)
         if (s->in.voice == NULL) {
             DPRINTF("[XBLC] Tried to get data from the input audio tream but "
                     "the audio stream is not initialized");
-            assert(false);
+            break;
         }
 
-        int available = SDL_GetAudioStreamAvailable(s->in.voice);
-        if(available < 0) {
-            DPRINTF("[XBLC] Error getting avaialable data from audio stream: %s\n", SDL_GetError());
-            assert(false);
-        }
+        to_process = p->iov.size;
 
-        to_process = MIN(p->iov.size, available);
-
-        // Read data from the stream
         while (to_process) {
             chunk_len = SDL_GetAudioStreamData(s->in.voice, s->in.packet,
                                                MIN(sizeof(s->in.packet), to_process));
@@ -302,11 +284,6 @@ static void usb_xblc_handle_data(USBDevice *dev, USBPacket *p)
             }
             usb_packet_copy(p, (void *)s->in.packet, chunk_len);
             to_process -= chunk_len;
-            available -= chunk_len;
-        }
-
-        if (available) {
-            DPRINTF("[XBLC] %d bytes remaining in the stream\n", available);
         }
 
         break;
@@ -317,14 +294,13 @@ static void usb_xblc_handle_data(USBDevice *dev, USBPacket *p)
         if (s->out.voice == NULL) {
             DPRINTF("[XBLC] Tried to put data into the speaker audio stream "
                     "but the audio stream is not initialized");
-            assert(false);
+            return;
         }
 
         if (!SDL_PutAudioStreamData(s->out.voice, p->iov.iov->iov_base,
                                     p->iov.size)) {
             DPRINTF("[XBLC] Error putting data into output stream: %s\n",
                     SDL_GetError());
-            assert(false);
         }
 
         break;
@@ -340,13 +316,11 @@ static void usb_xbox_communicator_unrealize(USBDevice *dev)
     USBXBLCState *s = USB_XBLC(dev);
 
     if (s->in.voice) {
-        SDL_PauseAudioStreamDevice(s->in.voice);
         SDL_DestroyAudioStream(s->in.voice);
         s->in.voice = NULL;
     }
     
     if (s->out.voice) {
-        SDL_PauseAudioStreamDevice(s->out.voice);
         SDL_DestroyAudioStream(s->out.voice);
         s->out.voice = NULL;
     }
--- a/hw/xbox/xblc.c
+++ b/hw/xbox/xblc.c
@@ -2,6 +2,7 @@
  * QEMU USB Xbox Live Communicator (XBLC) Device
  *
  * Copyright (c) 2022 Ryan Wendland
+ * Copyright (c) 2025 faha223
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
@@ -58,26 +59,22 @@ static const uint16_t xblc_sample_rates[5] = {
     8000, 11025, 16000, 22050, 24000
 };
 
+typedef struct XBLCStream {
+    SDL_AudioDeviceID device_id;
+    SDL_AudioStream *voice;
+    SDL_AudioSpec spec;
+    uint8_t packet[XBLC_MAX_PACKET];
+    Fifo8 fifo;
+} XBLCStream;
+
 typedef struct USBXBLCState {
     USBDevice dev;
     uint8_t   device_index;
     uint8_t   auto_gain_control;
     uint16_t  sample_rate;
 
-    AudioBackend *audio_be;
-    struct audsettings as;
-
-    struct {
-        SWVoiceOut* voice;
-        uint8_t packet[XBLC_MAX_PACKET];
-        Fifo8 fifo;
-    } out;
-
-    struct {
-        SWVoiceIn *voice;
-        uint8_t packet[XBLC_MAX_PACKET];
-        Fifo8 fifo;
-    } in;
+    XBLCStream out;
+    XBLCStream in;
 } USBXBLCState;
 
 enum {
@@ -165,82 +162,119 @@ static void usb_xblc_handle_reset(USBDevice *dev)
     USBXBLCState *s = (USBXBLCState *)dev;
 
     DPRINTF("[XBLC] Reset\n");
+
+    if(s->in.voice != 0)
+        SDL_LockAudioStream(s->in.voice);
+    if(s->out.voice != 0)
+        SDL_LockAudioStream(s->out.voice);
+
     fifo8_reset(&s->in.fifo);
     fifo8_reset(&s->out.fifo);
+
+    if(s->in.voice != 0)
+        SDL_UnlockAudioStream(s->in.voice);
+    if(s->out.voice != 0)
+        SDL_UnlockAudioStream(s->out.voice);
 }
 
-static void output_callback(void *opaque, int avail)
+static void output_callback(void *userdata, SDL_AudioStream *stream, int additional_amount, int total_amount)
 {
-    USBXBLCState *s = (USBXBLCState *)opaque;
+    USBXBLCState *s = (USBXBLCState *)userdata;
     const uint8_t *data;
-    uint32_t processed, max_len;
+    uint32_t max_len;
 
     // Not enough data to send, wait a bit longer, fill with silence for now
     if (fifo8_num_used(&s->out.fifo) < XBLC_MAX_PACKET) {
-        do {
-            processed = AUD_write(s->out.voice, (void *)silence, ARRAY_SIZE(silence));
-            avail -= processed;
-        } while (avail > 0 && processed >= XBLC_MAX_PACKET);
-        return;
-    }
-
-    // Write speaker data into audio backend
-    while (avail > 0 && !fifo8_is_empty(&s->out.fifo)) {
-        max_len   = MIN(fifo8_num_used(&s->out.fifo), avail);
-        data      = fifo8_pop_bufptr(&s->out.fifo, max_len, &max_len);
-        processed = AUD_write(s->out.voice, (void *)data, max_len);
-        avail    -= processed;
-        if (processed < max_len) return;
+        SDL_PutAudioStreamData(stream, (void*)silence, MIN(total_amount, ARRAY_SIZE(silence)));
+    } else {
+        // Write speaker data into audio backend
+        while (total_amount > 0 && !fifo8_is_empty(&s->out.fifo)) {
+            max_len = MIN(fifo8_num_used(&s->out.fifo), (uint32_t)total_amount);
+            data    = fifo8_pop_bufptr(&s->out.fifo, max_len, &max_len);
+            SDL_PutAudioStreamData(stream, data, max_len);
+            total_amount -= max_len;
+        }
     }
 }
 
-static void input_callback(void *opaque, int avail)
+static void input_callback(void *userdata, SDL_AudioStream *stream, int additional_amount, int total_amount)
 {
-    USBXBLCState *s = (USBXBLCState *)opaque;
-    uint32_t processed, max_len;
+    USBXBLCState *s = (USBXBLCState *)userdata;
+    static uint8_t buffer[XBLC_FIFO_SIZE];
+    
+    // Don't try to put more into the queue than will fit
+    uint32_t total_bytes_read = 0;
+    uint32_t max_len = MIN(total_amount, fifo8_num_free(&s->in.fifo));
+    if (max_len > 0) {
+        memset(buffer, 0, max_len);
+        int bytes_read = SDL_GetAudioStreamData(stream, buffer, max_len);
+        if(bytes_read > 0) {
+            total_bytes_read += bytes_read;
+            fifo8_push_all(&s->in.fifo, buffer, bytes_read);
+        } else if(bytes_read < 0) {
+            DPRINTF("[xblc] Error Getting audio stream data: %s\n", SDL_GetError());
+        }
+    }
+            
+    // Clear out the remainder of the input buffer
+    int bytes_read = 1;
+    while(bytes_read > 0 && total_bytes_read < total_amount) {
+        bytes_read = SDL_GetAudioStreamData(stream, buffer, MIN(XBLC_FIFO_SIZE, total_amount - total_bytes_read));
+        if(bytes_read > 0)
+            total_bytes_read += bytes_read;
+    }
+}
 
-    // Get microphone data from audio backend
-    while (avail > 0 && !fifo8_is_full(&s->in.fifo)) {
-        max_len   = MIN(sizeof(s->in.packet), fifo8_num_free(&s->in.fifo));
-        processed = AUD_read(s->in.voice, s->in.packet, max_len);
-        avail    -= processed;
-        fifo8_push_all(&s->in.fifo, s->in.packet, processed);
-        if (processed < max_len) return;
+static void xblc_audio_channel_init(USBXBLCState *s, bool capture)
+{
+    XBLCStream *channel = capture ? &s->in : &s->out;
+    SDL_AudioDeviceID device_id = capture ?
+        SDL_AUDIO_DEVICE_DEFAULT_RECORDING :
+        SDL_AUDIO_DEVICE_DEFAULT_PLAYBACK;
+    
+    if(channel->voice != 0) {
+        SDL_PauseAudioStreamDevice(channel->voice);
+        SDL_DestroyAudioStream(channel->voice);
+        channel->voice = 0;
     }
 
-    // Flush excess/old data - this can happen if the user program stops the iso transfers after it
-    // has setup the xblc.
-    while (avail > 0)
-    {
-        processed = AUD_read(s->in.voice, s->in.packet, XBLC_MAX_PACKET);
-        avail -= processed;
-        if (processed == 0) break;
-    }
+    channel->device_id = device_id;
+    
+    fifo8_reset(&channel->fifo);
+    
+    channel->spec.channels = 1;
+    channel->spec.freq = s->sample_rate;
+    channel->spec.format = SDL_AUDIO_S16LE;
+
+    channel->voice = SDL_OpenAudioDeviceStream(device_id,
+                                               &channel->spec, 
+                                               capture ? input_callback : output_callback,
+                                               (void*)s);
+    
+    SDL_ResumeAudioStreamDevice(channel->voice);
 }
 
 static void xblc_audio_stream_init(USBDevice *dev, uint16_t sample_rate)
 {
     USBXBLCState *s = (USBXBLCState *)dev;
+    bool init_input_stream = false, 
+         init_output_stream = false;
 
-    AUD_set_active_out(s->out.voice, FALSE);
-    AUD_set_active_in(s->in.voice, FALSE);
+    if(s->sample_rate != sample_rate) {
+        init_input_stream = true;
+        init_output_stream = true;
+        s->sample_rate = sample_rate;
+    }
 
-    fifo8_reset(&s->in.fifo);
-    fifo8_reset(&s->out.fifo);
+    if(s->in.voice == 0)
+        init_input_stream = true;
+    if(s->out.voice == 0)
+        init_output_stream = true;
 
-    s->as.freq = sample_rate;
-    s->as.nchannels = 1;
-    s->as.fmt = AUDIO_FORMAT_S16;
-    s->as.endianness = 0;
-
-    s->out.voice = AUD_open_out(s->audio_be, s->out.voice, TYPE_USB_XBLC "-speaker",
-                                s, output_callback, &s->as);
-
-    s->in.voice = AUD_open_in(s->audio_be, s->in.voice, TYPE_USB_XBLC "-microphone",
-                                s, input_callback, &s->as);
-
-    AUD_set_active_out(s->out.voice, TRUE);
-    AUD_set_active_in(s->in.voice, TRUE);
+    if(init_input_stream || init_output_stream) {
+        xblc_audio_channel_init(s, true);
+        xblc_audio_channel_init(s, false);
+    }
     DPRINTF("[XBLC] Init audio streams at %d Hz\n", sample_rate);
 }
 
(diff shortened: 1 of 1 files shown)


hw/xbox/xblc.c
```diff
@@ -165,82 +162,119 @@ static void usb_xblc_handle_reset(USBDevice *dev)
...
+static void input_callback(void *userdata, SDL_AudioStream *stream, int additional_amount, int total_amount)
+{
+    USBXBLCState *s = (USBXBLCState *)userdata;
+    static uint8_t buffer[XBLC_FIFO_SIZE];
+    
+    // Don't try to put more into the queue than will fit
+    uint32_t total_bytes_read = 0;
+    uint32_t max_len = MIN(total_amount, fifo8_num_free(&s->in.fifo));
+    if (max_len > 0) {
+        memset(buffer, 0, max_len);
```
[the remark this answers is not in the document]
faha223: I implemented the suggested change

hw/xbox/xblc.c
```diff
@@ -165,82 +162,119 @@ static void usb_xblc_handle_reset(USBDevice *dev)
...
+        init_input_stream = true;
+    if(s->out.voice == 0)
+        init_output_stream = true;
 
-    AUD_set_active_out(s->out.voice, TRUE);
-    AUD_set_active_in(s->in.voice, TRUE);
+    if(init_input_stream || init_output_stream) {
+        xblc_audio_channel_init(s, true);
+        xblc_audio_channel_init(s, false);
+    }
```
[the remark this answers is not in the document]
faha223: I implemented the suggested change

hw/xbox/xblc.c
```diff
@@ -313,26 +349,22 @@ static void usb_xblc_handle_data(USBDevice *dev, USBPacket *p)
...
+    
     fifo8_destroy(&s->out.fifo);
     fifo8_destroy(&s->in.fifo);
 
-    AUD_close_out(s->audio_be, s->out.voice);
-    AUD_close_in(s->audio_be, s->in.voice);
+    SDL_DestroyAudioStream(s->in.voice);
+    s->in.voice = 0;
+    SDL_DestroyAudioStream(s->out.voice);
+    s->out.voice = 0;
```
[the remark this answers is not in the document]
faha223: I implemented the suggested change

hw/xbox/xblc.c
```diff
@@ -165,82 +162,119 @@ static void usb_xblc_handle_reset(USBDevice *dev)
...
+    channel->spec.channels = 1;
+    channel->spec.freq = s->sample_rate;
+    channel->spec.format = SDL_AUDIO_S16LE;
+
+    channel->voice = SDL_OpenAudioDeviceStream(device_id,
+                                               &channel->spec, 
+                                               capture ? input_callback : output_callback,
+                                               (void*)s);
+    
+    SDL_ResumeAudioStreamDevice(channel->voice);
```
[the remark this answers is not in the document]
faha223: I implemented the suggested change

hw/xbox/xblc.c
```diff
@@ -165,82 +162,124 @@ static void usb_xblc_handle_reset(USBDevice *dev)
...
-    while (avail > 0 && !fifo8_is_empty(&s->out.fifo)) {
-        max_len   = MIN(fifo8_num_used(&s->out.fifo), avail);
-        data      = fifo8_pop_bufptr(&s->out.fifo, max_len, &max_len);
-        processed = AUD_write(s->out.voice, (void *)data, max_len);
-        avail    -= processed;
-        if (processed < max_len) return;
+static void input_callback(void *userdata, SDL_AudioStream *stream, int additional_amount, int total_amount)
+{
+    USBXBLCState *s = (USBXBLCState *)userdata;
+    static uint8_t buffer[XBLC_FIFO_SIZE];
```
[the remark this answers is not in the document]
faha223: I moved this buffer into the USBXBLCState structure because the output stream doesn't need it

hw/xbox/xblc.c
```diff
@@ -165,82 +164,130 @@ static void usb_xblc_handle_reset(USBDevice *dev)
     USBXBLCState *s = (USBXBLCState *)dev;
 
     DPRINTF("[XBLC] Reset\n");
+
+    if (s->in.voice != 0) {
```
mborgerson: nit: Since this is a pointer, prefer checking against `NULL` instead

hw/xbox/xblc.c
```diff
@@ -2,6 +2,7 @@
  * QEMU USB Xbox Live Communicator (XBLC) Device
  *
  * Copyright (c) 2022 Ryan Wendland
+ * Copyright (c) 2025 faha223
```
mborgerson: If you add your name here, prefer to use your real name

hw/xbox/xblc.c
```diff
@@ -165,82 +164,130 @@ static void usb_xblc_handle_reset(USBDevice *dev)
...
+}
 
-    // Write speaker data into audio backend
-    while (avail > 0 && !fifo8_is_empty(&s->out.fifo)) {
-        max_len   = MIN(fifo8_num_used(&s->out.fifo), avail);
-        data      = fifo8_pop_bufptr(&s->out.fifo, max_len, &max_len);
-        processed = AUD_write(s->out.voice, (void *)data, max_len);
-        avail    -= processed;
-        if (processed < max_len) return;
+static void input_callback(void *userdata, SDL_AudioStream *stream, int additional_amount, int total_amount)
```
mborgerson: SDL3 enables us to simplify this code and drop these extra fifos and callbacks. Let's refactor this code to use the direct stream approach instead of callbacks

hw/xbox/xblc.c
```diff
@@ -165,82 +164,130 @@ static void usb_xblc_handle_reset(USBDevice *dev)
...
+            fifo8_push_all(&s->in.fifo, s->input_buffer, bytes_read);
+        } else if(bytes_read < 0) {
+            DPRINTF("[xblc] Error Getting audio stream data: %s\n", SDL_GetError());
+        }
+    }
+            
+    // Clear out the remainder of the input buffer
+    int bytes_read = 1;
+    while (bytes_read > 0 && total_bytes_read < total_amount) {
+        bytes_read = SDL_GetAudioStreamData(stream, s->input_buffer, MIN(XBLC_FIFO_SIZE, total_amount - total_bytes_read));
```
mborgerson: If `SDL_GetAudioStreamData` is returning <0, something is wrong. We should not hammer on it here

hw/xbox/xblc.c
```diff
@@ -165,82 +164,130 @@ static void usb_xblc_handle_reset(USBDevice *dev)
...
-    s->as.nchannels = 1;
-    s->as.fmt = AUDIO_FORMAT_S16;
-    s->as.endianness = 0;
-
-    s->out.voice = AUD_open_out(s->audio_be, s->out.voice, TYPE_USB_XBLC "-speaker",
-                                s, output_callback, &s->as);
+    if (s->sample_rate != sample_rate) {
+        init_input_stream = true;
+        init_output_stream = true;
+        s->sample_rate = sample_rate;
```
mborgerson: We shouldn't need to recreate the stream to change the sample rate. SDL3 offers an API for this [SDL_SetAudioStreamFormat](https://wiki.libsdl.org/SDL3/SDL_SetAudioStreamFormat)

hw/xbox/xblc.c
faha223: If `SDL_GetAudioStreamData` returns <0 then it exits the loop. I'm not sure what else to do when that happens.

hw/xbox/xblc.c
faha223: I changed this function so that it uses `SDL_SetAudioStreamFormat` if the streams are not NULL instead of calling `xblc_audio_channel_init`

hw/xbox/xblc.c
faha223: There were a lot of instances where that was being compared or assigned 0 instead of NULL. I went through and fixed them

hw/xbox/xblc.c
faha223: I've refactored the xblc to remove the fifo queues and callbacks, and I'm not directly calling `SDL_PutAudioStreamData` and `SDL_GetAudioStreamData` inside the `usb_xblc_handle_data` function. It appears that it still works. There's slightly less latency now when testing with N.U.D.E.

hw/xbox/xblc.c
faha223: Changed to my real name

hw/xbox/xblc.c
```diff
@@ -283,56 +268,67 @@ static void usb_xblc_handle_control(USBDevice *dev, USBPacket *p,
...
-        fifo8_push_all(&s->out.fifo, s->out.packet, to_process);
+
+        // Process the desired amount of data
+        to_process = p->iov.size;
+
+        // In case the out packet is smallerthan the iov size, do it in chunks
+        while(to_process) {
+            chunk_len = MIN(p->iov.size, sizeof(s->out.packet));
+            usb_packet_copy(p, (void *)s->out.packet, chunk_len);
+            SDL_PutAudioStreamData(s->out.voice, s->out.packet, chunk_len);
```
mborgerson: * Is it possible that `s->out.voice` is not initialized here? E.g. if `xblc_audio_stream_init` is never called
* `SDL_PutAudioStreamData` may fail
* Why do we need this intermediate buffer `s->out.packet`?

hw/xbox/xblc.c
```diff
@@ -283,56 +268,67 @@ static void usb_xblc_handle_control(USBDevice *dev, USBPacket *p,
...
         assert(p->ep->nr == XBLC_EP_IN);
-        to_process = MIN(fifo8_num_used(&s->in.fifo), p->iov.size);
+        to_process = p->iov.size;
         chunk_len = 0;
 
-        // fifo may not give us a contiguous packet, so may need multiple calls
         while (to_process) {
-            const uint8_t *packet = fifo8_pop_bufptr(&s->in.fifo, to_process, &chunk_len);
-            usb_packet_copy(p, (void *)packet, chunk_len);
+            chunk_len = SDL_GetAudioStreamData(s->in.voice, s->in.packet, MIN(to_process, p->iov.size));
```
mborgerson: * Is it possible that `s->in.voice` is not initialized here? E.g. if `xblc_audio_stream_init` is never called
*  `to_process` will always be <= `p->iov.size`

hw/xbox/xblc.c
```diff
@@ -283,56 +268,67 @@ static void usb_xblc_handle_control(USBDevice *dev, USBPacket *p,
...
+            if(chunk_len < 0)
```
mborgerson: nit: Needs braces

hw/xbox/xblc.c
```diff
@@ -283,56 +268,67 @@ static void usb_xblc_handle_control(USBDevice *dev, USBPacket *p,
...
     case USB_TOKEN_OUT:
-        // Speaker data - get data from usb packet then push to fifo.
+        // Speaker data - get data from usb packet then put directly into audio stream.
         assert(p->ep->nr == XBLC_EP_OUT);
-        to_process = MIN(fifo8_num_free(&s->out.fifo), p->iov.size);
-        usb_packet_copy(p, s->out.packet, to_process);
-        fifo8_push_all(&s->out.fifo, s->out.packet, to_process);
+
+        // Process the desired amount of data
+        to_process = p->iov.size;
```
mborgerson: Comment is redundant

hw/xbox/xblc.c
```diff
@@ -283,56 +268,67 @@ static void usb_xblc_handle_control(USBDevice *dev, USBPacket *p,
...
 static void usb_xbox_communicator_unrealize(USBDevice *dev)
 {
     USBXBLCState *s = USB_XBLC(dev);
 
-    AUD_set_active_out(s->out.voice, false);
-    AUD_set_active_in(s->in.voice, false);
-
-    fifo8_destroy(&s->out.fifo);
-    fifo8_destroy(&s->in.fifo);
+    if (s->in.voice) {
```
mborgerson, line 339: Why not also just merge conditions and handle destruction here too

hw/xbox/xblc.c
```diff
@@ -283,56 +268,67 @@ static void usb_xblc_handle_control(USBDevice *dev, USBPacket *p,
...
+        // Speaker data - get data from usb packet then put directly into audio stream.
         assert(p->ep->nr == XBLC_EP_OUT);
-        to_process = MIN(fifo8_num_free(&s->out.fifo), p->iov.size);
-        usb_packet_copy(p, s->out.packet, to_process);
-        fifo8_push_all(&s->out.fifo, s->out.packet, to_process);
+
+        // Process the desired amount of data
+        to_process = p->iov.size;
+
+        // In case the out packet is smallerthan the iov size, do it in chunks
```
mborgerson: Redundant comment

hw/xbox/xblc.c
```diff
@@ -283,56 +268,67 @@ static void usb_xblc_handle_control(USBDevice *dev, USBPacket *p,
...
+        while(to_process) {
```
mborgerson: nit: Spaces. In this case feel free to just run clang-format on this file to normalize whitespace, clang-format config in root dir

hw/xbox/xblc.c
faha223: - I'm not sure
- That's true. I'm not sure what to do when it fails though
- It turns out we don't. I can just call `SDL_PutAudioStreamData` using `p->iov.iov->iov_base` directly

faha223: I suppose if it fails I should log the error and `assert(false);`

hw/xbox/xblc.c
faha223: I hadn't noticed that they could be merged when I removed the fifo code here. I've merged the if statements

hw/xbox/xblc.c
```diff
@@ -165,115 +158,107 @@ static void usb_xblc_handle_reset(USBDevice *dev)
...
-
-    // Not enough data to send, wait a bit longer, fill with silence for now
-    if (fifo8_num_used(&s->out.fifo) < XBLC_MAX_PACKET) {
-        do {
-            processed = AUD_write(s->out.voice, (void *)silence, ARRAY_SIZE(silence));
-            avail -= processed;
-        } while (avail > 0 && processed >= XBLC_MAX_PACKET);
-        return;
+    if (s->in.voice != NULL) {
+        SDL_LockAudioStream(s->in.voice);
```
mborgerson: Why is it necessary to lock the audio streams?

According to the docs for [SDL_ClearAudioStream](https://wiki.libsdl.org/SDL3/SDL_LockAudioStream):

> It is safe to call this function from any thread.

hw/xbox/xblc.c
```diff
@@ -165,115 +158,107 @@ static void usb_xblc_handle_reset(USBDevice *dev)
...
-        processed = AUD_read(s->in.voice, s->in.packet, max_len);
-        avail    -= processed;
-        fifo8_push_all(&s->in.fifo, s->in.packet, processed);
-        if (processed < max_len) return;
+    XBLCStream *channel = capture ? &s->in : &s->out;
+    SDL_AudioDeviceID device_id = capture ? SDL_AUDIO_DEVICE_DEFAULT_RECORDING :
+                                            SDL_AUDIO_DEVICE_DEFAULT_PLAYBACK;
+
+    if (channel->voice != NULL) {
+        SDL_PauseAudioStreamDevice(channel->voice);
```
mborgerson: Why are we pausing the stream before destruction? I think SDL_DestroyAudioStream should implicitly pause a stream

hw/xbox/xblc.c
```diff
@@ -283,65 +268,97 @@ static void usb_xblc_handle_control(USBDevice *dev, USBPacket *p,
...
+        // Microphone Data - Get data from audio stream and copy into usb packet
         assert(p->ep->nr == XBLC_EP_IN);
-        to_process = MIN(fifo8_num_used(&s->in.fifo), p->iov.size);
         chunk_len = 0;
 
-        // fifo may not give us a contiguous packet, so may need multiple calls
+        if (s->in.voice == NULL) {
+            DPRINTF("[XBLC] Tried to get data from the input audio tream but "
+                    "the audio stream is not initialized");
+            assert(false);
```
mborgerson: Why `assert`? xblc emulation should still work, even if we failed to bring up audio streams

hw/xbox/xblc.c
```diff
@@ -283,65 +268,97 @@ static void usb_xblc_handle_control(USBDevice *dev, USBPacket *p,
...
+
+        int available = SDL_GetAudioStreamAvailable(s->in.voice);
+        if(available < 0) {
+            DPRINTF("[XBLC] Error getting avaialable data from audio stream: %s\n", SDL_GetError());
+            assert(false);
+        }
+
+        to_process = MIN(p->iov.size, available);
+
+        // Read data from the stream
```
mborgerson: Redundant comment

hw/xbox/xblc.c
```diff
@@ -283,65 +268,97 @@ static void usb_xblc_handle_control(USBDevice *dev, USBPacket *p,
...
         chunk_len = 0;
 
-        // fifo may not give us a contiguous packet, so may need multiple calls
+        if (s->in.voice == NULL) {
+            DPRINTF("[XBLC] Tried to get data from the input audio tream but "
+                    "the audio stream is not initialized");
+            assert(false);
+        }
+
+        int available = SDL_GetAudioStreamAvailable(s->in.voice);
```
mborgerson: Do we need to check if there are data available? What if we just attempt to directly SDL_GetAudioStreamData instead

hw/xbox/xblc.c
```diff
@@ -283,65 +268,97 @@ static void usb_xblc_handle_control(USBDevice *dev, USBPacket *p,
...
+        // Speaker data - put data from usb packet directly into audio stream.
         assert(p->ep->nr == XBLC_EP_OUT);
-        to_process = MIN(fifo8_num_free(&s->out.fifo), p->iov.size);
-        usb_packet_copy(p, s->out.packet, to_process);
-        fifo8_push_all(&s->out.fifo, s->out.packet, to_process);
+
+        if (s->out.voice == NULL) {
+            DPRINTF("[XBLC] Tried to put data into the speaker audio stream "
+                    "but the audio stream is not initialized");
+            assert(false);
```
mborgerson: Same comment as above. We should still be able to emulate an XBLC even if SDL3 streams aren't available

hw/xbox/xblc.c
```diff
@@ -283,65 +268,97 @@ static void usb_xblc_handle_control(USBDevice *dev, USBPacket *p,
...
+        }
+
+        if (!SDL_PutAudioStreamData(s->out.voice, p->iov.iov->iov_base,
+                                    p->iov.size)) {
+            DPRINTF("[XBLC] Error putting data into output stream: %s\n",
+                    SDL_GetError());
+            assert(false);
```
mborgerson: Same comment as above

hw/xbox/xblc.c
```diff
@@ -165,115 +158,107 @@ static void usb_xblc_handle_reset(USBDevice *dev)
...
     }
 
-    // Write speaker data into audio backend
-    while (avail > 0 && !fifo8_is_empty(&s->out.fifo)) {
-        max_len   = MIN(fifo8_num_used(&s->out.fifo), avail);
-        data      = fifo8_pop_bufptr(&s->out.fifo, max_len, &max_len);
-        processed = AUD_write(s->out.voice, (void *)data, max_len);
-        avail    -= processed;
-        if (processed < max_len) return;
+    SDL_ClearAudioStream(s->in.voice);
```
mborgerson: * Let's check that the `stream != NULL` before calling `SDL_ClearAudioStream`
* ~~Should we be clearing input here? See comment below~~ Actually clearing here probably makes sense to discard queued data that might increase latency

hw/xbox/xblc.c
```diff
@@ -165,115 +158,107 @@ static void usb_xblc_handle_reset(USBDevice *dev)
...
+    if (s->out.voice == NULL) {
+        xblc_audio_channel_init(s, false);
+    } else if (s->out.spec.freq != sample_rate) {
+        s->out.spec.freq = sample_rate;
+        SDL_SetAudioStreamFormat(s->out.voice, &s->out.spec, &s->out.spec);
+    }
 
-    s->in.voice = AUD_open_in(s->audio_be, s->in.voice, TYPE_USB_XBLC "-microphone",
-                                s, input_callback, &s->as);
+    SDL_ClearAudioStream(s->in.voice);
```
mborgerson: * Should only clear a stream if it exists. If we initialized a new one, we probably don't need to clear it?
* According to https://wiki.libsdl.org/SDL3/SDL_SetAudioStreamFormat it appears you do not need to clear streams when the format changes, so maybe we can just drop the clear

hw/xbox/xblc.c
```diff
@@ -283,65 +268,97 @@ static void usb_xblc_handle_control(USBDevice *dev, USBPacket *p,
...
-    AUD_set_active_out(s->out.voice, false);
-    AUD_set_active_in(s->in.voice, false);
-
-    fifo8_destroy(&s->out.fifo);
-    fifo8_destroy(&s->in.fifo);
-
-    AUD_close_out(s->audio_be, s->out.voice);
-    AUD_close_in(s->audio_be, s->in.voice);
+    if (s->in.voice) {
+        SDL_PauseAudioStreamDevice(s->in.voice);
```
mborgerson: Same comment as above. Do we really need to pause before destroying a stream?

hw/xbox/xblc.c
faha223: We don't need to check. It's just useful info to have when debugging. I'll remove it

hw/xbox/xblc.c
faha223: It appears that we don't, so I removed the call to SDL_PauseAudioStreamDevice

hw/xbox/xblc.c
faha223: There were 2 reasons why I added this:

- When calling SDL_SetAudiostreamFormat, any samples already in the queue will be in the previous format
- This solves a problem I was seeing in N.U.D.E. where the queue was filling up with data while the mic was "off" (not being read, but still active). I noticed that every time the mic starts to be used a command is sent to set the sample rate, so I use that as a cue to clear the queue

hw/xbox/xblc.c
faha223: This was leftover from when we were using the callbacks and managing the queued data ourselves. This isn't necessary so I removed it.

hw/xbox/xblc.c
faha223: Leftover from SDL2 and from when we were using callbacks. I've removed these locks

hw/xbox/xblc.c
mborgerson: > When calling SDL_SetAudiostreamFormat, any samples already in the queue will be in the previous format

According to [SDL_SetAudioStreamFormat](https://wiki.libsdl.org/SDL3/SDL_SetAudioStreamFormat) docs:

> Future calls to and [SDL_GetAudioStreamAvailable](https://wiki.libsdl.org/SDL3/SDL_GetAudioStreamAvailable) and [SDL_GetAudioStreamData](https://wiki.libsdl.org/SDL3/SDL_GetAudioStreamData) will reflect the new format, and future calls to [SDL_PutAudioStreamData](https://wiki.libsdl.org/SDL3/SDL_PutAudioStreamData) must provide data in the new input formats.

So I don't think we need to worry about old format of samples--SDL should handle this for us.

> his solves a problem I was seeing in N.U.D.E. where the queue was filling up with data while the mic was "off" (not being read, but still active)

This makes sense. But isn't this going to be done in `usb_xblc_handle_reset`?

hw/xbox/xblc.c
```diff
@@ -283,65 +255,98 @@ static void usb_xblc_handle_control(USBDevice *dev, USBPacket *p,
...
+        }
+
+        to_process = p->iov.size;
+
         while (to_process) {
-            const uint8_t *packet = fifo8_pop_bufptr(&s->in.fifo, to_process, &chunk_len);
-            usb_packet_copy(p, (void *)packet, chunk_len);
+            chunk_len = SDL_GetAudioStreamData(s->in.voice, s->in.packet,
+                                               MIN(sizeof(s->in.packet), to_process));
+            if (chunk_len < 0) {
```
mborgerson, line 293: * If we break out here, should we be providing silence to the guest as before?
* If SDL_GetAudioStreamData doesn't have enough data to provide, it will spin here and block emulation. Previous behavior was to pull as much as possible from the FIFO and provide silence for the rest of the packet. Blocking might be better to introduce some latency in order to avoid underruns, but totally blocking emulation here is not ideal. A simple solution to both underruns and blocking could be to just provide silence for a while until the stream has buffered at least a full packets worth of data plus some headroom on the input stream.

hw/xbox/xblc.c
```diff
@@ -41,43 +42,38 @@
...
-    8000, 11025, 16000, 22050, 24000
-};
+static const uint16_t xblc_sample_rates[5] = { 8000, 11025, 16000, 22050,
+                                               24000 };
+
+typedef struct XBLCStream {
+    SDL_AudioDeviceID device_id;
+    SDL_AudioStream *voice;
+    SDL_AudioSpec spec;
+    uint8_t packet[XBLC_MAX_PACKET];
```
mborgerson: Do we need the `packet` field anymore? Looks like it's only used `usb_xblc_handle_data` `USB_TOKEN_IN` case, so we can probably replace with a local buffer there

hw/xbox/xblc.c
```diff
@@ -350,12 +355,11 @@ static void usb_xbox_communicator_realize(USBDevice *dev, Error **errp)
...
 
-    if (!AUD_backend_check(&s->audio_be, errp)) {
-        return;
-    }
+    s->in.voice = NULL;
+    s->out.voice = NULL;
 
-    fifo8_create(&s->in.fifo, XBLC_FIFO_SIZE);
-    fifo8_create(&s->out.fifo, XBLC_FIFO_SIZE);
+    // What sample rate does the physical XBLC default to?
```
mborgerson: Is this a `TODO` or `FIXME` or something else? If so, we should indicate as such. Any reason for picking the sample rate at index 1, or was this an arbitrary choice? Why not highest or lowest?

hw/xbox/xblc.c
```diff
@@ -41,43 +42,38 @@
...
+static const uint8_t silence[256] = { 0 };
 
-static const uint16_t xblc_sample_rates[5] = {
-    8000, 11025, 16000, 22050, 24000
-};
+static const uint16_t xblc_sample_rates[5] = { 8000, 11025, 16000, 22050,
+                                               24000 };
+
+typedef struct XBLCStream {
+    SDL_AudioDeviceID device_id;
```
mborgerson: Is the `device_id` field needed? It is set, but never read, from `xblc_audio_channel_init`

hw/xbox/xblc.c
faha223: We only need it for the USB_TOKEN_IN case, so it can be replaced with a local buffer. It's only 48 bytes

hw/xbox/xblc.c
faha223: Yeah, I don't see any harm in filling the buffer with silence if we don't have enough data to fill the buffer. We definitely don't want to wait for new data if we can avoid it

hw/xbox/xblc.c
faha223: It'll be needed for the UI changes, but those aren't in this PR so I'll remove it

hw/xbox/xblc.c
faha223: this was a `FIXME`, but I went looking and, according to the source code for Ryzee119's HAWK XBLC replacement, the XBLC appears to default to 16 KHz, so I changed it to 16KHz here, and added a comment explaining where I got that number.

mborgerson:
Thanks!

PR #2703, opened 2026-01-26
xemu-snapshots: fix assert

oltolm:
`bdrv_unref(bs_ro);` frees `bs_ro` so the `assert(bs_ro->refcnt == 0);` in next line is a use after free. Move the assert before the `free`.


mborgerson:
Thanks! I adjusted commit message to be more descriptive.

PR #2704, opened 2026-01-28
ui: Allow Alt to trigger RenderDoc+PGRAPH caps

abaire:
It appears that `KeyCtrl` is no longer being set for the control key on macOS. This change allows either ctrl or alt/option to trigger PGRAPH traces along with F10 RenderDoc captures.

### DIFF
--- a/ui/xui/menubar.cc
+++ b/ui/xui/menubar.cc
@@ -79,7 +79,7 @@ void ProcessKeyboardShortcuts(void)
     if (ImGui::IsKeyPressed(ImGuiKey_F10) && nv2a_dbg_renderdoc_available()) {
         ImGuiIO& io = ImGui::GetIO();
         int num_frames = io.KeyShift ? 5 : 1;
-        nv2a_dbg_renderdoc_capture_frames(num_frames, io.KeyCtrl);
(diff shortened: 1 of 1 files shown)


mborgerson:
I think Dear ImGui just maps Ctrl to Cmd key on macOS to conform to platform standards?

abaire:
> I think Dear ImGui just maps Ctrl to Cmd key on macOS to conform to platform standards?

Apparently so. It definitely did not do this in the past (I haven't used RenderDoc on Linux/Windows in years now). 

Digging around, it seems that when they implemented this they forgot to update the comments and left these completely misleading names in place in the struct. Even better, their fix lists the triggering key for `KeySuper` on macOS as "Ctrl", because who would ever be confused that a key named `Ctrl` does not set `KeyCtrl`... :)  https://github.com/ocornut/imgui/commit/da6c97203e4f80a5ff93197709b504ec9d4a4116 

I'll update my wiki docs; if we end up bringing them into the xemu-site we will need to make sure it's explained there as well.

mborgerson:
Yeah.. they should probably avoid names like Ctrl and just use Meta or something

PR #2720, opened 2026-02-03
nv2a: Fix R12 input in paired/multiout ops writing to oPos

### LINKED ISSUE #1864: Prince of Persia: The Sands of Time: Incorrect rendering of water surface
### Title

https://xemu.app/titles/5553001d/#Prince-of-Persia-The-Sands-of-Time

### Bug Description

As mentioned in last test (Jun 29, 2021) by Thrimbor#0666, water surface/reflections is broken. It looks like the image from the screen is superimposed on the surface of the water. There is no transparency of the water.
image

### Expected Behavior

Correct rendering of water surfaces and reflections, transparency of water.
image

### xemu Version

0.8.8

### System Information

CPU:          Intel(R) Core(TM) i5-9500 CPU @ 3.00GHz
OS Platform:  Windows
OS Version:    22H2
GPU Model:    NVIDIA GeForce GTX 1650/PCIe/SSE2
Driver:       4.0.0 NVIDIA 560.94

### Additional Context

_No response_

abaire:
Paired MAC+ILU and duplicate output instructions can write to oPos while reading from R12. This leads to a case where xemu's serialized emulation erroneously uses the output of a previous instruction when calculating the value of a later one.

E.g., in

```
 /* 0x00000000 0x0080201A 0xC4002868 0x7CB0E800 */
 MAD oPos.xyz, R12.xyz, R1.x, C[1].xyz
   +  MAD R11.xy, R12.xyz, R1.x, C[1].xyz
```

the value of oPos prior to the first instruction should be used for both MAD calculations.

This could alternatively be fixed by writing oPos to a temp vector and deferring the output vector update until after the token is fully processed.

[Tests](https://github.com/abaire/nxdk_pgraph_tests/blob/1745a45290a5d607f9582303d6882cbf62f003de/src/tests/vertex_shader_independence_tests.cpp#L125)
[HW results](https://abaire.github.io/nxdk_pgraph_tests_golden_results/results/Vertex_shader_independence_tests/index.html)
[PR results](https://abaire.github.io/xemu-dev_pgraph_test_results/fix_1864_r12_input_parallel_with_opos_write/index.html)

Fixes #1864

### DIFF
--- a/hw/xbox/nv2a/pgraph/glsl/vsh-prog.c
+++ b/hw/xbox/nv2a/pgraph/glsl/vsh-prog.c
@@ -209,6 +209,7 @@ static bool ilu_force_scalar[] = {
     false,
 };
 
+#define OUTPUT_REG_R12 0
 #define OUTPUT_REG_FOG 5
 
 static const char* out_reg_name[] = {
@@ -296,14 +297,14 @@ static MString *decode_swizzle(const uint32_t *shader_token,
 
 static MString *decode_opcode_input(const uint32_t *shader_token,
                                     VshParameterType param,
-                                    VshFieldName neg_field, int reg_num)
+                                    VshFieldName neg_field, int reg_num,
+                                    bool *uses_r12_latch)
 {
     /* This function decodes a vertex shader opcode parameter into a string.
      * Input A, B or C is controlled via the Param and NEG fieldnames,
      * the R-register address for each input is already given by caller. */
 
     MString *ret_str = mstring_new();
-
 
     if (vsh_get_field(shader_token, neg_field) > 0) {
         mstring_append_fmt(ret_str, "-");
@@ -315,6 +316,9 @@ static MString *decode_opcode_input(const uint32_t *shader_token,
     switch (param) {
     case PARAM_R:
         snprintf(tmp, sizeof(tmp), "R%d", reg_num);
+        if (reg_num == 12) {
+            *uses_r12_latch = true;
+        }
         break;
     case PARAM_V:
         reg_num = vsh_get_field(shader_token, FLD_V);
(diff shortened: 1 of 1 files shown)


viniciusol263:
Its fixed indeed, thanks!
image

Triticum0:
Test a few game don't think any known issues are fixed or have same behaviour.

mborgerson:
Thanks for the patch!

> This could alternatively be fixed by writing oPos to a temp vector and deferring the output vector update until after the token is fully processed.

The extra register and additional latching/stale checks are adding more complexity. I think this alternative suggestion sounds simple and more maintainable--not limited to oPos, but analyze inputs/outputs to mitigate the hazard generally. We've encountered this issue already, so we should unify the approaches.

abaire:
> The extra register and additional latching/stale checks are adding more complexity. I think this alternative suggestion sounds simple and more maintainable--not limited to oPos, but analyze inputs/outputs to mitigate the hazard generally. We've encountered this issue already, so we should unify the approaches.

I don't think it can be trivially unified; the current temp register usage will only work for the MAC+ILU case.
The least complex approach is to detect writes to oPos and always defer them until after the instruction is fully processed with an additional special temp register. E.g.,

```
tempPos.mask = something;
oPos.mask = tempPos.mask
```

Then we'd be covered for the most complicated case, which would be a MAC + ILU pairing where one or the other writes to oPos + a temp reg and either/both of them read from R12. 

The downside is that we don't (currently) parse the ILU args until after the MAC is fully processed, so there may be a situation where we assign to the temp unnecessarily since a MAC write to oPos would have to cover the worst case of an R12 ILU read.

abaire:
Committed a version that piggybacks on the existing suffix workaround to conservatively defend modifications of oPos.

I think ideally this code would be refactored to have a bit more state tracking so that all inputs could be resolved with special cases (like R12 use) flagged as they're expanded into strings (we can also drop a couple string copies by threading the buffer through). Then we could reserve this hackery for the cases where `oPos` is modified before a use of `R12`. For now I'm optimistic that this simple approach won't have much perf impact since it's only triggered in MAC+ILU or multi-write situations that output position.

PR #2751, opened 2026-02-21
meson: Bump SDL3 to 3.4.2

mborgerson:
* Vendors SDL3 wrap for now. Submitted wrapdb update at https://github.com/mesonbuild/wrapdb/pull/2666


Bobble80:
Just did a quick test and the over brightened lighting issue in Xemu seems to be fixed in that new SDL update.

mborgerson:
Not gonna wait for wrapdb. We'll handle it independently for now.

PR #2786, opened 2026-03-10
ui: Do not pass missing path to SDL_ShowOpenFileDialog

### LINKED ISSUE #2785: If games directory becomes invalid, 'Load Disc..' crashes Xemu
### Bug Description

In Xemu, I never explicitly set the 'General > Games directory' option but it got set automatically when I loaded a disc. I decided to change the location where I store my disc images, so the destination specified in this option was no longer valid. Later on if I decide to 'Load Disc...', immediately upon closing the Windows file selector dialog, Xemu will either close automatically or give an application error citing memory can not be written.

1. Set 'General > Games directory' to some valid directory.
2. Remove that directory outside of Xemu
3. Start Xemu and use 'Load Disc...'
4. Close file selector dialog either by selecting a file and open, pressing cancel, or closing the window.
5. Crash occurs

### Expected Behavior

There should be an error check at some point to ensure that the value in the option is valid. If it is not valid, Xemu should automatically set the option to a fallback directory.

Also there should be extra changes in the code that loads discs that ensures that even if that option is set to an invalid directory, an attempt to access some unset value will gracefully generate an exception to be handled or error message for the user.

### xemu Version

0.8.134 Release

### System Information

OS: Windows 10

### Additional Context

_No response_

abaire:
Passing two levels of invalid paths to `SDL_ShowOpenFileDialog` on Windows causes a crash.  E.g., given a directory `d:\foo` that exists, passing `d:\foo\bar\baz` will trigger the crash.

This appears to be because the `FileDialogCallbackWrapper` callback is invoked twice in this case, leading to a double-free.

Fixes #2785

### DIFF
--- a/ui/xui/misc.cc
+++ b/ui/xui/misc.cc
@@ -70,6 +70,10 @@ static std::string NormalizeDefaultLocation(const char *default_location)
         if (std::filesystem::is_directory(path)) {
             return (path / "").string();
         }
+        // Prevent a crash in SDL3 file dialog
+        if (!std::filesystem::exists(path)) {
+            return {};
+        }
 #endif
     } catch (...) {
         // Fall through to return original path


mborgerson:
Thanks!

SDL should probably handle this

abaire:
https://github.com/libsdl-org/SDL/issues/15194 though I'd argue that the bug you filed about inconsistent behavior should cover this as well.

PR #2768, opened 2026-03-02
fix(vulkan): Add MoltenVK/Metal compatibility for macOS

CosmicSnow:
This commit introduces comprehensive fixes to ensure the Vulkan renderer works correctly on macOS through MoltenVK (Vulkan-over-Metal translation). MoltenVK has several limitations compared to native Vulkan drivers that require workarounds for proper Xbox NV2A emulation.

## Motivation

MoltenVK (the Vulkan portability implementation for macOS/iOS via Metal) has the following critical limitations that broke xemu's Vulkan renderer:

1. No native Geometry Shader support (Metal API limitation)
2. No VK_EXT_provoking_vertex extension (Metal uses fixed "first" provoking)
3. No VK_KHR_external_memory native support
4. Requires explicit VK_KHR_portability_subset extension enrollment

### 1. Instance Creation & Device Enumeration (hw/xbox/nv2a/pgraph/vk/instance.c)

- Added `VK_KHR_PORTABILITY_SUBSET_EXTENSION_NAME` to required device extensions on `__APPLE__`. Without this, MoltenVK hides the physical device and Vulkan initialization fails on Apple Silicon.

- Added `VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME` support with `VK_INSTANCE_CREATE_ENUMERATE_PORTABILITY_BIT_KHR` flag. This allows MoltenVK devices to be enumerated during instance creation.

- Defined fallback macro for `VK_KHR_PORTABILITY_SUBSET_EXTENSION_NAME` for older Vulkan headers that may lack it despite runtime support.

### 2. External Memory Disable (hw/xbox/nv2a/pgraph/vk/renderer.h)

- Set `HAVE_EXTERNAL_MEMORY` to 0 on macOS. MoltenVK does not support native external memory extensions, and attempting to use them causes crashes during device initialization. This disables the `VK_KHR_external_memory_*` code paths on Apple platforms.

### 3. Geometry Shader Feature Detection (hw/xbox/nv2a/pgraph/vk/instance.c)

- Query and store `geometryShader` physical device feature in `supports_geometry_shaders` flag. MoltenVK reports this as false since Metal lacks geometry shader support. This flag drives the primitive emulation fallback logic.

### 4. Primitive Topology Mapping without Geometry Shaders (hw/xbox/nv2a/pgraph/vk/draw.c)

- Modified `get_primitive_topology()` to map NV2A primitive types directly to Vulkan topologies when geometry shaders are unavailable:
[...]

PR #2799, opened 2026-03-17
macOS/MoltenVK: Vulkan renderer support for ARM Macs

t-chamb:
Reworked per the split discussion: this PR now contains **only** the macOS/MoltenVK enablement (1 commit, 10 files, +174/−1). The platform-neutral pipeline/SPIR-V cache persistence moved to #2993.

What this does:

- Enables `VK_KHR_portability_enumeration` on the instance and accepts `VK_KHR_portability_subset` devices, so MoltenVK is usable as the Vulkan implementation on Apple Silicon
- Bundles the Vulkan loader and MoltenVK into the app bundle with an ICD manifest; `VK_DRIVER_FILES` is set programmatically in `instance.c` so Finder launches (which don't inherit terminal env) find the bundled ICD, with an `Info.plist` `LSEnvironment` fallback
- Adapts buffer, surface, texture and display paths for MoltenVK limitations
- Adds the Homebrew pkg-config path on macOS so the Vulkan SDK is found at configure time

Tested on an M3 Ultra (macOS 26): renderer initializes and boots titles; Halo 2 measured notably smoother and more consistent than the GL 4.1 path.

@mborgerson I know you mentioned on #2768 you plan to handle macOS Vulkan properly — happy to adapt this to whatever direction you take, or treat it as a reference. The IOSurface question from earlier in this thread still stands, since my frame-interpolation experiments hinge on what the display path exposes.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -75,6 +75,47 @@ package_macos() {
     plutil -replace CFBundleShortVersionString -string "${xemu_version}" dist/xemu.app/Contents/Info.plist
     plutil -replace CFBundleVersion            -string "${xemu_version}" dist/xemu.app/Contents/Info.plist
 
+    # Bundle Vulkan/MoltenVK runtime if available
+    if command -v brew &>/dev/null; then
+        brew_prefix="$(brew --prefix)"
+        mvk_lib="${brew_prefix}/lib/libMoltenVK.dylib"
+        vk_loader="${brew_prefix}/lib/libvulkan.1.dylib"
+
+        if [ -f "$mvk_lib" ] && [ -f "$vk_loader" ]; then
+            echo "Bundling Vulkan/MoltenVK runtime..."
+            vk_lib_dest="dist/xemu.app/Contents/Libraries/${target_arch}"
+            vk_icd_dest="dist/xemu.app/Contents/Resources/vulkan/icd.d"
+
+            # Copy Vulkan loader and MoltenVK
+            cp "$vk_loader" "${vk_lib_dest}/libvulkan.1.dylib"
+            cp "$mvk_lib" "${vk_lib_dest}/libMoltenVK.dylib"
+            codesign -s - -f "${vk_lib_dest}/libvulkan.1.dylib"
+            codesign -s - -f "${vk_lib_dest}/libMoltenVK.dylib"
+
+            # Create ICD manifest pointing to bundled MoltenVK
+            mkdir -p "$vk_icd_dest"
+            cat > "${vk_icd_dest}/MoltenVK_icd.json" << 'ICDJSON'
+{
+    "file_format_version" : "1.0.0",
+    "ICD": {
+        "library_path": "../../../Libraries/__ARCH__/libMoltenVK.dylib",
+        "api_version" : "1.2.0",
+        "is_portability_driver" : true
+    }
+}
+ICDJSON
+            sed -i '' "s/__ARCH__/${target_arch}/g" "${vk_icd_dest}/MoltenVK_icd.json"
+
+            # Set VK_ICD_FILENAMES in Info.plist so the Vulkan loader finds MoltenVK.
+            # Note: LSEnvironment values are resolved relative to /, so we cannot
+            # use @executable_path here. The programmatic setenv in instance.c
+            # handles the Finder launch case; this is a fallback for edge cases.
+            plutil -replace LSEnvironment -json "{\"VK_DRIVER_FILES\": \"../Resources/vulkan/icd.d/MoltenVK_icd.json\"}" dist/xemu.app/Contents/Info.plist
+        else
+            echo "Warning: MoltenVK/Vulkan loader not found, Vulkan renderer will not be available"
+        fi
+    fi
+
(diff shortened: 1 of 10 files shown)


t-chamb:
Separately — I saw you mentioned on #2768 that you plan to handle macOS Vulkan properly. Is the approach going to expose IOSurface from the display path? Asking because I've been working on VTFrameRateConversion (Apple Neural Engine frame interpolation) locally and it needs an IOSurface to feed frames to the ANE. Happy to adapt to whatever display surface API you land on.

mborgerson:
@t-chamb Are you an autonomous LLM agent, such as openclaw?

t-chamb:
😂🤣😂...💀 no, no full auto bot here & no hit piece website coming your way if you don't approve the PR.

-Todd-

t-chamb:
Addressed — updated both files:

- `draw.c`: check `fwrite` return + `fclose` before marking saved
- `glsl.c`: check all `fwrite` returns, stop on first failure, `remove()` partial file on error, defer `g_free(path)` to end of function

Force-pushed.

jessedye:
@t-chamb I opened #2989 covering the same two caches before spotting how much of this PR it overlapped, and have closed it in favour of yours — this predates it by five months.

If it helps with the split abaire suggested, the cache half is already isolated against master at `jessedye/xemu:vk-persist-pipeline-cache` (2 commits, +170/-3, no macOS changes). Take the commits, cherry-pick, or ignore them entirely.

A few findings from testing on a Raspberry Pi 5 with Mesa V3DV, where pipeline compilation is slow enough to measure, in case they're useful either way:

- Your `atexit` handler is load-bearing rather than belt-and-braces. `finalize_pipeline_cache` runs from the renderer's `ops.finalize`, which fires on a renderer switch; a normal quit exits without tearing the device down, so a save placed only there never runs. I had to learn that the hard way after shipping a version that silently wrote nothing.
- `fopen`/`fwrite` truncates an existing good cache before writing, so a crash mid-save leaves a corrupt file that the next launch still feeds to the driver. `g_file_set_contents` writes to a temp file and renames, which avoids that.
- Gating both caches on `g_config.perf.cache_shaders` keeps the existing "Cache shaders to disk" toggle honest — otherwise files keep appearing after a user switches it off.

On the value: the benefit is real but confined to startup. Boot went 20.5 -> 43.7 fps with the caches seeded, while steady-state gameplay measured inside noise — only 27 shaders and 73 pipelines are created across 3089 frames of Morrowind, and a 45-second CPU profile of gameplay contains no compilation samples at all.

t-chamb:
@jessedye Thanks for closing #2989 in favor of this one, and especially for the V3DV testing — a platform where compilation is actually slow enough to measure is exactly the data this PR was missing.

Both suggestions are now in (10ebf9bb3b):

- **Atomic writes**: both caches now go through `g_file_set_contents` (temp file + rename), so a crash mid-save can't clobber a good cache with a truncated one. This also let me drop the manual `fwrite`-checking/`remove()` cleanup from the last revision — the SPIR-V cache is serialized into a buffer and written in one shot.
- **Config gating**: load and save of both caches are now gated on `perf.cache_shaders`, so the existing "Cache shaders to disk" toggle controls them like it does the GL cache. No more files reappearing after switching it off.

Good to have independent confirmation that the `atexit` path is load-bearing — I hit the same thing during development (normal quit never reaches `ops.finalize`), which is why it's registered unconditionally rather than as a backstop.

And agreed on the scope of the win: this is a startup/first-boot improvement, not a steady-state one. Your 20.5 → 43.7 fps boot number on the Pi 5 is a nice concrete data point for that — thanks for measuring it.

I'll keep your `vk-persist-pipeline-cache` branch in mind if a split ends up being the way forward here.

t-chamb:
Adding a macOS data point alongside the Pi 5 numbers, measured with temporary counters around `pgraph_vk_compile_glsl_to_spv` and `vkCreateGraphicsPipelines` (not part of the PR). Two identical 120-second no-input boots of Halo 2 on an M3 Ultra via MoltenVK — cold (both caches deleted) vs. seeded (caches from the cold run):

| | cold | seeded |
|---|---|---|
| SPIR-V compiles | 106 (299.1 ms) | 0 (0.0 ms) |
| `vkCreateGraphicsPipelines` calls | 124 (258.2 ms) | 123 (46.1 ms) |
| total compile stall | ~557 ms | ~46 ms |

Consistent with @jessedye's findings: the seeded run eliminates GLSL→SPIR-V work entirely and cuts driver pipeline creation ~5.6×, but the absolute win on fast desktop hardware is around half a second spread across boot — the benefit scales with how slow the platform's compiler is, so V3DV-class hardware sees the dramatic end of it. Since each miss is a stall on the render path at the moment a new shader/pipeline first appears, this mostly buys smoother first-encounter frames rather than throughput.

t-chamb:
Split done: the platform-neutral cache persistence is now #2993, rebased onto current master as two clean commits (+262/−6, `draw.c`/`glsl.c` only), with the review feedback and benchmark data from this thread folded in.

I'll rework this PR to contain only the macOS/MoltenVK enablement once #2993 settles, to avoid churn from rebasing it twice.

PR #2805, opened 2026-03-23
Cci

EqUiNoX-Labs:


MasonT8198:
I do think that some of the issues that Gemini is throwing a tantrum about are valid, specially the OOB error related issue in particular. Secondly, are there alternate tools to make a CCI file, other than just Repackinator, similar to Antangelos XISO tool or similar?

Also, I presume that CCI still only works with Cerbios, if the proposed PR is decided to be merged, it would be nice to have a write-up of the format on the xemu-website github repository for end-users to read, and some of the benefits over an xiso (compression, etc)

Also, the format will need to be tested with various games to ensure there is no degradation in emulation that isn't experienced on real HW etc...

@Triticum0 would probably be a good person to maybe help get some headway on that, as they've been an invaluable resource when it comes to managing issues.

mborgerson:
There's no PR description and I'm not familiar with this file format. Before this patch is reviewed or considered for merge, a few questions:

* What's the motivation? Saving some disk space? Anything else?
* Who manages the CCI format spec?
* What's the maintenance plan for this code? Do you plan to continue maintaining it?
* What's the existing tooling ecosystem around this file format?
* What does existing file format adoption look like?
* How does this format compare to others, like CHD? Why is CCI better than other options (in terms of Xbox preservation, compatibility, usability, etc)?
* Are there any docs on the CCI format? e.g.
   * It's clear that LZ4 is used for compression, what else does it support?
   * Does it support multi-partition images?
   * Does it embed additional disc security/manufacturing data?
   * Does the format embed any other metadata?
   * etc

Issue #2578, opened 2025-12-01
Meta: In-game Cut-scenes running too fast


Triticum0:
### Title

https://xemu.app/titles/545400ab/#The-Da-Vinci-Code
https://xemu.app/titles/41560045/#True-Crime-New-York-City (Occured on one of my dumps. When redumped it didn't occur.)
https://xemu.app/titles/4947007d/#Marc-Ecko-s-Getting-Up-Contents-Under-Pressure

### Bug Description

These game have cut-scenes that are extremely fast than they should be.
Example.

https://github.com/user-attachments/assets/5146c50c-6848-45c5-b06f-ebacef388e10

### Expected Behavior

https://youtu.be/cmP2PrlRGMg?t=273

### xemu Version

v0.8.118

### System Information

Field | Value
-- | --
OS | Windows 11
CPU | AMD Ryzen 5 5700x Eight-Core Processoe
Graphics Device | NVIDIA GeForce RTX 3060 Ti/PCIe/SSE2
Graphics Driver | 4.0.0 NVIDIA 581.29
Render | Opengl

### Additional Context

Been an issue since at least 2021 jan when last reported

RinMaru:
Can you retest this? I didnt have issues at least on VK. also are all the vids fast or just specific ones? Only Tested TC New York

Triticum0:
Fixed by https://github.com/xemu-project/xemu/pull/2945

PR #2809, opened 2026-03-26
Add CHD (Compressed Hunks of Data) support

EqUiNoX-Labs:
## Add CHD (Compressed Hunks of Data) support
This adds support for **CHD** (style compressed disc images) as a QEMU block format so xemu can mount **single `.chd` files** on the DVD drive like ISO/XISO.
### User-facing
- Disc picker / game list treats **`.chd`** like other disc images (e.g. alongside `.iso` / `.xiso`).
- **Load disc** uses `qmp_blockdev_change_medium` with format **`chd`** when the path ends in `.cci` (case-insensitive), and **`raw`** otherwise—same tray/medium flow as existing images.
### Implementation
- New **`chd`** block driver (`block/chd.c`): CCIM header, sector index, raw or LZ4-compressed 2048-byte sectors; **read-only** (`pwritev` returns `-EROFS`).
### Testing
- [ ] Mount a `.chd` via the UI and confirm the title boots.
- [ ] Confirm `.iso` / `.xiso` behavior unchanged.

### DIFF
--- a/block/cci.c
+++ b/block/cci.c
@@ -1,0 +1,484 @@
+/*
+ * CCI Xbox disc container (xemu-cci)
+ *
+ * Format per Team-Resurgent XboxToolkit:
+ *   https://github.com/Team-Resurgent/XboxToolkit/blob/main/XboxToolkit/CCIContainerReader.cs
+ *
+ * Header (32 bytes LE): magic "CCIM" (0x4D494343), hdr_size 32, uncompressed_size,
+ * index_offset, block_size 2048, version 1, index_alignment 2.
+ * Sectors = uncompressed_size / 2048; index has sectors+1 uint32 entries:
+ *   position = (entry & 0x7FFFFFFF) << alignment; LZ4 = entry & 0x80000000.
+ * Compressed sector blob (matches XboxToolkit CCISectorDecoder): byte0 = trailer
+ * padding length; bytes [1 .. span-1-trailer] are one LZ4 block (not raw bytes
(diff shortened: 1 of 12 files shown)


Triticum0:
One thing I like to mention is some game are sensitive when comes to the type dump used I had issue with True Crime: New York City where only redump played at correct speed and my other dump didn't and would run cutscenes too fast so I don't know if it would behaviour similar on the chd and cci formats.

https://github.com/xemu-project/xemu/issues/2578

GXTX:
See: #238

MasonT8198:
On this: I think the consensus in discord was that there needs to be meaningful proof as to why going with **any** method of compression would be beneficial beyond just and arbitrary space difference, and if it were to be implemented Matt said something along the lines of we should actively look for the best option to do disc compression from a space standpoint as well as other factors.

Issue #238 indicates that this should probably be discussed thoroughly in discord, with the final decision/updates posted here so we don't lose track of what the decisions/updates are

RinMaru:
> On this: I think the consensus in discord was that there needs to be meaningful proof as to why going with **any** method of compression would be beneficial beyond just and arbitrary space difference, and if it were to be implemented Matt said something along the lines of we should actively look for the best option to do disc compression from a space standpoint as well as other factors.
> 
> Issue #238 indicates that this should probably be discussed thoroughly in discord, with the final decision/updates posted here so we don't lose track of what the decisions/updates are

He's said he would maintain it and CCI if it gets merged. Not sure what else is needed

PR #2817, opened 2026-04-11
Added path to 'Xemu' text

### LINKED ISSUE #1181: Incorrect xemu SVG icon
### Bug Description

When seeing https://github.com/mborgerson/xemu/blob/master/ui/icons/xemu.svg, we can notice that the `xemu` name is on the right side, going through the circle, instead of being centered.

### Expected Behavior

The xemu icon should look like https://github.com/mborgerson/xemu/blob/master/ui/icons/xemu_512x512.png.

### xemu Version

All

### System Information

_No response_

### Additional Context

_No response_

Sjondepon:
I converted the xemu text to a path to avoid the (unlisted) ubuntu font family dependency. This way every device should see the right picture, when opening the xemu.svg

This resolves #1181


MasonT8198:
Works on my end as well, thanks for working on this!

Issue #2736, opened 2026-02-10
Clean up use of `assert(false)` throughout `hw/xbox` code


abaire:
### Bug Description

There are a number of places where we call `assert(false)`, generally in paths that are never expected to be called.

In the case where these do trigger, the user is presented with a non-informative error pinned to a line number that often changes over time. This is both a suboptimal user experience and makes it harder to debug/correct the issue for contributors.

### Expected Behavior

Instead of `assert(false)` we should always `assert(!"Some informative string")`. 

Functionally this is identical (apart from the space for the string) but provides the user and future contributor with a better explanation of why the assert was never expected to happen.

### xemu Version

n/a affects all versions

### System Information

n/a

mborgerson:
Issue should clarify 'xemu-owned'

abaire:
Hopefully clear now. I haven't checked for uses outside the Xbox dies but figure it's not worth any merge hassle to fix things that could be fixed upstream

mborgerson:
Let's just say ./hw/xbox, for those taking this on as their first issue

Reedzit:
I'm looking into this now as my first issue!

Reedzit:
Quick question: Should the "SET_MASK" call be placed before the "assert(false)" line [here](https://github.com/xemu-project/xemu/blob/master/hw/xbox/nv2a/pfifo.c#L300)? I've seen a couple instances of the assert being placed both [before ](https://github.com/xemu-project/xemu/blob/master/hw/xbox/nv2a/pfifo.c#L423)and after SET_MASK, however, I am not super familiar with the behavior of GET/SET MASK.

abaire:
> Quick question: Should the "SET_MASK" call be placed before the "assert(false)" line [here](https://github.com/xemu-project/xemu/blob/master/hw/xbox/nv2a/pfifo.c#L300)? I've seen a couple instances of the assert being placed both [before ](https://github.com/xemu-project/xemu/blob/master/hw/xbox/nv2a/pfifo.c#L423)and after SET_MASK, however, I am not super familiar with the behavior of GET/SET MASK.

Either is fine, the assert will stop execution anyway so it won't change the behavior in practice. 

For the sake of consistency, let's just make them match in that use. I'd vote for having the `assert` before since there's no reason to bother changing the state just before forcibly crashing out.

mborgerson:
@Reedzit Let's keep the change minimal. Don't worry about reordering asserts, just give them better messages as described in issue text

[...]

PR #2823, opened 2026-04-15
hw/xbox: Clean up use of `assert(false)` 

Reedzit:
This PR addresses issue https://github.com/xemu-project/xemu/issues/2736. Each of the assert(false) statements in the hw/xbox directory were modified to give a better description of the error in the event that the program crashes.

The original PR was closed after renaming the branch on my fork...this is a new one replacing the old one that closed.


hw/xbox/mcpx/apu/dsp/dsp_emu.c.inc
```diff
@@ -32,7 +32,7 @@ static void emu_undefined(dsp_core_t* dsp)
     /* Add some artificial CPU cycles to avoid being stuck in an infinite loop */
     dsp->instr_cycle += 100;
     if (dsp->exception_debugging) {
-        assert(false);
+        assert(!"Illegal dsp instruction - see error above");
```
abaire: nit: In cases where there is no console bound (e.g., Windows when not launched through a shell), there won't be a visible error above. Maybe "see error log" or just omit the trailing part of the message?

hw/xbox/mcpx/apu/vp/vp.c
```diff
@@ -565,11 +565,11 @@ static void fe_method(MCPXAPUState *d, uint32_t method, uint32_t argument)
             DPRINTF("idle voice %d\n", argument);
             d->set_irq = true;
         } else {
-            assert(false);
+            assert(!"SE2FE_IDLE_VOICE called but NV_PAPU_FETFORCE1_SE2FE_IDLE_VOICE not enabled");
         }
         break;
     default:
-        assert(false);
+        assert(!"Unrecognized VP method - could be unimplemented or invalid");
```
abaire, line 573: Should we fprintf or DPRINT `method` before this assert so we can figure out what was emitted?

hw/xbox/nv2a/pfifo.c
```diff
@@ -440,7 +440,7 @@ static void pfifo_run_pusher(NV2AState *d)
     uint32_t error = GET_MASK(*dma_state, NV_PFIFO_CACHE1_DMA_STATE_ERROR);
     if (error) {
         NV2A_DPRINTF("pb error: %d\n", error);
-        assert(false);
+        assert(!"Error occurred getting mask for dma state - check logs for more info");
```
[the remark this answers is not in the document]
abaire: Looks like this suggestion is still valid; changing "occurred getting mask for" to "reported in" also seems reasonable.

hw/xbox/nv2a/pgraph/vk/texture.c
```diff
@@ -80,7 +80,7 @@ static enum S3TC_DECOMPRESS_FORMAT kelvin_format_to_s3tc_format(int color_format
     case NV097_SET_TEXTURE_FORMAT_COLOR_L_DXT45_A8R8G8B8:
         return S3TC_DECOMPRESS_FORMAT_DXT5;
     default:
-        assert(false);
+        assert(!"Invalid color kelvin format");
```
abaire: nit: I think "Invalid texture color format" would be clearer

hw/xbox/nv2a/pgraph/gl/texture.c
```diff
@@ -664,7 +664,7 @@ static TextureBinding* generate_texture(const TextureShape s,
             case 2: gl_target = GL_TEXTURE_2D; break;
             case 3: gl_target = GL_TEXTURE_3D; break;
             default:
-                assert(false);
+                assert(!"Invalid texture dimensionality");
```
abaire, line 668: This is another case where it would be useful to fprintf `s.dimensionality` so we can figure out what it was. If this assert triggers it probably means that there's a case of PFIFO memory being stomped (e.g., in #2666)

hw/xbox/nv2a/user.c
```diff
@@ -57,11 +57,11 @@ uint64_t user_read(void *opaque, hwaddr addr, unsigned int size)
             }
         } else {
             /* ramfc */
-            assert(false);
+            assert(!"Invalid channel id");
```
abaire: I think this one (and the change on line 103) would be clearer if it just repeated the `if` condition:

`assert(channel_id == cur_channel_id)`

Or at least made reference to it:
`assert(!"Unsupported: channel_id != cur_channel_id")`

hw/xbox/nv2a/pgraph/glsl/psh.c
```diff
@@ -424,7 +424,7 @@ static MString* get_input_var(struct PixelShader *ps, struct InputInfo in, bool
         res = mstring_from_fmt("-%s", mstring_get_str(reg));
         break;
     default:
-        assert(false);
+        assert(!"Invalid PS_INPUTMAPPING mode");
```
abaire, line 430: For most of these it'd be useful to also fprintf the value before blowing up. I wonder if having a `LOG_FATAL` macro that fprintf's a message before forcibly asserting would be useful?

The messages are already better than nothing, so I wouldn't block on it, but might be a good followup.

Reedzit: I added fprintf statements just before the messages for this file.

hw/xbox/nv2a/user.c
Reedzit: I changed this message and line 103 to refer to the if statement that failed.

mborgerson:
Thanks! Edited your commit message to give a bit more context

PR #2843, opened 2026-04-30
nv2a: Fix epoxy flake using debug funcs on macOS

abaire:
Epoxy sometimes loses track of the `GL_KHR_debug` extension's availability on macOS. This leads to a race condition where the app may report the extension as available during startup, but then fail when epoxy attempts to resolve the actual debug functions later on.

It is unclear why the extension's presence is being lost; I've confirmed that the same GL context is active during the init and later use of the methods but sometimes the extension goes from present to not, generally fairly early during startup when launched via RenderDoc. This may be exacerbated or caused by RenderDoc; I'd previously become used to occasional random launch failures in the context of RD but migrating to an M5 MacBook seems to trigger these failures far more often than before.

This change simply maintains alternate function pointers. Since these aren't likely to be useful outside of this debug function, this feels like the simplest workaround and is aligned with our existing Apple-specific code.

### DIFF
--- a/hw/xbox/nv2a/pgraph/gl/debug.c
+++ b/hw/xbox/nv2a/pgraph/gl/debug.c
@@ -35,6 +35,8 @@
 #include "thirdparty/renderdoc_app.h"
 #endif
 
+#include <SDL3/SDL.h>
+
 #define CHECK_GL_ERROR() do { \
   GLenum error = glGetError(); \
   if (error != GL_NO_ERROR) {  \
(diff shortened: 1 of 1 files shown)


hw/xbox/nv2a/pgraph/gl/debug.c
```diff
@@ -63,10 +66,41 @@ void gl_debug_initialize(void)
          * debug functions which we depend on will still work as expected,
          * so skip the call for this platform.
          */
+        glEnable(GL_DEBUG_OUTPUT_SYNCHRONOUS);
```
mborgerson, line 69: According to [OpenGL reference pages](https://registry.khronos.org/OpenGL-Refpages/gl4/html/glEnable.xhtml):

>  GL_DEBUG_OUTPUT and GL_DEBUG_OUTPUT_SYNCHRONOUS are available only if the GL version is 4.3 or greater. 

But macOS supports only OpenGL <=4.1?

mborgerson:
> This may be exacerbated or caused by RenderDoc; I'd previously become used to occasional random launch failures in the context of RD but migrating to an M5 MacBook seems to trigger these failures far more often than before.

Does this happen when you're not using RenderDoc?

hw/xbox/nv2a/pgraph/gl/debug.c
abaire: Yeah, it only fully supports 4.1 but it seems that they partially adopted the mainstream adopted names for some settings (at least these ones).

abaire:
> Does this happen when you're not using RenderDoc?

I've occasionally seen random crashes during startup without using RenderDoc, but they're not at all reliably reproducible. Even this issue isn't consistently reproducible, though with the M5 it happens more times than not. Anecdotally w/ the M5 I had it fail nine times in ten attempts, with the M3 it probably failed closer to one in twenty.

PR #2857, opened 2026-05-10
Add stale issues workflow to manage inactive issues

MasonT8198:


MasonT8198:
Can be tweaked to desired params to start

.github/workflows/stale-issues.yml
```diff
@@ -0,0 +1,47 @@
...
+        with:
+          days-before-stale: 90
+          days-before-close: 30
+          stale-issue-message: >
+            This issue has been automatically marked as stale due to inactivity.
+            It will be closed unless further activity occurs.
+          close-issue-message: >
+            Closing this issue due to prolonged inactivity.
+          stale-issue-label: "stale"
+          exempt-issue-labels: "graphics,input,enhancement,audio"
```
abaire: Can we do the opposite and limit these auto closes/nags to things flagged as pending author response? 

There are a lot of issues that are totally valid and can easily go for 6 months or more between responses (e.g., things like signed texture support that are root caused but require somewhat complex fixes, things that are non-obvious and will require weeks of reversing to determine, etc...). There are also times when the tiny pool of people knowledgeable enough to do something with reports are busy with other things.

It also can take a long time for PRs to get reviewed, I can imagine it would be annoying to have a PR out to fix an issue and then get nagged for not bumping threads.

MasonT8198: that's fair, I can make the stale issue label just "pending author response" and it'll pick that up

abaire:
Generally looks good to me. Maybe we should add "invalid" and "needs investigation" as well, though I don't know if we bother to use invalid (I think we'd just close things) and am not sure if we use NI to mean that a contributor should investigate or that the author should investigate. Perhaps we should just remove those labels instead?

Might want to bump up the numbers a bit; it sometimes takes a month or two for a contributor to look at a bug, it feels fair to give reporters a similar period to respond given that they may have long moved on to other things since reporting.

Issue #2738, opened 2026-02-10
GitHub Issue bot


mborgerson:
### Feature Request

There seems to be a problem with people filing issues and not following the template. e.g. Including various tags or other info in the title, not including affected title names or URLs, picking the wrong template, not providing enough detail, filing duplicates, etc. Let's add an issue bot to help manage these issues.

Related #473

PR #2858, opened 2026-05-10
Add issue template validation workflow

MasonT8198:
This'll need tweaked to desired parameters OFC, I added N/A in there but it'll probably need removed, there does need to be a way to ignore certain parameters of the templates that may not be applicable to the issue

### DIFF
--- a/.github/workflows/issue-template-validator.yml
+++ b/.github/workflows/issue-template-validator.yml
@@ -29,8 +29,6 @@ jobs:
             // placeholder rejection phrases
             const placeholders = [
               "fill this out",
-              "n/a",
-              "none",
               "no description provided"
             ];
 
--- a/.github/workflows/issue-template-validator.yml
+++ b/.github/workflows/issue-template-validator.yml
@@ -1,0 +1,72 @@
+name: Validate Issue Template Completeness
+
+on:
+  issues:
+    types: [opened, edited]
+
+permissions:
+  issues: write
+
+jobs:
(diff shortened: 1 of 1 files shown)


MasonT8198:
> Note
> 
> Gemini is unable to generate a review for this pull request due to the file types involved not being currently supported.

shame, since you use markdown to function.

abaire:
Does this apply to all issues? E.g., feature requests don't use a template so they wouldn't have existing markdown.

I'm of the opinion that users will always fail to fill out templates and a lot of the info is available in xemu; I think something like #2456 would be more effective in getting the baseline info filled out consistently.

MasonT8198:
> Does this apply to all issues? E.g., feature requests don't use a template so they wouldn't have existing markdown.
> 
> I'm of the opinion that users will always fail to fill out templates and a lot of the info is available in xemu; I think something like #2456 would be more effective in getting the baseline info filled out consistently.

there is *kind of* a template request, this was meant to be marked as a draft, I just wanted initial thoughts, it does occur occasionally that people post issues with nothing inherently meaningful that can just be closed. This is meant to solve one of the nits raised with https://github.com/xemu-project/xemu/issues/2738 which is why it exists in the first place. if this isn't something to go forward with maybe that should be tweaked or closed?

this really is only meant to exist to discourage low quality issues

MasonT8198:
I'm more than willing to tweak as needed ofc

PR #2864, opened 2026-05-13
Let SMC handle actual tray eject/load

### LINKED ISSUE #1563: Request: Option to open the emulator without automatically running the last game

BiatuAutMiahn:
This is a remake of #1564

hw/xbox/smc: Centralize disc tray control and fix scratch register handling

Route Xemu UI disc eject/load commands through the emulated SMC's `SMC_REG_TRAYEJECT` register. This improves emulation accuracy by ensuring the SMC's state is always synchronized with disc operations, mimicking how a real Xbox interacts with its hardware.

Add a new machine property `eject-after-boot` and corresponding command-line option `-eject_after_boot` to set the SMC's scratch register flag `SMC_REG_SCRATCH_EJECT_AFTER_BOOT`.

Correctly use bitwise OR assignment (`|=`) when setting SMC scratch register flags (like `SMC_REG_SCRATCH_SHORT_ANIMATION`) to avoid unintentionally overwriting other flags.

Fixes #1563

### DIFF
--- a/hw/xbox/smbus_xbox_smc.c
+++ b/hw/xbox/smbus_xbox_smc.c
@@ -152,7 +152,7 @@ static int smc_write_data(SMBusDevice *dev, uint8_t *buf, uint8_t len)
                                        &error);
         } else {
             xemu_settings_set_string(&g_config.sys.files.dvd_path, "");
-            qmp_eject(true, "ide0-cd1", false, NULL, true, false, &error)
+            qmp_eject(true, "ide0-cd1", false, NULL, true, false, &error);
         }
         xbox_smc_update_tray_state();
         break;
@@ -370,10 +370,12 @@ void xbox_smc_eject_button(void)
 void xbox_smc_tray_eject(uint8_t val)
 {
     Object *obj = object_resolve_path_type("", TYPE_XBOX_SMC, NULL);
-    uint8_t buf[2];
-    buf[0] = 0x0c;
-    buf[1] = val;
-    smc_write_data(obj, buf, 2);
+    if (obj) {
+        uint8_t buf[2];
+        buf[0] = 0x0c;
+        buf[1] = val;
+        smc_write_data(obj, buf, 2);
+    }
 }
 
 // FIXME: Ideally this would be called on a tray state change callback (see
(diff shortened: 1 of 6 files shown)


hw/xbox/smbus_xbox_smc.c
```diff
@@ -141,6 +144,19 @@ static int smc_write_data(SMBusDevice *dev, uint8_t *buf, uint8_t len)
         }
         break;
 
+    case SMC_REG_TRAYEJECT:
```
GXTX, line 149: Let's first check if the device here is actually CD/DVD player.

BiatuAutMiahn: Not sure what you mean

GXTX: What happens if `ide0-cd1` doesn't exist?

BiatuAutMiahn: I don't think it will matter. `ide0-cd1` is created as a part of xemu bootstrap. I suppose if `ide0-cd1` didn't exist in the current implementation it would fail here https://github.com/xemu-project/xemu/blob/59c5631e6054a031be24432ea3ccd7fa404c0bed/ui/xemu.c#L1405

BiatuAutMiahn: Looks like if `ide0-cd1` is absent the bios/kernel will throw an exception (error 10,11 or 12) on bootup before it get's this far.

GXTX: Think of xemu as a *virtual xbox*, with certain kernels you can run without a DVD drive. Ignore the fact it's attached early in start-up, it's possible for that to change.

BiatuAutMiahn: I argue that the current implementation in master will have the issue too is the drive isn't installed. I believe that it would be out of scope for this PR.

BiatuAutMiahn: If you guys insist I'll work it out.

Issue #108, opened 2020-12-24
Implement missing DSP instructions


mborgerson:
Enabling "Dolby Surround" causes games to crash

```
c1890 - extractu #CO, S2, D
Dsp: 0x0519: 0x0c1890 Illegal instruction
**
ERROR:D:/a/xemu/xemu/hw/xbox/mcpx/dsp/dsp_emu.inl:40:emu_undefined: assertion failed: (0)
```

_Originally posted by @GXTX in https://github.com/mborgerson/xemu/issues/103#issuecomment-749892207_

https://xemu.app/titles/41540003/#Pro-Fishing-Challenge
https://xemu.app/titles/4d530064/#Halo-2

Triticum0:
This affects Pro Fishing Challenge
https://xemu.app/titles/41540003/#Pro-Fishing-Challenge

Triticum0:
@mborgerson add pro fishing challenge to issue so it shows up on website comp page #1968

D3nt0n1418:
Also affects Halo 2. Game will not start if Dolby Surround is enabled in the dashboard.

blakeplusplus:
@mborgerson please add Pro Fishing Challenge link above to issue so it shows up

Maxkatsur:
Did it also affect _Project Gotham Racing_ (2001)? Yes, it did!

PR #2801, opened 2026-03-20
apu: Add new DSP56300 emulator with JIT execution engine

mborgerson:
- Fixes a few bugs in the existing C interpreter
- Adds a state abstraction layer to switch between engines at runtime and maintain snapshot compatibility

TODO:
- [x] Fix xemu test cases
- [x] Fix macOS Build (target arch specification)
- [x] Fix Windows build (Rust toolchain)
- [x] Add dsp56300 license details to xemu's license gen script
- [x] Pin version

Fix #108

PR #2865, opened 2026-05-13
hw/xbox/mcpx/dsp: implement DSP56300 surround/AC-3 opcodes

Caesar0007:
## Summary

This PR adds clean-room implementations of **9 DSP56300 opcode families** that the Xbox MCPX GP DSP firmware emits in its surround / AC-3 code paths.

Without these handlers, the moment the firmware encounters `extractu` / `brkcc` / `plock` / `insert` / `debug` etc. it hits `dsp_set_interrupt(DSP_INTERRUPT_ILLEGAL)` and the APU thread halts. This blocks every title whose EEPROM AudioFlags select a surround-capable firmware path (`XC_AUDIO_FLAGS = 0x10002`, i.e. SURROUND + ENABLE_AC3 — the "Digital 5.1 Surround Sound" dashboard option).

## Why this matters

Today xemu's audio path produces **stereo only**, with no reverb tail, for every dsstdfx-based title. Two distinct reasons:

1. The GP DSP firmware aborts on these missing opcodes the moment it tries to run anything beyond the boot/idle path (this PR addresses this).
2. The monitor tap reads only mixbins 0/1 — the reverb tail at mixbins 6..9 is silently dropped (separate follow-up PR).

This PR alone gets us past (1). The (2) fix is a ~30-LOC follow-up PR that completes the picture: after both land, every dsstdfx title will have audible reverb in xemu for the first time.

## Opcodes added

Per **NXP DSP56300 Family Manual** ([DSP56300FM.pdf](https://www.nxp.com/docs/en/reference-manual/DSP56300FM.pdf)):

| Mnemonic | Manual ref | Forms |
|----------|-----------|-------|
| `extractu` | §A.87 | register-source + immediate |
| `extract` | §A.86 | register-source + immediate |
| `insert` | §A.108 | register-source + immediate |
| `brkcc` | §A.32 | (single form) |
| `plock ea` / `punlock ea` | §A.149 / §A.151 | (single forms) |
| `debug` / `debugcc` | §A.71 / §A.72 | unconditional + conditional |
| `bscc` | §A.35 | xxxx / xxx / Rn (3 forms) |
| `bcc Rn` / `bra Rn` / `bsr Rn` | §A.16 / §A.27 / §A.36 | register-indirect |
| `lsr #ii, D` | §A.122 | immediate shift |

Each handler has a comment block citing its DSP56300FM section.

## Robustness fixes (bundled)

The patch also softens 4 hard asserts in the DSP execution path to log + recover, so that a single anomalous instruction or memory address doesn't bring down the host process:

- `emu_undefined()` — log + advance PC by 1 word instead of `assert(false)`
- `lookup_opcode_slow()` — return `NULL` on unmatched bit-pattern; caller routes through `emu_undefined()`
- `read_memory_p()` — soft-fail (log + return 0) on out-of-range P-RAM reads
- `dsp56k_read_memory()` — same softening for X/Y/L space

These are conservative changes: each only triggers on an otherwise-fatal condition that would have aborted the host. None affect stereo titles.

## Stereo-title impact

**None.** The new opcode handlers are only reached when the surround firmware is loaded, which requires `XC_AUDIO_FLAGS = 0x10002` in the user's EEPROM. Stereo-mode firmware paths never touch these opcodes. The robustness changes also only trigger on conditions that would have caused an immediate crash before; healthy execution paths are unchanged.

## Testing

- ✅ Compile-verified (`gcc -fsyntax-only`) on the mingw64 win32 build
- ✅ Snapshot-driven dynamic test: ran the patched DSP CPU against a captured `dsstdfx.bin` (UC2 5849.17, SHA1 `df4bc8c95263…`) with surround firmware loaded; previously aborted on first `extractu`, now executes the full reverb24k FX graph to completion
- ✅ Tested under real Xbox EEPROM dump with `XC_AUDIO_FLAGS = 0x10002`

If reviewers want me to split this into per-opcode-family commits, I can do so for v2.

## References

- NXP DSP56300 Family Manual: https://www.nxp.com/docs/en/reference-manual/DSP56300FM.pdf
- DSP56362 User Manual: https://www.nxp.com/docs/en/user-guide/DSP56362UM.pdf
- xboxdevwiki / DSP: https://xboxdevwiki.net/DSP
- xboxdevwiki / APU: https://xboxdevwiki.net/APU

## Closes

Part of #108 (surround audio support).

### DIFF
--- a/hw/xbox/mcpx/apu/dsp/dsp_cpu.c
+++ b/hw/xbox/mcpx/apu/dsp/dsp_cpu.c
@@ -164,7 +164,7 @@ static const OpcodeEntry nonparallel_opcodes[] = {
     { "0000110000011110011SsssD", "asr S1, S2, D", NULL, NULL },
     { "00001101000100000100CCCC", "bcc xxxx", dis_bcc_long, emu_bcc_long }, //??
     { "00000101CCCC01aaaa0aaaaa", "bcc xxx", dis_bcc_imm, emu_bcc_imm },
-    { "0000110100011RRR0100CCCC", "bcc Rn", NULL, NULL },
+    { "0000110100011RRR0100CCCC", "bcc Rn", NULL, emu_bcc_reg },
     { "0000101101MMMRRR0S00bbbb", "bchg #n, [X or Y]:ea", dis_bchg_ea, emu_bchg_ea, match_MMMRRR },
     { "0000101100aaaaaa0S00bbbb", "bchg #n, [X or Y]:aa", dis_bchg_aa, emu_bchg_aa },
     { "0000101110pppppp0S00bbbb", "bchg #n, [X or Y]:pp", dis_bchg_pp, emu_bchg_pp },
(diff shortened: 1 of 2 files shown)


mborgerson:
Thanks for the contribution, but I do not plan to continue improving the C emulator implementation any longer. It is being replaced by a new emulator in #2801.

Caesar0007:
Thanks for the quick review and the §A.87 clamp catch — that's a real bug in my implementation; I'll keep it in mind. Understood about the C interpreter freeze; happy to leave this here as reference for anyone who needs to keep stereo+reverb working in the interpreter path until the JIT lands.

Adjacent findings from the same research effort (sharing in case useful for #2801):

The dsstdfx.bin FX graph routes the I3DL2 reverb wet output to XTLK_FL/FR (mixbins 6/7), not FL/FR (0/1) — currently the GP monitor tap only reads 0/1, so reverb is silently dropped even when the firmware runs to completion. ~30 LOC fix in gp_ep.c.
The EP DSP's output FIFO is currently overwritten with ep_silence[] in ep_sink_samples() (gp_ep.c:228) regardless of what the EP emitted — looks like a leftover from before the EP was wired up.
The MCPX EP firmware (the DOLBY blob in dsound.lib's epdsp.obj, 29 KB, SHA1 60abf96b...) is byte-identical across XDKs 5344-5933 and can be runtime-extracted from any user's own Xbox game image without redistribution.
Happy to share my detailed notes / write up specific findings if any of this would be useful for the JIT work.

Caesar0007:
For the record — Gemini's review caught five real bugs that I've now corrected in my local fork (SSS mapping inversion, INSERT sign-extension, LSR off-by-one, BSCC absolute addressing, EXTRACT/EXTRACTU offset clamp). These would have produced subtle DSP corruption the moment the surround firmware actually used these opcodes — the automated review saved my local users from the consequences. Thank you both.

PR #2871, opened 2026-05-19
build: Configure meson to optimize for x86_64-v3 by default

GXTX:
Follow-up on #1557 

After some further off-platform discussions and research I believe x86-64-v3 march will be a good point between *usable* CPUs with xemu, and supporting *modern* instructions (i.e. AVX2) which should allow for better compiler optimizations.

The decision around this is based on Steam Hardware Survey (with AVX2 being supported on 95% of CPUs), and Ubuntu's inclusion of an `amd64v3` optimized repo in their 26.04 LTS release.

This will be a hard limit of Intel Haswell+ (2013), and AMD Excavator+ (2015).

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -196,6 +196,9 @@ case "$platform" in # Adjust compilation options based on platform
     Linux)
         echo 'Compiling for Linux...'
         sys_cflags='-Wno-error=redundant-decls'
+        if [ "$target_arch" == "x86_64" ]; then
+            sys_cflags="$sys_cflags -march=x86-64-v3"
+        fi
         opts="$opts --disable-werror"
         postbuild='package_linux'
         ;;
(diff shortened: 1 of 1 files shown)


mborgerson:
It turns out that meson.build inherited a configuration option for this already a couple of years ago (with some [discussion](https://lists.gnu.org/archive/html/qemu-devel/2024-06/msg03766.html) about why `-march=x86-64-v*` wasn't used):

https://github.com/xemu-project/xemu/blob/master/meson.build#L448-L454

So we could simplify this down a bit by just changing this config option's default value: https://github.com/xemu-project/xemu/blob/6318bb112091635ef908255019e4d42956bc5fa8/meson_options.txt#L390

GXTX:
Sure. Changed to that.

GXTX:
Changed back to original. Would require more work -- QEMU's current meson set-up appears not to *only* select this when the target is x86, thus trying to enable AVX when compiling for arm.

mborgerson:
I resolved the issues discovered with the x86_version approach (among some other things) in #2890, so I'll close this PR in favor of that. Thank you for championing the optimizations!

Issue #2852, opened 2026-05-08
Could not select file location under System->Files using xemu installed on Archlinux Hyprland window manager


jrullo:
### Bug Description

Could not choose a file location to set where roms, bios, disk, or eprom are to be located. 

1) Installed xemu via "paru -S xemu" command
2) run xemu from app runner in Hyprland
3) window with xemu opens
4) navigated to Settings and System section
5) Was able to change System memory and click AV Pack to show choices. 
6) under Files heading the next line MCPX Boot ROM file selector line has (None Selected) under the title. I click on the line expecting to get a file picker dialog of some kinds. 
7) tried to click any blank area on the line. highlighted display line just blinks momentarily and does nothing
8) tried to click the little icon of a white page at the end of the line, same behavior
9) clicked the x icon at the end of the line and get a message at top of window "Application restart required to apply settings" with an x to clear the message. I made no changes but see this message. nothing has changed on screen. 
10) same behavior for any line in the Files area.

### Expected Behavior

I expected a file chooser of some type to open where I could click a suggested directory, or "..." to choose my own directory. 

### xemu Version

xemu_version: 0.8.134-31-g7c7196b8a4
xemu_commit: 7c7196b8a4149c653e85a6b7cbac108eef022d91
xemu_date: Wed May  6 01:12:17 UTC 2026

### System Information

OS    ➜  Garuda Linux x86_64
 ├   ➜  Linux 7.0.3-zen1-2-zen
 ├ 󰏖  ➜  3023 (pacman)
 └   ➜  bash 5.3.9

WM    ➜  Hyprland 0.54.3 (Wayland)
 ├ 󰀻  ➜  Obsidian-Sand [GTK2/3/4]
 ├   ➜  Big-Banana (48px)
 ├   ➜  kitty 0.46.2
 └   ➜  NotoSansMono-Regular (14pt)

PC    ➜  Gigabyte Technology Co., Ltd. B550 MB Type B550 AORUS PRO V2
 ├   ➜  AMD Ryzen 9 5900X (12) @ 4.95 GHz GHz
 ├ 󰢮  ➜  AMD AMD Radeon RX 6800 XT @  GHz
 ├   ➜  8.26 GiB / 62.69 GiB (13%)
 ├ 󰓡  ➜  0 B / 62.69 GiB (0%)
 ├ 󰋊  ➜  336.59 GiB / 465.27 GiB (72%) - btrfs
[...]

PR #2873, opened 2026-05-22
ui: Add fallback FileDialog for linux

SyberHexen:
- Fixes the issues with setcap permissions killing FileDialog. [#2852](https://github.com/xemu-project/xemu/issues/2852)

### DIFF
--- a/ui/xui/misc.cc
+++ b/ui/xui/misc.cc
@@ -56,7 +56,9 @@ static void RunOnMainThread(std::function<void()> &&func)
 // Check if running as root or with elevated file capabilities (setcap)
 static bool ShouldBypassXDGPortal()
 {
-    // Force reset process dumpability to handle privilege drop state transitions cleanly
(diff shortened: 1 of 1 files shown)


SyberHexen:
/gemini review
While I'm happy with this on linux, gemini wanted to change (fix) those mac/windows specific functions so they should be tested or CR by someone with more innate understanding of SDL interactions.

mborgerson:
It looks like SDL3 has a Zenity driver (https://wiki.libsdl.org/SDL3/SDL_HINT_FILE_DIALOG_DRIVER). According to the docs:

> By default, SDL will try all available dialog backends in a reasonable order until it finds one that can work, but this hint allows the app or user to force a specific target.

I think SDL3 should be falling back to Zenity if portal is non-functional?

mborgerson:
Waiting on author response. Will close for now.

Issue #1837, opened 2025-01-02
Vulkan: Frame scaling/resolution issues causing distortion in some games


ehw:
### Title

This issue affects multiple titles, here are a few that I've noticed but there are definitely more:

https://xemu.app/titles/4343000e/#Auto-Modellista CC-014 [4343000e]
https://xemu.app/titles/41430007/#Aggressive-Inline AC-007 [41430007]
https://xemu.app/titles/45410083/#Black EA-131 [45410083]
https://xemu.app/titles/4b4e002e/#Karaoke-Revolution-Party KN-046 [4b4e002e]

### Bug Description

Some games seem to produce vertically distorted outputs when using the Vulkan backend that does not appear in the OpenGL one. The distortion appears over all graphics and elements in the frame, and can also affect full motion video. The distortion appears consistently and in the same areas in all content while playing.

Steps to reproduce:
1.) You just have to set Vulkan as your backend while a game is running, games affected by this issue appear to be affected the moment the games are booted.

All screenshots here were taken with the Vulkan backend at native resolution.

Aggressive Inline:
Notice how there are some parts of the Akklaim logo that look edged and uneven. This isn't how it's supposed to look and it looks fine in OpenGL
![image](https://github.com/user-attachments/assets/c587fbfc-b274-4a95-8a18-943cc5602114)

Black:
The A/B button icons appear uneven and seem to be distorted. The "A" in the word "PAUSE" should be symmetrical but looks uneven here.
![image](https://github.com/user-attachments/assets/3c21d938-7e27-4d01-9d6a-6134b5174b03)

Karaoke Revolution Party:

### Expected Behavior

The video output of the Vulkan backend should be on par with the OpenGL backend. In this case, the frame resolution/scaling should be identical between the two.

All screenshots here were taken with the OpenGL backend.

Aggressive Inline:
In comparison to the image posted above, take note that the Akklaim logo looks even and lacks the hard edges as seen in the screenshot from the Vulkan backend.
![image](https://github.com/user-attachments/assets/b94f01fc-096b-4e23-b43d-73c8a8134eff)

Black:
In comparison to the image posted above, notice that the A/B button icons look fine and don't seem as distorted. The "A" in the word "PAUSE" seems more symmetrical. 
![image](https://github.com/user-attachments/assets/ff3049dd-3e56-4165-b5c7-b286bd1e5e00)

### xemu Version

Version:      0.8.2
Branch:       master
[...]

Issue #1833, opened 2025-01-01
Vulkan: Broken Texture Filtering on some games


ehw:
### Title

This issue affects multiple titles:
https://xemu.app/titles/4d530006/#NFL-Fever-2002 MS-006 [4d530006]
https://xemu.app/titles/4156002b/#Spider-Man-2 AV-043 [4156002b]
https://xemu.app/titles/4c410011/#Star-Wars-Battlefront LA-017 [4c410011]
https://xemu.app/titles/4c41001a/#Star-Wars-Battlefront-II LA-026 [4c41001a]

### Bug Description

Sometimes it appears as though the texture filtering on distant polygons seems broken or at the very least not on par with the OpenGL backend. I'm not sure if this is all related or even if it's all related to texture filtering, but it definitely looks off in comparison to the OpenGL backend which looks more accurate.

Screenshots are from the Vulkan backend.

NFL Fever 2002:
1.) Make sure Vulkan is set as your video backend.
2.) Start the game.
3.) Press start to skip the opening FMV and wait at the loading screen for the title screen to appear.
4.) Press Start at the title screen.
5.) Go to PRACTICE
6.) Select OPEN PRACTICE
7.) Select Offense
8.) Select a Team
9.) Press A to "Kickoff" at the Controller Select menu
10.) Once the game has loaded in, press the A button twice to skip the form menu.
11.) You should now be looking behind your player waiting to kick off.

Screenshot:
Red arrows are from me. Notice how theres an uneven break on the lines on the floor the further away towards the wall.
![image](https://github.com/user-attachments/assets/678d2d59-be95-4112-a075-ba7601634d4e)

Spider-Man 2:
1.) Make sure Vulkan is set as your video backend.
2.) Start the game.
3.) Press start to skip the opening FMV and then press Start at the title screen.
4.) Press A at the name entry screen.
5.) Select a swing type and wait for the game to load.
6.) Press A to skip the opening cutscene.

Screenshot:
Red arrows are from me. Pause the game and alternate between the backends to see the texture filtering on the road in the distance change.
![image](https://github.com/user-attachments/assets/826fd644-19e5-4179-ba5e-e9e427e50e4c)

Star Wars: Battefront:

Screenshot:
Red arrows are from me. The shadows for the trees in the distance don't seem visible with the Vulkan backend but appear in the OpenGL one
![image](https://github.com/user-attachments/assets/a53eae76-745b-4eb9-8f0b-a2336e7e3c4f)

[...]

PR #2874, opened 2026-05-23
nv2a/vk: Fix blurry 2D/UI rendering at internal resolutions > 1x

stianeklund:
I believe this is a bug introduced when the Vulkan renderer was initially added. Please let me know if the PR format is not acceptable.

When the CPU writes directly to a surface in VRAM, `pgraph_vk_upload_surface_data()` is called to sync that data to the GPU.

At internal resolutions above 1x, upscaling is done via `vkCmdBlitImage`. Color surfaces were explicitly using `VK_FILTER_LINEAR` for this blit. AFAIK linear filtering will ultimately introduce blur, and I'm not sure that's something you want for upscaling? Maybe I'm naive here but I don't see why we can't simply simply replicate pixels, (which is also what the GL renderer does via `surface_copy_expand()`, so hopefully fairly safe to do?). This fix switches the blit to `VK_FILTER_NEAREST` unconditionally.

In practice this affects content that writes its framebuffer from the CPU. I haven't found a ton of good repro scenarios for this just yet; mainly the XDK dashboard is _very_ obvious but I'll try to find a different example.. I suspect some game menus and other applications may that use CPU-rendered content. Pure 3D rendering is unaffected as it never goes through this path.

Edit: I've decided to change the downscale path to work the same way.

---

Also as an aside out of curiosity:

The downscale / download path at [line 251](https://github.com/xemu-project/xemu/blob/e72e76bd1f12ce45c6c63b4196c6918ea2cc705d/hw/xbox/nv2a/pgraph/vk/surface.c#L251-L256) has the same `surface->color ? VK_FILTER_LINEAR : VK_FILTER_NEAREST` logic:

```c
        vkCmdBlitImage(cmd, surface->image,
                       VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL,
                       surface->image_scratch,
                       surface->image_scratch_current_layout, 1, &blit_region,
                       surface->color ? VK_FILTER_LINEAR : VK_FILTER_NEAREST);
```

Wouldn't the CPU expect accurate pixels from a rendered framebuffer in this case; not blended values? If this is also a bug I can include this in the PR or create a separate one if you like. I'm genuinely not sure about the downscaling part as it's a bit over my head.

### DIFF
--- a/hw/xbox/nv2a/pgraph/vk/surface.c
+++ b/hw/xbox/nv2a/pgraph/vk/surface.c
@@ -252,7 +252,7 @@ static void download_surface_to_buffer(NV2AState *d, SurfaceBinding *surface,
                        VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL,
                        surface->image_scratch,
                        surface->image_scratch_current_layout, 1, &blit_region,
-                       surface->color ? VK_FILTER_LINEAR : VK_FILTER_NEAREST);
+                       VK_FILTER_NEAREST);
 
         pgraph_vk_transition_image_layout(pg, cmd, surface->image_scratch,
                                           surface->host_fmt.vk_format,
--- a/hw/xbox/nv2a/pgraph/vk/surface.c
+++ b/hw/xbox/nv2a/pgraph/vk/surface.c
@@ -1236,7 +1236,7 @@ void pgraph_vk_upload_surface_data(NV2AState *d, SurfaceBinding *surface,
         vkCmdBlitImage(cmd, surface->image_scratch,
                        surface->image_scratch_current_layout, surface->image,
                        VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL, 1, &blitRegion,
-                       surface->color ? VK_FILTER_LINEAR : VK_FILTER_NEAREST);
+                       VK_FILTER_NEAREST);
     } else {
         // Note: We should be able to vkCmdCopyBufferToImage directly into
         // surface->image, but there is an apparent AMD Windows driver


abaire:
It makes sense to me that it'd use nearest for integer scaling.

It would be pretty straightforward to craft tests for this but since it's not something that HW supports I guess the best we could do would be to render something and then check access the surface via system RAM to ensure that it matches expectations.

You could use https://github.com/abaire/pbkitplusplus/blob/e91d509e4f75d6d815807d80443e4e56aab89772/src/nv2astate.h#L912 or just render to the back buffer normally and access it via CPU after the render. There are some examples in https://github.com/abaire/nxdk_pgraph_tests that you could inform how to read from the back buffer.

stianeklund:
@abaire Thanks for the input. `pbkitplusplus` and the pgraph tests (just wow, very comprehensive and impressive tbh), very helpful. I'm not sure if you hinted towards contributing to nxdk here but I'm not sure why or how considering this is a xemu bug? I might have misinterpreted / misunderstood.

Photo of two screenshots side by side, ignore the middle bar as this is just two side by side photos in Windows (left this PR, right xemu 0.8.135):

Photos_Zg9beJnGTp

I tried a few different variants. My first idea was to render narrow bands, like in the first image, but the issue is hard to see unless you zoom in quite a bit.

In the end, I switched to a grid where the right half of the image uses the GPU path as a reference.

Since this bug presumably only affects one specific path inside xemu, (the upload blit), I made a test where the two sides of the screen take different routes through the renderer:

| Screen region | Path | Purpose | What it does |
| --- | --- | --- | --- |
| Left half | CPU path | Path under test | Writes pixel data directly into VRAM via `pb_back_buffer()`. On the next GPU command, xemu calls `pgraph_vk_upload_surface_data()`, which blits the native-resolution surface up to the internal scaled resolution. |
| Right half | GPU path | Reference path | Calls `pb_fill()`, which sends pushbuffer commands to the GPU. xemu executes this at the already-scaled resolution, so there is no upload blit involved. `pb_fill()` produces pixel-perfect output regardless of the upload-blit filter setting. |

Latest xemu `v0.8.135`, Vulkan, 4x upscale:

xemu-2026-05-24-09-01-33-vulkan-4x-v0 8 135

This PR, Vulkan, 4x upscale:

xemu-2026-05-24-09-01-18-vulkan-4x

abaire:
> ... I'm not sure if you hinted towards contributing to nxdk here but I'm not sure why or how considering this is a xemu bug? I might have misinterpreted / misunderstood.

I think there is at least a partial misunderstanding; the title of the pgraph test repo is just making note of the fact that it uses the nxdk. The primary purpose is to exercise the nv2a to produce golden images (https://github.com/abaire/nxdk_pgraph_tests_golden_results) that can be compared to emulator output (e.g., https://github.com/abaire/xemu-nxdk_pgraph_tests_results). Generally this doesn't involve changes to the nxdk, though sometimes new commands/bits/features are needed and eventually upstreamed.

What I was saying is aligned with your point - it's not clear that a test used to validate upscale filtering fits this mold since the hardware is always 1x, but that it'd be good to at least verify the behavior w/ the PR (as you've done).

We could also validate the downscaling by having the CPU read a row of pixels after the GPU blit to verify that they're not filtered, but we'd need to update the test harness to run tests at scales > 1x, which is (intentionally) not done today. It'd be useful to write the test anyway if you'd like to also address the downscaling in this PR.

> Photo of two screenshots side by side, ignore the middle bar as this is just two side by side photos in Windows (left xemu 0.8.135, right this PR):

Are you sure about the order? It looks like the right side is blurry so I'd expect that to be 0.8.135.

> I tried a few different variants. My first idea was to render narrow bands, like in the first image, but the issue is hard to see unless you zoom in quite a bit.
> 
> In the end, I switched to a grid where the right half of the image uses the GPU path as a reference.

This pair looks like I'd expect (diagonal lines are great for visualizing linear vs pixel-perfect scaling).

Triticum0:
Might be related #1837, #1833

stianeklund:
@Triticum0 I don't think so, I tested Aggressive Inline and I'm not able reproduce what is described in the first issue tbh, these fixes are only affected by surfaces written to by CPU; I have no clue what games do this :P 

In fact I even tested it using OpenGL 1x, 2x and Vulkan 1x,2x etc just to see if there was any noticeable difference, and there isn't to my eye.

stianeklund:
@abaire Ah, yes, sorry about that had the order wrong on the first image, fixed, thanks. I created a PR with what I think is along the lines of what you had in mind: https://github.com/abaire/nxdk_pgraph_tests/pull/306

I've tested this quite a bit now and it seems to work as intended. I hope this actually is useful in some way. As mentioned this is _way_ out of my league. 😅 

Also updated the PR to not use linear for downscaling as well as I think this is the correct approach; kept this a separate commit in case it's wrong so it can be dropped.

mborgerson:
Nearest sampling makes sense for an image of black & white checkerboard pattern as in your demonstration, but I'm not sure it's best in most cases. Can you provide some other examples where this change to nearest does look better than linear?

Linear filter was intentionally chosen at this point to try to make things look "better" when scaling (e.g. downscaling a 2x->1x with a linear filter is an antialiasing technique), but may not not be the right choice for all cases. Nearest does make more sense for upscaling surfaces when sharper, pixelated appearance is desired--it's not unreasonable to have that be the default.

For the GL backend's surface copy method--I wrote that for quick scaling, not for quality--eventually it will be replaced by hardware scaling.

This could be a matter of preference--in which case a configuration option might make sense.

stianeklund:
@mborgerson Interesting. I honestly just thought this was a bug, but I see your points, ultimately I guess it depends on what you're wanting to upscale (or downscale), but from my perspective it makes sense to use nearest when the source is a complete framebuffer?

Regarding having an option for it: The programmer side of me thinks this is a good idea, but as a user I think it'd be confusing seeing there already is a filtering option.  As I thought this was initially a bug I now realize I may have jumped the gun on this a little bit. Please let me know if you think it's better to approach this in a different way.


I haven't noticed any actual difference in games or most applications, but the clearest example I have is the XDK Launcher. I was hoping to find some other examples but they're nowhere close to being as noticeable as this. Let me know if you want me to remove these screenshots & find other examples, I'm happy to try to do so.

| Build | Renderer | Scale | Result |
|---|---|---:|---|
| xemu 0.8.135 | Vulkan | 2x | xemu 0.8.135 Vulkan 2x |
| This branch | Vulkan | 2x | this branch Vulkan 2x |
| xemu 0.8.135 | OpenGL | 2x | xemu 0.8.135 OpenGL 2x |

stianeklund:
@mborgerson would you rather see this as a configurable option in GUI? You mention potentially replacing these with hardware scaling, were you thinking of using shaders for this? I think the shader-approach is a good idea personally.

Let me know, I can close this PR and try to contribute with something else instead; the main reason I created this is because I thought it was a bug :P

mborgerson:
@stianeklund I have not forgotten about this PR! Let me think on it a bit more and I'll either merge it or have some actionable feedback

mborgerson:
For now it makes sense to align with the GL renderer, and use nearest. We can make it configurable in the future if need be

PR #2880, opened 2026-05-25
Fix font issue with icon

commie-caden:
Previously, the icon used a text object which loaded a font to render the name "Xemu". This relied on the "Ubuntu Mono" font being installed. If it wasn't, the icon looked like this for users:

image

Well, it's looked like this for several years, so I finally got annoyed enough to do something about it! All I did was convert the text object into a path, so the font is now baked into icon as hard-coded vector paths.

### DIFF
--- a/ui/icons/xemu.svg
+++ b/ui/icons/xemu.svg
@@ -2,23 +2,23 @@
 <!-- Created with Inkscape (http://www.inkscape.org/) -->
 
 <svg
-   xmlns:dc="http://purl.org/dc/elements/1.1/"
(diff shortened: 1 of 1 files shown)


mborgerson:
Thanks! I appreciate this PR but I think this is a duplicate of #2817 (which should have been merged a while back)

commie-caden:
> Thanks! I appreciate this PR but I think this is a duplicate of #2817 (which should have been merged a while back)

Why's it still busted? It didn't look merged when I edited the file.

mborgerson:
@TriVoxel The PR isn't merged yet. I'll merge it soon

PR #2921, opened 2026-07-05
CHD support

ghost:
I added CHD support. XISO and Redump ISOs are recognized as CHD. The library used is https://github.com/rtissera/libchdr.

### DIFF
--- a/block/chd.c
+++ b/block/chd.c
@@ -1,7 +1,10 @@
 /*
  * QEMU Block driver for CHD images
  *
- * This work is licensed under the terms of the GNU GPL, version 2 or later.
+ * Copyright (c) 2026 JBW89
+ *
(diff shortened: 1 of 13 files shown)


ghost:
tenor

MasonT8198:
#2809

RinMaru:
I figured I should add context here but I tried this build and it works perfectly. Swapping disc work, loading up last played at startup works. Redump style CHDs work and Trimmed (Padding Partion Removed) work. There was no Diff in performance in my testing and I ran around in a few games for 2 hours or so without issue. Everything runs as it would an avarage xiso file.
Great work by the auther and I hope this can be merged. Id hate to have to revert to plain isos again after reducing the size of everything as a whole by 75% if not more. Good thing is in CHD DVD format which im using converts back to uncompressed xiso if needed.

PR #2929, opened 2026-07-07
nv2a: defer PGRAPH IRQ raise from the pfifo thread to a bottom half (fixes random load-time hang)

flowa1911you:
Taking the BQL from the pfifo thread (NO_OPERATION and context-switch interrupt raising) can deadlock: a BQL holder may itself be blocked waiting on the pfifo thread.

**Concrete scenario, captured with gdb on Windows:** IDE DMA loading game data into RAM that overlaps a GPU-watched range holds the BQL inside the block-layer completion callback and waits in `mem_check_access_callback_ramaddr` for the flush that only the pfifo thread performs — while the pfifo thread sits in `bql_lock()` inside the `NO_OPERATION` handler. Both threads wait on each other forever and the emulator hard-freezes.

Thread stacks at the moment of the hang:
- pfifo thread: `qemu_mutex_lock(bql)` ← `pgraph_NV097_NO_OPERATION_handler` ← `pgraph_method` ← `pfifo_run_pusher`
- BQL holder: `qemu_event_wait` ← `mem_check_access_callback_ramaddr` ← `flatview_write` ← `pci_dma_read` ← `bmdma_rw_buf` ← `ide_atapi_cmd_read_dma_cb` ← `blk_aio_complete`

Observed in the wild as a long-standing random hang when starting games; with Silent Scope Complete it reproduces every few boots during the disc-heavy loading phase.

**Fix:** schedule the `nv2a_update_irq()` call on a main-loop bottom half (which runs with the BQL already held) instead of acquiring the BQL from the pfifo thread. `pending_interrupts` is set before scheduling, and `nv2a_update_irq()` recomputes from pending & enabled, so the deferral is safe and idempotent.

### DIFF
--- a/hw/xbox/nv2a/pgraph/pgraph.c
+++ b/hw/xbox/nv2a/pgraph/pgraph.c
@@ -177,6 +177,23 @@ void pgraph_write(void *opaque, hwaddr addr, uint64_t val, unsigned int size)
     qemu_mutex_unlock(&d->pfifo.lock);
 }
 
+static void pgraph_update_irq_bh(void *opaque)
+{
+    NV2AState *d = opaque;
+    nv2a_update_irq(d);
+}
+
+/* Raising a PGRAPH interrupt needs the BQL, but taking the BQL from the
+ * pfifo thread can deadlock: a BQL holder may itself be blocked waiting on
+ * this very thread. For example, IDE DMA loading game data into RAM that
+ * overlaps a GPU-watched range holds the BQL and waits for the
+ * memory-access-callback flush that only the pfifo thread performs. Defer
+ * the IRQ raise to a main-loop bottom half, which runs with the BQL held. */
+static void pgraph_schedule_irq_update(NV2AState *d)
+{
+    aio_bh_schedule_oneshot(qemu_get_aio_context(), pgraph_update_irq_bh, d);
+}
+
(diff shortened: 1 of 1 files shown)


Triticum0:
@flowa1911you just to let you know Tovarichtch on github was working on light gun support xemu while working on chihiro implementation, might be on his personal repo if interested.

jessedye:
Tested this on a Raspberry Pi 5 (aarch64, Mesa V3DV) and it costs frame time here.

Tony Hawk's Pro Skater 3, same gameplay snapshot, same 150s duration, only the binary differing:

| | without | with |
|---|---|---|
| fps | 37.79 | 34.51 (-9%) |
| p50 frame | 26.13 ms | 32.55 ms (+25%) |
| gpu_busy | 10.28 ms | 10.04 ms (flat) |

GPU work is unchanged, so the cost is host-side latency, consistent with each PGRAPH interrupt now waiting for a main-loop iteration instead of being raised inline. Reverting it alone recovered fps to within noise (-1%).

We do hit the lock-ordering problem this fixes, so this is not an objection to the approach - but on a slower host the added latency per interrupt is significant. Would it be possible to raise inline when the BQL is already held (or uncontended) and only defer on the path that would deadlock? That would keep the fix without paying a main-loop turn for every interrupt.

PR #2930, opened 2026-07-07
nv2a: consume INLINE_ARRAY vertex data in bulk

flowa1911you:
`NV097_INLINE_ARRAY` was defined through the NON_INC wrapper, which invokes the handler once per pushbuffer word: bulk geometry paid a dispatch-loop iteration (plus the method logging hook) for every single 32-bit word.

This PR defines the handler directly and lets it consume the whole run of available words at once, reporting the amount through `*num_words_consumed` — the same batching other handlers already use. The unusual `inc` case keeps the previous word-at-a-time behavior.

Draw-call-heavy titles that submit geometry inline benefit substantially. Measured with Silent Scope Complete (~5000 inline-array draws per frame): several milliseconds per frame less method-processing time on the pfifo thread. This was part of a series of local fixes that took the game from ~15 fps to its full frame-rate cap on the same machine (Intel iGPU laptop).

### DIFF
--- a/hw/xbox/nv2a/pgraph/pgraph.c
+++ b/hw/xbox/nv2a/pgraph/pgraph.c
@@ -2753,11 +2753,28 @@ DEF_METHOD(NV097, DRAW_ARRAYS)
     pg->draw_arrays_prevent_connect = false;
 }
 
-DEF_METHOD_NON_INC(NV097, INLINE_ARRAY)
+/* Defined without the NON_INC wrapper on purpose: bulk vertex data would
+ * otherwise pay a dispatch-loop iteration per word. Consume the whole run
+ * at once instead. */
+DEF_METHOD(NV097, INLINE_ARRAY)
 {
     pgraph_check_within_begin_end_block(pg);
     assert(pg->inline_array_length < NV2A_MAX_BATCH_LENGTH);
-    pg->inline_array[pg->inline_array_length++] = parameter;
+
+    if (inc) {
+        pg->inline_array[pg->inline_array_length++] = parameter;
+        return;
+    }
+
+    size_t count = MIN(num_words_available,
+                       (size_t)NV2A_MAX_BATCH_LENGTH -
+                           (size_t)pg->inline_array_length);
+    for (size_t i = 0; i < count; i++) {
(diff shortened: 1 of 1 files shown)


flowa1911you:
The suggested overflow can't happen as written: the batched path already clamps the copy with `count = MIN(num_words_available, NV2A_MAX_BATCH_LENGTH - pg->inline_array_length)`, so it never writes past the buffer.

On the "release builds where assertions are disabled" premise: QEMU/xemu builds ship with assertions enabled (`b_ndebug` stays off; assertion failures do fire in release binaries in the wild), so the entry assert is active — and it's the exact same guard, with the same semantics, that the previous per-word handler had. A guest pushing more than `NV2A_MAX_BATCH_LENGTH` words hits that assert both before and after this change, so this PR doesn't alter that contract.

If maintainers would rather harden this path to discard excess data instead of asserting, I'm happy to do that — but since it would change behavior inherited from the existing handler, it feels like a separate change.

hw/xbox/nv2a/pgraph/pgraph.c
```diff
@@ -2753,11 +2753,28 @@ DEF_METHOD(NV097, DRAW_ARRAYS)
     pg->draw_arrays_prevent_connect = false;
 }
 
-DEF_METHOD_NON_INC(NV097, INLINE_ARRAY)
+/* Defined without the NON_INC wrapper on purpose: bulk vertex data would
+ * otherwise pay a dispatch-loop iteration per word. Consume the whole run
+ * at once instead. */
+DEF_METHOD(NV097, INLINE_ARRAY)
```
abaire, line 2759: While the effect of this approach makes sense, it feels potentially confusing and duplicative to reuse DEF_METHOD this way.

I wonder if we should add a new `DEF_METHOD_CUSTOM` macro that avoids the `pgraph_method_inc`/`pgraph_method_non_inc` entirely instead.

abaire:
FYI, #3023 is a (problematic, wierd AI generated) change that overlaps with yours.  The one interesting thing it does is to cover ARRAY_ELEMENTS in addition to INLINE_ARRAY which may be worth bringing into this PR if you're interested (or if you're interested in doing a followup).

PR #2932, opened 2026-07-11
nv2a: Add driver version to shader cache validation.

### LINKED ISSUE #2597: Tighten FS shader cache validation
### Bug Description

It looks like cached shader loading wasn't as strict as I remembered. Apparently we are only checking xemu version and GL vendor, but not driver version, which explains failure after driver updates in #2370.

https://github.com/xemu-project/xemu/blob/b27f54789ca97b928030f5221cc07ba99ceef295/hw/xbox/nv2a/pgraph/gl/shaders.c#L622-L626

### Expected Behavior

xemu probably shouldn't be loading cached shaders across driver versions.

### xemu Version

b27f54789ca97b928030f5221cc07ba99ceef295

### System Information

N/A

### Additional Context

_No response_

Reedzit:
I added GL_VERSION to the already existing cache verification methods. This should prevent the need to manually delete the cache after updating drivers.
This should fix xemu-project#2597

I did not experience any slow down or issues after initial cache warm up.

### DIFF
--- a/hw/xbox/nv2a/pgraph/gl/shaders.c
+++ b/hw/xbox/nv2a/pgraph/gl/shaders.c
@@ -252,6 +252,7 @@ static void generate_shaders(PGRAPHGLState *r, ShaderBinding *binding)
 }
 
 static const char *shader_gl_vendor = NULL;
+static const char *shader_gl_version = NULL;
 
 static void shader_create_cache_folder(void)
 {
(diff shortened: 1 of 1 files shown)


mborgerson:
Review bot left some good feedback which applies to existing code which this patch mirrors. Let's address it in a follow up PR

PR #2933, opened 2026-07-12
Lightgun support with per-device pointer input (Raw Input / evdev)

flowa1911you:
This adds lightgun support to xemu: an emulated **EMS TopGun II** (`usb-xbox-lightgun`, VID `0x0b9a` PID `0x016b`, XID subtype `0x50`) selectable as the emulated device on any controller port, aimed with real lightguns (Sinden and anything that presents itself as a pointer device) or a regular mouse.

Games genuinely detect a lightgun on the port, **including in-game calibration**: the `SET_REPORT` calibration report is accepted and its offsets are applied to the aim, matching real hardware behavior (`XInputSetLightgunCalibration`).

## Why a per-device input backend

Aiming requires telling multiple pointer devices apart: every system has at least one regular mouse, and 2-gun games need two guns on separate ports. SDL merges all mice into one virtual pointer, so this PR adds a small per-platform enumeration backend that exposes each pointer device individually as a bindable input device (listed by product name in the Input menu, DemulShooter-style):

- **Windows — Raw Input**: per-device `WM_INPUT` via the SDL message hook, absolute coordinates as reported by guns like the Sinden, hotplug via `WM_INPUT_DEVICE_CHANGE` (deferred to the main loop, since the hook runs outside the iothread lock).
- **Linux — evdev**: `/dev/input/event*` enumeration (absolute or relative pointers), hotplug via inotify. Saved port bindings also accept an `evdev:<node>` alias so frontends can assign detected guns to ports at launch time — [Batocera](https://batocera.org) integration uses this. Gun extra buttons that arrive as keyboard keys or low `BTN_*` codes on the gun's own device are read directly from it, because X11 ignores joystick-classified devices and those events never reach SDL (this is what makes e.g. the Sinden D-pad work out of the box).
- **Other hosts**: stubs — no pointer devices are listed, everything else is unaffected.

The platform-independent aim pipeline (`ui/xemu-pointer-aim.c.inc`, shared by both backends) maps the device position onto the actual game render area, reports the `ONSCREEN` wButtons flag accordingly, and offers optional adaptive smoothing (1-euro filter: steady crosshair when still, no added latency on fast moves) and sensitivity scaling, both exposed as sliders in the Input menu.

Buttons follow a fixed default mapping (trigger=A, right=B, middle=Start, X1=Back, X2=X) and every function (17 slots, including D-pad, White/Black and analog triggers for service menus) can be remapped to any detected pointer-device button or keyboard key via the `input.lightgun_mapping.*` settings.

## Compatibility

- No behavior changes for existing users: mice are **never auto-bound** to a port (unlike gamepads, every system has a mouse — grabbing a port with it would be surprising), and the two new UI-shortcut guards (`input.disable_dblclick_fullscreen`, `input.disable_rclick_menu`, useful so a gun trigger doesn't toggle fullscreen) **default to off**.
- The XID gamepad path is untouched; the lightgun device embeds and reuses the gamepad report handling.

## Testing

- Windows 11: Sinden Lightgun and regular mice, Silent Scope Complete (SS1/SS2/SS3 all playable, in-game calibration verified end-to-end)
- Batocera Linux 43 (x86_64): two Sinden guns simultaneously on ports 1/2, frontend-assigned via the `evdev:` alias, D-pad and extra buttons via direct device reads
- A build of this feature has been distributed to the lightgun community for several days (hundreds of downloads) without regressions reported on pad-only setups

Happy to split, rename options or adjust the approach as preferred.

### DIFF
--- a/config_spec.yml
+++ b/config_spec.yml
@@ -55,6 +55,44 @@ input:
   auto_bind:
     type: bool
     default: true
+  # Keep mouse clicks away from UI shortcuts while using a mouse/lightgun
+  disable_dblclick_fullscreen:
+    type: bool
+    default: false
+  disable_rclick_menu:
+    type: bool
+    default: false
+  # Lightgun aim feel: smoothing tames hand/camera jitter (0 = raw),
+  # sensitivity scales aim range around the screen center (1.0 = 1:1)
+  lightgun_smoothing:
+    type: number
+    default: 0.2
+  lightgun_sensitivity:
+    type: number
+    default: 1.0
+  # Optional custom bindings for lightgun ports. Empty = default mouse
+  # mapping (LEFT=trigger/A, RIGHT=B, MIDDLE=Start, X1=Back, X2=X).
+  # Formats: "mouse:<hash>:<left|right|middle|x1|x2>",
+  # "mouse:<hash>:k<input-event code>" (raw device key, e.g. Linux guns)
+  # or "key:<SDL key name>"
+  lightgun_mapping:
+    trigger: string
+    b: string
+    x: string
+    start: string
+    back: string
+    dpad_up: string
+    dpad_down: string
+    dpad_left: string
+    dpad_right: string
+    y: string
+    white: string
+    black: string
+    lstick: string
+    rstick: string
+    ltrig: string
+    rtrig: string
+    guide: string
   # Deprecated: use 'enable_rumble' on the controller mapping settings instead
   allow_vibration:
     type: bool
--- a/hw/xbox/meson.build
+++ b/hw/xbox/meson.build
@@ -16,6 +16,7 @@ specific_ss.add(files(
 	'xid.c',
 	'xblc.c',
 	'xid-gamepad.c',
+	'xid-lightgun.c',
 	))
 subdir('nv2a')
 subdir('mcpx')
--- a/hw/xbox/xid-lightgun.c
+++ b/hw/xbox/xid-lightgun.c
@@ -1,0 +1,369 @@
+/*
+ * QEMU USB XID Lightgun (EMS TopGun II)
+ *
+ * Copyright (c) 2013 espes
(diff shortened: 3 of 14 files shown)


Tovarichtch:
Is this work based on https://github.com/xemu-project/xemu/pull/1875 ?

flowa1911you:
Thanks — addressed in the updated commit:

- **VMState**: good catch, the calibration offsets are now serialized (`offset_x/y`, `offset_upp_x/y` added to `vmstate_usb_xbox_lightgun`).
- **`g_new0`/`g_free`**: switched in both new backends (note: `xemu-input.c` itself uses `malloc` for its `ControllerState`s, which is what I had mirrored — the new files now use GLib allocators as suggested).
- **`CreateFileW`**: switched, with UTF-8 → UTF-16 conversion of the device path.
- **`gunichar2`**: left as is — it *is* the standard GLib type for UTF-16 code units, and it's the exact pointer type `g_utf16_to_utf8()` takes (`const gunichar2 *str`). Casting to `guint16 *` would introduce a type mismatch with the GLib prototype.

flowa1911you:
@Tovarichtch No — I wasn't aware of #1875 when I built this; it was developed independently, using Cxbx-Reloaded's EMS TopGun II implementation as the *hardware reference* (VID/PID, XID subtype 0x50, the ONSCREEN wButtons flag and the calibration report format).

Looking at #1875 now, the main differences: that PR aims with the single SDL mouse/keyboard, while this one enumerates pointer devices **individually** (Raw Input on Windows, evdev on Linux) so multiple guns can be bound to different ports (2-player games), absolute-coordinate guns like the Sinden work natively, in-game calibration (`SET_REPORT`) is honored, and gun extra buttons are mappable. It also lets frontends assign guns to ports (the Batocera integration uses this).

That said, I'm happy to align with #1875's XID structure split or converge the two approaches if the maintainers prefer.

PR #2952, opened 2026-07-27
nv2a: Back PRAMIN with unified VRAM

korzewarrior:
PRAMIN is currently modeled as an independent RAM block. As a result, memory
reserved by `MmClaimGpuInstanceMemory` and accessed through system RAM is not
coherent with the NV2A PRAMIN aperture.

On the Xbox, PRAMIN exposes the final MiB of unified VRAM in reverse 16-byte
blocks. This change translates PRAMIN MMIO accesses into that VRAM range,
marks writes dirty for NV2A consumers, and routes internal RAMHT, DMA, and
object reads through the same mapping.

Live inspection confirms the expected mapping:

```
PRAMIN 0x10010 == physical 0x03feffe0
PRAMIN 0x10020 == physical 0x03feffd0
PRAMIN 0x10030 == physical 0x03feffc0
PRAMIN 0x10040 == physical 0x03feffb0
```

The four dwords within each block retain their order.

Testing:

- Clean `./build.sh -j16`
- `./scripts/checkpatch.pl --no-tree --strict --branch origin/master..HEAD`
- Live PRAMIN/VRAM coherence check during NV2A initialization

### DIFF
--- a/hw/xbox/nv2a/nv2a.c
+++ b/hw/xbox/nv2a/nv2a.c
@@ -22,6 +22,9 @@
 #include "hw/xbox/nv2a/nv2a_int.h"
 #include "qemu/main-loop.h"
 
+#define NV2A_PRAMIN_SIZE 0x100000
+#define NV2A_PRAMIN_BLOCK_SIZE 16
+
 void nv2a_update_irq(NV2AState *d)
 {
     /* PFIFO */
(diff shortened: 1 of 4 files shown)


hw/xbox/nv2a/pfifo.c
```diff
@@ -523,10 +523,9 @@ static RAMHTEntry ramht_lookup(NV2AState *d, uint32_t handle)
 
     assert(ramht_address + hash * 8 < memory_region_size(&d->ramin));
```
abaire, line 524: nit: there are a lot of `ram_address + hash * 8` recalculations here; consider assigning this to a const local var to make sure they stay in sync.

hw/xbox/nv2a/nv2a.c
```diff
@@ -53,14 +56,85 @@ void nv2a_update_irq(NV2AState *d)
     }
 }
 
+/*
+ * RAMIN occupies the final MiB of VRAM. Its 16-byte blocks are exposed in
```
abaire, line 60: Have you checked this on a devkit / 128 MiB modded console? Is it always top of memory or is it the last of the standard 64 MiB?

hw/xbox/nv2a/nv2a.c
```diff
@@ -53,14 +56,85 @@ void nv2a_update_irq(NV2AState *d)
...
+           (addr & (NV2A_PRAMIN_BLOCK_SIZE - 1));
+}
+
+static uint64_t nv2a_pramin_read(void *opaque, hwaddr addr, unsigned size)
+{
+    NV2AState *d = opaque;
+    uint64_t value = 0;
+
+    assert(addr + size <= NV2A_PRAMIN_SIZE);
+    for (unsigned i = 0; i < size; i++) {
```
abaire, line 81: nit: Is it necessary (in practice) to do this byte-by-byte conversion? 

I would guess that any PRAMIN reads are going to be aligned within a single block; would it make sense to assert that and then memcpy?

PR #2956, opened 2026-07-30
nv2a/vk: only finish when a vertex RAM update affects a pending draw

MattPonton:
## Problem

`sync_vertex_ram_buffer()` rounds each vertex attribute range outward to page boundaries before testing guest dirty state:

```c
hwaddr start_addr = r->vertex_ram_buffer_syncs[i].addr & TARGET_PAGE_MASK;
hwaddr end_addr = r->vertex_ram_buffer_syncs[i].addr +
                  r->vertex_ram_buffer_syncs[i].size;
end_addr = ROUND_UP(end_addr, TARGET_PAGE_SIZE);
```

`pgraph_vk_update_vertex_ram_buffer()` then tests for write-after-read hazards against `uploaded_bitmap`, which is also page-granular.

The combination means a guest write *anywhere* in a page that happens to contain vertex data forces a full-page re-upload, collides with the bitmap, and submits the command buffer via `VK_FINISH_REASON_VERTEX_BUFFER_DIRTY` — even when the bytes those pending draws will read are all unchanged. Games that keep per-object state (matrices, animation, particle data) adjacent to vertex data in the same page pay this on nearly every draw.

## Measurements

I instrumented the hazard path to compare incoming bytes against the current contents of `BUFFER_VERTEX_RAM` — i.e. what the not-yet-submitted draws will actually read. Dead or Alive 2 Ultimate, ~98,000 reported conflicts:

| Metric | Value |
| --- | --- |
| Average upload per reported conflict | ~9,200 B (≈2.25 pages) |
| Average distance from first to last differing byte | ~3,682 B |
| Average bytes actually differing | ~754 B (~8% of the upload) |
| Uploads byte-for-byte identical to buffer contents | 2.9% |
| Smallest differing span observed | 1 byte |

So roughly **92% of every "vertex data changed" upload rewrites identical bytes**, and the changed bytes that do exist are frequently outside the ranges any pending draw sources vertices from.

## Change

Track the exact, pre-alignment ranges that draws already recorded into the current command buffer will read (`pgraph_vk_mark_vertex_ram_referenced()`). These are registered *after* each draw's own uploads complete, so a draw is never treated as conflicting with itself.

`pgraph_vk_update_vertex_ram_buffer()` keeps the page bitmap as a cheap first-level filter, then compares only the intersection of the pending write with those referenced ranges, finishing only when a byte inside one of them is genuinely about to change.

This narrows the test rather than removing it, and remains conservative in three ways:

1. The comparison is on actual bytes, not just addresses.
2. The recorded ranges are a superset of what the GPU indexes (they come from the same `update_memory_buffer()` calls that drive the existing sync logic).
3. If a command buffer needs more than `VK_MAX_VERTEX_RAM_REFERENCED` (256) distinct ranges, an overflow flag restores the previous behavior for that command buffer.

## Results

Dead or Alive 2 Ultimate, the stage where this was originally observed:

- Before: 40-47 fps (MSPF ~1), `FINISH_VERTEX_BUFFER_DIRTY` continuously non-zero (49-57)
- After:  59-60 fps (MSPF ~14), `FINISH_VERTEX_BUFFER_DIRTY: 0`

## Testing notes and limitations

- Tested on Windows (mingw cross build), NVIDIA RTX 3080 Ti, Vulkan renderer.
- `DRAW_ARRAYS` was 0 in every scene I captured; the affected stages use the inline-elements and inline-arrays paths. **The `draw_arrays` path is therefore not exercised by my testing** and would benefit from review or testing by someone with a title that uses it.
- Game coverage is limited. A missed referenced range would present as flickering or stale geometry rather than a crash.
- Setting `VK_MAX_VERTEX_RAM_REFERENCED` to 0 forces the overflow path and restores the old behavior, which is a convenient way to A/B any suspected regression.

## Relationship to #2345

#2345 proposed adding `&& r->in_draw` to the same condition. That was rejected as removing too much of the safety net, and I think that read was correct — in fact it removes all of it. `r->in_draw` is only true between `begin_draw()` and `end_draw()`, all three callers of `pgraph_vk_update_vertex_ram_buffer()` run with it false, and `pgraph_vk_finish()` asserts `!r->in_draw`, so the condition can never be true. The branch becomes unreachable and the hazard check is disabled outright.

The review comment there asked what data was actually changing. The table above is that answer, and this change acts on it by making the test precise instead of bypassing it.

### DIFF
--- a/hw/xbox/nv2a/pgraph/vk/draw.c
+++ b/hw/xbox/nv2a/pgraph/vk/draw.c
@@ -1234,6 +1234,8 @@ void pgraph_vk_finish(PGRAPHState *pg, FinishReason finish_reason)
                                 BUFFER_VERTEX_INLINE);
         sync_staging_buffer(pg, cmd, BUFFER_UNIFORM_STAGING, BUFFER_UNIFORM);
         bitmap_clear(r->uploaded_bitmap, 0, r->bitmap_size);
+        r->num_vertex_ram_referenced = 0;
+        r->vertex_ram_referenced_overflowed = false;
         flush_memory_buffer(pg, cmd);
         VK_CHECK(vkEndCommandBuffer(r->aux_command_buffer));
         r->in_aux_command_buffer = false;
@@ -1565,6 +1567,17 @@ static void sync_vertex_ram_buffer(PGRAPHState *pg)
         return;
     }
 
+    /* Preserve the exact (unaligned) ranges before they are rounded out to page
+     * boundaries below. These are the bytes the upcoming draw will actually
+     * source vertices from, and are registered as referenced once this draw's
+     * own uploads are done.
+     */
+    MemorySyncRequirement precise[NV2A_VERTEXSHADER_ATTRIBUTES];
+    size_t num_precise = r->num_vertex_ram_buffer_syncs;
+
+    memcpy(precise, r->vertex_ram_buffer_syncs,
+           num_precise * sizeof(precise[0]));
+
     // Align sync requirements to page boundaries
     NV2A_VK_DGROUP_BEGIN("Sync vertex RAM buffer");
 
@@ -1635,6 +1648,15 @@ static void sync_vertex_ram_buffer(PGRAPHState *pg)
     }
 
     r->num_vertex_ram_buffer_syncs = 0;
+
+    /* Register after the uploads above: the draw that reads these bytes has not
+     * been recorded into the command buffer yet, so its own upload is not a
+     * conflict with itself.
+     */
+    for (size_t i = 0; i < num_precise; i++) {
+        pgraph_vk_mark_vertex_ram_referenced(pg, precise[i].addr,
+                                             precise[i].size);
+    }
 
     NV2A_VK_DGROUP_END();
 }
--- a/hw/xbox/nv2a/pgraph/vk/renderer.h
+++ b/hw/xbox/nv2a/pgraph/vk/renderer.h
@@ -51,6 +51,9 @@ typedef struct MemorySyncRequirement {
     hwaddr addr, size;
 } MemorySyncRequirement;
 
+/* Max distinct vertex RAM byte ranges tracked per command buffer. */
+#define VK_MAX_VERTEX_RAM_REFERENCED 256
+
 typedef struct RenderPassState {
     VkFormat color_format;
     VkFormat zeta_format;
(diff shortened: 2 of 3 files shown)


jessedye:
We have an overlapping change in #2988 (guarding vertex RAM writes on what draws actually read). Rather than duplicate it, one datum from running this class of fix on a Raspberry Pi 5 (Mesa V3DV, aarch64) that may affect sizing here.

`VK_MAX_VERTEX_RAM_REFERENCED = 256` is sized against submissions per frame. Across our real gameplay logs we measure **1.7-2.2 `queue_submit` per frame** with 374-918 draw batches per frame — an order of magnitude fewer submissions than the tracking budget assumes. At that ratio the referenced set overflows and the optimisation silently disables itself rather than degrading, which is easy to miss because nothing in the perflog says so.

Happy to run the `DRAW_ARRAYS` coverage you flag as untested, on V3DV/aarch64 with per-frame counters for hazards, stalls and pass breaks, if that would help move this along.

MattPonton:
> Happy to run the `DRAW_ARRAYS` coverage you flag as untested, on V3DV/aarch64 with per-frame counters for hazards, stalls and pass breaks, if that would help move this along.

I welcome the help.

MattPonton:
@jessedye 
To follow up with something more useful than my last reply as I've now looked over your PR (#2988).

You're right that the fixed cap is the weak point I think. Your bitmap approach doesn't have such a weak point at all from what I see, which is the better shape over mine which can hit the ceiling and leave no signal when it does.

Here's some more data in regard to the ratio, from my workload as I've been profiling DOA2 Ultimate on a Steam Deck. In the heaviest scene I get the following: `QUEUE_SUBMIT: 1`, `BEGIN_ENDS: ~3430` on a given frame. That's one submission per frame against approximately 3400 draw batches, so even more than the 256 assumes from your raspberry pi numbers. In my scene, its almost entirely inline vertex data (`INLINE_ARRAYS: 2728`, `INLINE_ELEMENTS: ~708`, and `DRAW_ARRAYS: 0`), so I'm not certain how much of it actually reaches the vertex RAM path. Rather than infer overflow from the ratio, I'll look into adding a counter to see how often the set genuinely hits the cap and report back.

On ordering of our PRs: since #2988 fixes which set gets tested and mine is merely an optimization sitting on top of that same guardrail, it probably makes sense for yours to go first and for me to then rebase onto the project once merged, rather than us maintaining two separate versions for tracking.

Lastly, unrelated to our PR conflict, but possibly just interesting data given your Morrowind numbers: I see the same shape on the Steam Deck where my DOA2U test instance's Guest CPU thread gets pinned at ~94%, GPU at 8%, and the renderer thread with idle to spare. That aligns with your "vCPU pinned, GPU busy 12 ms of a 53 ms frame" find pretty closely, on completely separate hardware. I've been looking into this on the side as well since I'm trying to get XEMU DOA2U performance on the deck upgraded to 60 fps similar to your Morrowind goal.

MattPonton:
Marking this as draft while I continue to work on it. I'm doing this primarily for two reasons: the branch is about a month behind master, and my original measurements were taken against that older base, so they need redoing regardless.

Also, #2988 corrects which set the guard tests, and since this change sits on top of that same guard it makes sense for that to land first. Once it does I'll rebase onto it and re-measure to see whether this still adds anything — if it doesn't, I'll close this out.

I also added the Agent declaration added to the description per AGENTS markdown in the repo.

PR #2965, opened 2026-08-11
nv2a/vk: Flush queue when reusing surface memory

### LINKED ISSUE #2902: Crimson Skies: Menus and text completely broken
### Title

Title: https://xemu.app/titles/4d530021/#Crimson-Skies-High-Road-to-Revenge

### Bug Description

Image

Image

Image

Image

the game also gives off a damaged disc error unless you use a fresh hdd image, though that is probably unrelated

issue is present on every version from 0.8.134 onward, also the game has extreme framerate issues on at least 0.8.133 onwards. havent tested further. framerate is extremely bad regardless of resolution settings (tried from 4x, my usual, to 1x with no change) even though my system should be plenty strong for it.

EDIT: Menu issue isn't there when using OpenGL, only exists in Vulkan. Extreme framerate issues are still there regardless of settings 

### Expected Behavior

Image

### xemu Version

xemu v0.8.136

### System Information

OS: Windows 11
CPU: Intel Core I9 13900H
GPU: Nvidia RTX 4070
GPU Driver Version: 596.36

### Additional Context

_No response_

abaire:
PR #2839 removed an unnecessary flush in the VK pipeline to align it with GL behavior. Unfortunately this exposed a case where titles that reuse a surface to texture geometry repeatedly while mutating using the CPU during a single frame end up causing host to overwrite the texture before it is ever utilized by the host GPU (typically resulting in the last modification being applied to all geometry).

This change attempts to detect cases where a texture is reused during a single command buffer and forces a flush before uploading the new texture data.

Fixes #2902

Tests: https://github.com/abaire/nxdk_pgraph_tests/blob/4dcb3b0eec3f84a04a32a4cf46906af36dc2d6b6/src/tests/texture_cpu_update_tests.cpp#L145
HW results: https://abaire.github.io/nxdk_pgraph_tests_golden_results/results/Texture_CPU_Update/index.html

NOTE: This change doesn't precisely match HW behavior. The test currently uses an NV semaphore to await draw completion before mutating the texture. This matches the observed behavior of Crimson Skies and xemu works correctly (with this PR) when the semaphore is used. On hardware, removing the semaphore still allows the test to pass whereas on xemu it fails again, even with this PR. My guess is that there may be some implicit sync point that xemu is not currently implementing but from my standpoint this is a low priority and can be tracked as a known issue.


abaire:
It'd be great if some folks could test this to verify that it doesn't cause performance regressions. I added a new  counter to track this flush path so we can see if it's overtriggering but won't have time to do testing across titles myself for at least the next month or two.

sargashC:
I actually just booted up Crimson skies to play it and have this issue, how would I go about getting this patch to fix it?

abaire:
> I actually just booted up Crimson skies to play it and have this issue, how would I go about getting this patch to fix it?

You can download the artifact from the "Checks" header at the top of this page. Feel free to ask on the Discord if you need more assistance. 

If you test it and find any issues related to this change, please report them here.

PR #2968, opened 2026-08-13
Fix 2906 memory leak in slow vga path

### LINKED ISSUE #2906: null render memory leak
### Bug Description

start with or switch to NULL render on xubuntu linux  
ram usage with spike until overflow and the app crashes

Image

### Expected Behavior

ram usage should drop slightly from render being null'd

### xemu Version

0.8.134 offending build

0.8.133 last working

### System Information

PU:          Intel(R) Core(TM) i5-4210U CPU @ 1.70GHz
OS Platform:  Linux
OS Version:   Ubuntu 26.04 LTS
Manufacturer: Mesa
GPU Model:    llvmpipe (LLVM 21.1.8, 256 bits)
Driver:       4.5 (Core Profile) Mesa 26.0.3-1ubuntu1
Shader:       4.50

### Additional Context

(edit this does happen on windows albeit a lot slower)
repro'd on 134 in virtual box ubuntu 24 LTS

dinky pc specs issue was discovered on

PU:          Intel(R) Core(TM) i5-4210U CPU @ 1.70GHz
OS Platform:  Linux
OS Version:   Ubuntu 26.04 LTS
Manufacturer: Mesa
GPU Model:    llvmpipe (LLVM 21.1.8, 256 bits)
Driver:       4.5 (Core Profile) Mesa 26.0.3-1ubuntu1
Shader:       4.50

abaire:
PR #2691 decoupled the guest vblank handling from xemu GUI updates. This introduced a condition where the vblank timer may fire during a call to  `gl_render_frame`.
    
In cases where `gl_render_frame` falls back to creating a local texture (e.g., direct writes to the framebuffer by the guest or every frame when using the null renderer backend) this creates a race where the surface initiated by the slow path fallback may be mutated by the vblank timer thread before the paired call to `xb_surface_gl_destroy_texture`, leaking the created texture.
    
When using the null renderer with the UI at 60 fps, this happens extremely often, creating a very noticeable memory leak.
    
This PR ensures that the texture is always deleted, even if the surface is updated before the destroy call.
    
As a bonus, the second commit creates a blank texture in the null renderer which is reused for every frame, bypassing the churn of creating and immediately destroying textures in `gl_render_frame`'s fallback path. For clarity, I've verified that the first commit alone resolves the memory leak, this is just a further optimization and is not masking the leak.

Fixes #2906

### DIFF
--- a/hw/xbox/nv2a/pgraph/null/renderer.c
+++ b/hw/xbox/nv2a/pgraph/null/renderer.c
@@ -17,10 +17,30 @@
  * License along with this library; if not, see <http://www.gnu.org/licenses/>.
  */
 
-#include "qemu/osdep.h"
+#include "renderer.h"
+
 #include "qemu/thread.h"
-#include "hw/hw.h"
 #include "hw/xbox/nv2a/nv2a_int.h"
+
+static void pgraph_null_init(NV2AState *d, Error **errp)
+{
+    PGRAPHState *pg = &d->pgraph;
+
+    pg->null_renderer_state = g_malloc0(sizeof(*pg->null_renderer_state));
+}
+
+static void pgraph_null_finalize(NV2AState *d)
+{
+    PGRAPHState *pg = &d->pgraph;
+    PGRAPHNullState *r = pg->null_renderer_state;
+    if (r) {
+        if (r->blank_texture) {
+            glDeleteTextures(1, &r->blank_texture);
+        }
+        g_free(r);
+        pg->null_renderer_state = NULL;
+    }
+}
 
 static void pgraph_null_sync(NV2AState *d)
 {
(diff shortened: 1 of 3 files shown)


GiladN1:
Is there a way to make the UI FPS match the guest FPS? framepacing feels uneven to me cause Xemu reports much higher FPS than the game's guest FPS, which also makes VRR not engage properly.

abaire:
> Is there a way to make the UI FPS match the guest FPS? framepacing feels uneven to me cause Xemu reports much higher FPS than the game's guest FPS, which also makes VRR not engage properly.

That is unrelated to this PR, you're welcome to ask on Discord or to create or comment on an existing relevant GitHub issue. 

I don't want to have a discussion on an unrelated topic making this thread hard to read for future reviewers and future maintainers but I can say there are compelling reasons to disconnect the GUI and guest framerates, happy to explain/discuss more in a more appropriate channel.

PR #2829, opened 2026-04-21
ui: Force NVIDIA present method to native

abaire:
Setting the NVIDIA driver Vulkan/OpenGL present method to "Prefer layerd on DXGI Swapchain" with vsync enabled within xemu causes framerate to be ~halved. This change overrides the driver setting to force the legacy/native value to be used.

Resolves one identified issue with #2790

PR #2972, opened 2026-08-15
ui: Use DXGI to perform swapping on Windows

### LINKED ISSUE #2790: Performance regression between 0.8.133 and 0.8.134 releases caused by PR #2691. Related to vsync being enabled and influenced by display refresh rate, seems specific to AMD drivers on Windows.
### Bug Description

I have tested several games and there's a clear degradation in performance between 0.8.133 and 0.8.134 releases (traced specifically to #2691). It's immediately noticeable during the booting animation. In 0.8.133 the animation is basically locked at 60 fps where 0.8.134 runs around 30 fps. Here are a few examples but it affects almost all the games I tried. Specially on OpenGL.

Image
Image
Image
Image
Image
Image

### Expected Behavior

It should perform the same as 0.8.133.

### xemu Version

0.8.134

Last working in 0.8.133

### System Information

CPU: AMD Ryzen 9 6900HX with Radeon Graphics        
OS Platform: Windows
OS Version: 25H2
Manufacturer: ATI Technologies Inc.
GPU Model: AMD Radeon RX 6600M
Driver: 4.0.0 Core Profile Context 26.2.2.260218
Shader: 4.60

### Additional Context

I did notice that using RTSS to limit the framerate at 60 fixes the issue. So the performance regression is probably related to the refresh rate (120 Hz in my case).

#### Approximate FPS values against display refresh rate (vsync on)

* MAT2M: Midway Arcade Treasures 2, main menu, on `A.P.B.`
* CB: Conker, bar scene displayed after the intro videos, before entering the game

||60 Hz| 120Hz|
|---|---|---|
|0.8.133|MAT2M: 60<br>CB: 30 |MAT2M: 60<br>CB: 30 |
|0.8.134|MAT2M: 10<br>CB: 6 |MAT2M: 20<br>CB: 12 |

### New Findings
It seems that my initial assumption of the performance regression happening only when the refresh rate was set to 120 Hz or more was flawed. I just tested setting the refresh rate to 60 Hz on Windows and the performance regression was even worse that what I saw when it was at 120 Hz.

[Note from abaire] - I've written up some analysis and attempted solutions here for context: https://gist.github.com/abaire/531dfd73b34f682c2feb5b1626e06853

abaire:
In #2691, xemu was updated to decouple the guest vblank interrupt from the host refresh. This allows the UI thread to run at host FPS while the guest runs at 60 fps, matching HW behavior.

On Window machines using AMD drivers, this led to a major decrease in guest performance when vsync is enabled, apparently due to locking behavior in the driver holding the guest pfifo thread while awaiting vblank. This manifests as reduced perf that gets worse when running at a lower display refresh rate.

This PR bypasses the GL swap entirely, utilizing Windows DXGI to perform swaps and allowing the pfifo operations to proceed without blocking.

https://github.com/abaire/xemu-perf-tests/blob/4115d262c5e80d6b71234f8df65fa78c8ef15e12/src/tests/surface_rendering_tests.cpp#L167 reproduces the issue seen with Conker L&R (slow glo_readpixels) but it is somewhat subtle at the given tuning. Modifying this test to perform more draws per frame will likely produce a more obvious result w/ the pre-PR builds as there are more opportunities for glReadPixels to stall on main thread vblank waits.

Fixes #2790

Note that this does _not_ allow us to revert PR #2829, affected NVIDIA users report that they still have poor performance with this PR with 2829 reverted.

### DIFF
--- a/ui/xemu.c
+++ b/ui/xemu.c
@@ -535,11 +535,27 @@ static void handle_windowevent(SDL_Event *ev)
 
 #ifdef _WIN32
             if (win32_dxgi_present_is_active()) {
-                win32_dxgi_present_resize(ev->window.data1, ev->window.data2);
+                int width;
+                int height;
+                if (!SDL_GetWindowSizeInPixels(scon->real_window, &width,
+                                               &height)) {
+                    fprintf(stderr, "SDL_GetWindowSizeInPixels failed "
+                                    "responding to resize event.\n");
+                } else {
+                    win32_dxgi_present_resize(width, height);
+                }
             }
 #endif
         }
         break;
+#ifdef _WIN32
+    case SDL_EVENT_WINDOW_PIXEL_SIZE_CHANGED:
+        if (win32_dxgi_present_is_active()) {
+            win32_dxgi_present_resize(ev->window.data1, ev->window.data2);
+        }
+        break;
+#endif
+
     case SDL_EVENT_WINDOW_FOCUS_GAINED:
     case SDL_EVENT_WINDOW_MOUSE_ENTER:
         if (!gui_grab && (qemu_input_is_absolute(scon->dcl.con) || absolute_enabled)) {
@@ -883,8 +899,10 @@ static void gl_render_frame(struct xemu_console *scon)
     } else {
         static bool warned = false;
         if (!warned) {
-          fprintf(stderr, "win32_dxgi present failed, falling back to SDL_GL_SwapWindow\n");
-          warned = true;
+            fprintf(stderr,
+                    "win32_dxgi present failed or unavailable, falling back to "
+                    "SDL_GL_SwapWindow\n");
+            warned = true;
         }
         SDL_GL_SwapWindow(scon->real_window);
     }
@@ -907,7 +925,15 @@ static bool event_watch_callback(void *userdata, SDL_Event *event)
     if (event->type == SDL_EVENT_WINDOW_RESIZED) {
 #ifdef _WIN32
         if (win32_dxgi_present_is_active()) {
-            win32_dxgi_present_resize(event->window.data1, event->window.data2);
+            int width;
+            int height;
+            if (!SDL_GetWindowSizeInPixels(scon->real_window, &width,
+                                           &height)) {
+                fprintf(stderr, "SDL_GetWindowSizeInPixels failed "
+                                "responding to resize event.\n");
+            } else {
+                win32_dxgi_present_resize(width, height);
+            }
         }
 #endif
         gl_render_frame(scon);
--- a/ui/xemu.c
+++ b/ui/xemu.c
@@ -60,6 +60,10 @@
 #include <math.h>
 #include <SDL3/SDL.h>
 
+#ifdef _WIN32
+#include "xui/win32-dxgi-present.h"
+#endif
+
 #ifndef DEBUG_XEMU_C
 #define DEBUG_XEMU_C 0
 #endif
@@ -528,6 +532,12 @@ static void handle_windowevent(SDL_Event *ev)
                 g_config.display.window.last_width = ev->window.data1;
                 g_config.display.window.last_height = ev->window.data2;
             }
+
+#ifdef _WIN32
+            if (win32_dxgi_present_is_active()) {
+                win32_dxgi_present_resize(ev->window.data1, ev->window.data2);
+            }
+#endif
         }
         break;
     case SDL_EVENT_WINDOW_FOCUS_GAINED:
@@ -836,6 +846,12 @@ static void gl_render_frame(struct xemu_console *scon)
         xemu_main_loop_unlock();
     }
 
+#ifdef _WIN32
+    if (win32_dxgi_present_is_active()) {
+        win32_dxgi_present_begin_frame();
+    }
+#endif
+
     glClearColor(0, 0, 0, 0);
     glClear(GL_COLOR_BUFFER_BIT);
     xemu_snapshots_set_framebuffer_texture(tex, flip_required);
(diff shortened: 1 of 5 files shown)


ui/xemu.c
```diff
@@ -860,7 +876,21 @@ static void gl_render_frame(struct xemu_console *scon)
...
     nv2a_release_framebuffer_surface();
+
+#ifdef _WIN32
+    if (win32_dxgi_present_is_active()) {
+        win32_dxgi_present_end_frame(g_config.display.window.vsync);
+    } else {
+        static bool warned = false;
+        if (!warned) {
+          fprintf(stderr, "win32_dxgi present failed, falling back to SDL_GL_SwapWindow\n");
+          warned = true;
```
mborgerson: If present fails once should we permanently fall back to SDL_GL_SwapWindow?

ui/xui/win32-dxgi-present.cc
```diff
@@ -0,0 +1,679 @@
...
+    ComPtr<IDXGIFactory2> dxgi_factory;
+    HRESULT res = CreateDXGIFactory1(IID_PPV_ARGS(&dxgi_factory));
+    if (FAILED(res)) {
+        fprintf(stderr,
+                "win32_dxgi_present: CreateDXGIFactory1 failed (hr=0x%08lX)\n",
+                (unsigned long)res);
+        return false;
+    }
+
+    HMONITOR target_monitor = MonitorFromWindow(hwnd, MONITOR_DEFAULTTONEAREST);
```
mborgerson, line 369: What if the xemu window is moved to another monitor? Do we need to reinitialize?

ui/xui/win32-dxgi-present.cc
```diff
@@ -0,0 +1,679 @@
+//
+// xemu User Interface
+//
+// Copyright (C) 2026 Matt Borgerson
```
mborgerson, line 4: I didn't write this code :)

abaire: Should we start a "xemu-project contributors" or something like that? I feel like anytime personal names get put into copyright messages it invites arguments about if/when people should add themselves, etc... 

My not-a-lawyer understanding is that (in the USA) we don't have to have a specific author name to assert copyright, so I'd also be fine leaving it blank.

ui/xui/win32-dxgi-present.cc
abaire: I believe player1 on Discord tested connecting and disconnecting an external display live and it worked for them. 

I think as long as both monitors are served by the same adapter this will be fine (I added this code initially when debugging failure in docked mode, thinking that maybe player1 had an egpu and we were grabbing the wrong adapter). If we want to support changing adapters on the fly we would have to catch that and pipe through a reinit.

At the moment I'm of the opinion that this addition is already somewhat complicated and would prefer to have it soak for a release or two to see what real-world issues pop up. If it turns out that on-the-fly adapter switching is a real use case I'd imagine one of the DXGI calls will fail and we can (hopefully) use the resulting log to inform exactly what error needs to be handled.

ui/xui/win32-dxgi-present.cc
```diff
@@ -0,0 +1,679 @@
...
+}
+
+void Win32DxgiPresenter::EndFrame(bool vsync)
+{
+    if (!m_active.load(std::memory_order_acquire) || !m_wgl_object ||
+        !m_render_fbo || !m_interop_fbo) {
+        return;
+    }
+
+    m_wgl_dx_lock_objects_nv(m_wgl_device, 1, &m_wgl_object);
```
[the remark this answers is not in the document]
abaire: Good call, done.

ui/xui/win32-dxgi-present.cc
```diff
@@ -0,0 +1,679 @@
...
+    ComPtr<ID3D11Texture2D> back_buffer;
+    HRESULT res = m_swap_chain->GetBuffer(0, IID_PPV_ARGS(&back_buffer));
+    if (SUCCEEDED(res)) {
+        m_d3d_context->CopyResource(back_buffer.Get(), m_shared_texture.Get());
+    }
+
+    UINT sync_interval = vsync ? 1 : 0;
+    UINT present_flags =
+        (!vsync && m_allow_tearing) ? DXGI_PRESENT_ALLOW_TEARING : 0;
+    m_swap_chain->Present(sync_interval, present_flags);
```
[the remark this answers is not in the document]
abaire: Done.

ui/xemu.c
```diff
@@ -528,6 +532,12 @@ static void handle_windowevent(SDL_Event *ev)
                 g_config.display.window.last_width = ev->window.data1;
                 g_config.display.window.last_height = ev->window.data2;
             }
+
+#ifdef _WIN32
+            if (win32_dxgi_present_is_active()) {
+                win32_dxgi_present_resize(ev->window.data1, ev->window.data2);
```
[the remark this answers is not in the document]
abaire: Done.

ui/xemu.c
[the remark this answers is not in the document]
abaire: Fixed.

ui/xemu.c
```diff
@@ -874,8 +904,14 @@ static bool event_watch_callback(void *userdata, SDL_Event *event)
 {
     struct xemu_console *scon = (struct xemu_console *)userdata;
 
-    if (event->type == SDL_EVENT_WINDOW_EXPOSED ||
-        event->type == SDL_EVENT_WINDOW_RESIZED) {
+    if (event->type == SDL_EVENT_WINDOW_RESIZED) {
+#ifdef _WIN32
+        if (win32_dxgi_present_is_active()) {
+            win32_dxgi_present_resize(event->window.data1, event->window.data2);
```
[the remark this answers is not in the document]
abaire: Fixed

ui/xemu.c
abaire: In practice I think that's what it does now (including the missing error check changes suggested by copilot), if `Present` fails it'll emit an error message and `Cleanup` the DXGI path.

Based on the docs for `Present` it looks like the only time it would return an error (besides being unsupported which almost certainly would've been caught by one of the setup calls failing earlier) would be due to a disconnected monitor. I'm not sure if it's worth the complexity to pipe this particular failure through to attempt to trigger a reinit though (nor am I sure that a reinit wouldn't just deathloop, e.g. if the monitor was disconnected via a KVM and then reappeared some time later).

mborgerson:
Thanks for investigating this and figuring out a solution

mirh:
Just throwing it out there for the records, but SDL [would love](https://github.com/libsdl-org/SDL/issues/15036) if this could be implemented internally.

PR #2974, opened 2026-08-17
nv2a: Improve code quality by patching edge cases.

Reedzit:
This PR addresses the bot feedback given in #2932 stemming from issue #2597 
Specifically, the changes include the following: 
- Add null-termination validation for cached strings to prevent overruns when calling strcmp()
- Add bounds checking for cached string lens
- Add fallback values for glGetString() in the case that it returns NULL
- Replace static variables with locally scoped variables in `shader_write_to_disk()` to prevent potential data races and undefined behavior. Since `strlen` is comparatively small, this change should not introduce any significant overhead.

I rebuilt and tested out the changes with no noticeable issues.

### DIFF
--- a/hw/xbox/nv2a/pgraph/gl/shaders.c
+++ b/hw/xbox/nv2a/pgraph/gl/shaders.c
@@ -421,27 +421,43 @@ static void shader_load_from_disk(PGRAPHState *pg, uint64_t hash)
 
     READ_OR_ERR(&cached_xemu_version_len, sizeof(cached_xemu_version_len));
 
-    cached_xemu_version = g_malloc(cached_xemu_version_len +1);
+    if (cached_xemu_version_len == 0 || cached_xemu_version_len > 1024) {
+        fclose(shader_file);
+        goto error;
+    }
+
+    cached_xemu_version = g_malloc(cached_xemu_version_len + 1);
     READ_OR_ERR(cached_xemu_version, cached_xemu_version_len);
-    if (strcmp(cached_xemu_version, xemu_version) != 0) {
+    if (cached_xemu_version[cached_xemu_version_len - 1] != '\0' || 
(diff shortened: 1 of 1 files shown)


hw/xbox/nv2a/pgraph/gl/shaders.c
```diff
@@ -421,27 +421,43 @@ static void shader_load_from_disk(PGRAPHState *pg, uint64_t hash)
 
     READ_OR_ERR(&cached_xemu_version_len, sizeof(cached_xemu_version_len));
 
-    cached_xemu_version = g_malloc(cached_xemu_version_len +1);
+    if (cached_xemu_version_len == 0 || cached_xemu_version_len > 1024) {
```
abaire, line 424: There's a lot of duplication in the error checking, I wonder if we should roll this into a macro as well? We could combine most of these into a `READ_PASCAL_STRING` or similar macro that handles the size prefix + string read w/ the null terminator sanity check, then we're just doing the `strcmp` & bail (or even that could be included).

hw/xbox/nv2a/pgraph/gl/shaders.c
```diff
@@ -617,20 +639,11 @@ static void *shader_write_to_disk(void *arg)
     char *shader_bin = shader_get_bin_directory(binding->node.hash);
     char *shader_path = shader_get_binary_path(shader_bin, binding->node.hash);
 
-    static uint64_t gl_vendor_len;
-    if (gl_vendor_len == 0) {
-        gl_vendor_len = (uint64_t) (strlen(shader_gl_vendor) + 1);
-    }
+    uint64_t gl_vendor_len = (uint64_t)(strlen(shader_gl_vendor) + 1);
```
abaire, line 642: Why don't we keep these as static and initialize them the one and only time the strings are set, around line 575 above?

PR #2977, opened 2026-08-19
nv2a/vk: fix infinite loop selecting compute workgroup size

jessedye:
The workgroup-size selection loop in `surface-compute.c` starts at 1024 and
halves until it divides the output unit count:

```c
while (group_size > 1) {
    if (group_size > r->device_props.limits.maxComputeWorkGroupSize[0]) {
        continue;   /* skips the only decrement */
    }
    ...
    group_size /= 2;
}
```

When the device reports `maxComputeWorkGroupSize[0]` below the starting
value, the `continue` bypasses the only `group_size /= 2`, so the loop never
terminates and the emulator hangs at surface conversion. This is reachable on
real hardware: V3DV (Raspberry Pi 5) reports 256 for both
`maxComputeWorkGroupSize[0]` and `maxComputeWorkGroupInvocations`.

Clamp the starting size to both device limits before searching. On desktop
drivers that report 1024 or more, the clamp is a no-op and behaviour is
unchanged; on smaller GPUs the loop terminates and the compute path works.

Found and verified on a Raspberry Pi 5 (V3DV / Mesa Vulkan 1.2), where Halo 2
reaches gameplay with this fix and hangs at first surface download without it.

### DIFF
--- a/hw/xbox/nv2a/pgraph/vk/renderer.h
+++ b/hw/xbox/nv2a/pgraph/vk/renderer.h
@@ -331,6 +331,7 @@ typedef struct PGRAPHVkState {
     VkPhysicalDevice physical_device;
     VkPhysicalDeviceFeatures enabled_physical_device_features;
     VkPhysicalDeviceProperties device_props;
+    uint32_t max_1d_compute_group_size;
     VkDevice device;
     VmaAllocator allocator;
     uint32_t allocator_last_submit_index;
--- a/hw/xbox/nv2a/pgraph/vk/surface-compute.c
+++ b/hw/xbox/nv2a/pgraph/vk/surface-compute.c
@@ -334,19 +334,13 @@ void pgraph_vk_compute_finish_complete(PGRAPHVkState *r)
 
 static int get_workgroup_size_for_output_units(PGRAPHVkState *r, int output_units)
 {
-    int group_size = 1024;
-
     // FIXME: Smarter workgroup size calculation could factor in multiple
     //        submissions. For now we will just pick the highest number that
     //        evenly divides output_units.
 
-    while (group_size > 1) {
-        if (group_size > r->device_props.limits.maxComputeWorkGroupSize[0]) {
-            continue;
-        }
-        if (output_units % group_size == 0) {
-            break;
-        }
+    int group_size = 1024;
+    while (group_size > 1 && (group_size > r->max_1d_compute_group_size ||
+                              output_units % group_size != 0)) {
         group_size /= 2;
     }
 
--- a/hw/xbox/nv2a/pgraph/vk/instance.c
+++ b/hw/xbox/nv2a/pgraph/vk/instance.c
@@ -419,6 +419,9 @@ static bool select_physical_device(PGRAPHState *pg, Error **errp)
     }
 
     vkGetPhysicalDeviceProperties(r->physical_device, &r->device_props);
+    r->max_1d_compute_group_size =
+        MIN(r->device_props.limits.maxComputeWorkGroupSize[0],
+            r->device_props.limits.maxComputeWorkGroupInvocations);
     xemu_settings_set_string(&g_config.display.vulkan.preferred_physical_device,
                              r->device_props.deviceName);
     r->vk_api_version = MIN(r->vk_api_version, r->device_props.apiVersion);


hw/xbox/nv2a/pgraph/vk/surface-compute.c
```diff
@@ -334,19 +334,23 @@ void pgraph_vk_compute_finish_complete(PGRAPHVkState *r)
...
     //        evenly divides output_units.
 
-    while (group_size > 1) {
-        if (group_size > r->device_props.limits.maxComputeWorkGroupSize[0]) {
-            continue;
-        }
-        if (output_units % group_size == 0) {
-            break;
-        }
+    // Clamp to what the device actually supports before searching. Desktop
```
abaire: Should we calculate this once when `device_props` is initialized and store it?

hw/xbox/nv2a/pgraph/vk/surface-compute.c
```diff
@@ -334,19 +334,23 @@ void pgraph_vk_compute_finish_complete(PGRAPHVkState *r)
...
+    // GPUs report 1024 here, but smaller parts report less (V3D reports 256),
+    // and dispatching a workgroup larger than the limit is invalid.
+    int max_group_size = r->device_props.limits.maxComputeWorkGroupSize[0];
+    if (r->device_props.limits.maxComputeWorkGroupInvocations < max_group_size) {
+        max_group_size = r->device_props.limits.maxComputeWorkGroupInvocations;
+    }
+
+    int group_size = 1024;
+    while (group_size > 1 && group_size > max_group_size) {
```
abaire: Combine the two loops with an `||` ?

jessedye:
Both applied: the device limit is now computed once when device properties are read, and the two loops are combined. Thanks for the review.

hw/xbox/nv2a/pgraph/vk/renderer.h
```diff
@@ -331,6 +331,9 @@ typedef struct PGRAPHVkState {
     VkPhysicalDevice physical_device;
     VkPhysicalDeviceFeatures enabled_physical_device_features;
     VkPhysicalDeviceProperties device_props;
+    /* Largest 1-D compute workgroup the device accepts: the smaller of
```
abaire: Generally xemu style is to omit comments unless they're explaining something especially tricky. In this case, the intent of the var could be fully expressed with a name tweak. 

Maybe `max_1d_compute_group_size`?

hw/xbox/nv2a/pgraph/vk/surface-compute.c
```diff
@@ -334,19 +334,16 @@ void pgraph_vk_compute_finish_complete(PGRAPHVkState *r)
...
     //        evenly divides output_units.
 
-    while (group_size > 1) {
-        if (group_size > r->device_props.limits.maxComputeWorkGroupSize[0]) {
-            continue;
-        }
-        if (output_units % group_size == 0) {
-            break;
-        }
+    // Clamp to what the device actually supports before searching. Desktop
```
abaire: This is an example of why xemu style generally leaves out comments - this comment no longer makes sense, since the loop is doing both the clamping and the searching, so it's now confusing as it's unclear if there's a bug in the code or an issue with the comment. 

I don't think this comment adds value in general, it's obvious that the code is clamping to the previously calculated max supported size and is also looking for the highest appropriate group size based on output_units.

jessedye:
Applied: renamed to `max_1d_compute_group_size` and dropped both comments.

jessedye:

jessedye:
Apologies, I made these changes but never replied to say so. All four are in the current branch:

**Compute once and store it** — `max_1d_compute_group_size` is now read from the device properties once during instance setup (`instance.c:422`) and kept on the renderer state, rather than being queried per call.

**Combine the two loops** — done, they are one loop now:

```c
int group_size = 1024;
while (group_size > 1 && (group_size > r->max_1d_compute_group_size ||
                          output_units % group_size != 0)) {
    group_size /= 2;
}
```

**The comment that no longer made sense, and the one on the struct field** — both dropped. The field is just `uint32_t max_1d_compute_group_size;` with no comment, and what remains in the function is the pre-existing FIXME about smarter sizing, which is still accurate.

Point taken on the comment style generally — I have been keeping new comments out of these since.

Triticum0:
Quick test with Terminator 3: Rise of the Machines and morrowind and seem the build runs fine on my hardware using the vk renderer.

Issue #2918, opened 2026-07-03
Bypass OpenGL 4 check for Raspberry Pi 5


varkanoid:
### Feature Request

Xemu actually works on a Raspberry Pi 5. You can run the Flatpak, xemu-0.8.136-aarch64.AppImage or you can build from source. However due to the OpenGL 4 check that is shown and quits the xemu due to Pi 5 not using OpenGL 4 I have to force it to use software rendering to bypass this check. 

LIBGL_ALWAYS_SOFTWARE=true xemu

This allows the xemu to run on a Pi 5.

Ideally would like it to use Vulcan Hardware Acceleration but I have found no way to do this. If I run it with Software Rendering its really slow as you can imagine. If I then try and change it to Vulcan in the Display menu it just reverts back to OpenGL.

Is there any way you could produce a Pi 5 version that throws up the OpenGL 4 box but as a warning and does not close xemu but allow it to proceed. Then maybe Vulcan can be selectable and stay selectable in the Display settings ? Or in general can the OpenGL 4 check be just put as a warning but allows you to proceed further ?

abaire:
I believe xemu currently uses GL for the UI even when the Vulkan backend is enabled.

You can try patching the check out here https://github.com/xemu-project/xemu/blob/295c13e972338879ff74a3e9f0dd57b35e148c60/ui/xemu.c#L1046

but it's unlikely that xemu will run properly on an older version of GL. If you can get it working please post a comment here or in the Discord and we can discuss whether providing some advanced/hidden option to bypass the check would make sense.

varkanoid:
Thanks been on holiday but will give it a go when I have chance.

varkanoid:
Hi

unfortunately it fails with these errors.

GL_VENDOR: (null)
GL_RENDERER: (null)
GL_VERSION: (null)
GL_SHADING_LANGUAGE_VERSION: (null)
glo_context_create: Failed to creatre GL context.

Although I've just realised thats not the actual issue. As you say OpenGL is for the UI and the UI runs okay if I use software rendering. If I set the backend to OpenGL games run but too slow. If I set the backend to Vulkan xemu crashes out with a Vulkan error - this is probably the error I need to try and solve but I think I need to accept that the Pi5 wont run with Vulkan enabled. Trixie OS now comes with MESA (Vulkan) 26.2.0 but that still fails.

I need to try and solve this error

[...]

PR #2979, opened 2026-08-19
ui/gl: survive limited desktop OpenGL

jessedye:
Two small robustness fixes for hosts with limited desktop OpenGL:

- The GPU property probe (geometry shader winding detection) aborted startup
  when its GLSL 4.00 shaders failed to compile. Fall back to defaults and
  continue; drivers capped below GL 4 (e.g. Mesa V3D at 3.1) can otherwise
  never reach the renderer selection.
- Shader program link status was unchecked, so a failed link surfaced later
  as a crash with no diagnostic. Check it and report the info log.

No behaviour change on drivers where the probe and links succeed.

### DIFF
--- a/hw/xbox/nv2a/pgraph/gl/gpuprops.c
+++ b/hw/xbox/nv2a/pgraph/gl/gpuprops.c
@@ -109,7 +109,8 @@ static GLuint compile_shader(GLenum type, const char *source)
         log[sizeof(log) - 1] = '\0';
         fprintf(stderr, "GL shader type %d compilation failed: %s\n", type,
                 log);
-        assert(!"GL shader compilation failed");
+        glDeleteShader(shader);
+        return 0;
     }
 
     return shader;
@@ -121,12 +122,22 @@ static GLuint create_program(const char *vert_source, const char *geom_source,
     GLuint vert_shader = compile_shader(GL_VERTEX_SHADER, vert_source);
     GLuint geom_shader = compile_shader(GL_GEOMETRY_SHADER, geom_source);
     GLuint frag_shader = compile_shader(GL_FRAGMENT_SHADER, frag_source);
+    if (!vert_shader || !geom_shader || !frag_shader) {
+        glDeleteShader(vert_shader);
+        glDeleteShader(geom_shader);
+        glDeleteShader(frag_shader);
+        return 0;
+    }
 
     GLuint shader_prog = glCreateProgram();
     glAttachShader(shader_prog, vert_shader);
     glAttachShader(shader_prog, geom_shader);
     glAttachShader(shader_prog, frag_shader);
     glLinkProgram(shader_prog);
+
+    glDeleteShader(vert_shader);
+    glDeleteShader(geom_shader);
+    glDeleteShader(frag_shader);
 
     GLint success;
     glGetProgramiv(shader_prog, GL_LINK_STATUS, &success);
@@ -135,12 +146,9 @@ static GLuint create_program(const char *vert_source, const char *geom_source,
         glGetProgramInfoLog(shader_prog, sizeof(log), NULL, log);
         log[sizeof(log) - 1] = '\0';
         fprintf(stderr, "GL shader linking failed: %s\n", log);
-        assert(!"GL shader linking failed");
+        glDeleteProgram(shader_prog);
+        return 0;
     }
-
-    glDeleteShader(vert_shader);
-    glDeleteShader(geom_shader);
-    glDeleteShader(frag_shader);
 
     return shader_prog;
 }
@@ -181,7 +189,12 @@ static uint8_t *render_geom_shader_triangles(int width, int height)
 
     GLuint shader_prog = create_program(
         vertex_shader_source, geometry_shader_source, fragment_shader_source);
-    assert(shader_prog != 0);
+    if (shader_prog == 0) {
+        glBindFramebuffer(GL_FRAMEBUFFER, 0);
+        glDeleteRenderbuffers(1, &rbo);
+        glDeleteFramebuffers(1, &fbo);
+        return NULL;
+    }
 
     glUseProgram(shader_prog);
     check_gl_error("glUseProgram");
@@ -339,6 +352,12 @@ void pgraph_gl_determine_gpu_properties(void)
     const int height = 480;
 
     uint8_t *pixels = render_geom_shader_triangles(width, height);
+    if (pixels == NULL) {
+        fprintf(stderr,
+                "GL geometry shader probe unavailable; using defaults\n");
+        return;
+    }
+
     determine_triangle_winding_order(pixels, width, height,
                                      &pgraph_gl_gpu_properties);
     g_free(pixels);
--- a/hw/xbox/nv2a/pgraph/gl/gpuprops.c
+++ b/hw/xbox/nv2a/pgraph/gl/gpuprops.c
@@ -353,10 +353,11 @@ void pgraph_gl_determine_gpu_properties(void)
 
     uint8_t *pixels = render_geom_shader_triangles(width, height);
     if (pixels == NULL) {
-        fprintf(stderr,
-                "GL geometry shader probe unavailable; using defaults\n");
+        fprintf(stderr, "GL geometry shader probe failed\n");
+        pgraph_gl_gpu_properties.valid = false;
         return;
     }
+    pgraph_gl_gpu_properties.valid = true;
 
     determine_triangle_winding_order(pixels, width, height,
                                      &pgraph_gl_gpu_properties);
--- a/hw/xbox/nv2a/pgraph/gl/gpuprops.c
+++ b/hw/xbox/nv2a/pgraph/gl/gpuprops.c
@@ -25,7 +25,7 @@
 static GPUProperties pgraph_gl_gpu_properties;
 
 static const char *vertex_shader_source =
-    "#version 400\n"
+    PROBE_GLSL_VERSION_DIRECTIVE
     "out vec3 v_fragColor;\n"
     "\n"
     "vec2 positions[11] = vec2[](\n"
@@ -62,7 +62,7 @@ static const char *vertex_shader_source =
     "}\n";
 
 static const char *geometry_shader_source =
-    "#version 400\n"
+    PROBE_GLSL_VERSION_DIRECTIVE
     "layout(triangles) in;\n"
     "layout(triangle_strip, max_vertices = 3) out;\n"
     "out vec3 fragColor;\n"
@@ -87,7 +87,7 @@ static const char *geometry_shader_source =
     "}\n";
 
 static const char *fragment_shader_source =
-    "#version 400\n"
+    PROBE_GLSL_VERSION_DIRECTIVE
     "out vec4 outColor;\n"
     "in vec3 fragColor;\n"
     "\n"
--- a/hw/xbox/nv2a/pgraph/gl/renderer.c
+++ b/hw/xbox/nv2a/pgraph/gl/renderer.c
@@ -46,6 +46,13 @@ static void pgraph_gl_init(NV2AState *d, Error **errp)
 {
     PGRAPHState *pg = &d->pgraph;
 
+    if (!pgraph_gl_get_gpu_properties()->valid) {
+        error_setg(errp,
+                   "OpenGL renderer unavailable: GPU capability probe failed "
+                   "(OpenGL 4.0 support is required)");
+        return;
+    }
+
     pg->gl_renderer_state = g_malloc0(sizeof(*pg->gl_renderer_state));
     PGRAPHGLState *r = pg->gl_renderer_state;
 
--- a/hw/xbox/nv2a/pgraph/gl/renderer.c
+++ b/hw/xbox/nv2a/pgraph/gl/renderer.c
@@ -49,7 +49,7 @@ static void pgraph_gl_init(NV2AState *d, Error **errp)
     if (!pgraph_gl_get_gpu_properties()->valid) {
         error_setg(errp,
                    "OpenGL renderer unavailable: GPU capability probe failed "
-                   "(OpenGL 4.0 support is required)");
+                   "(OpenGL " PROBE_GL_VERSION_STRING " support is required)");
         return;
     }
 
--- a/ui/xui/gl-helpers.cc
+++ b/ui/xui/gl-helpers.cc
@@ -240,8 +240,8 @@ uniform sampler2D tex;
 uniform uint palette[256];
 float gamma_ch(int ch, float col)
 {
-    uint packed = palette[uint(col * 255.0)];
-    return float((packed >> uint(ch * 8)) & 0xFFu) / 255.0;
+    uint entry = palette[uint(col * 255.0)];
+    return float((entry >> uint(ch * 8)) & 0xFFu) / 255.0;
 }
 
 vec4 gamma(vec4 col)
--- a/ui/xui/gl-helpers.cc
+++ b/ui/xui/gl-helpers.cc
@@ -293,6 +293,19 @@ void main() {
     glAttachShader(s->prog, frag);
     glBindFragDataLocation(s->prog, 0, "out_Color");
     glLinkProgram(s->prog);
+
+    GLint link_status;
+    glGetProgramiv(s->prog, GL_LINK_STATUS, &link_status);
+    if (link_status != GL_TRUE) {
+        char link_err[1024];
+        glGetProgramInfoLog(s->prog, sizeof(link_err), NULL, link_err);
+        fprintf(stderr,
+                "NewDecalShader: shader program link failed: %s\n"
+                "--- vertex source ---\n%s\n"
+                "--- fragment source ---\n%s\n",
+                link_err, vert_src, frag_src ? frag_src : "(null)");
+        assert(!"NewDecalShader: shader program link failed");
+    }
     glUseProgram(s->prog);
 
     // Flag shaders for deletion when program is deleted
--- a/ui/xui/gl-helpers.cc
+++ b/ui/xui/gl-helpers.cc
@@ -207,7 +207,7 @@ DecalShader *NewDecalShader(enum ShaderType type)
     s->time = 0;
 
     const char *vert_src = R"(
-#version 150 core
+#version 140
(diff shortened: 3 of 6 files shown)


jessedye:

hw/xbox/nv2a/pgraph/gl/gpuprops.c
```diff
@@ -121,6 +122,9 @@ static GLuint create_program(const char *vert_source, const char *geom_source,
     GLuint vert_shader = compile_shader(GL_VERTEX_SHADER, vert_source);
     GLuint geom_shader = compile_shader(GL_GEOMETRY_SHADER, geom_source);
     GLuint frag_shader = compile_shader(GL_FRAGMENT_SHADER, frag_source);
+    if (!vert_shader || !geom_shader || !frag_shader) {
+        return 0;
```
abaire, line 129: Looks like it's now possible to exit this function without cleaning up the shaders when returning 0 to indicate an error. Please fix.

ui/xui/gl-helpers.cc
```diff
@@ -293,6 +293,15 @@ void main() {
...
     glBindFragDataLocation(s->prog, 0, "out_Color");
     glLinkProgram(s->prog);
+
+    GLint link_status;
+    glGetProgramiv(s->prog, GL_LINK_STATUS, &link_status);
+    if (link_status != GL_TRUE) {
+        char link_err[1024];
+        glGetProgramInfoLog(s->prog, sizeof(link_err), NULL, link_err);
+        fprintf(stderr, "Shader program link failed: %s\n", link_err);
+        assert(0);
```
abaire: I think we're trying to move away from "assert(false)" behavior since it generates opaque output from a user perspective. Let's `assert(!"NewDecalShader: link failed");` or something along those lines instead.

ui/xui/gl-helpers.cc
```diff
@@ -293,6 +293,15 @@ void main() {
     glAttachShader(s->prog, frag);
     glBindFragDataLocation(s->prog, 0, "out_Color");
     glLinkProgram(s->prog);
+
+    GLint link_status;
+    glGetProgramiv(s->prog, GL_LINK_STATUS, &link_status);
+    if (link_status != GL_TRUE) {
+        char link_err[1024];
+        glGetProgramInfoLog(s->prog, sizeof(link_err), NULL, link_err);
+        fprintf(stderr, "Shader program link failed: %s\n", link_err);
```
abaire: There may be other places where we should check for link errors, let's clarify that this is specifically in `NewDecalShader` in the error message.

I'd argue that it may also be worthwhile dumping the raw `vert_src` and `frag_src` as well so it's clear exactly what was being compiled. We're already in about-to-die mode so, in my opinion, the more info we can give devs to support users the better.

hw/xbox/nv2a/pgraph/gl/gpuprops.c
```diff
@@ -339,6 +349,12 @@ void pgraph_gl_determine_gpu_properties(void)
     const int height = 480;
 
     uint8_t *pixels = render_geom_shader_triangles(width, height);
+    if (pixels == NULL) {
+        fprintf(stderr,
+                "GL geometry shader probe unavailable; using defaults\n");
```
abaire: Are we sure that the default values will actually produce correct output on devices that fall back here? I wonder if we should instead use this as an indication that the GL renderer should be made unavailable, maybe running an additional test to make sure the GL usage (currently) required by the GUI will still operate correctly.

In general, I'm worried that silently supporting fallbacks on low-spec devices is going to lead to more users demanding help or reporting issues that exist specifically because they're running HW that xemu has not historically supported.

jessedye:
All three code comments addressed in the force-push (leak on the early bail, named assert, NewDecalShader named in the message plus both sources dumped).

On the fallback question — I looked at who can actually reach it:

- The only drivers that fail the probe's compile are the ones that cannot
  compile the nv2a GL renderer's `#version 400` shaders either, so the GL
  renderer was never viable there. Before this change they crashed at the
  probe, which is *before* renderer selection — a user with
  `renderer = VULKAN` configured could never reach it.
- Agreed that a GL4-capable device that fails the probe for some other
  reason would silently get possibly-wrong winding defaults, and that's
  worth closing off.

The shape you suggest already has machinery in-tree: `pgraph_gl_init` takes
an `Error **`, `attempt_renderer_init` surfaces it to the user, and
`init_renderer` retries the default renderer. Probe failure could set a flag
that `pgraph_gl_init` turns into `error_setg(...)` — GL becomes unavailable
loudly instead of degraded silently.

One catch found while checking that path: `get_default_renderer()` prefers
OPENGL, so on a config with `renderer = OPENGL` a declining GL init falls
through to the fatal "cannot initialize renderer" exit rather than trying
Vulkan. Making decline-on-probe-failure safe needs that fallback to consider
the Vulkan renderer. Happy to do both in this PR or split the fallback-order
change out — your call.

hw/xbox/nv2a/pgraph/gl/gpuprops.c
```diff
@@ -135,7 +148,8 @@ static GLuint create_program(const char *vert_source, const char *geom_source,
         glGetProgramInfoLog(shader_prog, sizeof(log), NULL, log);
         log[sizeof(log) - 1] = '\0';
         fprintf(stderr, "GL shader linking failed: %s\n", log);
-        assert(!"GL shader linking failed");
+        glDeleteProgram(shader_prog);
```
abaire, line 149: This still bails without deleting shaders, doesn't it?

`If a program object to be deleted has shader objects attached to it, those shader objects will be automatically detached but not deleted unless they have already been flagged for deletion by a previous call to [glDeleteShader](https://registry.khronos.org/OpenGL-Refpages/gl4/html/glDeleteShader.xhtml).`

Given that we flag them all for deletion just below, perhaps we can flag simply them before checking the link status?

abaire (commented):
Personally I think we should fix all of the fallback behavior in a single PR, I'd argue that we're not really "surviving limited desktop OpenGL" if we're really just allowing incorrect behavior or a crash slightly later down the line.

I'll defer to @mborgerson though. You can either wait for Matt's input or go ahead with the larger change if you're happy potentially dropping it or splitting to a separate PR.

jessedye:
Went ahead with the full fallback fix in this PR, per the discussion — third
commit:

- The probe records whether it succeeded, and `pgraph_gl_init` now refuses
  to initialize with a user-visible reason when it did not. No more guessed
  winding defaults, and no deferred crash at the first `#version 400`
  compile: on a device that fails the probe, the GL renderer is simply not
  offered.
- That decline used to be fatal when `renderer = OPENGL` was configured,
  because the fallback only retried the default renderer — the GL renderer
  itself. This resolves the `// FIXME: Try others` in `init_renderer`: it
  now walks every registered renderer (OPENGL, VULKAN, NULL), so such a
  device lands on Vulkan with a "Switched to Vulkan renderer" notification.

Behaviour on drivers where the probe succeeds is unchanged. The three
commits are independent if @mborgerson prefers the fallback rework split
out — the first two stand alone.

The first two commits are what has been running on the motivating hardware
(Raspberry Pi 5, Mesa V3D GL 3.1). I'm putting the new decline-and-fallback
path through a build and run on that device now and will confirm here.

jessedye:
Done — shaders are now flagged for deletion right after attach, so both the link-failure bail and the success path release them through the same mechanism.

hw/xbox/nv2a/pgraph/gl/gpuprops.c
jessedye: Fixed — each compiled shader is deleted before the early return.

hw/xbox/nv2a/pgraph/gl/gpuprops.c
jessedye: Done — all three shaders are flagged with glDeleteShader right after glLinkProgram, before the status check, so both exit paths release them.

ui/xui/gl-helpers.cc
jessedye: Done: `assert(!"NewDecalShader: shader program link failed")`.

ui/xui/gl-helpers.cc
jessedye: Done — the message names NewDecalShader and dumps both shader sources.

hw/xbox/nv2a/pgraph/gl/gpuprops.c
jessedye: Addressed in the third commit: probe failure now makes the GL renderer decline init with a visible reason, and renderer fallback tries every registered renderer, so such devices land on Vulkan instead of running with guessed defaults.

jessedye:
Simplified the early bail to three unconditional `glDeleteShader` calls — deleting shader 0 is silently ignored per spec, confirmed on the motivating hardware (Mesa V3D, GL 3.1): the GLSL 4.00 compile fails as expected and the triple delete leaves `glGetError()` clean.

jessedye:
Hardware confirmation, and one caveat worth flagging.

With `renderer = OPENGL` configured on Mesa V3D (GL 3.1), the fallback works as intended: probe fails, the GL renderer declines, and the Vulkan renderer initializes and runs in its place.

The caveat supports your point about not just deferring the failure. On that same device xemu still aborts *before* the renderer thread starts, in the GUI: `ui/xui/gl-helpers.cc` compiles `#version 150 core` shaders and V3D tops out at GLSL 1.40, so `Shader()` asserts during HUD init. My confirmation above only reaches the fallback because I bypass the GL HUD locally.

So this PR makes the renderer path survive limited GL, but a device like this still needs the GUI's shaders lowered (or a non-GL UI path) before it is genuinely supported. Happy to take that on separately if it is wanted — it is a larger change and out of scope here.

jessedye:
Added the GUI shader lowering here rather than separately, since the two halves only complete each other on a GL 3.1 device.

Verified on Mesa V3D (GL 3.1, GLSL 1.40) with `renderer = OPENGL`:

- before: probe survives and the GL renderer declines, then xemu aborts in HUD init compiling a `#version 400` shader
- after: runs, no abort, no shader compilation failures

The 4.00 shader needed only `bitfieldExtract`, replaced by a shift and a mask; the two 1.50 shaders use nothing newer than 1.30. The logo shader was already 1.40, so this makes the interface internally consistent.

hw/xbox/nv2a/pgraph/gl/renderer.c
```diff
@@ -46,6 +46,13 @@ static void pgraph_gl_init(NV2AState *d, Error **errp)
 {
     PGRAPHState *pg = &d->pgraph;
 
+    if (!pgraph_gl_get_gpu_properties()->valid) {
+        error_setg(errp,
+                   "OpenGL renderer unavailable: GPU capability probe failed "
+                   "(OpenGL 4.0 support is required)");
```
abaire: I'm worried that some future update could require a newer GL (I think we're only held to 4.0 for macOS support; if we get MoltenVK/KosmicKrisp/whatever working we might want to bump the required version. If we do bump it, the chances that we forget about this string feel pretty high.

I think we should either use a constant someplace that controls the required version and this string or make this more vague about the requirement.

ui/xui/gl-helpers.cc
```diff
@@ -235,12 +235,13 @@ void main() {
...
 
     const char *image_gamma_frag_src = R"(
-#version 400 core
+#version 140
 uniform sampler2D tex;
 uniform uint palette[256];
 float gamma_ch(int ch, float col)
 {
-    return float(bitfieldExtract(palette[uint(col * 255.0)], ch*8, 8)) / 255.0;
+    uint packed = palette[uint(col * 255.0)];
```
abaire: Do we know if there's any measurable performance cost of doing the manual shift & mask versus `bitfieldExtract`? Would it be worth dynamically choosing between generating the fast version and using  the fallback only if our previous probe has shown that GL 4.0 is unavailable?

varkanoid:
Hi,

Eric Abair sent me your way from Issue #2918 I have been dabbling getting xemu running on a Pi5. I have been looking into getting emulators working on the Pi5 that previously people said were "impossible". I did a Youtube video a while ago where I got xemu running on my Pi5 8gb with a few games. Poor framerates but at least they run. 

https://youtu.be/WJALdBkPTrQ?si=WKFQLgLCgzbWPGYt

My findings have been if I start xemu with

`LIBGL_ALWAYS_SOFTWARE=1 ./xemu `

it bypasses the OpenGL check and shows the GUI running using software rendering. With the backend set as OpenGL I can then launch games. However OpenGL is really poor so I have been trying to get Vulcan to work but I get this error when switching to it.

`xemu: ../hw/xbox/nv2a/pgraph/vk/buffer.c:32 create_buffer: Assertion 'vk_result == VK_SUCCESS && "vk check failed"' `

This is on Trixie OS with MESA 26.2.0.

Not sure if I can be of any help as I`m not a coder

jessedye:
Thanks for trying it, and the video is a nice datapoint.

Your Vulkan assertion is a separate problem from this PR, and it is a fixable one. `buffer.c:32` is `create_buffer`, and the sizes it is asked for are fixed at build time:

```c
.buffer_size = (1024 * 10) * (1024 * 10) * 8,   // 800 MB, and again for COMPUTE_SRC
.buffer_size = 4096 * 4096 * 4,                 // 64 MB, and again for STAGING_SRC
```

That is roughly 1.7 GB of buffers requested up front. On a discrete GPU with its own VRAM that is fine; on V3D the heap is system memory and shared with everything else, so `vmaCreateBuffer` fails and the `VK_CHECK` assertion fires. It will fail the same way on an 8 GB Pi, because the limit is the reported heap size and `maxStorageBufferRange`, not how much RAM the board has.

I size these from the device instead, which is what got Vulkan running here:

- take the largest reported memory heap, budget a fraction of it for these buffers and leave the rest for surfaces and textures
- clamp storage buffers to `maxStorageBufferRange`, which V3D reports well below the 800 MB request
- keep a floor so tiny heaps still get something workable

That plus the GLSL fixes in this PR is what took a Pi 5 from "aborts at startup" to playable. For reference, on a 2 GB Pi 5 with V3D 7.1.10.2 under Vulkan: Halo 1 around 19-21 fps, Halo 2 around 12, Morrowind 15-17, THPS3 around 45.

Two notes that may save you time:

- `LIBGL_ALWAYS_SOFTWARE=1` gets you past the GUI, but it makes llvmpipe do the GUI rendering, and with the OpenGL backend it will be doing the game too. That is why the framerate is poor. The path worth chasing is Vulkan for the renderer, which needs the buffer sizing above.
- The GLSL changes in this PR matter for you regardless of backend: xemus own GUI shaders were `#version 150`/`400`, and V3D caps at GLSL 1.40, so the overlay could not compile even when the renderer was fine.

Happy to open the buffer sizing as its own PR if maintainers would like it separate from this one - it is self-contained and does not touch behaviour on hosts that can satisfy the original request.

jessedye:
@abaire thanks, both fair.

**Version string** - agreed, and fixed in 7f4d0608bb. The directive and the human-readable version are now defined once in `renderer.h`, beside the `pgraph_gl_get_gpu_properties()` declaration both files already share, so the probe shaders and the error message cannot drift apart. Raising the requirement is now a one-line change.

**bitfieldExtract vs shift and mask** - I have not measured it, and I want to be straight that I cannot: my test hardware is Mesa V3D, which caps at GL 3.1 / GLSL 1.40, so it only ever runs the fallback. Claiming "no cost" without a GL 4.0 device to measure on would be overclaiming.

On mechanism I would expect no difference. `bitfieldExtract` with compile-time-constant offset and width is a canonical pattern that lowers to the same one or two ALU ops as `(x >> c) & 0xFF` on every backend I am aware of; it is a spelling convenience rather than access to a hardware feature. And this is the UI decal gamma shader, a handful of quads a frame for xemus own overlay, not a shader in a games hot path.

On dynamically choosing: doable, but the cost/benefit looks poor to me. `ui/xui/gl-helpers.cc` has no access to `pgraph_gl_get_gpu_properties()` today, so it would need plumbing across that boundary. And the other two shaders in the same file, `vert_src` and `mask_frag_src`, had to drop from `#version 150` to `140` for the same driver and do not use `bitfieldExtract` at all - so a GL 4.0 fast path means maintaining two variants of all three shaders to save at most a couple of ALU ops in overlay code.

If you would still like it, say the word and I will add it - I would rather you decide than have me guess at where the project wants that line. If a maintainer with GL 4.0 hardware wants to measure first, I am happy to be shown wrong.

jessedye:
@varkanoid opened the buffer sizing as its own PR: #2996. It is self-contained and does not touch behaviour on hosts that can satisfy the original allocation, so it should be reviewable independently of this one.

Between the two, this PR gets the GUI and the capability probe working on a GL 3.1 driver, and #2996 gets the Vulkan renderer past startup. You need both on a Pi 5.

jessedye:
Sorry, I left both of these hanging.

**On the version string going stale** — agreed, and it is done. The requirement now comes from a single definition and the message is built from it, so bumping the required version moves the string with it rather than leaving it to be noticed later:

```c
error_setg(errp,
           "OpenGL renderer unavailable: GPU capability probe failed "
           "(OpenGL " PROBE_GL_VERSION_STRING " support is required)");
```

**On `bitfieldExtract` versus the shift and mask** — I did not know, so I measured it rather than guess. A standalone benchmark running each variant as a full-screen 1080p pass, 3000 passes, on a 4090:

| variant | run 1 | run 2 |
|---|---|---|
| `bitfieldExtract` | 50.5 ms | 52.1 ms |
| shift & mask | 49.7 ms | 52.0 ms |

The run-to-run spread on the *same* shader (50.5 → 52.1) is larger than the gap between the two, so there is no measurable cost. In absolute terms the whole pass is about 0.017 ms at 1080p, and both variants are one or two ALU ops behind a texture fetch that dominates them.

So I would rather not add the dynamic selection: it would mean carrying two shader sources and a branch on the probe result to buy something I cannot measure on hardware far faster than the devices this matters for. Happy to be overruled if you would rather keep `bitfieldExtract` where it is available on principle.

One caveat on the measurement: it is a desktop GPU, since a 3.1 device cannot compile the `bitfieldExtract` variant at all to compare against. The ratio should hold — both are trivial next to the texture fetch — but I have not measured it on V3DV and cannot.

PR #2980, opened 2026-08-19
tcg/aarch64: emulate guest TSO with acquire/release accesses

jessedye:
Emulating x86's memory model on aarch64 emits a `dmb` before every guest
load and store. Measured on a Raspberry Pi 5 running an x86 guest, those
fences cost roughly a quarter of end-to-end throughput.

This keeps the ordering and drops the fences: guest loads emit LDAPR and
guest stores STLR (RCpc acquire / release, exactly the semantics x86
accesses need). Since both forms are base-register-only and fault on
unaligned addresses without FEAT_LSE2, wider-than-byte accesses test
alignment at run time and fall back to barrier + plain access; the 16-byte
path and TLB-miss helpers keep explicit barriers.

Measured, same binary, fenced vs `XEMU_TSO=1` (Halo 2, bounded runs, equal
guest progress): median frame time −10%, p99 −9%, 1%-low +11%, stutters
−17%. Also validated on GTA San Andreas (steady 60 guest fps) and
Morrowind.

Opt-in via `XEMU_TSO=1` for now; generated code is unchanged without it.
Happy to rework the gate into a config option if there's interest in
promoting it.

### DIFF
--- a/accel/tcg/translate-all.c
+++ b/accel/tcg/translate-all.c
@@ -344,6 +344,24 @@ TranslationBlock *tb_gen_code(CPUState *cpu, TCGTBCPUState s)
     tcg_ctx->addr_type = target_long_bits() == 32 ? TCG_TYPE_I32 : TCG_TYPE_I64;
     tcg_ctx->guest_mo = cpu->cc->tcg_ops->guest_default_memory_order;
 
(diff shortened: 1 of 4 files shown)


jessedye:

PR #2981, opened 2026-08-19
i386: enable the hard FPU helpers on aarch64 hosts

jessedye:
The hard FPU path (helpers recompiled with native-double arithmetic in place
of floatx80 softfloat) was gated to x86-64 hosts. Its only real x86-64
dependency is that `long double` is the 80-bit format there, so the value
and the register bits coincide for free.

On aarch64 the register layout cannot change — the FXSAVE image is
guest-visible and exactly 512 bytes — so values stay in the 80-bit format
and convert to/from a host double around each operation (a few dozen
integer ops, against the much larger softfloat routines replaced). Same
accepted tradeoffs as the x86-64 hard path: subnormals flush, 80-bit
precision narrows to double.

`perf.hard_fpu` also selects the inline TCG-FP tier, whose opcodes only the
i386 backend implements; on aarch64 that tier stays off and the setting
drives the helper variants alone (the enum in translate.c is split
accordingly, with x86-64 behaviour unchanged).

Measured on a Raspberry Pi 5, where x87 softfloat was ~8.5% of the vCPU
thread: median frame time −5%, p99 −8%, 1%-low +14%, stutters −11%, and
Halo 2 reaches its native 30 fps cap. Also validated on GTA San Andreas
and Morrowind.

### DIFF
--- a/target/i386/helper.h
+++ b/target/i386/helper.h
@@ -101,7 +101,7 @@ DEF_HELPER_FLAGS_3(write_crN, TCG_CALL_NO_RWG, void, env, int, tl)
 #endif /* !CONFIG_USER_ONLY */
 
 /* x86 FPU */
-#if defined(XBOX) && defined(__x86_64__)
+#if defined(XBOX) && (defined(__x86_64__) || defined(__aarch64__))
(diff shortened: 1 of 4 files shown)


jessedye:

PR #2984, opened 2026-08-20
nv2a/gl: support render-to-vertex-buffer (e.g. DisplacementMap)

EqUiNoX-Labs:
Titles that render into a surface and then point a vertex array at that same VRAM (render-to-vertex-buffer) rendered incorrectly:

1. The vertex fetch reads guest VRAM directly, bypassing the CPU-access callback that normally triggers a surface download, so the freshly rendered data was never flushed to VRAM. update_memory_buffer now downloads any dirty surface overlapping the vertex range first (and marks it NV2A-dirty so the geometry buffer re-uploads it).

2. When surface scaling (>1x) is enabled, such a surface was rendered at the scaled resolution and point-sampled back down, corrupting the exact per-texel values the vertex fetch relies on. Track which VRAM addresses get vertex-fetched and render/download those surfaces at native 1x (per-surface scale), while the rest of the frame keeps the configured upscaling. Threaded through surface create/upload/download, the surface-to-texture paths, and the draw viewport/scissor.

Fixes the DisplacementMap sample (torn geometry, and at >1x the streaked mesh plus the black Weight/Index map views) at any render scale.

Before...

image

After...

image

### DIFF
--- a/hw/xbox/nv2a/pgraph/gl/draw.c
+++ b/hw/xbox/nv2a/pgraph/gl/draw.c
@@ -100,8 +100,10 @@ void pgraph_gl_clear_surface(NV2AState *d, uint32_t parameter)
                       scissor_width >= pg->surface_binding_dim.width &&
                       scissor_height >= pg->surface_binding_dim.height;
 
-    pgraph_apply_scaling_factor(pg, &xmin, &ymin);
-    pgraph_apply_scaling_factor(pg, &scissor_width, &scissor_height);
+    unsigned int sf = r->color_binding ? r->color_binding->scale :
+                                         pg->surface_scale_factor;
+    xmin *= sf; ymin *= sf;
+    scissor_width *= sf; scissor_height *= sf;
 
     /* FIXME: Respect window clip?!?! */
     glEnable(GL_SCISSOR_TEST);
@@ -294,9 +296,12 @@ void pgraph_gl_draw_begin(NV2AState *d)
         glDisable(GL_POLYGON_SMOOTH);
     }
 
+    unsigned int sf = r->color_binding ? r->color_binding->scale :
+                                         pg->surface_scale_factor;
+
     unsigned int vp_width = pg->surface_binding_dim.width,
                  vp_height = pg->surface_binding_dim.height;
-    pgraph_apply_scaling_factor(pg, &vp_width, &vp_height);
+    vp_width *= sf; vp_height *= sf;
     glViewport(0, 0, vp_width, vp_height);
 
     /* Surface clip */
@@ -309,8 +314,8 @@ void pgraph_gl_draw_begin(NV2AState *d)
 
     pgraph_apply_anti_aliasing_factor(pg, &xmin, &ymin);
     pgraph_apply_anti_aliasing_factor(pg, &scissor_width, &scissor_height);
-    pgraph_apply_scaling_factor(pg, &xmin, &ymin);
-    pgraph_apply_scaling_factor(pg, &scissor_width, &scissor_height);
+    xmin *= sf; ymin *= sf;
+    scissor_width *= sf; scissor_height *= sf;
 
     glEnable(GL_SCISSOR_TEST);
     glScissor(xmin, ymin, scissor_width, scissor_height);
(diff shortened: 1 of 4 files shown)


hw/xbox/nv2a/pgraph/gl/draw.c
```diff
@@ -100,8 +100,10 @@ void pgraph_gl_clear_surface(NV2AState *d, uint32_t parameter)
                       scissor_width >= pg->surface_binding_dim.width &&
                       scissor_height >= pg->surface_binding_dim.height;
 
-    pgraph_apply_scaling_factor(pg, &xmin, &ymin);
-    pgraph_apply_scaling_factor(pg, &scissor_width, &scissor_height);
+    unsigned int sf = r->color_binding ? r->color_binding->scale :
+                                         pg->surface_scale_factor;
+    xmin *= sf; ymin *= sf;
```
abaire, line 105: nit: I don't think we do multiple statements per line elsewhere in the code. Even within this PR it looks like this is done inconsistently.

As a meta-comment, having `pgraph_apply_scaling_factor` exist but then be the wrong thing to use in certain situations is almost certainly going to lead to maintenance headaches. If there is no way to deterministically choose the correct scaling factor we should get rid of the function and choose the situation-appropriate scale everywhere or modify the helper to provide a hint as to what scale should be used.

That said, if a surface is always responsible for carrying its own scale, why not have a helper that takes a `SurfaceBinding` instead of `PGRAPHState`? Then it'd naturally use the correct value.

hw/xbox/nv2a/pgraph/gl/surface.c
```diff
@@ -632,6 +634,50 @@ void pgraph_gl_surface_download_if_dirty(NV2AState *d,
...
+                         (unsigned int)ARRAY_SIZE(r->vertex_fetched_addrs));
+    for (unsigned int i = 0; i < n; i++) {
+        if (r->vertex_fetched_addrs[i] == addr) {
+            return true;
+        }
+    }
+    return false;
+}
+
+void pgraph_gl_download_surfaces_in_range_if_dirty(NV2AState *d, hwaddr addr,
```
abaire, line 663: This naming mirrors the VK usage but it looks like it always marks the region as "vertex fetched". If there is a need to download dirty surfaces in a region outside of the context of `update_memory_buffer` is it going to be problematic that it is marked as vertex fetched?

hw/xbox/nv2a/pgraph/gl/renderer.h
```diff
@@ -185,6 +186,9 @@ typedef struct PGRAPHGLState {
 
     QTAILQ_HEAD(, SurfaceBinding) surfaces;
     SurfaceBinding *color_binding, *zeta_binding;
+
+    hwaddr vertex_fetched_addrs[16];
```
abaire, line 190: Why 16?

hw/xbox/nv2a/pgraph/gl/surface.c
```diff
@@ -632,6 +634,50 @@ void pgraph_gl_surface_download_if_dirty(NV2AState *d,
     }
 }
 
+static void mark_vram_vertex_fetched(PGRAPHGLState *r, hwaddr addr)
```
abaire, line 637: Looks like there's no way to clear the flag; if a memory region happens to be used for vertices early on and is then always used as a texture/whatever, will it cause issues?

hw/xbox/nv2a/pgraph/gl/surface.c
```diff
@@ -632,6 +634,50 @@ void pgraph_gl_surface_download_if_dirty(NV2AState *d,
...
+                                                   hwaddr size)
+{
+    PGRAPHGLState *r = d->pgraph.gl_renderer_state;
+    SurfaceBinding *surface;
+    QTAILQ_FOREACH(surface, &r->surfaces, entry) {
+        if (check_surface_overlaps_range(surface, addr, size)) {
+            mark_vram_vertex_fetched(r, surface->vram_addr);
+            if (surface->draw_dirty) {
+                surface_download(d, surface, true);
+                memory_region_set_client_dirty(d->vram, surface->vram_addr,
```
abaire, line 673: Is this the only situation in which the dirty bitmap needs to be set after a download?

abaire (commented):
Presumably similar changes would be needed for Vulkan?

abaire:
FYI, https://github.com/abaire/nxdk_pgraph_tests/pull/314 adds tests to the pgraph suite for this case. This PR makes things better, but there still appears to be an issue the first time the tests are run. Note that the test intentionally varies the surface address each run to defeat the persistent flagging in the current version of this PR.

PR #2986, opened 2026-08-20
nv2a/gl: Clear PGRAPH register dirty bitmap when binding shader

### LINKED ISSUE #2983: pgraph `regs_dirty` bitmap is never cleared in GL renderer
### Bug Description

It looks like the [`regs_dirty` bitmap](https://github.com/xemu-project/xemu/blob/298e4edbbd2adfc1f6ad1e0f0a1d672b1eb88529/hw/xbox/nv2a/pgraph/pgraph.h#L241) is used in the GL renderer (e.g., in [pgraph_glsl_check_shader_state_dirty](https://github.com/xemu-project/xemu/blob/298e4edbbd2adfc1f6ad1e0f0a1d672b1eb88529/hw/xbox/nv2a/pgraph/glsl/shaders.c#L56)) but is never actually cleared in the OpenGL renderer.

This means that the fast path in [gl/shaders.c](https://github.com/xemu-project/xemu/blob/298e4edbbd2adfc1f6ad1e0f0a1d672b1eb88529/hw/xbox/nv2a/pgraph/gl/shaders.c#L797) is essentially never taken in practice.

### Expected Behavior

We should presumably be clearing the dirty bits for anything consumed during the bind.

### xemu Version

I think everything since the regs_dirty bitmap was introduced in v0.7.143

### System Information

n/a affects all systems

### Additional Context

_No response_

abaire:
Fixes #2983

### DIFF
--- a/hw/xbox/nv2a/pgraph/gl/shaders.c
+++ b/hw/xbox/nv2a/pgraph/gl/shaders.c
@@ -833,6 +833,9 @@ void pgraph_gl_bind_shaders(PGRAPHState *pg)
         glUseProgram(r->shader_binding->gl_program);
     }
 
+    // FIXME: Clear only the registers that are consumed by this binding.
+    pgraph_clear_dirty_reg_map(pg);
+
     NV2A_GL_DGROUP_END();
 
 update_uniforms:


mborgerson:
Oops! Thanks

PR #2987, opened 2026-08-20
i386/tcg: stub rdpmc to return 0 instead of raising #UD

EqUiNoX-Labs:
The rdpmc helper inherited from upstream QEMU raised EXCP06_ILLOP (illegal instruction) because no performance-monitoring counters are emulated. On real hardware rdpmc never raises #UD, so guests that read the Pentium III PMCs faulted only under emulation -- e.g. the FastCPU XDK sample crashes with 0xC000001D right after reaching its render loop.

Return EAX:EDX = 0 instead, mirroring the policy the WHPX accelerator already uses for unsupported MSR reads. The CPL/CR4.PCE #GP guard is kept. The helper now writes EAX/EDX like rdtsc, so drop noreturn and the TCG_CALL_NO_WG flag from its declaration and let gen_RDPMC fall through instead of ending the block with DISAS_NORETURN.

image

### DIFF
--- a/target/i386/helper.h
+++ b/target/i386/helper.h
@@ -71,7 +71,7 @@ DEF_HELPER_1(rechecking_single_step, void, env)
 DEF_HELPER_1(cpuid, void, env)
 DEF_HELPER_FLAGS_1(rdpid, TCG_CALL_NO_WG, tl, env)
 DEF_HELPER_1(rdtsc, void, env)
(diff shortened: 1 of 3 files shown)


abaire:
This seems like it should be submitted upstream; patching qemu in xemu means maintaining the patch on every update.

Issue #1838, opened 2025-01-02
Vulkan: Performance Issues


ehw:
### Title

This issue affects several games, here are some that I have found that are consistently reproducable.

(Updating when more are discovered from other users too);.

https://xemu.app/titles/54540001/#4x4-Evo-2 TT-001 [54540001]
https://xemu.app/titles/4d530013/#Blinx-The-Time-Sweeper MS-019 [4d530013]
https://xemu.app/titles/4d530065/#Blinx-2-Battle-of-Time-Space-Blinx-2-Masters-of-Time-Space MS-101 [4d530065]
https://xemu.app/titles/53450004/#Crazy-Taxi-3-High-Roller SE-004 [53450004]
https://xemu.app/titles/42550004/#Double-S-T-E-A-L-The-Second-Clash BU-004 [42550004]
https://xemu.app/titles/46490002/#Futurama FI-002 [46490002]
https://xemu.app/titles/43430005/#Group-S-Challenge CC-005 [43430005]
https://xemu.app/titles/56550021/#Jurassic-Park-Operation-Genesis VU-033 [56550021]
https://xemu.app/titles/4947001f/#Mission-Impossible-Operation-Surma IG-031 [4947001f]
https://xemu.app/titles/55530016/#Murakumo-Renegade-Mech-Pursuit US-022 [55530016]
https://xemu.app/titles/5443000d/#Ninja-Gaiden-Black TC-013 [5443000d]
https://xemu.app/titles/4541003c/#SSX-3 EA-060 [4541003c]
https://xemu.app/titles/4c410005/#Star-Wars-Jedi-Starfighter LA-005 [4c410005]
https://xemu.app/titles/4c410002/#Star-Wars-Starfighter-Special-Edition LA-002 [4c410002]
https://xemu.app/titles/49470016/#Superman-The-Man-of-Steel IG-022 [49470016]
https://xemu.app/titles/41540001/#Touge-R AT-001 [41540001]
https://xemu.app/titles/41430014/#XGRA-Extreme-G-Racing-Association AC-020 [41430014]

### Bug Description

There are some games that perform worse while running with the Vulkan backend in comparison to the OpenGL backend, even at native resolution. The performance difference in this case is great enough to cause the games to become unplayable. Despite the low performance, visually the games look identical to the OpenGL backend with no added advantages. This issue seems to occur when there are mostly 3D graphics on the screen, and don't always plague menus or static screens.

Steps to reproduce:
[...]

PR #2988, opened 2026-08-20
nv2a/vk: guard vertex RAM writes on what draws read

jessedye:
The guard before overwriting the vertex RAM mirror tests `uploaded_bitmap`, which records pages uploaded since the last submission. That is not the set that matters.

A page uploaded while an earlier command buffer was being built, and read by a draw recorded into the current one without needing a re-upload, carries no bit. Its contents are then overwritten while a recorded but unexecuted draw still refers to it.

`sync_vertex_ram_buffer` already computes the exact ranges each draw needs, so this marks those instead, and clears the set where `uploaded_bitmap` is cleared — once the work referring to them has been submitted and waited for.

**Measured**, instrumented on Morrowind at 1080p from a snapshot-resumed gameplay scene: the existing guard lets through **3.36 such writes per frame**.

Frame timings are unchanged. The tracked set is larger, but it is the set correctness requires, and no measurable cost came with it:

| | before | after |
|---|---|---|
| fps | 15.4 | 15.2 |
| median frame | 107.8 ms | 109.5 ms |

Both within run-to-run noise on the same scene with the same guest workload.

Tested on a Raspberry Pi 5 with Mesa V3DV only; I have no desktop GPU to hand, so a second pair of eyes on a discrete card would be welcome.

Part of a small series making xemu run well on Raspberry Pi 5 (aarch64 + V3DV): #2977, #2979, #2980, #2981. Each stands alone and they merge in any order.

### DIFF
--- a/hw/xbox/nv2a/pgraph/vk/draw.c
+++ b/hw/xbox/nv2a/pgraph/vk/draw.c
@@ -1233,7 +1233,6 @@ void pgraph_vk_finish(PGRAPHState *pg, FinishReason finish_reason)
         sync_staging_buffer(pg, cmd, BUFFER_VERTEX_INLINE_STAGING,
                                 BUFFER_VERTEX_INLINE);
         sync_staging_buffer(pg, cmd, BUFFER_UNIFORM_STAGING, BUFFER_UNIFORM);
-        bitmap_clear(r->uploaded_bitmap, 0, r->bitmap_size);
         bitmap_clear(r->referenced_bitmap, 0, r->bitmap_size);
         flush_memory_buffer(pg, cmd);
         VK_CHECK(vkEndCommandBuffer(r->aux_command_buffer));
--- a/hw/xbox/nv2a/pgraph/vk/draw.c
+++ b/hw/xbox/nv2a/pgraph/vk/draw.c
@@ -1234,6 +1234,7 @@ void pgraph_vk_finish(PGRAPHState *pg, FinishReason finish_reason)
                                 BUFFER_VERTEX_INLINE);
         sync_staging_buffer(pg, cmd, BUFFER_UNIFORM_STAGING, BUFFER_UNIFORM);
         bitmap_clear(r->uploaded_bitmap, 0, r->bitmap_size);
+        bitmap_clear(r->referenced_bitmap, 0, r->bitmap_size);
         flush_memory_buffer(pg, cmd);
         VK_CHECK(vkEndCommandBuffer(r->aux_command_buffer));
         r->in_aux_command_buffer = false;
@@ -1632,6 +1633,11 @@ static void sync_vertex_ram_buffer(PGRAPHState *pg)
             pgraph_vk_update_vertex_ram_buffer(pg, addr, d->vram_ptr + addr,
                                                size);
         }
+    }
+
+    for (int i = 0; i < num_syncs; i++) {
+        bitmap_set(r->referenced_bitmap, merged[i].addr / TARGET_PAGE_SIZE,
+                   merged[i].size / TARGET_PAGE_SIZE);
     }
 
     r->num_vertex_ram_buffer_syncs = 0;
--- a/hw/xbox/nv2a/pgraph/vk/vertex.c
+++ b/hw/xbox/nv2a/pgraph/vk/vertex.c
@@ -62,7 +62,6 @@ void pgraph_vk_update_vertex_ram_buffer(PGRAPHState *pg, hwaddr offset,
     nv2a_profile_inc_counter(NV2A_PROF_GEOM_BUFFER_UPDATE_1);
     memcpy(r->storage_buffers[BUFFER_VERTEX_RAM].mapped + offset, data, size);
 
-    bitmap_set(r->uploaded_bitmap, start_bit, nbits);
 }
 
 static void update_memory_buffer(NV2AState *d, hwaddr addr, hwaddr size)
--- a/hw/xbox/nv2a/pgraph/vk/vertex.c
+++ b/hw/xbox/nv2a/pgraph/vk/vertex.c
@@ -53,8 +53,7 @@ void pgraph_vk_update_vertex_ram_buffer(PGRAPHState *pg, hwaddr offset,
     size_t end_bit = TARGET_PAGE_ALIGN(offset + size) / TARGET_PAGE_SIZE;
     size_t nbits = end_bit - start_bit;
 
-    if (find_next_bit(r->uploaded_bitmap, start_bit + nbits, start_bit) <
-        end_bit) {
+    if (find_next_bit(r->referenced_bitmap, end_bit, start_bit) < end_bit) {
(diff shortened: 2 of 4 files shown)


Triticum0:
@jessedye Haven't got your profile to give a accurate comparison but does really seem to be bottleneck on my system.

before 1x res 480p
Screenshot_20260820_170828

after 1x res 480p
Screenshot_20260820_170717

I tested without Hard FPU emulation and it doubled performace. so most likey cpu bottlenecked. 15 to 35 fps.

Let me know if you want me to test any of your other commits on my system.

also can test in more heavy area of the game. Also heard that Morrowind has a tendency to crash so be warned

jessedye:
> @jessedye Haven't got your profile to give a accurate comparison but does really seem to be bottleneck on my system.
> 
> before 1x res 480p Screenshot_20260820_170828
> 
> after 1x res 480p Screenshot_20260820_170717
> 
> I tested without Hard FPU emulation and it doubled performace. so most likey cpu bottlenecked. 15 to 35 fps.
> 
> Let me know if you want me to test any of your other commits on my system.
> 
> also can test in more heavy area of the game. Also heard that Morrowind has a tendency to crash so be warned

I apprecate you testing this! I've mostly been focused on Morrowind because it runs the worst, a lot of games (Halo, Halo 2, Tony Hawk 3, and GTA San Andreas) I've been able to track down and fix in my other PRs.

From my tests on a Pi 5 for comparison: mttcg_cpu_thread_fn 66% of samples with the vCPU thread pinned at ~100% of one core, GPU busy only 12 ms of a 53 ms frame — so CPU-bound agrees with what you're seeing.
Testing offer very welcome: #2977, #2979, #2980, #2981 are the others.

Triticum0:
There is a vulkan and opengl performace Issue meta if you want more test cases for badly performing games #1838, #1098.

Also don't know if virtual CPU overclock would help as its very heavy on original hardware. https://youtu.be/ekCq5K9cm54?t=39https://youtu.be/ekCq5K9cm54?t=39

hw/xbox/nv2a/pgraph/vk/vertex.c
```diff
@@ -53,8 +53,7 @@ void pgraph_vk_update_vertex_ram_buffer(PGRAPHState *pg, hwaddr offset,
     size_t end_bit = TARGET_PAGE_ALIGN(offset + size) / TARGET_PAGE_SIZE;
     size_t nbits = end_bit - start_bit;
 
-    if (find_next_bit(r->uploaded_bitmap, start_bit + nbits, start_bit) <
-        end_bit) {
+    if (find_next_bit(r->referenced_bitmap, end_bit, start_bit) < end_bit) {
```
mborgerson: We want to batch updates to the GPU copy of vertex data. The purpose of this code is to detect the case when the memory is changed while we build the command buffer of batched draws.

current:
if dirty now, and dirty before, bail

proposed:
if dirty now, and referenced before, bail


* if referenced at all in the draw, we need to flush the current command buffer before updating. Proposal looks good (even if description misses the original point).
* We sync all ram updates before executing the draw list. If we have a pending draw (OLD) that was not dirty, then one that is dirty (NEW), we end up using the NEW data for the OLD draw.
* this change leaves uploaded_bitmap vestigil. Just change where the bit gets set?
* what was the symptom of the issue which the PR is fixing?

mborgerson:
What was the symptom of this bug? Was there a particular issue with Morrowwind that is fixed by the PR?

The current implementation checks for the case when memory is dirty twice during draw batching into a command list, but fails to account for early draws from non-dirty memory that get batched along with draws from dirty memory--the proposed change looks correct.

However, this leaves uploaded_bitmap vestigial. Let's drop uploaded_bitmap now.

hw/xbox/nv2a/pgraph/vk/draw.c
```diff
@@ -1634,6 +1635,11 @@ static void sync_vertex_ram_buffer(PGRAPHState *pg)
         }
     }
 
+    for (int i = 0; i < num_syncs; i++) {
```
mborgerson, line 1637: This could be merged with the above loop

jessedye:
Thanks — both points addressed, and sorry for the slow reply.

**`uploaded_bitmap` is dropped.** You are right that it was left vestigial: after the check moves to `referenced_bitmap` it is allocated, set, cleared and freed without anything ever reading it. Removed in a second commit. I also rebased the branch, which drops an unrelated `bump-subproject-wraps.yml` change that had been riding along in the diff. It is now four files, +12/-9.

**On the symptom** — it showed up as a stall rather than as visible corruption, so I want to be precise about what I can and cannot attest to.

The check keys on "have we already uploaded any of these pages during this command buffer", which is a much wider set than "does any draw currently batched actually read these pages". Vertex RAM gets written continuously as the guest walks its data, so on the titles I profile it fired constantly and each one is a full submit-and-wait. `VK_FINISH_REASON_VERTEX_BUFFER_DIRTY` was among the largest single per-frame costs I measured, and it tracked how badly a game ran almost perfectly:

| title | vertex_buffer_dirty, ms/frame |
|---|---|
| Morrowind | 31.84 |
| GTA San Andreas | 18.74 |
| Halo | 13.17 |
| Tony Hawk's Pro Skater 3 | 0.00 |

THPS3 being the one title that ran well on that hardware, and the only one at zero.

What I should be careful about: my local tree also carries a separate change that redirects hot vertex regions, and I cannot cleanly separate the two contributions from those logs. So I would rather describe the mechanism — the trigger condition is far broader than the hazard it guards — than claim a specific delta for this patch alone.

The direction you identified is the other half of it, and I had not thought about it that way: a draw reading pages that were uploaded during an *earlier* command buffer has no bit set here, so the old check would miss that hazard entirely. I have not reproduced visible corruption from it, so I cannot claim it as an observed bug — but it is a real gap and `referenced_bitmap` closes it by construction, since it records what draws actually read rather than what happened to be uploaded.

hw/xbox/nv2a/pgraph/vk/vertex.c
```diff
@@ -63,7 +62,6 @@ void pgraph_vk_update_vertex_ram_buffer(PGRAPHState *pg, hwaddr offset,
     nv2a_profile_inc_counter(NV2A_PROF_GEOM_BUFFER_UPDATE_1);
     memcpy(r->storage_buffers[BUFFER_VERTEX_RAM].mapped + offset, data, size);
 
-    bitmap_set(r->uploaded_bitmap, start_bit, nbits);
```
mborgerson, line 66: nit: Trim empty line above

mborgerson:
> The check keys on "have we already uploaded any of these pages during this command buffer", which is a much wider set than "does any draw currently batched actually read these pages"

I'm not sure about this claim. The set of memory detected dirty should be a subset of all memory referenced by the command buffer

> What I should be careful about: my local tree also carries a separate change that redirects hot vertex regions

I'm not sure why performance data were presented here, if this change was not profiled in isolation

PR #2989, opened 2026-08-20
nv2a/vk: persist the pipeline and SPIR-V caches to disk

jessedye:
The Vulkan renderer creates its `VkPipelineCache` empty on every launch and
destroys it without reading it back, and recompiles identical GLSL through
glslang each run. Both caches are rebuilt from scratch every time xemu starts.

Two commits:

- **Pipeline cache** — load the blob at init, write it back on exit. The cache
  header carries vendor ID, device ID and a driver UUID, so data from a
  different device or an updated driver is rejected before the driver sees it,
  and the write goes through `g_file_set_contents` so an interrupted save
  cannot truncate a good cache.
- **SPIR-V cache** — key compiled SPIR-V on source, stage, the debug-shader
  flag and the xemu version, stored with the same sharded layout the GL shader
  cache already uses. Entries are rejected unless the SPIR-V magic word and
  word alignment hold.

Both honour the existing "Cache shaders to disk" setting.

One detail worth calling out: `finalize_pipeline_cache` alone is not enough.
It runs from the renderer's `ops.finalize`, which fires on a renderer switch,
while a normal quit exits without tearing the device down — so a save placed
only there never runs. The exit handler covers that, and a live flag keeps it
from touching a cache `finalize` already destroyed.

Measured on a Raspberry Pi 5 with Mesa V3DV, where pipeline compilation is
slow enough to be visible. The gain is confined to startup: the boot window
went 20.5 -> 43.7 fps and p99 1282 -> 309 ms once the caches were seeded.
**There is no steady-state gameplay gain and I am not claiming one** — only 27
shaders and 73 pipelines are created across 3089 frames of Morrowind, and a
45-second CPU profile of gameplay contains no compilation samples at all.

#2799 also implements disk caching for both, as part of a larger change that
adds macOS/MoltenVK bundling and loader plumbing. This is only the cache half,
against master, for whichever is easier to take.

### DIFF
--- a/hw/xbox/nv2a/pgraph/vk/draw.c
+++ b/hw/xbox/nv2a/pgraph/vk/draw.c
@@ -19,6 +19,7 @@
 
 #include "qemu/osdep.h"
 #include "qemu/fast-hash.h"
+#include "ui/xemu-settings.h"
 #include "renderer.h"
 #include <math.h>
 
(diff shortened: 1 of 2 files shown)


abaire:
Please see #2799. I think it'd be preferred to work with the author on an existing PR if it doesn't work for your needs rather than create an overlapping one.

abaire:
Just saw from the tail of your comment that you're aware, I still think it'd be better to have them split into two than to create a PR that'd either be obviated or require them to rebase.

jessedye:
Agreed — closing in favour of #2799. I've offered the isolated commits there in case they're useful for the split.

PR #3003, opened 2026-08-22
ui: fall back to a GL 3.1 context when 4.0 is unavailable

jessedye:
xemu asks for a GL 4.0 context and exits when it does not get one, in two places: the main window in `ui/xemu.c` and the offscreen context in `gloffscreen/sdl.c`. The window only composites a textured quad and the interface, and the offscreen context exists for display interop, so neither needs 4.0. A device that tops out earlier is turned away even when the Vulkan renderer is selected and fully supported.

V3DV on the Raspberry Pi 5 reports GL 3.1, so xemu cannot start there at all — the first thing a user sees is "graphics device on this system does not support OpenGL 4.0. xemu cannot continue and will now exit."

This asks for 4.0 first and keeps it wherever it is offered, so nothing changes for devices that have it, and retries at 3.1 only when it is not. In `gloffscreen/sdl.c` the window is recreated for the retry, because the GL attributes bind to the window when it is created and setting them afterwards has no effect on an existing one. Both failure paths now report `SDL_GetError()`, which is what made this diagnosable.

`ui/shader/xemu-logo.frag` is lowered to `#version 140` alongside, since it is the one interface shader still asking for 330 and a 3.1 context is useless while it fails to compile.

Testing:

- x86-64 / NVIDIA: still reports `GL_VERSION: 4.0.0`, so the 4.0 path is taken unchanged.
- aarch64 / V3DV on a Raspberry Pi 5: both contexts are now created at 3.1 (`GL_VERSION: 3.1 Mesa 24.2.8`) and startup proceeds past the interface into renderer initialisation, where master previously exited at the message box.

This removes the GL barrier but is not on its own sufficient for that hardware — with #2979 and #2996 also applied, startup now reaches Vulkan buffer creation and fails there with `VK_ERROR_OUT_OF_DEVICE_MEMORY`, which is a separate issue I am still looking at. Submitting this on its own because refusing to start on a 3.1 device is wrong regardless of what comes after it.

### DIFF
--- a/hw/xbox/nv2a/pgraph/thirdparty/gloffscreen/sdl.c
+++ b/hw/xbox/nv2a/pgraph/thirdparty/gloffscreen/sdl.c
@@ -48,28 +48,48 @@ GloContext *glo_context_create(void)
     SDL_GL_SetAttribute(SDL_GL_DEPTH_SIZE, 24);
     SDL_GL_SetAttribute(SDL_GL_STENCIL_SIZE, 8);
 
-    // Initialize rendering context
-    SDL_GL_SetAttribute(SDL_GL_SHARE_WITH_CURRENT_CONTEXT, 1);
-    SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 4);
-    SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 0);
-    SDL_GL_SetAttribute(
-        SDL_GL_CONTEXT_PROFILE_MASK,
-        SDL_GL_CONTEXT_PROFILE_CORE);
+    static const struct {
+        int major, minor;
+    } versions[] = { { 4, 0 }, { 3, 1 } };
 
-    // Create main window
-    context->window = SDL_CreateWindow(
-        "SDL Offscreen Window",
-        640, 480,
-        SDL_WINDOW_OPENGL | SDL_WINDOW_HIDDEN);
+    context->window = NULL;
+    context->gl_context = NULL;
+
+    for (unsigned i = 0; i < sizeof(versions) / sizeof(versions[0]); i++) {
+        SDL_GL_SetAttribute(SDL_GL_SHARE_WITH_CURRENT_CONTEXT, 1);
+        SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, versions[i].major);
+        SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, versions[i].minor);
+        SDL_GL_SetAttribute(
+            SDL_GL_CONTEXT_PROFILE_MASK,
+            SDL_GL_CONTEXT_PROFILE_CORE);
+
+        context->window = SDL_CreateWindow(
+            "SDL Offscreen Window",
+            640, 480,
+            SDL_WINDOW_OPENGL | SDL_WINDOW_HIDDEN);
+        if (context->window == NULL) {
+            continue;
+        }
+
+        context->gl_context = SDL_GL_CreateContext(context->window);
+        if (context->gl_context != NULL) {
+            break;
+        }
+
+        SDL_DestroyWindow(context->window);
+        context->window = NULL;
+    }
+
     if (context->window == NULL) {
-        fprintf(stderr, "%s: Failed to create window\n", __func__);
+        fprintf(stderr, "%s: Failed to create window: %s\n", __func__,
+                SDL_GetError());
         SDL_Quit();
         exit(1);
     }
 
-    context->gl_context = SDL_GL_CreateContext(context->window);
     if (context->gl_context == NULL) {
-        fprintf(stderr, "%s: Failed to create GL context\n", __func__);
+        fprintf(stderr, "%s: Failed to create GL context: %s\n", __func__,
+                SDL_GetError());
         SDL_DestroyWindow(context->window);
         SDL_Quit();
         exit(1);
--- a/ui/shader/xemu-logo.frag
+++ b/ui/shader/xemu-logo.frag
@@ -1,4 +1,4 @@
-#version 330
+#version 140
 uniform sampler2D tex;
 uniform vec4 in_ColorPrimary;
 uniform vec4 in_ColorFill;
--- a/ui/xemu.c
+++ b/ui/xemu.c
@@ -1050,10 +1050,22 @@ static void display_very_early_init(DisplayOptions *o)
     }
 
     if (m_context == NULL) {
+        SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 3);
+        SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 1);
+        m_context = SDL_GL_CreateContext(m_window);
+
+        if (m_context != NULL && epoxy_gl_version() < 31) {
+            SDL_GL_MakeCurrent(NULL, NULL);
+            SDL_GL_DestroyContext(m_context);
+            m_context = NULL;
+        }
+    }
+
+    if (m_context == NULL) {
         SDL_ShowSimpleMessageBox(SDL_MESSAGEBOX_ERROR,
             "Unable to create OpenGL context",
             "Unable to create OpenGL context. This usually means the\r\n"
-            "graphics device on this system does not support OpenGL 4.0.\r\n"
+            "graphics device on this system does not support OpenGL 3.1.\r\n"
             "\r\n"
             "xemu cannot continue and will now exit.",
             m_window);
--- a/ui/xui/main.cc
+++ b/ui/xui/main.cc
@@ -145,7 +145,8 @@ void xemu_hud_init(SDL_Window* window, void* sdl_gl_context)
 
     // Setup Platform/Renderer bindings
     ImGui_ImplSDL3_InitForOpenGL(window, sdl_gl_context);
-    ImGui_ImplOpenGL3_Init("#version 150");
+    ImGui_ImplOpenGL3_Init(epoxy_gl_version() >= 32 ? "#version 150" :
+                                                     "#version 140");
     ImPlot::CreateContext();
 
 #if defined(_WIN32)


abaire:
I'm worried that most users are going to be on the 4.0 path, leaving us without any pre-release guards to prevent the 3.1 path from breaking in the future since it will probably receive little to no manual testing.

[RenderDoc only supports 3.2 and later](https://renderdoc.org/docs/behind_scenes/opengl_support.html) so I'd be very concerned about dropping support, but if we only use 4.x for debug features, perhaps we should consider flipping things around so that debug/developer-focused builds use 4.x and the main release is limited to 3.1? This would mean we'd have a fast failure if somebody/some agent tried to introduce features that would not run on the lowest supported spec and could have a discussion about how to handle it.

Alternatively, can you suggest a mechanism to enforce 3.1 compliance automatically (e.g., via a GH workflow)?

jessedye:
That is a fair worry and I think it is the right one — an untested fallback rots. Two clarifications and then a concrete mechanism.

**This does not drop 4.x.** It asks for 4.0 first and keeps it wherever the driver offers it; 3.1 is only reached when 4.0 is refused. On any 3.2+ GPU nothing changes and RenderDoc is unaffected. So the risk is not lost debuggability, it is exactly what you said: a second path almost nobody exercises.

**On flipping the default**, I would push back gently. It would mean developers debug a configuration users do not run, which trades one untested path for another, and it does not help where it matters most — the GL renderer's own requirements are separate from the window's, so capping release builds at 3.1 could take the GL renderer away from people who can run it. I would rather test the low path than ship it to everyone.

**The enforcement you asked about is straightforward, because Mesa can lie about its version.** `llvmpipe` reports 4.5, and capping it reproduces the Pi's capability profile exactly on an ordinary runner:

```
$ glxinfo | grep 'OpenGL version'
OpenGL version string: 4.5 (Core Profile) Mesa 25.3.6

$ MESA_GL_VERSION_OVERRIDE=3.1 MESA_GLSL_VERSION_OVERRIDE=140 glxinfo | grep -E 'OpenGL (version|shading)'
OpenGL version string: 3.1 Mesa 25.3.6
OpenGL shading language version string: 1.40
```

Starting xemu under that cap, the log reports `GL_VERSION: 3.1 (Core Profile)` and every GLSL failure surfaces as a normal compile error, so a grep is enough to gate on. I ran it against two builds of this branch — one missing the offscreen-context fallback and one with it:

| build | detections |
|---|---|
| without the `gloffscreen` fallback | 1 (`Failed to create GL context`) |
| with it | 0 |

so it does fail when the floor is broken, rather than passing quietly.

A workflow along these lines:

```yaml
name: GL 3.1 floor
on: [push, pull_request]
jobs:
  gl31:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
        with: { submodules: recursive }
      - run: sudo apt-get update && sudo apt-get install -y mesa-utils libgl1-mesa-dri mesa-vulkan-drivers xvfb
      - run: ./build.sh -j$(nproc)
      - env:
          MESA_GL_VERSION_OVERRIDE: "3.1"
          MESA_GLSL_VERSION_OVERRIDE: "140"
          LIBGL_ALWAYS_SOFTWARE: "1"
          __GLX_VENDOR_LIBRARY_NAME: mesa
          SDL_AUDIODRIVER: dummy
        run: xvfb-run -s "-screen 0 1280x960x24" timeout 60 ./dist/xemu -config_path ci/gl31.toml > gl31.log 2>&1 || true
      - run: |
          if grep -qiE 'is not supported\. Supported versions|failed to compile|failed to link|Failed to create GL context|does not support OpenGL|uncompiled/unspecialized' gl31.log; then
            echo "::error::xemu used something unavailable at GL 3.1 / GLSL 1.40"
            grep -iE 'is not supported|failed to compile|failed to link|Failed to create' gl31.log | head -20
            exit 1
          fi
```

Same trick works for the Vulkan floor if that is ever wanted, since `lavapipe` can stand in for a modest device.

**One thing I should say plainly**: run that gate against this branch today and it still fails, later than before — past context creation and shader compilation, at an `glGetError()` assertion in `gl_render_frame`. The same assertion reproduces on a real Pi, so it is a genuine gap at 3.1 and not an artifact of the software stack. That is your point demonstrating itself, and I would rather show it than have the gate go in green and hide it. I am happy to keep this PR open until the floor is actually clean, or split the CI job out first so the failure is visible in the tree before the fallback lands — whichever you prefer.

jessedye:
Follow-up: I put the gate together and used it to chase the failure I mentioned, so there is now a precise answer rather than "it still fails somewhere".

Setting `MESA_DEBUG=1` alongside the cap makes Mesa name the offending call, and the whole thing is one root cause:

```
Mesa: error: GL_INVALID_ENUM in glCreateShader(GL_GEOMETRY_SHADER)
Mesa: error: GL_INVALID_VALUE in glShaderSourceARB
Mesa: error: GL_INVALID_VALUE in glCompileShader
Mesa: error: GL_INVALID_VALUE in glGetShaderiv
Mesa: error: GL_INVALID_VALUE in glGetShaderInfoLog(shader)
```

Geometry shaders are 3.2, so at 3.1 `GL_GEOMETRY_SHADER` is not a valid enum, `glCreateShader` returns 0, and the four calls that follow all fail against shader 0. The GPU-properties probe in `gpuprops.c` runs unconditionally. #2979 already handles the probe *failing* — `render_geom_shader_triangles` returning NULL is checked and the properties marked invalid — but nothing stops it setting five GL errors on the way, and `gl_render_frame` asserts on `glGetError()` at the end of the frame, so that is what takes it down.

The fix is a version gate at the top of `render_geom_shader_triangles`, before any GL call:

```c
if (epoxy_gl_version() < 32) {
    return NULL;
}
```

I tried it and it clears all five, including a `glPolygonMode(face)` error that turned out to be downstream of the same mess. **It belongs in #2979 rather than here**, though: the NULL return is only safe because #2979 added the check at the call site, and master passes the pointer straight to `determine_triangle_winding_order`. Putting it in this PR would introduce a null dereference on master, which I found the direct way by doing exactly that and watching it segfault.

So the sequence that makes the floor clean is #2979 (with the gate added) and then this. I have not pushed the gate to #2979 — your call whether you want it there or as its own change.

Worth noting for the original question: this is the kind of thing the gate catches that manual testing would not. The probe already "failed cleanly" and printed a friendly message; nothing was visibly wrong unless you were on a 3.1 device, and the actual crash was an assertion three subsystems away from the cause.

jessedye:
I ran the sequence end to end to see whether the floor actually comes clean, rather than leaving it as a claim. Building #2979 + the probe gate + this branch and running under the cap takes the errors from five to one:

| build | errors at the 3.1 floor |
|---|---|
| this branch alone | 5 |
| with #2979 and the probe gate | 1 |

The last one is not ours:

```
Mesa: error: GL_INVALID_ENUM in glPolygonMode(face)
```

`subprojects/imgui/backends/imgui_impl_opengl3.cpp:661`, in the state restore:

```c
if (bd->GlVersion <= 310 || bd->GlProfileIsCompat) {
    glPolygonMode(GL_FRONT, (GLenum)last_polygon_mode[0]);
    glPolygonMode(GL_BACK,  (GLenum)last_polygon_mode[1]);
} else {
    glPolygonMode(GL_FRONT_AND_BACK, (GLenum)last_polygon_mode[0]);
}
```

The comment above it says "Desktop OpenGL 3.0 and OpenGL 3.1 had separate polygon draw modes", which is true of the compatibility profile. In a 3.1 **core** context only `GL_FRONT_AND_BACK` is accepted, so `GlVersion <= 310` takes the wrong branch — the condition wants to be the profile check on its own. It is harmless in itself, being only a state restore, but `gl_render_frame` asserts on `glGetError()` at the end of the frame, so it is still fatal for us.

That leaves three ways forward and I do not think it is my call which:

1. Report it upstream to ImGui and carry a patch on the subproject until it lands.
2. Have `ImGui_ImplOpenGL3_Init` tell ImGui it is a core profile, if that is enough to steer the condition — I have not checked whether `GlProfileIsCompat` is inferred or settable.
3. Leave it, and have the CI job allow this single known ImGui error while failing on anything else. Least satisfying, but it would still catch every regression in our own code, which is what you were asking for.

Either way the useful part stands: from a bare "it asserts somewhere in `gl_render_frame`" to naming a line in a vendored dependency, in a handful of 35-second runs on a desktop, without touching a Pi. That is the argument for the gate better than anything I could assert about it.

jessedye:
Checking the option I left open above: telling ImGui it is a core profile is not available to us.

`GlProfileIsCompat` is derived, not settable, and ImGui deliberately skips the query below 3.2:

```c
if (bd->GlVersion >= 320)
    glGetIntegerv(GL_CONTEXT_PROFILE_MASK, &bd->GlProfileMask);
bd->GlProfileIsCompat = (bd->GlProfileMask & GL_CONTEXT_COMPATIBILITY_PROFILE_BIT) != 0;
```

That guard is their own fix for #6539, since the query is not valid below 3.2 — so at 3.1 the mask stays 0 and `GlProfileIsCompat` is false. The front/back branch is then taken purely on `GlVersion <= 310`, and there is no state we can set from `ImGui_ImplOpenGL3_Init` that changes it. It needs either a patch on the subproject or an upstream fix narrowing that condition to the profile check.

So it is options 1 or 3 from my previous comment, not 2.

jessedye:
Two useful things since the last comment.

**The ImGui fix has a home already.** `subprojects/imgui.wrap` points at `xemu-project/imgui`, so this does not need to wait on upstream ImGui — the condition can be narrowed in the fork and the wrap revision bumped:

```c
-if (bd->HasPolygonMode) { if (bd->GlVersion <= 310 || bd->GlProfileIsCompat) {
+if (bd->HasPolygonMode) { if (bd->GlProfileIsCompat) {
```

Since ImGui does not query the profile mask below 3.2 at all, `GlProfileIsCompat` is false there, so this takes the `GL_FRONT_AND_BACK` path — legal in both profiles. The behaviour it gives up is restoring different front and back polygon modes on a 3.0/3.1 *compatibility* context, which seems a fair trade for not emitting an illegal enum on every core-profile frame.

**Verified, with that applied:** #2979 + the probe gate + this branch + the ImGui narrowing leaves no GL-version errors at the floor. The `glPolygonMode` error is gone and nothing replaces it.

What is left in my capped run is `GL_OUT_OF_MEMORY`, which is llvmpipe's own limit under Xvfb rather than anything about 3.1 — it does not occur on the actual Pi, and #2996/#3004 do not touch it because it is a GL error rather than a Vulkan allocation. A CI job would want to either give the software rasterizer more room or exclude that one code, which is worth knowing before wiring the gate up as blocking.

So the full sequence for a clean floor is: #2979 (plus a version gate in the probe), the ImGui narrowing in the fork, and this. Happy to open the fork PR if you want it, or leave it to you — it is your repo and a one-line change.

PR #3009, opened 2026-08-25
ui: Log joystick devices that are not gamecontrollers, implement `gamecontrollerdb_path` config handling

abaire:
Hoping to give some debugging info/pointers for users with unmapped controllers, also implements loading the custom DB using the config param that already exists.

Reminder: We should add a section to the website docs (maybe cross linked from troubleshooting) if this gets merged.

### DIFF
--- a/ui/xemu-input.h
+++ b/ui/xemu-input.h
@@ -125,7 +125,7 @@ extern "C" {
 extern int *g_keyboard_scancode_map[25];
 
 void xemu_input_init(void);
-void xemu_input_process_sdl_events(const SDL_Event *event); // SDL_EVENT_GAMEPAD_ADDED, SDL_EVENT_GAMEPAD_REMOVED
+void xemu_input_process_sdl_events(const SDL_Event *event);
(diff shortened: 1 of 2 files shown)


ui/xemu-input.h
```diff
@@ -125,7 +125,12 @@ extern "C" {
 extern int *g_keyboard_scancode_map[25];
 
 void xemu_input_init(void);
-void xemu_input_process_sdl_events(const SDL_Event *event); // SDL_EVENT_GAMEPAD_ADDED, SDL_EVENT_GAMEPAD_REMOVED
+// Handles SDL events:
```
abaire: I'd be happy to remove this comment altogether; it already missed `_REMAPPED` and I would wager that some future additions will be unaware of this comment as well.

mborgerson: Yeah let's drop it

abaire: Done.

mborgerson:
Thanks!

PR #3010, opened 2026-08-26
nv2a/gl: cut texture-streaming cost (native S3TC upload + GL object recycle)

xionic:
# nv2a/gl: cut texture-streaming cost (native S3TC upload + GL object recycle)

## The problem

In PGR2, driving through certain sections of some tracks (e.g. Sydney Dawes
Point) spikes frame time hard, up to ~80ms/frame. That's where the game
streams in a lot of new textures in a single frame, and the texture upload
path is doing more work than it needs to:

1. Every S3TC/DXT texture gets decompressed on the CPU before upload, even
   though the renderer already requires `GL_EXT_texture_compression_s3tc`
   on the host (asserted at renderer init in `renderer.c`).
2. Every texture create/destroy does a fresh `glGenTextures` /
   `glDeleteTextures`, so the driver reallocates storage from scratch even
   when a same-sized texture was freed a moment ago.

I believe there's a general performance improvement too, though that's more
subjective currently.

## What I did

- Upload S3TC textures straight to `glCompressedTexImage2D` instead of
  decompressing on the CPU first. The renderer already asserts
  `GL_EXT_texture_compression_s3tc` is present at init; before this change
  we were just decompressing something we'd already confirmed the host
  could take compressed.
- Keep a small pool (256) of just-destroyed GL texture objects, keyed by
  their storage layout, and reuse a match on create so the driver can
  respecify in place instead of reallocating. `generate_texture` still
  writes every level regardless, so a mismatched or missing entry is always
  safe — the pool is purely a perf win, never a correctness dependency.

One gotcha worth flagging for review: the S3TC path has to be all-or-nothing
per texture, not per mip level — mixing compressed and uncompressed formats
across the mip chain leaves the texture incomplete and it renders black.

## How I tested it

The worst texture-burst frame I could find in PGR2 (vsync off, ~385 textures
in one frame): 82.8ms → 34.1ms, 12fps → 29fps.

I also profiled this branch against an unmodified master build on the same
scene: on heavy frames (2500+ draw calls) it's ~25% faster on average, and
the tail drops from 33ms to 14ms (p95) — the streaming stalls are basically
gone, and light frames are unchanged.

## How you can test it

Boot PGR2 and drive through Sydney Dawes Point — there's a certain corner
you'll come across which lags badly. Frame time should be noticeably
smoother with this patch.

I haven't tested other titles; this should be safe generally (the pool only
helps when a matching freed texture exists; otherwise it behaves exactly as
before), but PGR2 is the only thing I've actually driven through.

Full disclosure — I investigated this with Claude Code, and it also wrote
the patch.

### DIFF
--- a/hw/xbox/nv2a/pgraph/gl/renderer.h
+++ b/hw/xbox/nv2a/pgraph/gl/renderer.h
@@ -81,7 +81,22 @@ typedef struct TextureBinding {
     bool border_color_set;
     GLenum gl_target;
     GLuint gl_texture;
+    uint64_t storage_sig; /* recycle pool key: gl_target/format/dims/levels */
 } TextureBinding;
+
+/*
+ * Recycled GL texture objects. When a title streams new content it destroys
+ * and re-creates hundreds of same-sized textures per frame; freeing and
+ * re-allocating the GL object each time is a large per-frame cost. Keep a
+ * bounded pool of destroyed objects keyed by their storage signature and
+ * reuse a matching one, so the driver can respecify it in place rather than
+ * reallocate. Accessed only from the single pfifo/pgraph thread; no lock.
+ */
+#define NV2A_GL_TEX_POOL_SIZE 256
+typedef struct TexPoolEntry {
+    uint64_t sig;
+    GLuint gl_texture;
+} TexPoolEntry;
 
 typedef struct ShaderModuleCacheKey {
     GLenum kind;
@@ -193,6 +208,9 @@ typedef struct PGRAPHGLState {
     TextureBinding *texture_binding[NV2A_MAX_TEXTURES];
     Lru texture_cache;
     TextureLruNode *texture_cache_entries;
+
+    TexPoolEntry tex_pool[NV2A_GL_TEX_POOL_SIZE];
+    unsigned int tex_pool_count;
 
     Lru shader_cache;
     ShaderBinding *shader_cache_entries;
--- a/hw/xbox/nv2a/pgraph/gl/texture.c
+++ b/hw/xbox/nv2a/pgraph/gl/texture.c
@@ -671,15 +671,65 @@ static void upload_gl_texture(GLenum gl_target,
     }
 }
 
+/*
+ * Signature identifying a GL texture's storage layout, used to recycle a
+ * matching object from the pool so the driver can respecify it in place
+ * instead of reallocating. Only affects performance: generate_texture always
+ * fully respecifies every level, so a mismatched (or absent) object is still
+ * correct - the driver just reallocates as before.
+ */
+static uint64_t tex_storage_sig(GLenum gl_target, const TextureShape *s,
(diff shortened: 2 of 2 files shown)


hw/xbox/nv2a/pgraph/gl/renderer.h
```diff
@@ -81,8 +81,23 @@ typedef struct TextureBinding {
...
 
+/*
+ * Recycled GL texture objects. When a title streams new content it destroys
+ * and re-creates hundreds of same-sized textures per frame; freeing and
+ * re-allocating the GL object each time is a large per-frame cost. Keep a
+ * bounded pool of destroyed objects keyed by their storage signature and
+ * reuse a matching one, so the driver can respecify it in place rather than
+ * reallocate. Accessed only from the single pfifo/pgraph thread; no lock.
+ */
+#define NV2A_GL_TEX_POOL_SIZE 256
```
abaire, line 95: Why 256 and not 4 or 4096 or any other particular number?

hw/xbox/nv2a/pgraph/gl/renderer.h
```diff
@@ -81,8 +81,23 @@ typedef struct TextureBinding {
     bool border_color_set;
     GLenum gl_target;
     GLuint gl_texture;
+    uint64_t storage_sig; /* recycle pool key: gl_target/format/dims/levels */
```
abaire, line 84: xemu style is to use comments only for things that are particularly confusing. Since this PR is AI generated it'd be a good idea to go through and do your own evaluation of whether comments really add value and whether they're likely to be kept up to date as the code changes.

This comment: 
1) Summarizes the code in a way that is likely to go stale if somebody modifies the key calculation function.
2) Feels like a good indication that `storage_sig` would be better named `recycle_pool_key` (using "key" here and everywhere)

hw/xbox/nv2a/pgraph/gl/texture.c
```diff
@@ -639,15 +671,65 @@ static void upload_gl_texture(GLenum gl_target,
...
+            r->tex_pool[i] = r->tex_pool[--r->tex_pool_count];
+            return t;
+        }
+    }
+    return 0;
+}
+
+static void tex_pool_put(PGRAPHGLState *r, uint64_t sig, GLuint tex)
+{
+    if (r->tex_pool_count < NV2A_GL_TEX_POOL_SIZE) {
```
abaire, line 718: Doesn't this mean that the pool could plausibly become choked by a bunch of esoteric texture formats and become useless? Maybe an LRU cache would be better so old types can be evicted by newer ones?

Also, there's some work being done to support xemu on lower spec systems (granted it's Vulkan only) - given that we support texture scaling, are you concerned about the pool potentially holding on to a lot of VRAM?

abaire (commented):
I know it's in draft but had a couple quick comments from glancing through

PR #3013, opened 2026-08-29
Improve controller responsiveness and other hanging issues In DJ - Fight for New York

SkillerCMP:
Cleaned up seems my last version had my version of the ptimer Storm cleaner that you have also adjusted in the latest build . rebased and added the change back , also build and ran. on windows . Works good

### DIFF
--- a/hw/xbox/nv2a/nv2a.c
+++ b/hw/xbox/nv2a/nv2a.c
@@ -39,12 +39,12 @@ void nv2a_update_irq(NV2AState *d)
     }
 
     /* PGRAPH */
-    if (d->pgraph.pending_interrupts & d->pgraph.enabled_interrupts) {
+    if (qatomic_read(&d->pgraph.pending_interrupts) &
+        qatomic_read(&d->pgraph.enabled_interrupts)) {
         d->pmc.pending_interrupts |= NV_PMC_INTR_0_PGRAPH;
     } else {
         d->pmc.pending_interrupts &= ~NV_PMC_INTR_0_PGRAPH;
     }
-
     /* PTIMER */
     if (d->ptimer.pending_interrupts & d->ptimer.enabled_interrupts) {
         d->pmc.pending_interrupts |= NV_PMC_INTR_0_PTIMER;
@@ -320,9 +320,9 @@ static void nv2a_reset(NV2AState *d)
     /* seems to start in color mode */
     d->vga.msr = VGA_MIS_COLOR;
 
-    d->pgraph.waiting_for_nop = false;
+    qatomic_set(&d->pgraph.waiting_for_nop, false);
     d->pgraph.waiting_for_flip = false;
-    d->pgraph.waiting_for_context_switch = false;
+    qatomic_set(&d->pgraph.waiting_for_context_switch, false);
 
     d->pmc.pending_interrupts = 0;
     d->pfifo.pending_interrupts = 0;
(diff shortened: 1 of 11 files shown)


abaire:
Some meta comments, I haven't looked at the change in detail as I don't think the PR would be mergeable in the current state if it comes out of draft:

I suggest decomposing this so the controller fixes are in a separate PR than the fixes needed to prevent the guest-side livelock.

You'll also need to fix all of your commit messages (or at least the first commit message if you want to have the changes squashed into one commit). You can run `git log -- <some file that you changed>` to see an example. qemu/xemu style requires something like "subsystem: description of change". You can put more detail after the first line, explaining why the change is necessary, how to test it, etc...

At first glance I suspect this is mostly LLM-written. If so, make sure you actually understand the change and have done a diligent code review. There are things in the comments that particularly stick out. xemu style is also to omit comments except when explaining very tricky code. I would suggest removing them and instead adding explanations in the PRs. You can add comments on the code yourself to guide reviewers through sections or write an overall description in the PR.

Your current description does not convey what the PR does and refers to discussions outside of the PR that people won't be able to find five years from now when they `git blame` the code and find their way back to the PR.

SkillerCMP:
there is no information on this format for notes and rules, when it comes to submitting . Yes its largly AI, i dont and never will hide that fact, im not a programer im a code maker "30 years"(i know ASM better then i know this), i made the Cheatcode that corrected this issue(Mainly), but that wont be use it so i desided to see if i could find the issue inside XEMU it self with the help of AI.  Do i understand what the code is doing, Yes, Would i know how to write it 100% with out the help of the AI no, 
Also i use Github Desktop. so not sure of the commands your starting . 
And yes few of my notes are because pulled the fix code from my debugger version . 

```
Your current description does not convey what the PR does and refers to discussions outside of the PR that people won't be able to find five years from now when they git blame the code and find their way back to the PR.
```
lol this makes sense, Sorry i dont tend to try and submit stuff i still have a bunch to learn when it comes to forks and this stuff .  this is why i tend to stick to the actual just making codes :P


Basics of the changes is this 3 fixs each one improving the issues. 
note there is still more issues inside DJ but this makes the game playable, the games very finiky when it comes to missing a timer. 
```
Controller fix = SDL dont block OHCI/BQL
PGRAPH fixes   = renderer dont block guest MMIO, IRQs, or flip progress
Atomic fixes   = threads dont miss each other's state changes
```


I will work on correcting things, thanks for the Guidince

abaire:
> there is no information on this format for notes and rules, when it comes to submitting . 

There may be docs, but I think most contributors just learn this via comments on their initial PR. Feel free to create a feature request for contributor docs or to ask on the Discord. I think these would be useful and probably also a good place to clarify expectations around AI use.

> Do i understand what the code is doing, Yes, Would i know how to write it 100% with out the help of the AI no, 

Generally that's fine. At this point I think it's expected that most developers will leverage LLMs, they're a great tool when used carefully. The key thing is that you understand the code you're proposing and are able to make adjustments when necessary.

> Also i use Github Desktop. so not sure of the commands your starting . And yes few of my notes are because pulled the fix code from my debugger version .

I'm fairly confident you can do a git blame or otherwise look at commits via the GUI tool. You can take a look at the commit history for any of the files you've changed to get an idea of what a commit message should look like.

> lol this makes sense, Sorry i dont tend to try and submit stuff i still have a bunch to learn when it comes to forks and this stuff . this is why i tend to stick to the actual just making codes :P

That's fine, feel free to ask for help if you get stuck. Generally you shouldn't have to close PRs, you can use force pushing and additional branches to split out your commits into separate PRs. I'd imagine that whatever LLM you're using to assist in coding can also provide insight on the best way to do this with the GitHub GUI tool.


> Basics of the changes is this 3 fixs each one improving the issues. note there is still more issues inside DJ but this makes the game playable, the games very finiky when it comes to missing a timer.

I'd probably split all three of those into separate PRs as I think they're somewhat independent. That will simplify the review process by keeping each change focused and succinct.

PR #3023, opened 2026-09-01
nv2a: process common PGRAPH inline packets in bulk

Mainkill1:
## Summary

Reduce PFIFO/PGRAPH CPU overhead by processing non-incrementing inline vertex and index packets as bounded spans instead of dispatching every dword through the generic method handler.

Large `INLINE_ARRAY`, `ARRAY_ELEMENT16`, and `ARRAY_ELEMENT32` packets contain repeated writes to one NV2A method. The generic path repeated method dispatch and safety checks for every word even though packet parsing had already established the span.

### Previous flow

```text
Decode non-incrementing packet of N words
        │
        ▼
For each word
        │
        ├── Dispatch generic method handler
        ├── Check begin/end state
        ├── Check output capacity
        ├── Check draw-array expansion state
        ├── Append one or two elements
        └── Update shared output length
```

The repeated control work scales with packet length.

### New flow

```text
Decode non-incrementing packet of N words
        │
        ▼
Recognized inline vertex/index method?
        │
        ├── No  ─► Generic scalar handler
        │
        └── Yes
              │
              ▼
        Validate packet state and total capacity once
              │
              ▼
        Process bounded word span
              │
              ├── ARRAY_ELEMENT16: low index, then high index
              ├── ARRAY_ELEMENT32: preserve dword order
              └── INLINE_ARRAY: preserve dword order
              │
              ▼
        Commit output length once
```

Incrementing packets and builds with per-method tracing enabled retain the scalar path so observability and uncommon semantics are unchanged.

## Performance effect

The bulk path removes per-word function dispatch, repeated capacity checks, repeated draw-array expansion checks, and repeated writes to the shared length field. The earlier instrumented inline-array smoke improved by approximately 1.4%. Production counter increments were subsequently removed, so the shipping path is leaner; revised dedicated timing is still required before claiming a larger number.

## Design rationale

All untraced non-incrementing packets use the bounded path, including one-word packets. There is no guessed packet-length cutoff.

`ARRAY_ELEMENT16` produces two indices per 32-bit packet word because that is the NV2A format. Capacity is derived from the existing destination storage contract. The helper first proves the complete packet fits, preventing a partially applied packet on overflow.

## Test statistics

- Isolated instrumented inline-array smoke: approximately 1.4% faster before shipping counters were removed.
- Combined-train PGR2 lagspot median changed -0.85% Vulkan 1x, +0.10% Vulkan 4x, +1.28% OpenGL 1x, and +15.02% OpenGL 4x. This broad record mixes renderer work, so it is supporting evidence rather than isolated attribution.
- Packet contract: exact ARRAY_ELEMENT16 low/high ordering and overflow rejection pass.
- Combined matrix: 1,008/1,008 records completed with no assert, crash, or hang.

### Complete PR train result

The final stacked head was compared with upstream `d73326b` on the dedicated Windows/NVIDIA rig across 504 matched record/cell pairs:

```text
equal-weight geometric time:  -1.20%
total guest-measured time:     451.58 s -> 447.70 s  (-0.86%)
PGRAPH fence polling:          33.71% to 48.44% faster
Vulkan surface download:       23.72% to 27.57% faster
Vulkan memory-pressure stages: 17.24% to 19.74% faster
CPU SSE scalar:                0.27% to 3.39% faster
```

CPU SSE attribution: PRs #2-#10 do not change `target/i386` or APU code. An identical-source A/B found the earlier approximately 23% SSE loss was caused by the lab Clang build versus xemu’s pinned upstream GCC toolchain; LTO alone was inconclusive. The small train-wide SSE movement above is supporting matrix data, not a source-level SSE optimization claim.

Per-cell aggregate result (negative time means faster):

| Renderer | Scale | Equal-weight geometric time | Total guest time |
| --- | ---: | ---: | ---: |
| Vulkan | 1x | -1.32% | -0.25% |
| OpenGL | 1x | -1.05% | +0.19% |
| Vulkan | 4x | -1.89% | -1.54% |
| OpenGL | 4x | -0.52% | -1.73% |

PR8 also reduces the Vulkan inline-vertex device/staging pair's initial requested capacity from 2,684,349,440 bytes to 16,777,216 bytes: 2,667,572,224 bytes saved (about 2.484 GiB, 99.375%). Buffers then grow to measured demand. This is allocation capacity, not a claim that every driver commits the same amount of physical VRAM.

All eight executions completed: baseline/candidate, OpenGL/Vulkan, and 1x/4x. OpenGL retained 126/126 baseline hashes in both cells. Vulkan retained 124/126; the two intentional differences changed from 13 failing source-precision tiles on upstream to zero failures on the train at both scales.

This is one diagnostic A/B pair per cell. It proves direction and combined correctness, not a confidence interval. S3TC fenced streaming measured 4.06% to 10.23% slower and requires focused repetition before a release-wide performance claim.

## Validation

The standalone packet contract passes, including exact 16-bit low/high ordering and overflow rejection. The final four-cell matrix completed 8/8 executions and 1,008/1,008 records without assert, crash, or hang. The final stacked Release configuration also compiles on Linux.

## Dependencies

None. The validation stack also carried xemu-project/xemu#2941, but this submission neither duplicates nor requires that change.

## Public validation suite

- Source and operator documentation: https://github.com/Mainkill1/xemu-perf-tests
- Ready-to-run ISO: https://github.com/Mainkill1/xemu-perf-tests/releases/tag/v1.0.0
- Full-suite gate: 141/141 records passed, zero oracle failures, and plain-disc boot verified on two Windows systems.
- ISO SHA-256: `7b8eb7af66a87fad93bcea019208457e42aaff2bd81b1353df2b6c378db7707c`


abaire:
This stomps on PR #2930 and creates weird paths that look unnecessarily complicated to avoid very minor duplication that would result from creating custom handlers for the affected methods and handling tracing in bulk rather than falling back.

If there's an actual human behind this, reach out to the author of PR #2930 to discuss merging efforts and in the future do a better job of evaluating the generated code before blitzing projects with multiple PRs.

Mainkill1:
I agree, and this is a little overkill/messy for the work being implemented and code layout.
I have upped the game and will make sure all releases are triple checked and harder testing gets performed.

All code is to improve the overall usage of the emulator and I have revised the assault paths to be more in-depth and harder hitting with more determined acceptance ratios.

I won't push anymore until the test finish up in a few days.
New tests have been update, baseline build failures have been patched out so my testing won't ever really be a 1 to 1 comparison but each and every performance branch after a fully patched baseline system will be viewed in depth... just with a more aggressive AI and then a final human look over. I Tend to keep busy so this is almost entirely automated across 3 rigs.

Test Suite
https://github.com/Mainkill1/xemu-perf-tests/tree/feature/eng-2026-523-report-query-capsule-u08aa8c76

Current builds/branches being tested:
https://github.com/Mainkill1/xemu/tree/Full-Speed

If you see something you like feel free to open a PR ahead of time to implement, any PR I would want to push will from now on be scanned for issues or existing next time as to not "stomp" on others work.

PR #3036, opened 2026-09-07
ci: Run unit tests alongside build

abaire:
At the moment this is mainly to keep bot-generated CLs from breaking any tested QEMU bits, but I think generally having more unit tests on the Xbox-specific codebase would be good (I'm hopefully adding some in #3035).

### DIFF
--- a/.github/workflows/build-tests.yml
+++ b/.github/workflows/build-tests.yml
@@ -1,0 +1,30 @@
+name: Unit tests
+
+on:
+  workflow_call:
+
+jobs:
+  test:
+    name: Run unit tests
+    runs-on: ubuntu-22.04
+    steps:
+    - name: Download source package
+      uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v4
+      with:
+        name: src
+    - name: Extract source package
+      run: |
+        mkdir src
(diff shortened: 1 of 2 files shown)


.github/workflows/build-tests.yml
```diff
@@ -0,0 +1,30 @@
...
+on:
+  workflow_call:
+
+jobs:
+  test:
+    name: Run unit tests
+    runs-on: ubuntu-22.04
+    steps:
+    - name: Download source package
+      uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v4
```
abaire, line 12: Do we want to keep these pinned? I think several of these actions are up to v7 already and presumably Dependabot respects commit pinning and doesn't bother to suggest things that might contain security/functionality improvements.

mborgerson: Yeah, we keep the hash so the tagged action doesn't move on us unexpectedly. Dependabot is supposed to update both the commit hash and tag comment when a new version of the Action is released, though it looks like some of the comments are stale for some reason

abaire: Ah, I trusted the comment and didn't realize that's the hash for v8. Maybe we should scrub the comments since it doesn't seem to fix them?

PR #3051, opened 2026-09-15
xbox: Add IO mechanism to interrogate xemu from guest

abaire:
Uses two ports to allow the guest to identify xemu as the host and fetch version information.

Demo: https://github.com/abaire/nxdk_low_level_nv2a_tests/blob/7b12f6d34492e790abf986a948e324d31b702ddc/src/xemu_guest_info.h#L1

This could also be expanded to provide more metadata. A concrete use case would be to skip tests that rely on some xemu-specific state (e.g., tests that only make sense at render scale > 1). Similarly it could be expanded to allow the guest to modify xemu state (e.g., having tests that set the render scale to verify correct operation).


abaire:
Just so it's clear, I won't be offended if this isn't something we want to incorporate. I think it is a useful feature for homebrew and especially for test suites but obviously it diverges from a pure focus on emulating the hardware.

